PR #1, opened 2019-07-16
mrc-361 create basic app

hillalex:
Here I've just changed the prototype to use Freemarker instead of Thymeleaf (I didn't like the latter, and we're already used to the former) plus jest instead of mocha, for all the reasons we've switched elsewhere. It still has a bunch of nonsense routes and vue components, but leaving them in for reference and we can update it as we go along


hillalex:
> Looks good! Are we just going to rename this repo? (once we have a name!)

I guess so!

PR #3, opened 2019-07-26
Mrc 368 Forms authentication

EmmaLRussell:
We originally thought to use Basic auth, but basic doesn't really support logging out, as the browser retains the Authorization header to pass on subsequent calls and workarounds to clear this are pretty flaky.  (See e.g. https://stackoverflow.com/questions/233507/how-to-log-out-user-from-web-site-using-basic-authentication). So I've used Forms authentication instead, which is very straightforward with pac4j.

This ticket is just the login/logout wiring and doesn't include the user database. Pac4j's built-in dummy SimpleTestUsernamePasswordAuthenticator is currently configured (login is successful if username and password are the same), to be replaced by DbProfileService in a separate ticket.


app/src/test/kotlin/org/imperial/mrc/modelserver/unit/templates/LoginTests.kt
```diff
@@ -0,0 +1,54 @@
+package org.imperial.mrc.modelserver.unit.templates
+
+import org.junit.jupiter.api.Test
+import org.springframework.ui.ConcurrentModel
+import org.assertj.core.api.Assertions.assertThat
+
+class LoginTests
```
EmmaLRussell: Rules have been deprecated in JUnit5 and replaced with extensions, but since we weren't really using the Freemarker Template Rule as much as a rule since it lives in the companion object of the test class, I've just converted it into an ordinary class.

app/src/test/kotlin/org/imperial/mrc/modelserver/integration/ModelserverApplicationTests.kt
```diff
@@ -16,8 +17,10 @@ class ModelserverApplicationTests(@Autowired val restTemplate: TestRestTemplate)
 	}
 
 	@Test
-	fun `Assert blog page title, content and status code`() {
+	fun `Assert redirects to login page`() {
```
EmmaLRussell: I'm not really sure quite how these integration tests need to work, can we talk about it on Monday?

app/build.gradle
```diff
@@ -33,10 +33,17 @@ dependencies {
 	implementation 'org.jetbrains.kotlin:kotlin-reflect'
 	implementation 'org.jetbrains.kotlin:kotlin-stdlib-jdk8'
 	implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.2.1'
+	implementation 'org.pac4j:spring-webmvc-pac4j:3.2.0'
+	implementation 'org.pac4j:pac4j-http:3.3.0'
+	implementation 'org.pac4j:pac4j-sql:3.3.0'
```
EmmaLRussell: Not actually using the sql package yet, but we will..

PR #4, opened 2019-07-29
Mrc 371 Create users db

EmmaLRussell:
Eventually there will probably be some other stuff in here, but for now we just need a sql db to save our users in.


app/src/test/kotlin/org/imperial/mrc/modelserver/database/UsersTests.kt
```diff
@@ -0,0 +1,47 @@
...
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.BeforeAll
+import org.junit.jupiter.api.Test
+import org.pac4j.core.credentials.password.JBCryptPasswordEncoder
+import org.pac4j.sql.profile.DbProfile
+import org.pac4j.sql.profile.service.DbProfileService
+import org.postgresql.jdbc3.Jdbc3PoolingDataSource
+import org.mindrot.jbcrypt.BCrypt
+
+//Tests that pac4j can access the database
```
EmmaLRussell: I just wanted to check that the pac4j DbProfileService class could access the table to add and remove users.

db/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM postgres:10.3
```
EmmaLRussell: Largely copied from the montagu-db configuration, not sure if it will all be relevant. 
I've ended with the database being called modelserver-db (with a dash) and the container being modelserver_db (with an underscore). Can consistify these if you prefer.

db/migrations/sql/V2019.07.29.1316__InitialSchema.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE users
```
EmmaLRussell: I think we can add other columns here like FirstName and LastName which would be applied to the profile as attributes

db/scripts/run
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+docker run --rm -d --network=host --name modelserver_db mrcide/modelserver_db:mrc-371
```
EmmaLRussell: This should be replaced by master, though the image is not currently being automatically built..

app/src/test/kotlin/org/imperial/mrc/modelserver/database/UsersTests.kt
hillalex: I think we'll want to use Libsodium, but we can raise a separate ticket for that.

db/Dockerfile
hillalex: `modelserver` was just a placeholder name for the prototype. I think makes sense for the container to be called `hint-db` and the db itself just be called `hint` (we need to rename this repo and the gradle project as well, will raise a ticket for that!)

EmmaLRussell:
> I was kind of assuming we'd have the db in its own repo like with Montagu. I guess it doesn't have to be but it might make CI nicer - with this setup we'll end up re-building the db images every time the Kotlin app changes and conversely, running all the java tests every time we add a new migration. I know we're doing that with OW (and the OW CI logic is a bit of nightmare) but my feeling is the de-coupled pattern is better. What do you think?

I think you're right, I dithered a bit about which to do. It was probably easier to get it basically working all in one repo, but I'll make a new repo, hint-db and move everything from the db folder into that.

EmmaLRussell:
Created a new ticket for using Libsodium, mrc-377

PR #6, opened 2019-07-30
add Travis and Jacoco

hillalex:
* Added Jacoco for Kotlin test coverage
* Added Travis config
* Added `run-development-dependencies.sh` for a convenient local dev script that tears down deps on ctrl-c


scripts/run-dependencies.sh
```diff
@@ -1,9 +1,25 @@
 #!/usr/bin/env bash
-docker run --rm -d --network=host --name modelapi_redis redis
-docker run --rm -d --network=host --name modelapi_model mrcide/modelapi:latest
```
hillalex: Removed these unused deps which were used for the prototype

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+          - $HOME/app/.gradle/caches/
+          - $HOME/app/.gradle/wrapper/
+          - $HOME/app/static/node_modules
+      jdk:
+      - openjdk8
+      install:
+      - npm install codecov -g
+      - ./app/gradlew -p app
+      before_script:
+      - sudo /etc/init.d/postgresql stop
```
hillalex: Postgres is running by default so stop it here so we can map port 5432 to our db container

scripts/run-dependencies.sh
```diff
@@ -1,9 +1,25 @@
...
+CONTAINER=hint_db
+docker network create $NETWORK
+
+docker run --rm -d \
+  --network=$NETWORK \
+  --name $CONTAINER \
+  -p 5432:5432 \
+  mrcide/hint-db:mrc-371
+
+# From now on, if the user presses Ctrl+C we should teardown gracefully
```
EmmaLRussell: I don't think the trap needs to be here as well as in the dev script?

hillalex: Not super important but allows the script to be stopped any point from here on, rather than just at the end.

PR #7, opened 2019-07-30
mrc-375 PJNZ upload

hillalex:
This PR makes the front end code a bit nicer by introducing a js api client to wrap shared logic, and adds functionality to get existing state on page load so that state persists as long as the session does, even if the user navigates away and back again.


src/app/static/src/app/store/baseline/baseline.ts
```diff
@@ -1,17 +1,19 @@
...
+import {actions, PJNZ} from './actions';
 import { mutations } from './mutations';
 import {RootState} from "../../main";
 
 export interface BaselineState {
     pjnzError: string
     country: string
-    complete: boolean
+    complete: boolean,
+    pjnzFilename: string
```
EmmaLRussell: I suspect we'll eventually restructure to have a 'FileUpload' type with error, filename properties, of which PJNZ will be one of several properties of this type in the state. But that doesn't need to go in now.

hillalex: Yeah this sounds sensible.

hillalex:
> All works like a dream. A red tick looks a bit odd to my eye (but that's how it is in the mockups)! What do you think?

Heh, good point. In general having red as a primary colour is somewhat confusing! But I guess in context, given the stepper goes red when complete, links are red etc., the red tick is ok?

PR #11, opened 2019-08-01
Mrc 373 User Repository

EmmaLRussell:


src/app/build.gradle
```diff
@@ -50,6 +51,10 @@ dependencies {
 	testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
 }
 
+configurations {
```
EmmaLRussell: Get errors about having multiple logging implementations without this

src/app/src/main/kotlin/org/imperial/mrc/hint/db/DataSourceConfig.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.boot.jdbc.DataSourceBuilder
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import javax.sql.DataSource
+
+@Configuration
+open class DataSourceConfig
+{
+    @Bean
+    @ConfigurationProperties(prefix="spring.datasource")
```
EmmaLRussell: This will pull in datasource configuration from the application properties files, from keys with the spring.datasource prefix

src/app/src/main/resources/application-test.properties
```diff
@@ -0,0 +1,2 @@
+spring.datasource.jdbc-url=jdbc:postgresql://localhost/hint
```
EmmaLRussell: Use this configuration in  tests, where the user CLI classes are not running within a container, hence can't access the hint db by the container name, needs to go to localhost instead,

src/userCLI/src/main/resources/application.properties
```diff
@@ -0,0 +1,6 @@
+spring.datasource.jdbc-url=jdbc:postgresql://hint_db/hint
+spring.datasource.username=hintuser
+spring.datasource.password=changeme
+spring.datasource.driver-class-name=org.postgresql.Driver
+
+logging.level.root=ERROR
```
EmmaLRussell: Keep Spring logging to a minimum when running userCLI

src/app/src/main/kotlin/org/imperial/mrc/hint/security/Pac4jProfileService.kt
```diff
@@ -0,0 +1,26 @@
...
+{
+    companion object {
+        val salt = "\$2a\$10\$whERBRv1HHJJ5/BEjtKYme"
+    }
+
+    @Bean
+    open fun profileService(): DbProfileService
+    {
+        //TODO: Use Libsodium
+        return DbProfileService(TransactionAwareDataSourceProxy(dataSource), JBCryptPasswordEncoder(salt))
```
EmmaLRussell: This TransactionAwareDataSourceProxy is the bit of magic you need to make the Pac4j DbProfileService work with test transactions declared via Spring's @Transactional

src/userCLI/src/test/kotlin/org/imperial/mrc/hint/userCLI/AppTests.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.jupiter.api.extension.ExtendWith
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.test.context.ActiveProfiles
+import org.springframework.test.context.junit.jupiter.SpringExtension
+import org.springframework.transaction.annotation.Transactional
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.ApplicationContext
+
+//The hint db container must be running to run these tests
+@ActiveProfiles(profiles=["test"])
```
EmmaLRussell: Picks up application-test.properties to augment/overwrite application.properties

src/app/src/main/resources/application-test.properties
hillalex: I guess this means the value set below in `application.properties` won't work for local development either. We'll have to figure out how to define one set of properties for dev, another for prod. Looks like there's some relevant documentation here: https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html

src/userCLI/build.gradle
```diff
@@ -36,6 +48,16 @@ if (hasProperty('docker_name')) {
...
+    push = false
     tag = cli_docker_name
     tagVersion = cli_docker_version
 }
+
+test {
+    useJUnitPlatform()
+}
+
+jar{
```
hillalex: This should already be enabled in the root gradle config's "subprojects" block; you should be able to move the `configurations` and `test` blocks into there as well since they're shared by both subprojects

src/app/src/main/kotlin/org/imperial/mrc/hint/security/Pac4jProfileService.kt
hillalex: I can see why it took a while to figure this out!

src/app/src/main/resources/application-test.properties
EmmaLRussell: We will - I thought we could cross that bridge when we put the Pac4 Db authentication into login in the next sprint...

PR #16, opened 2019-08-20
use a randomly generated salt for each user

hillalex:
Turns out BCrypt stores the salt alongside the password, so no extra work required to store salts in the db. Just needed to make sure we generate a new random salt for each password.


src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecureDbProfileService.kt
```diff
@@ -0,0 +1,20 @@
+package org.imperial.mrc.hint.security
+
+import org.pac4j.core.credentials.password.JBCryptPasswordEncoder
+import org.pac4j.core.credentials.password.PasswordEncoder
+
+class SecurePasswordEncoder : PasswordEncoder {
```
EmmaLRussell: Could add a test for these methods together I guess - test that matches(x, encode(x)) is true

PR #17, opened 2019-08-20
Mrc 402 Request password reset link front end

EmmaLRussell:
PR for front end only!


src/app/static/src/app/index.ts
```diff
@@ -1,19 +1,36 @@
...
-    },
-    render: h => h(Stepper),
-    methods: {
-        ...mapActions({loadBaseline: 'baseline/getBaselineData'})
-    },
-    beforeMount: function() {
-        this.loadBaseline()
-    }
-});
+export const app =
```
EmmaLRussell: So this might be completely the wrong approach! I wanted to do the ForgotPassword component as vue rather than a server template because I thought that would be nicer to show response messages in js than just using an html form. But then it didn't seem quite right to use the standard 'app' vue element that you get once you're logged, so I've done this, choosing the component to render based on element found in the document (as defined by the template). It does still use the same store type as before, but with a password module added. 

An alternative would be to have this render a single root component which conditionally renders a Stepper or a ForgotPassword - not sure what the most correct basis for this decision would be though..

src/app/static/src/app/components/password/ForgotPassword.vue
```diff
@@ -0,0 +1,64 @@
...
+            resetLinkRequested: state => state.resetLinkRequested
+        }),
+       components: {
+            ErrorAlert
+        },
+        methods: {
+            ...mapActions({requestResetLink: 'password/requestResetLink'}),
+            handleRequestResetLink: function(event: Event){
+                const form = this.$refs.forgotPasswordForm as HTMLFieldSetElement;
+                if (form.checkValidity()) {
```
EmmaLRussell: Built in form validation - not that pretty but does the job. Could put in nicer validation later if time..

hillalex: This should play nicely with bootstrap actually https://getbootstrap.com/docs/4.0/components/forms/?#validation

src/app/static/src/app/index.ts
hillalex: I'm not sure what advantage this exact approach offers because the same js bundle would have to be included on both the ForgottenPassword page and the regular app page. 

But I can see why splitting up the 2 into separate apps makes sense i.e. two separate entrypoints for webpack. So the forgotten password vue app would live in a different file and be bundled into it's own bundle, then each page only has to load the js it needs.

src/app/static/src/app/main.ts
```diff
@@ -6,9 +6,11 @@ import {
     surveyAndProgram,
     SurveyAndProgramDataState
 } from "./store/surveyAndProgram/surveyAndProgram";
+import {password, initialPasswordState, PasswordState} from "./store/password/password";
 
 export interface RootState {
     version: string;
+    password: PasswordState,
```
hillalex: Given the proposed separateness of the main app from the pw logic this wouldn't need to be here. Indeed the set password app would be much simpler - no modules, the root state for that would just be the password state.

src/app/static/webpack.config.js
```diff
@@ -70,6 +64,26 @@ module.exports = {
...
+    output: {
+        path: path.resolve(__dirname, './public/js'),
+        publicPath: '/public/js/',
+        filename: 'forgotPassword.js'
+    }
+};
+
+module.exports = [appConfig, forgotPasswordAppConfig];
+
 if (process.env.NODE_ENV === 'production') {
```
hillalex: Cool that this multiple bundle thing was so simple to configure! But now that an array is being returned instead of a single object do these production optional properties have to be applied explicitly to each item in the array? (realise we aren't yet setting this flag either, just added a ticket so we don't forget https://vimc.myjetbrains.com/youtrack/issue/mrc-425)

EmmaLRussell: Good point!

PR #18, opened 2019-08-20
mrc-415 backend for uploading survey and program files

hillalex:
1. Pull out file saving logic into FileManager class
2. Create API client for talking to hintr (and add hintr into run-dependencies script)
3. Add endpoints for uploading/validating survey and program data


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/DiseaseTests.kt
```diff
@@ -0,0 +1,66 @@
...
+        testFile.parentFile.mkdirs()
+        testFile.createNewFile()
+
+        val body = LinkedMultiValueMap<String, Any>()
+        body.add("file", FileSystemResource(testFile))
+        val headers = HttpHeaders()
+        headers.contentType = MediaType.MULTIPART_FORM_DATA
+
+        val entity = testRestTemplate.postForEntity<String>("/disease/survey/", HttpEntity(body, headers))
+        assertThat(entity.statusCode).isEqualTo(HttpStatus.BAD_REQUEST)
```
hillalex: This tests won't pass until the hintr responses are correctly formatted, so I've just stuck a `Disabled` flag on these for now.

scripts/run-dependencies.sh
```diff
@@ -11,10 +11,18 @@ docker run --rm -d \
...
+  -p 8888:8888 \
+  mrcide/hintr:latest
+
 # From now on, if the user presses Ctrl+C we should teardown gracefully
-trap on_interrupt INT
+trap cleanup INT
+trap cleanup ERR
 function on_interrupt() {
   docker stop $CONTAINER
+  docker stop hintr
```
EmmaLRussell: Should probably be consistent as to whether using variables for image names or not. Not clear here why hint_db should be $CONTAINER but hintr hardcoded.

src/app/build.gradle
```diff
@@ -16,6 +16,12 @@ jacocoTestReport {
 		xml.enabled = true
 		csv.enabled = false
 	}
+
+	afterEvaluate {
+		classDirectories = files(classDirectories.files.collect {
```
EmmaLRussell: Not sure what's going on here...?

src/app/src/main/kotlin/org/imperial/mrc/hint/APIClient.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.github.kittinunf.fuel.core.Response
+import com.github.kittinunf.fuel.httpPost
+import org.springframework.context.annotation.Configuration
+
+interface APIClient {
+    fun validate(path: String, type: FileType): Response
+}
+
+@Configuration
+open class HintAPIClient(
```
EmmaLRussell: I wonder whether this class should call the .asResponseEntity() method on its responses itself, and return spring ResponseEntities rather than Fuel responses - Fuel is really an internal implementation detail of the HintAPIClient so it might be nicer to keep everything outside it in Spring terms?

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/BaselineController.kt
```diff
@@ -1,34 +1,20 @@
...
-        val pjnzFile = File("${appProperties.uploadDirectory}/$id/pjnz/$fileName")
-
-        if (pjnzFile.parentFile.exists()) {
-            FileUtils.cleanDirectory(pjnzFile.parentFile)
-        } else {
-            FileUtils.forceMkdirParent(pjnzFile)
-        }
-
-        pjnzFile.writeBytes(file.bytes)
+        fileManager.saveFile(file, FileType.PJNZ)
```
EmmaLRussell: Does the pjnz not get validated too?

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/DiseaseController.kt
```diff
@@ -0,0 +1,33 @@
...
+@RestController
+@RequestMapping("/disease")
+class DiseaseController(private val fileManager: FileManager,
+                        private val apiClient: APIClient) {
+
+    @PostMapping("/survey/")
+    @ResponseBody
+    fun uploadSurvey(@RequestParam("file") file: MultipartFile): ResponseEntity<String> {
+
+        val path = fileManager.saveFile(file, FileType.Survey)
```
EmmaLRussell: Could pull these three lines out into a saveAndValidate(file, fileType) method, maybe in a FileValidationController base class for DiseaseController and BaselineController?

src/app/build.gradle
hillalex: I actually just copied this syntax from your exclusion of the SparkWrapper from test coverage in OW. (I didn't want it to complain about missing coverage for the API client, although once integration tests are in we will get coverage)

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/BaselineController.kt
hillalex: Not yet because hintr isn't working, hence you can never progress to step 2. So for dev purposed I'm mocking this out still.

hillalex:
> When I run the dependencies script, i don't seem to end up with a running hintr container, although it does appear to download the image and runs the docker run line without error. It's not in my list of exited containers either! Are you seeing the same behaviour?

Looks like hintr:latest is broken. Have changed to hintr:mrc-418 which is working!

src/app/src/main/kotlin/org/imperial/mrc/hint/APIClient.kt
hillalex: Seems reasonable

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/DiseaseController.kt
hillalex: Cool, have pulled out into private method and can pull out base class for the controllers when hintr is in a more usable state and the baseline methods get wired up properly.

src/app/build.gradle
hillalex: Although actually, not sure it's working! Codecov still seems to be complaining?

EmmaLRussell: Oh yeah, I've a feeling that didn't really work for me either, but I left it as a future exercise to look at.

PR #19, opened 2019-08-21
Mrc 402 Request password reset link - generate token

EmmaLRussell:
Contains changes from https://github.com/mrc-ide/hint/pull/17 which should be merged for a clean diff. 

This branch just contains code (copied from Montagu API) to generate the tokens - storing the tokens and sending the email to be implemented in separate branches.


src/app/src/main/kotlin/org/imperial/mrc/hint/security/OneTimeTokenGenerator.kt
```diff
@@ -0,0 +1,62 @@
...
+import java.time.Instant
+import java.util.*
+
+enum class TokenType
+{
+    ONETIME
+}
+
+@Configuration
+open class OneTimeTokenGenerator(
```
EmmaLRussell: This class combines functionality from montagu api's OneTimeTokenGenerator and WebTokenHelper, as I figured neither will have to be as complex as in montagu

src/app/src/main/kotlin/org/imperial/mrc/hint/security/OneTimeTokenManager.kt
```diff
@@ -0,0 +1,63 @@
...
+import java.time.Instant
+import java.util.*
+
+enum class TokenType
+{
+    ONETIME
+}
+
+@Configuration
+open class OneTimeTokenManager(
```
EmmaLRussell: This class combines functionality from montagu api's OneTimeTokenGenerator and WebTokenHelper, as I figured neither will have to be as complex as in montagu

src/app/src/main/kotlin/org/imperial/mrc/hint/AppProperties.kt
```diff
@@ -12,10 +12,10 @@ interface AppProperties {
 }
 
 @Configuration
-open class ConfiguredAppProperties: AppProperties {
+open class ConfiguredAppProperties(props: Properties = properties): AppProperties {
```
EmmaLRussell: This is kind of an ugly way to make this testable by allowing props other than those immediately read from disk by the companion object to be used.

src/app/src/main/kotlin/org/imperial/mrc/hint/security/tokens/GZipHelpers.kt
```diff
@@ -0,0 +1,73 @@
...
+import java.io.ByteArrayOutputStream
+import java.io.InputStreamReader
+import java.util.*
+import java.util.zip.GZIPInputStream
+import java.util.zip.GZIPOutputStream
+
+fun String.inflated() = inflate(this)
+fun String.deflated() = deflate(this)
+
+private fun deflate(uncompressed: String): String
```
hillalex: I don't think we need any of this deflating/inflating business. This was added in Montagu because test users who had access to all modelling groups ended up with tokens that exceeded maximum HTTP request header sizes. We aren't encoding roles/permissions here so tokens will be reasonably short.

src/app/src/main/kotlin/org/imperial/mrc/hint/security/tokens/OneTimeTokenAuthenticator.kt
```diff
@@ -0,0 +1,53 @@
...
+    private fun checkTokenAgainstRepository(credentials: TokenCredentials)
+    {
+        val compressedToken = credentials.token
+        if (!oneTimeTokenChecker.checkToken(compressedToken.inflated()))
+        {
+            throw CredentialsException("Token has already been used (or never existed)")
+        }
+    }
+
+    private fun handleUrlAttribute(credentials: TokenCredentials, jwt: JWT)
```
hillalex: We don't need this url check at all. This was added in Montagu to give onetime access to arbitrary endpoints, which we aren't doing here.

src/app/src/main/kotlin/org/imperial/mrc/hint/security/tokens/OneTimeTokenAuthenticator.kt
```diff
@@ -0,0 +1,53 @@
...
+    {
+        super.createJwtProfile(credentials, jwt)
+        val claims = jwt.jwtClaimsSet
+        val issuer = claims.issuer
+        if (issuer != tokenIssuer)
+        {
+            throw CredentialsException("Token was issued by '$issuer'. Must be issued by '${tokenIssuer}'")
+        }
+        val tokenType = claims.getClaim("token_type").toString()
+        if (tokenType != TokenType.ONETIME.toString())
```
hillalex: Will we be issuing any tokens other than these? If not we don't need to specify a token type

src/app/src/main/kotlin/org/imperial/mrc/hint/AppProperties.kt
hillalex: I can't think of a better way either and you're right it's good to test.

EmmaLRussell:
Ugh, code coverage is now moaning, though I can't see how it can be any different to before... :(

hillalex:
> Ugh, code coverage is now moaning, though I can't see how it can be any different to before... :(

Looks like it's complaining about the KeyHelper methods that load a key pair from disk, but it'd be a not worth it faff to test that.

PR #23, opened 2019-08-23
Mrc 412 Reset password

EmmaLRussell:


src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
```diff
@@ -24,4 +27,16 @@ class HintExceptionHandler : ResponseEntityExceptionHandler() {
...
         return ErrorDetail(status, e.message ?: "Something went wrong").toResponseEntity()
     }
+
+    @ExceptionHandler(HintException::class)
+    protected fun handleHintException(e: HintException): ResponseEntity<Any>
+    {
+        return ErrorDetail(e.httpStatus, e.message!!).toResponseEntity()
+    }
+
+    @ExceptionHandler(ConstraintViolationException::class)
```
EmmaLRussell: This is needed to handle the exception thrown when the controller method parameter validation fails

src/app/src/main/kotlin/org/imperial/mrc/hint/security/tokens/OneTimeTokenAuthenticator.kt
```diff
@@ -1,18 +1,23 @@
...
 
 import com.nimbusds.jwt.JWT
+import org.imperial.mrc.hint.AppProperties
 import org.pac4j.core.credentials.TokenCredentials
 import org.pac4j.core.exception.CredentialsException
 import org.pac4j.jwt.config.signature.SignatureConfiguration
 import org.pac4j.jwt.credentials.authenticator.JwtAuthenticator
+import org.springframework.context.annotation.Configuration
 
+@Configuration
```
EmmaLRussell: I rearranged the constructors of this and related classes to make better use of Spring IoC for testability

src/app/src/main/kotlin/org/imperial/mrc/hint/models/Response.kt
```diff
@@ -2,14 +2,16 @@ package org.imperial.mrc.hint.models
...
-data class SuccessResponse(val data: Any){
+data class SuccessResponse(val data: Any?){
     val errors = mapOf<Any, Any>()
     val status = "success"
 }
 
 
 fun SuccessResponse.toJsonString() = ObjectMapper().writeValueAsString(this)
 
+val EmptySuccessResponse = SuccessResponse(null)
```
hillalex: Null won't work so well for the front-end. Can we use `true` or the string "OK" instead?

src/app/src/main/kotlin/org/imperial/mrc/hint/security/tokens/SignatureConfig.kt
```diff
@@ -0,0 +1,18 @@
+package org.imperial.mrc.hint.security.tokens
+
+import org.pac4j.jwt.config.signature.SignatureConfiguration
+import org.pac4j.jwt.config.signature.RSASignatureConfiguration
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import java.security.KeyPair
+
+@Configuration
+class SignatureConfig
```
hillalex: This is a bit confusing - a class called `SignatureConfig` whose only method returns a class of type `SignatureConfiguration` :confused: 

In general I'm a bit unsure about this pattern - we seem to have a proliferation of classes whose only role is to expose a single Bean (https://github.com/mrc-ide/hint/blob/master/src/app/src/main/kotlin/org/imperial/mrc/hint/security/Pac4jConfig.kt#L20, https://github.com/mrc-ide/hint/blob/master/src/app/src/main/kotlin/org/imperial/mrc/hint/security/Pac4jProfileService.kt)

All these examples are security related, so one option would be to have a single SecurityConfig class that defines multiple Beans. But I notice that here you've defined a Bean without a wrapping class at all - https://github.com/mrc-ide/hint/pull/23/files#diff-c044958f508e9341914a1a4f2cbb9e84R16 That seems like a nicer pattern to me, we may as well take advantage of Kotlin not requiring functions to be class methods. What do you think?

src/app/src/main/kotlin/org/imperial/mrc/hint/db/JooqContext.kt
```diff
@@ -4,10 +4,23 @@ import java.sql.Connection
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
 import org.springframework.jdbc.datasource.TransactionAwareDataSourceProxy
+import org.springframework.web.context.annotation.RequestScope
 import javax.sql.DataSource
 
-class JooqContext(private val dataSource: DataSource) : AutoCloseable
+@Bean
+fun getDSL(@Autowired context: JooqContext): DSLContext
```
hillalex: This feels like a neat solution to get a single context per request!

src/app/static/src/app/store/password/actions.ts
```diff
@@ -2,27 +2,46 @@ import {ActionContext, ActionTree, Payload} from "vuex";
...
-                commit<PasswordActionPayload<null>>({type: "ResetLinkRequested", payload});
+                commit<PasswordActionPayload<null>>({type: "ResetLinkRequested", payload: null});
             })
             .catch((error: Error) => {
                 commit<PasswordActionPayload<string>>({type: "RequestResetLinkError", payload: error.message});
             });
+    },
+
+    resetPassword({commit}: ActionContext<PasswordState, PasswordState>, payload: ResetPasswordActionParams) {
+        let formData = new FormData();
```
hillalex: Can we avoid using `FormData` unless we absolutely have to? Because it's a browser API it's a massive pain to test. If we're just passing simple data it's easiest just to stringify it, as I've done here: https://github.com/mrc-ide/hint/pull/29/files#diff-0f73780710f9ddc4b1ffe4acd2956a6dR22

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/CleanDatabaseTests.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.test.context.ActiveProfiles
+import org.springframework.test.context.junit.jupiter.SpringExtension
+import java.io.File
+import javax.sql.DataSource
+
+@ActiveProfiles(profiles = ["dev"])
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ExtendWith(SpringExtension::class)
```
hillalex: What does this `SpringExtension` do?

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/PasswordTests.kt
```diff
@@ -41,6 +36,76 @@ class PasswordTests(@Autowired val restTemplate: TestRestTemplate): IntegrationT
...
+        val headers = HttpHeaders()
+        headers.contentType = MediaType.APPLICATION_FORM_URLENCODED
+
+        var entity = restTemplate.postForEntity<String>("/password/reset-password/",
+                HttpEntity(map, headers))
+        Assertions.assertThat(entity.statusCode).isEqualTo(HttpStatus.OK)
+
+        restTemplate.restTemplate.interceptors.add(AuthInterceptor(restTemplate, "newpassword"))
+        entity = restTemplate.getForEntity<String>("/")
+        Assertions.assertThat(entity.statusCode).isEqualTo(HttpStatus.OK)
```
hillalex: Can we also check that response bodies conform to the schema here? I know until https://vimc.myjetbrains.com/youtrack/issue/mrc-434 is done we can't do this programatically using the JSONValidator but just checking manually as per https://github.com/mrc-ide/hint/pull/28/files#diff-02791b2bdd6c77604f5d206a166f297bR63

src/app/src/main/kotlin/org/imperial/mrc/hint/security/tokens/SignatureConfig.kt
EmmaLRussell: It was originally called `SignatureConfigurationConfig` which was even worse! 

OK, that sounds good, I'll herd all those security beans into a single SecurtyConfig file, but have them as beans outside of classes.

src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
```diff
@@ -1,17 +1,45 @@
...
 import org.springframework.context.annotation.Configuration
+import org.springframework.jdbc.datasource.TransactionAwareDataSourceProxy
 import org.springframework.web.servlet.config.annotation.InterceptorRegistry
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer
+import javax.sql.DataSource
 
 @Configuration
-class SecurityConfig(val config: Config): WebMvcConfigurer {
+@ComponentScan(basePackages = ["org.pac4j.springframework.web", "org.pac4j.springframework.component"])
+class Pac4jConfig {
```
EmmaLRussell: So much for Beans outside classes - this has to be a class because it needs ComponentScan...

src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
```diff
@@ -1,17 +1,45 @@
...
+
+    @Bean
```
EmmaLRussell: ..this just didn't work at all as a function outside a class, couldn't be found, no idea why not, when the dsl one works...

src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
```diff
@@ -1,17 +1,45 @@
...
+        val formClient = FormClient("/login", profileService)
+        val clients = Clients("/callback", formClient)
+        return Config(clients).apply {
+            sessionStore = J2ESessionStore()
+        }
+    }
+}
+
+@Configuration
+class HintWebMvcConfig(val config: Config): WebMvcConfigurer {
```
EmmaLRussell: ..and this needs to be a real class as it does an override!

src/app/static/src/app/store/password/actions.ts
EmmaLRussell: Fair enough for the reset link request, but I wouldn't have thought that was ideal for passwords? Gets encrypted over https, but hangs around in plain text on client (browser history) and server logs.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/CleanDatabaseTests.kt
EmmaLRussell: Gives you access Autowiring etc, but I guess you don't need it if you're a SpringBootTest anyway..

src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
hillalex: As per our convo, figured out that the dsl bean is actually not working either, rather spring is providing it's own default bean.

PR #25, opened 2019-08-28
mrc-430 use kotlin-spring plugin

hillalex:
And remove all "open" modifiers


src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/HomeController.kt
```diff
@@ -17,56 +17,4 @@ class HomeController {
         model["title"] = "HINT"
         return "index"
     }
-
```
hillalex: Also just opportunistically removing these old methods from the first prototype

PR #26, opened 2019-08-28
mrc-435 security

hillalex:
Contains commits from https://github.com/mrc-ide/hint/pull/25

This PR:
* Adds integration test helpers for testing that endpoints are secure. For this I've used an interceptor for the RestTemplate which logs the user in and appends the session cookie to the request and junit5 parameterization to run tests twice, one with the interceptor enabled and once without, to test secure and unsecure requests.
* Fixes the security config to secure all endpoints except for those that should be freely accessible (i.e. login and forgotten password related ones)


src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
```diff
@@ -7,9 +7,11 @@ import org.springframework.web.servlet.config.annotation.InterceptorRegistry
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer
 
 @Configuration
-open class SecurityConfig(val config: Config): WebMvcConfigurer {
+class SecurityConfig(val config: Config): WebMvcConfigurer {
+
     override fun addInterceptors(registry: InterceptorRegistry) {
         registry.addInterceptor(SecurityInterceptor(config, "FormClient"))
-                .addPathPatterns("/")
```
hillalex: This only matches the exact path "/". Have added the general pattern "/**" with specific exclusions.

src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
```diff
@@ -7,9 +7,11 @@ import org.springframework.web.servlet.config.annotation.InterceptorRegistry
...
+                .addPathPatterns("/**")
+                .excludePathPatterns("/login", "/login/", "/password/**", "/callback", "/callback/")
```
hillalex: For the life of me could not get automatic trailing slash matching to work so gave up and specified with and without trailing slashes explicitly.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/IntegrationTests.kt
```diff
@@ -1,13 +1,97 @@
...
+}
+
+abstract class SecureIntegrationTests : IntegrationTests() {
+
+    @Autowired
+    lateinit var testRestTemplate: TestRestTemplate
+
+    @BeforeEach
+    fun beforeEach(info: TestInfo) {
+        val isAuthorized = info.displayName.contains("TRUE")
```
hillalex: This is the only way to get at a test's parameter values here - they are contained in the display name.

hillalex:
Looks like the tests that integrate with hintr have broken, see my comment here: https://dideimperial.slack.com/archives/GL5N84ZFU/p1567011635000900

I could ignore the tests or pin hintr to a previous version but think it's quite useful that we're picking up issue like this?

EmmaLRussell:
> Looks like the tests that integrate with hintr have broken, see my comment here: https://dideimperial.slack.com/archives/GL5N84ZFU/p1567011635000900
> 
> I could ignore the tests or pin hintr to a previous version but think it's quite useful that we're picking up issue like this?

Pinning to a previous version with a comment about why it's been done might be best, since at least we won't break out master build when we merge and it should be fairly difficult for us to forget to update it when hintr gets fixed?

Definitely good that this was picked up, hooray for tests!

hillalex:
> > Looks like the tests that integrate with hintr have broken, see my comment here: https://dideimperial.slack.com/archives/GL5N84ZFU/p1567011635000900
> > I could ignore the tests or pin hintr to a previous version but think it's quite useful that we're picking up issue like this?
> 
> Pinning to a previous version with a comment about why it's been done might be best, since at least we won't break out master build when we merge and it should be fairly difficult for us to forget to update it when hintr gets fixed?
> 
> Definitely good that this was picked up, hooray for tests!

Master is broken already for the same reason but I've just pinned master to Rich's new branch, and will do the same here...

PR #28, opened 2019-08-29
refactor api service and make use of async/await

hillalex:
This is a preliminary PR to adding integration tests. The code was getting littered with .then and .catch statements, so I've refactored the apiService to allow for some generic error handling and to make use of async/await rather than lots of promise chains.

See what you think about the new syntax for actions, e.g. here: https://github.com/mrc-ide/hint/pull/28/files#diff-879aab0773ac0cf78938796b522e20c7R22


src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/PasswordController.kt
```diff
@@ -36,9 +38,9 @@ class PasswordController(private val userRepository: UserRepository,
 
             emailManager.sendEmail(emailMessage, email)
 
-            return token
+            return SuccessResponse(token).toJsonString()
         }
 
-        return ""
+        return SuccessResponse(true).toJsonString()
```
hillalex: I know you're working on this code now so this is liable to change, just updated these for now so they conform to the response schema.

src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -19,24 +19,23 @@ export interface BaselineActions {
...
-        api.postAndReturn<PjnzResponse>("/baseline/pjnz/", formData)
-            .then((payload) => {
-                commit<BaselinePayload<PjnzResponse>>({type: "PJNZUploaded", payload});
-            })
-            .catch((error: Error) => {
-                commit<BaselinePayload<String>>({type: 'PJNZUploadError', payload: error.message});
-            });
+
+        const payload = await api()
+            .commitFirstErrorAsType(commit, "PJNZUploadError")
```
hillalex: Just occurred to me that we could extend this fluent interface further so we have something like:
```
await api(commit).commitFirstErrorAs("PJNZUploadError")
                .commitSuccessAs<PjnzResponse>("PJNZUploaded")
                .post("/baseline/pjnz/", formData)
```

EmmaLRussell: I think this would be good, to keep this all in one statement, it feels a little disjointed to have it in two. This would be really terse and clean. 

What I do prefer about the old way of doing things is the ordering:
1. what to do
2. what to do if 1. succeeds
3. what to do if 1. fails

Having the primary action come last in the statement feels a little bit backwards, and less readable. But I guess I could get used to it! Not sure how you could achieve it the other way round anyway, you'd need the initial 'post' to not actually do the post, but wait for the other commit setups. 
```
api(commit)
  .post(..)
  .commitSuccessAs(..)
  .commitFirstErrorAs(..)
  .go() 
``` 
would hardly be better!

src/app/static/src/app/apiService.ts
```diff
@@ -4,44 +4,95 @@ import {
...
-                const payload = response && response.data;
-                return payload.data as T
-            })
-            .catch((e: AxiosError<Failure>) => {
-                return this.handleError(e)
-            })
-    }
+export interface API {
+
+    commitFirstErrorAsType: (commit: Commit, type: string) => API
```
EmmaLRussell: If the aim is terseness, maybe we should drop the 'First' from this method name - I think we can take it as read that you'll get the first error only, and this method is going to be invoked lots of times.

src/app/static/src/app/apiService.ts
```diff
@@ -4,44 +4,95 @@ import {
...
+    };
+
+    private _ignoreErrors: Boolean = false;
+
+    static getFirstErrorFromFailure = (failure: Failure) => {
+        const firstError = failure.errors[0];
+        return firstError.detail ? firstError.detail : firstError.error;
+    };
+
+    private _notifyOnError: ((failure: Failure) => void) = (failure: Failure) => {
```
EmmaLRussell: We don't ever really want to do this, do we? Could we force the consumers of APIService to either choose to ignore errors or to commit them as mutations? Maybe it's an error if they haven't specified one or the other when they make the get or post? This contradicts my other comment about the method call ordering though!

hillalex: I've gone with a console.warn if no error handler is attached - https://github.com/mrc-ide/hint/pull/28/files#diff-0f0781cfcf5e78654409fce90c2c466fR100

src/app/static/src/app/apiService.ts
```diff
@@ -1,47 +1,122 @@
...
+        return this;
+    };
+
+    private _handleAxiosResponse(promise: Promise<AxiosResponse>) {
+        return promise.then((response: AxiosResponse) => {
+            const success = response && response.data;
+            if (this._onSuccess) {
+                this._onSuccess(success);
+            }
+            return success;
```
hillalex: I've left this still returning the success object here, in case for any reason a consumer wants to handle the response themselves

PR #29, opened 2019-08-29
mrc-433 add vue to hint integration tests

hillalex:
This PR:

* Adds a flag to config.properties to optionally turn auth off
* Adds script in /static  to run all dependencies including a dockerised hint app itself
* Adds integration tests for all vuex actions
* Moves the construction of the FormData object for POSTing files into the FileUpload component as suggested, to allow for testing of the actions with the node "form-data" package.

Should note that because auth is turned off for the integration tests there is no automated test that auth is working as expected (i.e. that axios is passing credentials back to the server as expected) - to do that you need to run up the app and confirm it works. But I'm not too bothered about this because every request should go via the ApiService class, so it's not like there are lots of potential failure modes, if it's working at all it must be working for all requests.


src/app/static/scripts/run-dependencies-for-integration-tests.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+set -ex
+
+HERE=$(readlink -f "$(dirname $0)")
+NETWORK=hint_nw
+HINT=hint
+$HERE/../../../../scripts/run-dependencies.sh
+HINT_VERSION=mrc-433_integrationtests
```
hillalex: This should be changed to master once this branch is merged. I think we want to pin the version like this (either hard-coded here or we could do it via a config file as we do for hintr) rather than get the current branch because otherwise we end up with a bit of an awkward dependency on Travis - the front-end build won't pass unless the back-end build has.

src/app/static/src/tests/integration/surveyAndProgram.itest.ts
```diff
@@ -0,0 +1,32 @@
...
+
+        const file = fs.createReadStream("fakesurvey");
+        const formData = new FormData();
+        formData.append('file', file);
+
+        await actions.uploadSurvey({commit} as any, formData);
+
+        expect(commit.mock.calls[0][0]).toStrictEqual({
+            type: "SurveyError",
+            payload: "could not find function \"validate_func\""
```
EmmaLRussell: Is this the current error coming out of hintr? we expect this to turn into something more meaningful when hintr is further along?

hillalex: This is the current error from hintr :/

src/app/static/README.md
```diff
@@ -6,12 +6,18 @@ Type definitions are auto-generated based on the
...
 
-### Test
-Run `npm test` to run tests locally using jest
+### Testing
+Tests are run with jest. Files with the suffix `.test.ts` are treated as unit tests, files 
+with the suffix `.itest.ts` treated as integration tests. Config for each can be found in 
+`jest.config.js` and `jest.integration.config.js`, respectively.
+- run unit tests with `npm test` 
+- run integration tests with `npm run integration-test` having first started the app and 
+all dependencies by running `./scripts/run-dependencies-for-integration-tests.sh` from this
```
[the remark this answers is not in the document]
hillalex: I think you more likely want to keep the dependencies running while you iterate on test failures rather than tearing down and re-running each time? But should probably have a `run-dependencies-dev-mode` with a nice ctrl-c exit. Will add that.

PR #36, opened 2019-09-02
Mrc 428 Error page if login attempt with blank username or password

EmmaLRussell:
The form now requires both inputs to have a value, but I also put in a custom DbProfileservice class to deal with this situation as a BadCredentials exception rather than a TechnicalException (which puts pac4j into an error state)

### DIFF
--- a/src/app/templates/login.ftl
+++ b/src/app/templates/login.ftl
@@ -21,12 +21,12 @@
                 <div class="form-group">
                     <label for="user-id">Username</label>
                     <input type="text" size="20" class="form-control" name="username" id="user-id" value="${username}" required>
-                    <div class="invalid-feedback">Please enter your username.</div>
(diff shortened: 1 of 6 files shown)


src/app/templates/login.ftl
```diff
@@ -4,18 +4,29 @@
     <title>${title}</title>
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
     <link rel="stylesheet" href="/public/css/style.css"/>
+    <script>
```
EmmaLRussell: Unfortunately this bit of javascript isn't tested, but I didn't think it was worth converting into a vue component just for that.. What do you think?

hillalex: Totally agree, this seems fine to me!

PR #37, opened 2019-09-03
add shape file endpoints

hillalex:
This adds 2 endpoints, one to POST a new shape file and validate it, one to GET an existing file from disk and return the validation result (on page load.) Also changes the existing endpoint `GET /baseline/` to be  `GET /baseline/pjnz/`, and updates the front-end correspondingly.

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
@@ -9,10 +9,11 @@ import java.io.File
 
 enum class FileType {
 
-    Survey,
+    ANC,
     Program,
     PJNZ,
-    ANC;
+    Shape,
+    Survey;
 
     override fun toString(): String {
         return this.name.toLowerCase()
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/BaselineController.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/BaselineController.kt
@@ -33,22 +33,29 @@ class BaselineController(private val fileManager: FileManager,
         return saveAndValidate(file, FileType.Shape)
     }
 
-    @GetMapping("/")
+    @GetMapping("/pjnz/")
     @ResponseBody
-    fun get(): String {
+    fun getPJNZ(): String {
 
         // TODO request serialised data for this id from the R API
         // for now just read basic file info from upload dir
         val file = fileManager.getFile(FileType.PJNZ)
 
-        val data = if (file != null) {
+        val pjnzResponse = if (file != null) {
             val fileName = file.name
             val countryName = fileName.split("_").first()
-            mapOf("pjnz" to buildPjnzResponse(fileName, countryName))
+            buildPjnzResponse(fileName, countryName)
         } else {
-            mapOf("pjnz" to null)
+            null
         }
-        return SuccessResponse(data).toJsonString()
+
+        return SuccessResponse(pjnzResponse).toJsonString()
+    }
+
+    @GetMapping("/shape/")
+    @ResponseBody
+    fun getShape(): ResponseEntity<String> {
+       return getIfExists(FileType.Shape)
     }
 
     private fun buildPjnzResponse(fileName: String, countryName: String): Map<String, Any> {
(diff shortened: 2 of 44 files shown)


src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/BaselineController.kt
```diff
@@ -24,22 +27,35 @@ class BaselineController(private val fileManager: FileManager) {
...
             val countryName = fileName.split("_").first()
-            mapOf("pjnz" to buildPjnzResponse(fileName, countryName))
+            buildPjnzResponse(fileName, countryName)
         } else {
-            mapOf("pjnz" to null)
+            null
         }
-        return SuccessResponse(data).toJsonString()
+
+        return SuccessResponse(pjnzResponse).toJsonString()
```
hillalex: Originally I was thinking we'd return all baseline data in one endpoint but that doesn't really make sense, so this is now just for pjnz, separate endpoint for shape below

src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
```diff
@@ -9,10 +9,11 @@ import java.io.File
 
 enum class FileType {
 
-    Survey,
+    ANC,
```
hillalex: moved these to being in alphabetical order

src/app/src/main/kotlin/org/imperial/mrc/hint/models/Response.kt
```diff
@@ -1,18 +1,20 @@
 package org.imperial.mrc.hint.models
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
 
 data class SuccessResponse(val data: Any?){
-    val errors = mapOf<Any, Any>()
+    val errors = listOf<ErrorDetail>()
```
hillalex: corresponds to the update schema here: https://github.com/mrc-ide/hintr/pull/23/files#diff-b13e27c6bcde278d556bb9d21fd628a0

src/app/static/scripts/run-dependencies-for-integration-tests.sh
```diff
@@ -5,9 +5,10 @@ HERE=$(readlink -f "$(dirname $0)")
 NETWORK=hint_nw
 HINT=hint
 $HERE/../../../../scripts/run-dependencies.sh
-HINT_VERSION=mrc-433_integrationtests
+HINT_VERSION=mrc-417_backend
```
hillalex: this is kind of annoyingly circular again, not sure how to handle this properly

src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -25,9 +24,7 @@ export const actions: ActionTree<BaselineState, RootState> & BaselineActions = {
     async getBaselineData({commit}) {
         await api<BaselineActionTypes, BaselineErrorActionTypes>(commit)
             .ignoreErrors()
-            .withSuccess("BaselineDataLoaded")
-            .get<BaselineData>("/baseline/");
+            .withSuccess("PJNZLoaded")
+            .get<PjnzResponse>("/baseline/pjnz/");
```
hillalex: Updated to return only the pjnz response. In the next PR which adds the shape file stuff to the front end will be getting the shape file here too.

src/app/static/scripts/run-dependencies-for-integration-tests.sh
EmmaLRussell: Should the hint version always be the current branch? No, because that only works if the image exists, and using master is fine, if there's only front end changes I guess. Hmm... It could push the current branch?

src/config/hintr_version
```diff
@@ -1 +1,2 @@
-master
+mrc-434
```
EmmaLRussell: Looks like this is merged in now

src/app/static/scripts/run-dependencies-for-integration-tests.sh
hillalex: We can build it and not push it :+1:

PR #38, opened 2019-09-03
mrc-443 Added uploadANC front end

EmmaLRussell:


src/app/static/src/app/main.ts
```diff
@@ -14,7 +14,7 @@ export interface RootState {
 }
 
 export interface StepState {
-    complete: boolean
+    complete: () => boolean
```
EmmaLRussell: I wondered if this would work better as a function, since it's likely to be a value derived from other properties in the state.

hillalex: Will this work? I don't think the function will be automatically executed when the underlying data changes. This could/should be tested by instantiating the component with incomplete data, then triggering the action/mutation which makes the stage complete, then checking whether the continue button is enabled.

It looks like vue does have a concept of "getters" aka computed properties for the store which I think are what you're after here: https://vuex.vuejs.org/guide/getters.html

EmmaLRussell:
I've tried using getters, which turned out to be slightly painful, primarily because I couldn't get mapGetters to work (compile kept failing as it wouldn't recognise the mapped getters as computed properties). Updated Babel to no avail. Possible that this is a bug and related to the fact that we're using modules. There's an open PR which may be relevant. https://github.com/vuejs/vuex/pull/1121
Would require declaring further interfaces. Anyway, I worked around it, but not sure how much it adds really!

src/app/static/src/tests/baseline/mutations.test.ts
```diff
@@ -16,11 +20,22 @@ describe("Baseline mutations", () => {
...
+            baseline: testState,
+            surveyAndProgram: {...initialSurveyAndProgramDataState}
+        };
+        const complete = testStore.getters!!.complete;
 
-        const testState = {...initialBaselineState};
         mutations.PJNZUploaded(testState, {payload: mockPJNZResponse({data: {country: "Malawi"}}), type: "PJNZLoaded"});
-        expect(testState.complete()).toBe(true);
+
+        expect(complete(testState, null, testRootState, null)).toBe(true);
```
EmmaLRussell: Getters really needs four parameters, but ours only use the first one!

src/app/src/test/kotlin/org/imperial/mrc/hint/helpers/JSONValidator.kt
```diff
@@ -72,7 +75,9 @@ class JSONValidator {
             "$name.schema.json"
         }
         val url = URL("https://raw.githubusercontent.com/mrc-ide/hintr/$hintrVersion/inst/schema/$path")
-        return url.openStream().use {
+
+        val conn = url.openConnection() as HttpURLConnection
+        return BufferedReader(InputStreamReader(conn.getInputStream())).use {
```
EmmaLRussell: Holy hell, this was annoying. This randomly started throwing errors when running locally, when attempting to validate the blameless ValidateInputResponse schema. Eventually worked out it appeared to always fail at the same character index, so presumably it wasn't managing to read the whole file over a certain length). Changing from url.openStream to url.openConnection seemed to fix this...

EmmaLRussell:
I regenerated the types from HINTR, as latest master is now returning different response to before (errors as array rather than object).

src/app/static/src/tests/components/stepper.test.ts
```diff
@@ -111,4 +114,23 @@ describe("Stepper component", () => {
         expect(steps.at(1).props().active).toBe(true);
     });
 
+    it("updates from completed state when active step data is populated", (done) => {
```
hillalex: nice :raised_hands:

PR #39, opened 2019-09-03
mrc-453 upload shape file front-end

hillalex:
Adds an upload widget tor the shape file to the front-end. 

Should be merged after https://github.com/mrc-ide/hint/pull/37 and https://github.com/mrc-ide/hint/pull/42


src/app/static/src/app/components/baseline/Baseline.vue
```diff
@@ -6,24 +6,28 @@
...
+                                 :error="shapeError"
+                                 :upload="uploadShape"
+                                 :existingFileName="shapeFileName"
+                                 accept="geojson,.geojson,GEOJSON,.GEOJSON"
+                                 name="shape">
                     </file-upload>
                 </form>
             </div>
         </div>
-        <div class="row">
```
hillalex: I moved this to appear next to the red tick after the PJNZ file is uploaded

src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -21,10 +22,22 @@ export const actions: ActionTree<BaselineState, RootState> & BaselineActions = {
...
     },
 
+    async uploadShape({commit}, formData) {
+        await api<BaselineActionTypes, BaselineErrorActionTypes>(commit)
+            .withSuccess("ShapeUploaded")
+            .withError("ShapeUploadError")
+            .postAndReturn<PjnzResponse>("/baseline/shape/", formData);
+    },
+
     async getBaselineData({commit}) {
```
EmmaLRussell: Could fire these off simultaneously rather than waiting for pjnz to complete before requesting shape?

hillalex: Great point

hillalex:
Updated the completion status and also mounted the upload directory as a volume so that you can really upload files and proceed to the next step. 

But this now contains commits from https://github.com/mrc-ide/hint/pull/42 so that one should be merged first!

PR #42, opened 2019-09-04
mount volume and use testdata for integration tests

hillalex:
* adds publically available testdata (the same files used in hintr, but just copied here for convenience)
* mounts upload volume when running dependencies
* update integration tests that were previously returning 400s, now 200s


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/DiseaseTests.kt
```diff
@@ -11,24 +11,24 @@ class DiseaseTests: SecureIntegrationTests() {
     @ParameterizedTest
     @EnumSource(IsAuthorized::class)
     fun `can upload survey file`(isAuthorized: IsAuthorized) {
-        val postEntity = createTestHttpEntity()
+        val postEntity = getTestEntity("survey.csv")
         val entity = testRestTemplate.postForEntity<String>("/disease/survey/", postEntity)
         assertSecureWithError(isAuthorized, entity, HttpStatus.BAD_REQUEST, "INVALID_FILE")
```
EmmaLRussell: The API is returning errors for valid programme and survey files?

hillalex: Survey data validation only just got implemented in hintr. Has now been merged, so have updated the test. I was assuming program validation hadn't been implemented yet but actually it has so this was a bug, good catch  - passing "program" instead of "programme". 

Added extra checking in the tests to make sure the correct types of data are being returned also:
https://github.com/mrc-ide/hint/pull/42/commits/77f69f703b0c0da7e29c256830d1c506a2a91e9c#diff-e84e9bec8188142e523c7b45b1bef5c0R19

Of course now the front-end integration tests started failing, but looks like hintr is now returning a different error message for non-existent files (not a nice human readable one, an R error) so I've also made these use the real files.

PR #43, opened 2019-09-05
Mrc 452 Filtered data options and selected data type in store

EmmaLRussell:
This pull request regenerates types for latest API, adds a new store component called filteredData, which has a selectedDataType property, which is currently just set on successful data upload (will also be set by the tabs component). 
Also added a getter to filteredData, to return the appropriate filter options, depending on the selected data type.


src/app/static/src/tests/integration/surveyAndProgram.itest.ts
```diff
@@ -25,8 +26,11 @@ describe("Survey and program actions", () => {
 
         expect(commit.mock.calls[0][0]).toStrictEqual({
             type: "SurveyError",
-            payload: "could not find function \"validate_func\""
+            payload: "File does not exist. Create it, or fix the path."
         });
+
+        //expectSelectedDataTypeUpdate(commit.mock.calls[1], DataType.Survey);
```
EmmaLRussell: These can be reinstated once we start getting 200s - once merged with the volume mount?

src/app/static/src/app/types.ts
```diff
@@ -1,25 +1,9 @@
 import {Payload} from "vuex";
+import {ProgrammeResponse, SurveyResponse} from "./generated";
 
 export interface PayloadWithType<T> extends Payload {
     payload: T
 }
 
-export type InternalResponse = SurveyResponse | ProgramResponse
+export type InternalResponse = SurveyResponse | ProgrammeResponse
```
hillalex: we don't need this type anymore if both these exist in the generated types now!

src/app/static/src/tests/surveyAndProgram/mutations.test.ts
```diff
@@ -6,21 +6,22 @@ import {mutations} from "../../app/store/surveyAndProgram/mutations";
...
+            data: [],
             filename: "somefile.csv"
         })
     };
 
     it("sets survey data and filename on SurveyLoaded", () => {
         const testState = {...initialSurveyAndProgramDataState};
         mutations.SurveyLoaded(testState, testPayload);
-        expect(testState.survey!!.data).toBe("SOME DATA");
+        expect(testState.survey!!.data.length).toBe(0);
```
hillalex: Feel like this check is kind of pointless, since many things could satisfy it. Can we leave these tests as they were and just use `data: "SOME DATA" as any` here: https://github.com/mrc-ide/hint/pull/43/files#diff-463a72978089f680367017681d7a669bL15

PR #44, opened 2019-09-06
Mrc 448 Filters component

EmmaLRussell:
Currently a partial implementation which only includes the sex filters in a vue-treeselect.

It should show filters appropriate to the current selectedDataType. We haven't yet implemented the tabs to change selected data type, but we do automatically select data type when a survey, programme or ANC file is uploaded.

### DIFF
--- a/src/app/static/package-lock.json
+++ b/src/app/static/package-lock.json
@@ -904,6 +904,15 @@
         "source-map-support": "^0.5.9"
       }
     },
+    "@babel/runtime": {
+      "version": "7.5.5",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.5.5.tgz",
+      "integrity": "sha512-28QvEGyQyNkB0/m2B4FU7IEZGK2NUrcMtT6BZEFALTguLk+AUT6ofsHtPk5QyjAdUkpMJ+/Em+quwz4HOt30AQ==",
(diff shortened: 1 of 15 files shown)


EmmaLRussell:
> Noticed when running this up that there is a small bug.
> 
> To reproduce from the "Upload survey and program data" section
> 
>     1. Upload a file, making sex filter appear
> 
>     2. Select a filter e.g. female
> 
>     3. Click the x in the right hand side of the field (not the x on the token itself)
> 
>     4. Add a different value e.g. male
> 
>     5. Both male & female will appear as tokens
> 
> 
> I'm guessing that maybe the x on the field itself is not updating the data in the background and only updating the UI?
> 
> Also as we talked about perhaps changing storage of selected filters from an array to a set will mean that we can remove some of the logic dealing with not adding duplicates.
> 
> This is the first time I've run up the app, it is looking great. Very exciting to see something!

Ooh, good spot! You're right, it's not handling the event where all values are removed with an action.

EmmaLRussell:
So I found that there was a sneaky vuex special event on the treeselect component, '@input' which gets triggered with the whole current list whenever anything changes in the component (select, deselect or delete all), fixing the bug. This also meant that the array logic went away, and so I decided it wasn't worth converting to Sets and back, since the treeselect uses arrays, and we can just pass those around wholesale.

PR #46, opened 2019-09-10
Mrc 448 allfilters

EmmaLRussell:
Adding the age, survey and and region filters. Some added complexity because the region filters are potentially nested, so some recursion required.

### DIFF
--- a/src/app/static/src/app/components/Filters.vue
+++ b/src/app/static/src/app/components/Filters.vue
@@ -26,6 +26,15 @@
                                 :normalizer = "treeselectNormalizer"
                                 @input="updateSurveyFilter"></treeselect>
                 </div>
+                <div class="col">
+                    <label>Region</label>
+                    <treeselect id="survey-filters" :multiple="true"
+                                :options="regionFilters.available"
+                                :value="regionFilters.selected"
+                                :normalizer = "treeselectNormalizer"
+                                :flat="true"
+                                @input="updateRegionFilter"></treeselect>
+                </div>
             </div>
         </div>
     </div>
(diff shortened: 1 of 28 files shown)


EmmaLRussell:
The code coverage is moaning because of the fake data that's being added if none comes back from the API. Not writing a test for that as it will be gone soon!

src/app/static/src/app/components/Filters.vue
```diff
@@ -1,10 +1,41 @@
...
+                    <label>Survey</label>
+                    <treeselect id="survey-filters" :multiple="true"
+                                :options="surveyFilters.available"
+                                :value="surveyFilters.selected"
+                                :normalizer = "treeselectNormalizer"
+                                @input="updateSurveyFilter"></treeselect>
+                </div>
+                <div class="col-sm-3">
+                    <label>Region</label>
+                    <treeselect id="survey-filters" :multiple="true"
```
r-ash: This ID should be `region-filters`?

src/app/static/src/scss/style.scss
```diff
@@ -108,8 +113,16 @@ a {
 }
 
 .vue-treeselect__label-container, .vue-treeselect__label-container:hover {
-  .vue-treeselect__checkbox--checked {
-    border-color: $theme-red;
-    background: $theme-red;
+  .vue-treeselect__checkbox--checked, .vue-treeselect__checkbox--indeterminate {
+    border-color: $theme-red !important;
```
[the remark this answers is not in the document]
EmmaLRussell: Seemed to be necessary here to overcome the treeselect component's urge to make everything blue. We don't define the innards of the components so we don't have the option of putting additional classes on these sub-elements, so not sure there's much alternative.

src/app/static/src/app/components/Filters.vue
EmmaLRussell: Yup!

EmmaLRussell:
>Should we hide the survey filter on all tabs except for the survey page? (at the moment these filters don't belong to a tab do they?)

We've got a separate ticket to make a visual indication of what filters are relevant for the currently displayed data. Maybe hiding the survey component, or maybe just greying it out (might be more odd if it just vanished).

EmmaLRussell:
> In the UI there isn't a way to switch between uploaded data - that will be a separate PR?

yep, tabs will be in a separate ticket

EmmaLRussell:
>  Or we should keep these filters but have some visual cue they are unavailable (grey them out or something)

Yeah, I thought we'd decided to keep any filters that the user has selected, even if some of them aren't currently relevant. (Don't lose them when switch back.) Future ticket to grey or otherwise indicate that these tickets aren't currently available.)

r-ash:
> > Or we should keep these filters but have some visual cue they are unavailable (grey them out or something)
> 
> Yeah, I thought we'd decided to keep any filters that the user has selected, even if some of them aren't currently relevant. (Don't lose them when switch back.) Future ticket to grey or otherwise indicate that these tickets aren't currently available.)

That sounds good.

EmmaLRussell:
Made the Programme change, updated the duff ID, made the stepper centred, with a horizontal rule, and tweaked the height of the filter controls, so hopefully it looks a little tidier!

r-ash:
Still reads "program" in the stepper (sorry)

EmmaLRussell:
> Still reads "program" in the stepper (sorry)

:woman_facepalming:

PR #48, opened 2019-09-16
update styles

hillalex:
* Updated the map view so filters are down the side. 
* Made the container width full for screens smaller than xl, fixed thereafter (and consistent between the stepper and the content)
* Removed the page headers and made the stepper names bolder/made the currently selected one more obvious 

Also fixed the issue with files not re-loading on page refresh, because it was annoying for iterating styles/refreshing the page a lot!

### DIFF
--- a/src/app/static/src/scss/style.scss
+++ b/src/app/static/src/scss/style.scss
@@ -1,8 +1,9 @@
 @import "partials/variables";
-@import "partials/maps";
 @import "../../node_modules/bootstrap/scss/bootstrap";
 @import "../../node_modules/leaflet/dist/leaflet";
 @import "../../node_modules/@riophae/vue-treeselect/dist/vue-treeselect";
+@import "partials/maps";
+@import "partials/stepper";
 
 a {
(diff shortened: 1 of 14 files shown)


src/app/static/src/scss/style.scss
```diff
@@ -102,38 +69,3 @@ a {
...
-
-.vue-treeselect__label-container, .vue-treeselect__label-container:hover {
-  .vue-treeselect__checkbox--checked, .vue-treeselect__checkbox--indeterminate {
-    border-color: $theme-red !important;
-    background: $theme-red !important;
-    width: auto;
-  }
-}
-
-.vue-treeselect {
```
hillalex: Mostly just moved these into the maps.scss partial, but removed this z-index rule and also the rule below - I couldn't see what the second one was doing but perhaps this was in error? What was the overflow-x auto rule doing? https://github.com/mrc-ide/hint/pull/48/files#diff-5a8656bcbf5cc2b00d87f69b9918fcc1L134

PR #49, opened 2019-09-16
mrc-484 tabs

hillalex:
Adds tabs to change between selected data type


src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
```diff
@@ -30,7 +30,18 @@
                 </form>
                 <filters></filters>
             </div>
-            <div class="col-md-9 sap-filters">
+            <div class="col-md-9">
+                <ul class="nav nav-tabs">
+                    <li class="nav-item">
+                        <a class="nav-link" :class="survey.tabClass" v-on:click="selectTab(2)">Survey</a>
```
hillalex: Note the calls to e.g. `selectTab(2)` - these have to be js rather than ts, hence not being able to use the enum. 

Alternatives: 
- create a js enum instead of a ts one (i.e. an object rather than a type)
- define 3 separate component methods "selectSurveyTab" etc. which can use the enum, and call them

I'm inclined to leave as is given it's the simplest solution and the tab changing functionality is under test

src/app/static/src/app/store/filteredData/filteredData.ts
```diff
@@ -5,7 +5,7 @@ import { getters } from './getters';
 import { RootState} from "../../root";
 import { FilterOption } from "../../generated";
 
-export enum DataType { ANC, Survey, Program }
+export enum DataType { ANC, Program, Survey }
```
hillalex: Changed these to be alphabetical so we can easily know which enum value corresponds to which number (see above comment)

hillalex:
> LGTM - one tiny styling thing - on hover, the text of the tab turns black, but other red-themed elements (the login button and the steps) have the background go darker but retain white text on hover. Might be more consistent (and a bit more readable) to do this on the tabs too..

Ah thanks, I had noticed that also and meant to fix.

PR #50, opened 2019-09-17
let user jump to run model after survey and program data upload

hillalex:
We're skipping the "review uploads" section for now, to work on model run, so just assume it's complete once the survey and program uploads are.


src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
```diff
@@ -1,38 +1,36 @@
 <template>
-    <div>
```
hillalex: Just randomly noticed there was an extra div here so removed it.

PR #51, opened 2019-09-17
mrc-469 loading spinner

hillalex:
Adds a loading spinner icon in place of the red tick while a file is uploading, cleared on success or error. Also noticed that errors weren't being cleared after a further successful upload, so have fixed that and put it under test.


src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -17,20 +17,23 @@ export interface BaselineActions {
 export const actions: ActionTree<BaselineState, RootState> & BaselineActions = {
 
     async uploadPJNZ({commit}, formData) {
+        commit({type: "PJNZLoaded", payload: null});
```
EmmaLRussell: Bit confused by these changes. Why is it commiting loaded/uploaded mutations before getting a response from the API? And is there a reason why this mutation is PJNZLoaded by Shape/Population*Up*loaded

hillalex: Once you've selected a new file, the old one should be deleted and the red "valid" tick should disappear. These commits just clear the existing data.

The naming I think is just an artefact of how it was originally implemented, pjnz was the first one and the others ended up being called "uploaded". We should make them consistent and perhaps in light of the fact that the resetting behaviour wasn't clear to you, perhaps we should go with a different name altogether, something just like "SetPJNZ" etc?

hillalex: Or "PJNZUpdated" maybe?

PR #53, opened 2019-09-18
mrc-473 update to new schema

hillalex:
This PR brings the app into line with the latest hintr schema, passing the shape file alongside the survey and programme files for validation.

### DIFF
--- a/src/app/build.gradle
+++ b/src/app/build.gradle
@@ -16,12 +16,6 @@ jacocoTestReport {
 		xml.enabled = true
 		csv.enabled = false
 	}
-
-	afterEvaluate {
-		classDirectories = files(classDirectories.files.collect {
-			fileTree(dir: it, exclude: '**/APIClient.kt')
-		})
-	}
 }
 
 group = 'org.imperial.mrc'
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/APIClient.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/APIClient.kt
@@ -6,17 +6,18 @@ import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Component
 
 interface APIClient {
-    fun validate(path: String, type: FileType): ResponseEntity<String>
+    fun validateBaseline(path: String, type: FileType): ResponseEntity<String>
+    fun validateSurveyAndProgramme(path: String, shapePath: String, type: FileType): ResponseEntity<String>
 }
 
 @Component
-class HintAPIClient(
+class HintrAPIClient(
         appProperties: AppProperties,
         private val objectMapper: ObjectMapper) : APIClient {
 
     private val baseUrl = appProperties.apiUrl
 
-    override fun validate(path: String, type: FileType): ResponseEntity<String> {
+    override fun validateBaseline(path: String, type: FileType): ResponseEntity<String> {
 
         val json = objectMapper.writeValueAsString(
                 mapOf("type" to type.toString().toLowerCase(),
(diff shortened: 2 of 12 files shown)


src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/BaselineController.kt
```diff
@@ -16,34 +14,14 @@ class BaselineController(private val fileManager: FileManager,
 
     @PostMapping("/pjnz/")
     @ResponseBody
-    fun uploadPJNZ(@RequestParam("file") file: MultipartFile): String {
-
-        val fileName = file.originalFilename!!
-        fileManager.saveFile(file, FileType.PJNZ)
-
-        // TODO request validation from R API and get back JSON
```
hillalex: Also replaced this TODO from ages ago now that the PJNZ endpoint exists in hintr.

src/app/build.gradle
```diff
@@ -16,12 +16,6 @@ jacocoTestReport {
 		xml.enabled = true
 		csv.enabled = false
 	}
-
-	afterEvaluate {
-		classDirectories = files(classDirectories.files.collect {
```
hillalex: I had previously excluded the APIClient from tests for reasons best known to past me. Have now added a test class for it.

src/app/static/src/tests/integration/surveyAndProgram.itest.ts
```diff
@@ -16,9 +16,12 @@ describe("Survey and program actions", () => {
 
         await actions.uploadSurvey({commit} as any, formData);
 
-
-        expect(commit.mock.calls[0][0]["type"]).toBe("SurveyLoaded");
-        expectSelectedDataTypeUpdate(commit.mock.calls[1], DataType.Survey);
+        // Unfortunately we can't test the success path because it relies on a persistent
+        // session between uploading a shape file and an anc file. But the success path is
```
r-ash: Super minor but, this should read "uploading a shape file and a survey file"?

src/app/static/src/tests/integration/surveyAndProgram.itest.ts
```diff
@@ -31,8 +34,12 @@ describe("Survey and program actions", () => {
 
         await actions.uploadProgram({commit} as any, formData);
 
-        expect(commit.mock.calls[0][0]["type"]).toBe("ProgramLoaded");
-        expectSelectedDataTypeUpdate(commit.mock.calls[1], DataType.Program);
+        // Unfortunately we can't test the success path because it relies on a persistent
+        // session between uploading a shape file and an anc file. But the success path is
```
r-ash: As above this should be "a shape file and a programme file"?

src/app/src/main/kotlin/org/imperial/mrc/hint/APIClient.kt
```diff
@@ -6,17 +6,18 @@ import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Component
 
 interface APIClient {
-    fun validate(path: String, type: FileType): ResponseEntity<String>
+    fun validateBaseline(path: String, type: FileType): ResponseEntity<String>
```
r-ash: Support for validation all baseline data together will come later? As a separate function called something like `validateBaselineAll`? I can update the endpoints on the R side from `validate/input` and `validate/baseline` to `validate/baseline` and `validate/baseline/all` so we are agreeing on our names.

hillalex: Yep, that all sound good to me

hillalex:
> Probably not a thing we should try and fix here but this raises an issue of how we're dealing with errors. It is good that we can return errors from the R API when it is a validation error. When it is an unexpected like the one attached in the image it is less useful. Really in this case we probably need some generic error message "Unexpected error please contact a system admin" or something and then log the error somewhere we can go and search for it. I think most of that handling will be done in the R API though. I have added a ticket for that https://vimc.myjetbrains.com/youtrack/issue/mrc-496
> 
> Selection_051

Yes, this would be nice. The pattern we have for handling this in Montagu api works quite well - all exceptions we throw ourselves are of a custom type, "MontaguError". We have a global error handler that catches all exceptions and if they are of an unknown type convert them into a "MontaguError" with a message like the one you suggest. We give these unexpected errors a random string id which gets returned to the user so they can quote it in a bug report and we can easily look for it in the logs.

PR #54, opened 2019-09-18
mrc-491 add model run endpoint

hillalex:
~~Contains commits from https://github.com/mrc-ide/hint/pull/53 so that should be merged first.~~

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ModelRunController.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ModelRunController.kt
@@ -12,7 +12,7 @@ class ModelRunController(val fileManager: FileManager, val apiClient: APIClient)
 
     @PostMapping("/run/")
     @ResponseBody
-    fun uploadSurvey(@RequestBody params: ModelRunParameters): ResponseEntity<String> {
+    fun run(@RequestBody params: ModelRunParameters): ResponseEntity<String> {
         val allFiles = fileManager.getAllFiles()
         return apiClient.submit(allFiles, params)
     }
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ModelRunController.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ModelRunController.kt
@@ -1,0 +1,19 @@
+package org.imperial.mrc.hint.controllers
+
+import org.imperial.mrc.hint.APIClient
+import org.imperial.mrc.hint.FileManager
+import org.imperial.mrc.hint.models.ModelRunParameters
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.*
+
+@RestController
+@RequestMapping("/model")
+class ModelRunController(val fileManager: FileManager, val apiClient: APIClient) {
+
+    @PostMapping("/run/")
+    @ResponseBody
+    fun uploadSurvey(@RequestBody params: ModelRunParameters): ResponseEntity<String> {
+        val allFiles = fileManager.getAllFiles()
+        return apiClient.submit(allFiles, params)
+    }
+}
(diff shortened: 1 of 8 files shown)


src/app/src/main/kotlin/org/imperial/mrc/hint/models/ModelRunParameters.kt
```diff
@@ -0,0 +1,6 @@
+package org.imperial.mrc.hint.models
+
+data class ModelRunParameters(val maxIterations: Int,
+                              val noOfSimulations: Int,
+                              val sleep: Int,
```
r-ash: So this sleep is an artefact of us not having Jeff's model yet. The model atm just sleeps for specified time and returns 2. If that adds debt I can update the model so that this isn't a required param.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ModelRunController.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.*
+
+@RestController
+@RequestMapping("/model")
+class ModelRunController(val fileManager: FileManager, val apiClient: APIClient) {
+
+    @PostMapping("/run/")
+    @ResponseBody
+    fun uploadSurvey(@RequestBody params: ModelRunParameters): ResponseEntity<String> {
```
r-ash: This function should be `runModel` or something?

r-ash: There is no test that calls this itself either? That will come when hook up the UI to this?

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ModelRunTests.kt
```diff
@@ -0,0 +1,29 @@
...
+        val params = ModelRunParameters(1,1,1,mapOf())
+        val headers = HttpHeaders()
+        headers.contentType = MediaType.APPLICATION_JSON
+        val jsonString = ObjectMapper().writeValueAsString(params)
+        return HttpEntity(jsonString, headers)
+    }
+
+    @ParameterizedTest
+    @EnumSource(IsAuthorized::class)
+    fun `can run model`(isAuthorized: IsAuthorized) {
```
r-ash: Do we need a test for when this fails to start a model running?

src/app/src/main/kotlin/org/imperial/mrc/hint/models/ModelRunParameters.kt
hillalex: I mean, this is going to have to change either way so I wouldn't worry for now. As and when we get the real interface I will update this accordingly.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ModelRunController.kt
hillalex: Oops, good catch! There is an integration test that calls this endpoint, I thought it was overkill to have a controller test as well given how simple it is...

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ModelRunTests.kt
hillalex: In general I've gone with our integration tests just testing the happy path for each endpoint. Which does mean we're not testing that errors from hintr are correctly formatted, but that should all be under test in hintr itself. We could test the error path here too as extra insurance, would that be helpful for you?

r-ash: No sorry that is good - we probably only need a test when front end part is in that errors are shown. You're right that we should be covering the error cases in hintr.

PR #56, opened 2019-09-18
mrc-494 model run frontend

hillalex:
Adds a new `modelRun` module and a big button for running the model with dummy params.

~~Contains commits from https://github.com/mrc-ide/hint/pull/54 so that should be merged first.~~


src/app/static/src/app/store/modelRun/modelRun.ts
```diff
@@ -0,0 +1,25 @@
...
+import {RootState} from "../../root";
+import {actions} from "./actions";
+import {mutations} from "./mutations";
+
+export interface ModelRunState {
+    modelRunId: string
+    status: ModelRunStatus
+}
+
+export type ModelRunStatus = "NotStarted" | "Started" | "Complete"
```
EmmaLRussell: Would this be more TypeScripty as an enum?

PR #58, opened 2019-09-18
mrc-497 load any existing survey and programme data on page load

hillalex:
- adds endpoints to retrieve existing survey and programme data just as we do for baseline
- calls these endpoints on page load
- wait til all endpoints have been hit to put the app in a ready state and show a loading spinner until then

The idea here is that if you are logged in you can navigate away and back again and your data will reload (rather than you having to uploading all your files afresh). Further PRs save portions of the vue state in localStorage (e.g. the step you were on) to fully restore the state.

(I realised that to completely restore the state it will be enough to save the active step, run id and run status in localStorage and use endpoints to reload the data that is too big for localStorage)


hillalex:
> Sorry, I don't think I have enough context to review this - I can sort of see how it's working but I've not worked through how the app is put together to see how it all meshes. @r-ash - could you take a look please?

No worries, I'll let this wait for Emma's review.

src/app/static/src/app/components/LoadingSpinner.vue
```diff
@@ -0,0 +1,103 @@
+<template>
```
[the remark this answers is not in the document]
hillalex: Noting this here so I don't have to look this up again and will update with a code comment later:
https://loading.io/spinner/spin/-ajax-spinner-gif

src/app/static/src/app/components/Stepper.vue
```diff
@@ -78,21 +84,24 @@
...
                     1: this.baselineComplete,
                     2: this.surveyAndProgramComplete,
                     3: this.surveyAndProgramComplete, // for now just mark as complete as soon as it's ready
                     4: false,
                     5: false
                 }
-            }
+            },
+            ...mapState<RootState>({
+                ready: state => state.surveyAndProgram.ready && state.baseline.ready
```
EmmaLRussell: I suppose ready could be a getter on the root state - to check the ready state of all the component states. Maybe not worth it if the stepper is the only thing that needs to use it.

src/app/static/src/app/store/surveyAndProgram/mutations.ts
```diff
@@ -37,5 +38,9 @@ export const mutations: MutationTree<SurveyAndProgramDataState> & SurveyAndProgr
 
     ANCError(state: SurveyAndProgramDataState, action: PayloadWithType<string>) {
         state.ancError = action.payload;
+    },
+
+    Ready(state: SurveyAndProgramDataState) {
```
EmmaLRussell: Since this mutation is going to be in (almost) every step, could we have a base mutation object with a Ready(state: InputState) function on it, which the other mutation objects 'inherit' from via {...baseMutations}?

src/app/static/src/app/components/Stepper.vue
hillalex: I think this is the only component that needs it

EmmaLRussell:
Good to merge once conflicts fixed.

PR #59, opened 2019-09-18
mrc-498 remember active step

hillalex:
1. Persist active step in local storage
1. On load (after app state marked as "ready") retrieve active step and check if it is valid. This is in case for some reason one of the data retrieval endpoint fails (e.g. because of a bad connection), rendering the active step invalid (because the user needs to re-upload data to proceed). 
1. If invalid just remove from local storage and do nothing. If valid re-populate the active step accordingly.
1. Clear local storage on the login screen because all app state should only be valid as long as the server session is.

~~Contains commits from https://github.com/mrc-ide/hint/pull/58 so that should be merged first~~


src/app/static/src/app/components/Stepper.vue
```diff
@@ -130,6 +132,24 @@
...
             ModelRun
+        },
+        watch: {
+            ready: function (newVal) {
+                if (newVal) {
+                    const activeStep = localStorageManager.getInt("activeStep");
+
+                    if (activeStep) {
+                        const invalidSteps = [1, 2, 3, 4, 5]
+                            .filter((i) => i < activeStep && !this.complete[i]);
```
EmmaLRussell: Could use Object.keys(this.complete) here to avoid the hardcoded array which assumes step number, though in that case you'd get them back as strings I think.

src/app/static/src/app/store/baseline/mutations.ts
```diff
@@ -3,6 +3,7 @@ import {BaselineState} from "./baseline";
 import {PjnzResponse, PopulationResponse, ShapeResponse} from "../../generated";
 import {PayloadWithType} from "../../types";
 import {readyStateMutations} from "../shared/readyStateMutations";
+import {ReadyState} from "../../root";
```
EmmaLRussell: Import not needed?

hillalex:
> I'm finding that if I reload from step 2 when I'd already uploaded at least one file, it correctly jumps to step 2 with the filenames shown, but the choropleth and filters are not visible. They reappear if I upload a further file. Suggests that selectedDataType isn't being persisted?

This literally just persists the active step. Further PRs for persisting e.g. model run state, but I hadn't considered filter state and you're right, that will need to be persisted also. Opened a ticket: https://vimc.myjetbrains.com/youtrack/issue/mrc-516

src/app/static/src/app/components/Stepper.vue
hillalex: realised could do `this.steps.map((s: Step) => s.number)` which seems nicer

PR #61, opened 2019-09-19
add state repository

hillalex:


src/app/static/src/tests/integration/baseline.itest.ts
```diff
@@ -44,7 +45,7 @@ describe("Baseline actions", () => {
 
         expect(commit.mock.calls[1][0]["type"]).toBe("ShapeUpdated");
         expect(commit.mock.calls[1][0]["payload"]["filename"])
-            .toBe("malawi.geojson");
+            .toBe("A59C122F27303D93467C0F2D09489878.geojson");
```
hillalex: Note this ticket to return original filenames rather than hashes: https://vimc.myjetbrains.com/youtrack/issue/mrc-513

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/HomeController.kt
```diff
@@ -1,16 +1,22 @@
...
 @Controller
-class HomeController(private val appProperties: AppProperties) {
+class HomeController(
+        private val stateRepository: StateRepository,
+        private val session: Session,
+        private val appProperties: AppProperties) {
 
     @GetMapping("/")
     fun index(model: Model): String {
+        stateRepository.saveSession(session.getId(), session.getUserProfile().id)
```
hillalex: I've gone with just saving the session to the db when hitting the index - does this seem sensible? It's a proxy for "once the user is logged in", can you think of a more direct way to get at that?

hillalex:
One thing I'm wondering about is how we decide when to delete old files. I guess something like, record which files get  "saved" (when the user downloads their session) and then regularly clean up any files that are orphaned in the sense that they are no longer associated with any session *and* have never been saved?

src/app/src/main/resources/application-node.properties
```diff
@@ -0,0 +1 @@
+server.servlet.session.cookie.http-only=false
```
EmmaLRussell: This is to prevent the cookie being accessed from the client?

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/HomeController.kt
EmmaLRussell: After this maybe?
https://github.com/mrc-ide/hint/blob/86f8dce9c5600ecfb2f3c9887d07ef0e8dea1700/src/app/src/main/kotlin/org/imperial/mrc/hint/security/HintDbProfileService.kt#L20

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/CleanDatabaseTests.kt
```diff
@@ -27,6 +28,11 @@ abstract class CleanDatabaseTests
 
         val tableFields = Tables::class.java.fields
 
+        // this table has foreign keys
+        // so has to be deleted first
+        dsl.deleteFrom(SESSION_FILE)
```
EmmaLRussell: It looks like Jooq should support truncate with cascade, which I think should avoid needing to sort out foreign keys first. 
https://www.jooq.org/javadoc/3.7.1/org/jooq/TruncateCascadeStep.html

src/app/src/main/kotlin/org/imperial/mrc/hint/db/StateRepository.kt
```diff
@@ -0,0 +1,93 @@
...
+import org.imperial.mrc.hint.FileType
+import org.imperial.mrc.hint.db.Tables.*
+import org.imperial.mrc.hint.models.SessionFile
+import org.jooq.DSLContext
+import org.jooq.Record
+import org.jooq.Record1
+import org.jooq.Result
+import org.springframework.stereotype.Component
+
+interface StateRepository {
```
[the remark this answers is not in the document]
hillalex: Yes I think you're right.

src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
```diff
@@ -29,46 +32,41 @@ interface FileManager {
...
 
     override fun getFile(type: FileType): File? {
-
-        val id = pac4jConfig.sessionStore.getOrCreateSessionId(context)
-
-        val pjnzFiles = File("${appProperties.uploadDirectory}/$id/$type")
-                .listFiles()
-
-        return pjnzFiles?.first()
+        val hash = stateRepository.getSessionFileHash(session.getId(), type)
```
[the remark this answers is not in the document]
hillalex: The db structure allows for multiple so this won't be a problem from that point of view. 

I think the only one that might be multiple is PJNZ? And yes if/when we want to support that then we'd need this method to return a list instead of a singleton. But a fair bit will have to change if/when we support multiple PJNZ but I don't think anything here precludes that.

Not sure what you mean about how to identify them - each file will be uniquely identified by its hash!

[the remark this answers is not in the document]
EmmaLRussell: This is true! I think I was thinking about what parameters you'd pass this method to get one hash out. But I guess you would just get a list of all hashes of that type instead.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/CleanDatabaseTests.kt
hillalex: ah brill, thanks

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/HomeController.kt
EmmaLRussell: Oh, so this (from the Session class) doesn't work yet?
`fun getId(): String = pac4jConfig.sessionStore.getOrCreateSessionId(webContext)`

hillalex: I don't think we can invoke the session class here because we'd get a circular dependency - session needs pac4jconfig which needs profileservice. I tried to read the session id directly from the context, but it seems like it's the old session id still, and a new one gets issued at some slightly later stage in the auth process... (a new session id is correctly issued after auth to avoid session hijacking exploit)

PR #63, opened 2019-09-24
mrc-471 Filter Choropleth

EmmaLRussell:


src/app/static/src/app/components/ChoroplethFilters.vue
```diff
@@ -0,0 +1,160 @@
+<template>
```
EmmaLRussell: I haven't tried to reuse code between this and the once-and-future Filters component, as it felt like they would diverge enough to not be worth it.

src/app/static/src/app/components/plots/Choropleth.vue
```diff
@@ -77,18 +84,10 @@
                 }
             },
             prevEnabled: function() {
-                const result = !!(this.indicatorData.prevRange.min || this.indicatorData.prevRange.max);
```
EmmaLRussell: Trying to derive whether a indicator should be enabled based on the range of the data returned isn't really going to work because it's now perfectly valid for there to be no data points in the data being displayed as the combination of selected filters may not map to any values (this is often the case in the sample data). Hardcoding it by dataType for now, rather than doing a separate sweep of the unfiltered data to see whether there are any values for the indicator. Suspect we'll revisit this at some point, might end up being metadata.

src/app/static/src/app/components/ChoroplethFilters.vue
```diff
@@ -0,0 +1,161 @@
...
+            },
+            updateSurveyFilter(id: string) {
+                this.updateFilter(FilterType.Survey, id, this.surveyFilters.available);
+            },
+            refreshSelectedChoroplethFilters(){
+                //if the selected data type has changed, we should update the choropleth filters if the dataset of that
+                //type does not include any of the selected filters as values. Set the filter to the first available value
+                const selectedChoroplethFilters = this.selectedChoroplethFilters;
+
+                const getNewFilter = function(filterName: string, available: FilterOption[]) {
```
hillalex: This could be it's own method no? Would make this long function easier to read. Also can it be called `getNewSelectedFilterOption`?

src/app/static/src/app/store/filteredData/filteredData.ts
```diff
@@ -58,9 +67,41 @@ export const initialSelectedFilters: SelectedFilters = {
     }
 };
 
+export const initialSelectedChoroplethFilters: SelectedChoroplethFilters = {
+    sex: null,
+    age: null,
+    survey: null,
+    updateByType: function(type: FilterType, value: FilterOption) {
```
hillalex: Shouldn't this be a mutation? Pretty sure you shouldn't be mutating state outside of mutations.

src/app/static/src/app/store/filteredData/filteredData.ts
```diff
@@ -58,9 +67,41 @@ export const initialSelectedFilters: SelectedFilters = {
...
+                break;
+            case (FilterType.Sex):
+                this.sex = value;
+                break;
+            case (FilterType.Survey):
+                this.survey = value;
+                break;
+        }
+    },
+    getByType: function(type: FilterType): FilterOption | null {
```
hillalex: Where is this used? Also pretty sure store state needs to be serialisable aka no functions.

hillalex:
Thanks, find this a lot clearer now. Just need to fix the conflict then merge.

PR #68, opened 2019-09-25
mrc-520 Tooltip in Choropleth

EmmaLRussell:


src/app/static/src/app/components/plots/Choropleth.vue
```diff
@@ -87,6 +88,20 @@
             artEnabled: function() {
                 const result = !!(this.indicatorData.artRange.min || this.indicatorData.artRange.max);
                 return result;
+            },
+            options: function() {
+                const indicatorData = this.indicatorData;
+                const indicator = this.indicator;
+                return {
+                    onEachFeature: function onEachFeature(feature: any, layer: any) {
```
hillalex: We have types for feature and layer, these don't have to be `any`.

hillalex: Edit: just checked and we don't have types for leaflet installed! My bad. After doing `npm install @types/leaflet` you can use `import {Layer} from "leaflet"` to get the correct type.

PR #69, opened 2019-09-25
mrc-513 restructure to pass original filename to hintr

hillalex:


hillalex:
schema has changed since this

PR #72, opened 2019-09-26
mrc-529 move stepper state to module

hillalex:
For now this just moves stepper state into it's own module. 

Later PR will change over the save/restore to/from local storage logic to save/load a partial global state object instead of just the active step.

I've not actually added any unit tests for the mutations/actions cos everything is still under test via the component tests. Do you think worth adding unit tests as well?

### DIFF
--- a/src/app/static/src/app/store/stepper/actions.ts
+++ b/src/app/static/src/app/store/stepper/actions.ts
@@ -1,13 +1,38 @@
 import {ActionContext, ActionTree} from "vuex";
 import {RootState} from "../../root";
-import {StepperState} from "./stepper";
+import {StepDescription, StepperState} from "./stepper";
+import {localStorageManager} from "../../localStorageManager";
 
 export interface StepperActions {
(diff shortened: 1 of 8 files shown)


src/app/static/src/app/store/stepper/actions.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export const actions: StepperActions & ActionTree<StepperState, RootState> = {
+    jump({commit}, step) {
+        commit({type: "Jump", payload: step});
+    },
+
+    next(store) {
+        const {state, getters} = store;
+        if (getters.complete[state.activeStep]) {
```
hillalex: Btw seems like it's perfectly good practice to access getters from an action - wasn't this what was needed in the filter flow to update the selected filter option when the selected data type changed?

EmmaLRussell: Hm. I think i was looking at the mutations for that rather than the actions, but maybe it would have made most sense in the actions anyway...

PR #73, opened 2019-09-26
mrc-531 auto-save and load global state from local storage

hillalex:
This adds the functionality to auto-save to local storage on every mutation, and reload from local storage on page load.


src/app/static/src/app/components/Stepper.vue
```diff
@@ -69,26 +69,21 @@
...
-                return this.activeStep == num;
+            ...mapActions(namespace, ["jump", "next"]),
+            ...mapActions(["reload"]),
+            isActive(num: number) {
+                return this.ready && this.activeStep == num;
             },
-            enabled(num: number) {
-                return this.steps.slice(0, num)
+            isEnabled(num: number) {
+                return this.ready && this.steps.slice(0, num)
```
hillalex: Changed the active/enabled/complete states so that they only show up once the data is loaded, to avoid odd user experience of some steps being coloured  before others.

src/app/static/src/app/root.ts
```diff
@@ -22,20 +24,36 @@ export interface ReadyState {
     ready: boolean
 }
 
+const persistState = (store: Store<RootState>) => {
```
hillalex: Turns out there is a notion of middleware, just as we needed!

src/app/static/src/app/store/root/actions.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+export const actions: ActionTree<RootState, RootState> & RootActions = {
+
+    reload(store) {
+        const {state, getters, commit} = store;
+        const invalidSteps = state.stepper.steps.map((s: StepDescription) => s.number)
+            .filter((i: number) => i < state.stepper!!.activeStep && !getters["stepper/complete"][i]);
+
+        if (invalidSteps.length > 0) {
+            commit({type: "Reset"});
```
hillalex: The flow I've gone with is;

1. initialise store states from local storage
2. wait for async data to be ready
3. this action fires which resets the state if in an invalid state

src/app/static/src/tests/components/choroplethFilter.test.ts
```diff
@@ -1,6 +1,6 @@
 import {createLocalVue, shallowMount} from '@vue/test-utils';
 import Vue from 'vue';
-import Vuex from 'vuex';
```
hillalex: Changes to this file are irrelevant to this ticket but I was seeing errors when running tests saying things like "action choroplethFilterUpdated not defined"/"mutation ChoroplethFilterUpdated not defined" so I just tidied this up to fix those.

src/app/static/src/app/localStorageManager.ts
```diff
@@ -1,30 +1,26 @@
...
-            window.localStorage.setItem(key, JSON.stringify(value));
-        } else {
-            this.removeItem(key);
-        }
-    };
+const appStateKey = "appState";
 
-    removeItem = (key: string) => {
-        window.localStorage.removeItem(key);
+export const serialiseState = (rootState: RootState): Partial<RootState> => {
```
hillalex: This function can get re-used when saving the state to file

src/app/static/src/app/store/root/actions.ts
EmmaLRussell: So this action isn't actually reloading anything, but should only be invoked by the stepper once the filles have been loaded from the backend (and the state should already have been initialised from local storage), and then this method checks that everything's good to go? Maybe it should be called 'validate' or something similar?

src/app/static/src/app/store/root/actions.ts
```diff
@@ -0,0 +1,19 @@
+import {ActionContext, ActionTree, Store} from "vuex";
+import {RootState} from "../../root";
+import {StepDescription} from "../stepper/stepper";
+
+export interface RootActions {
+    reload: (store: ActionContext<RootState, RootState>) => void;
+}
+export const actions: ActionTree<RootState, RootState> & RootActions = {
+
+    reload(store) {
```
EmmaLRussell: So this action isn't actually reloading anything, but should only be invoked by the stepper once the filles have been loaded from the backend (and the state should already have been initialised from local storage), and then this method checks that everything's good to go? Maybe it should be called 'validate' or something similar?

hillalex: Good shout, that's much better.

PR #75, opened 2019-09-30
Mrc 451 Get map bounds from selected regions

EmmaLRussell:
This includes when selected region is the default top level region

Contains changes from https://github.com/mrc-ide/hint/pull/71 which should be merged first


src/app/static/src/app/components/plots/Choropleth.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-    <l-map ref="map" :zoom="zoom" :center="center" style="height: 800px; width: 100%">
+    <l-map ref="map" v-bind:center="mapCenter" style="height: 800px; width: 100%">
+        <l-geo-json ref="selectedRegionGeoJson" :geojson="selectedRegionFeature"
```
EmmaLRussell: The easiest way to get the bounds from the selected feature seemed to be to render it as a component and then ask for its bounds (hence having to wait for the nextTick in the update method below). You can't assume that it will already be rendered as it may not be at the current selected level. Render this feature before the others so that it is behind, and the tooltips for the current features still work!

src/app/static/src/app/components/plots/Choropleth.vue
```diff
@@ -126,6 +150,34 @@
                 }
 
                 return data;
+            },
+            getFeatureFromAreaId(areaId: string){
+                return (this.features as any[]).filter(f => f.properties.area_id == areaId)[0];
+            },
+            updateBounds(){
+                Vue.nextTick().then(() => {
+                    const geoJsonArray = this.$refs.selectedRegionsGeoJson;
```
hillalex: I think this whole thing can be replaced with:
```
                Vue.nextTick().then(() => {
                    const map = this.$refs.map;                  
                    map.fitBounds(this.selectedRegionFeatures.map((f: Feature) => new GeoJSON(f).getBounds()));
                });
```
and the ghost features removed from the map

src/app/static/src/app/components/plots/Choropleth.vue
```diff
@@ -77,19 +94,25 @@
...
                         layer.bindPopup(`<div>
                                 <strong>${area_name}</strong>
                                 <br/>${value}
                             </div>`);
                     }
                 }
+            },
+            selectedRegionFeatures: function() {
+                if (this.selectedRegions && this.selectedRegions.length > 0) {
+                    return (this.selectedRegions as any[]).map(r => this.getFeatureFromAreaId(r.id));
```
hillalex: `as any[]` can be `as NestedFilterOption[]`

EmmaLRussell:
> I just noticed that the behaviour when you unselect all regions is to colour in everything, as though you'd really selected all regions. This feels weird to me, was this specifically requested behaviour?

Yes, as it's removing all filters. I think it seems weird particularly because it's being combined with filters where you can only select exactly one value..

hillalex:
> > I just noticed that the behaviour when you unselect all regions is to colour in everything, as though you'd really selected all regions. This feels weird to me, was this specifically requested behaviour?
> 
> Yes, as it's removing all filters. I think it seems weird particularly because it's being combined with filters where you can only select exactly one value..

Agreed. Ok well I guess leave as is and see what the users think.

src/app/static/src/app/components/plots/Choropleth.vue
hillalex: Oh I guess you don't need the nextTick bit at all either!

EmmaLRussell: :facepalm: Thank you! I tried and failed to get something like that working, think I was looking in the wrong place for the right GeoJSON class..

PR #77, opened 2019-10-02
Mrc-504_minmax Use min and max for chloropleth ranges from ploting metadata

EmmaLRussell:
Contains changes from https://github.com/mrc-ide/hint/pull/76 which should be merged first

### DIFF
--- a/src/app/static/src/app/store/filteredData/getters.ts
+++ b/src/app/static/src/app/store/filteredData/getters.ts
@@ -100,6 +100,7 @@ export const getters = {
                 indicators.prev.color = getColor(indicators.prev, getters.choroplethRanges.prev, getters.colorFunctions.prev);
             }
         }
+        alert(JSON.stringify(result));
         return result;
     },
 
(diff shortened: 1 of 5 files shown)


src/app/static/src/app/store/filteredData/getters.ts
```diff
@@ -73,8 +62,11 @@ export const getters = {
                     valueColumn = "current_art";
                     break;
                 case (DataType.ANC):
+                    //TODO: once using metadata, we need to allow multiple values per row, as in the case on ANC data
+                    //which contains both prevalence and art data in each row (wide format), unlike survey, which
+                    //is int long format, with an indicator column to show which indicator the value each row provides
                     indicator = "prev";
-                    valueColumn = "ancrt_test_pos";
+                    valueColumn = "prevalence";
```
EmmaLRussell: Using the column for prevalence which is actually in the metadata (though not using metadata to identify these columns yet..)

EmmaLRussell:
>ANC data does not contain ART coverage

Yup, planning on adding that when we're using the metadata for all indicator metadata

>Color scales on legend

Didn't realise that! Thanks!

>If no data for a filter combination the tooltip says undefined

I think I've just changed that to an empty string in the regions branch, as it also goes for unselected regions. But yes, maybe that's also confusing if it's another filter combination that has no rows. Maybe a more explicit message as you suggest would be better.

>app tries to restore state from some cache

In progress!

PR #85, opened 2019-10-08
mrc-549 state not being reset bug fix

hillalex:
When the auto-saved state is invalid for some reason and so gets reset, the baseline and survey/program module states weren't being reset. This was due to the initial state object being passed as a reference instead of copied into the module.

### DIFF
--- a/src/app/static/src/app/root.ts
+++ b/src/app/static/src/app/root.ts
@@ -28,6 +28,7 @@ export interface ReadyState {
 
 const persistState = (store: Store<RootState>) => {
     store.subscribe((mutation: MutationPayload, state: RootState) => {
+        console.log(mutation.type);
         localStorageManager.saveState(state);
     })
 };
@@ -42,13 +43,7 @@ export const emptyState = {
     stepper: initialStepperState
 };
 
-export const initialState = {
-    ...emptyState,
-    ...localStorageManager.getState()
-};
-
 export const storeOptions: StoreOptions<RootState> = {
-    state: initialState,
     modules: {
         baseline,
         metadata,
--- a/src/app/static/src/app/store/root/mutations.ts
+++ b/src/app/static/src/app/store/root/mutations.ts
@@ -8,5 +8,8 @@ export interface RootMutations {
 export const mutations: MutationTree<RootState> & RootMutations = {
     Reset(state: RootState) {
         Object.assign(state, emptyState);
+        state.surveyAndProgram.ready = true;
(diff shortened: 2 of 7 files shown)


src/app/static/src/app/root.ts
```diff
@@ -28,6 +28,7 @@ export interface ReadyState {
 
 const persistState = (store: Store<RootState>) => {
     store.subscribe((mutation: MutationPayload, state: RootState) => {
+        console.log(mutation.type);
```
hillalex: Just thought this was useful

src/app/static/src/app/root.ts
```diff
@@ -42,13 +43,7 @@ export const emptyState = {
     stepper: initialStepperState
 };
 
-export const initialState = {
-    ...emptyState,
-    ...localStorageManager.getState()
-};
-
 export const storeOptions: StoreOptions<RootState> = {
-    state: initialState,
```
hillalex: This is irrelevant but not needed - the state gets overwritten by the modules straight away.

src/app/static/src/app/store/root/mutations.ts
```diff
@@ -8,5 +8,8 @@ export interface RootMutations {
 export const mutations: MutationTree<RootState> & RootMutations = {
     Reset(state: RootState) {
         Object.assign(state, emptyState);
+        state.surveyAndProgram.ready = true;
```
hillalex: Realised that if the state is reset it also reverts to not ready and gets stuck in an endless loading spinner!

src/app/static/src/app/store/stepper/getters.ts
```diff
@@ -16,7 +16,7 @@ export const getters: StepperGetters & GetterTree<StepperState, RootState> = {
             1: rootGetters['baseline/complete'] && rootGetters['metadata/complete'],
             2: rootGetters['surveyAndProgram/complete'],
             3: rootGetters['surveyAndProgram/complete'], // for now just mark as complete as soon as it's ready
-            4: rootState.modelRun.success,
+            4: rootState.modelRun.success && rootState.modelRun.errors.length == 0,
```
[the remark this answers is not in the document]
EmmaLRussell: Maybe that could go in a getter for the modelRun state now, since the complete logic for the other modules lives in their 'complete' getters?

src/app/static/src/app/root.ts
EmmaLRussell: Definitely!

PR #89, opened 2019-10-09
mrc-552 FilterSelect shows placeholder if disabled

EmmaLRussell:
Avoid odd behaviour on disabled filters by setting value to null and showing a preset placeholder ("Not used").

### DIFF
--- a/src/app/static/src/app/components/ChoroplethFilters.vue
+++ b/src/app/static/src/app/components/ChoroplethFilters.vue
@@ -100,7 +100,7 @@
             buildViewFiltersForType(availableFilterOptions: FilterOption[],
                                     selectedFilterOption?: FilterOption) {
                 return {
-                    available: availableFilterOptions,
+                    available: availableFilterOptions ? availableFilterOptions : [],
                     selected: selectedFilterOption ? selectedFilterOption.id : null,
                     disabled: availableFilterOptions == undefined
                 }
--- a/src/app/static/src/app/components/FilterSelect.vue
+++ b/src/app/static/src/app/components/FilterSelect.vue
@@ -17,7 +17,25 @@
     import Treeselect from '@riophae/vue-treeselect';
     import {NestedFilterOption} from "../generated";
 
-    export default Vue.extend({
+    interface Methods {
+        treeselectNormalizer: (anyNode: any) => void;
+        select: (value: string[]) => void
+    }
+
+    interface Computed {
+        treeselectValue: string[] | string | null
+        placeholder: string
+    }
+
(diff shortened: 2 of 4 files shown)


src/app/static/src/app/components/FilterSelect.vue
```diff
@@ -25,24 +26,38 @@
...
+                return this.disabled ? "Not used" : "Select...";
+            }
+        },
         methods: {
             treeselectNormalizer(anyNode: any) {
-                //In the nested case, this gets called for the child nodes we add in below - just return these unchanged
-                if (anyNode.label) {
-                    return anyNode;
-                }
+                const normalize = (anyNode: any) => {
```
EmmaLRussell: This was an odd one - adding any computed property which accessed 'this' seemed to make the treeselectNormalizer method inaccessible on this from itself. Not sure why this should be??

src/app/static/src/app/components/FilterSelect.vue
```diff
@@ -25,24 +26,38 @@
...
-                return result;
+                    return result;
+                };
+
+                return normalize(anyNode);
             },
             select(value: string[]) {
-                this.$emit("select", value);
-            }
+                if (!this.disabled) {
```
EmmaLRussell: Prevent unwanted notifications of selecting null.

src/app/static/src/app/components/ChoroplethFilters.vue
```diff
@@ -100,7 +100,7 @@
             buildViewFiltersForType(availableFilterOptions: FilterOption[],
                                     selectedFilterOption?: FilterOption) {
                 return {
-                    available: availableFilterOptions,
+                    available: availableFilterOptions ? availableFilterOptions : [],
```
EmmaLRussell: Avoids Vue-treeselect warning when reload from later step

src/app/static/src/app/components/FilterSelect.vue
hillalex: Yeah I've noticed this too, seems the Vue type inference is a bit off. You can make it work by typing the component more explicitly:
```
    interface Methods {
        treeselectNormalizer: (anyNode: any) => void;
        select: (value: string[]) => void
    }

    interface Computed {
        treeselectValue: string[] | string | null
        placeholder: string
    }

    interface Props {
        multiple: boolean,
        label: string,
        disabled: boolean,
        options: any[],
        value: string[] | string
    }

    export default Vue.extend<{}, Methods, Computed, Props>({
... etc
```

PR #91, opened 2019-10-10
mrc-555 add typings

hillalex:
We were using `any` types quite liberally (I think a hangover from prototyping) - this PR adds types  so we can make better use of type checking.

There's a bug with Vuex typings (see https://github.com/ffxsam/vue-typescript-cookbook#im-using-vuex-mapstate-or-mapgetters-and-typescript-is-saying-the-mapped-stategetters-dont-exist-on-this) so I've defined some helper methods that wrap `mapState` and `mapGetters` to work around this.


hillalex:
I ended up merging mrc-466 in here because mrc-466 brings the generated types up to date. So this should now be reviewed after https://github.com/mrc-ide/hint/pull/95/files is merged.

src/app/static/src/app/store/filteredData/getters.ts
```diff
@@ -98,13 +94,9 @@ export const getters = {
...
 
-        for(const d of data) {
-            const row = d as any;
+        for (const row of data) {
 
-            if (!includeRowForSelectedChoroplethFilters(row,
-                state.selectedDataType,
-                state.selectedChoroplethFilters,
-                flattenedRegions)) {
+            if (getters.excludeRow(row)) {
```
hillalex: Sorry this was a bit of scope creep, but I thought this made the code neater - making the filter function a getter so we don't have to pass all the state related arguments, and inverting the logic.

PR #95, opened 2019-10-10
mrc-466 save button

hillalex:
Adds a dropdown to the header menu with save (implemented) and load (not yet implemented) links.

### DIFF
--- a/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintApplicationTests.kt
+++ b/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintApplicationTests.kt
@@ -41,7 +41,7 @@ class HintApplicationTests(@Autowired val restTemplate: TestRestTemplate) : Inte
         authorize()
         val rootEntity = restTemplate.getForEntity<String>("/")
         assertThat(rootEntity.statusCode).isEqualTo(HttpStatus.OK)
-        assertThat(rootEntity.body!!).contains("<a href=\"/logout\">Logout</a>")
+        assertThat(rootEntity.body!!).doesNotContain("<title>Login</title>")
     }
 
     @Test
--- a/src/app/static/src/app/components/UserHeader.vue
+++ b/src/app/static/src/app/components/UserHeader.vue
@@ -1,0 +1,25 @@
+<template>
+    <header class="mb-5">
+        <nav class="navbar navbar-dark bg-secondary">
+            <div class="container-fluid">
+                <div class="navbar-header text-light">
+                    {{title}}
+                </div>
+                <div class="logout">
+                    <a href="/logout">Logout</a>
+                </div>
+            </div>
+        </nav>
+    </header>
+</template>
+<script lang="ts">
+
+    import Vue from "vue";
+
+    export default Vue.extend({
+        props: ["title"],
+        created: void {
+            //console.log(this.title)
+        }
+    })
+</script>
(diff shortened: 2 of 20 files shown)


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintApplicationTests.kt
```diff
@@ -41,7 +41,7 @@ class HintApplicationTests(@Autowired val restTemplate: TestRestTemplate) : Inte
         authorize()
         val rootEntity = restTemplate.getForEntity<String>("/")
         assertThat(rootEntity.statusCode).isEqualTo(HttpStatus.OK)
-        assertThat(rootEntity.body!!).contains("<a href=\"/logout\">Logout</a>")
+        assertThat(rootEntity.body!!).doesNotContain("<title>Login</title>")
```
EmmaLRussell: This test could be reverted to previous stricter version, now that 'Logout' is being retained in the menu bar?

src/app/static/src/app/components/UserHeader.vue
```diff
@@ -0,0 +1,92 @@
...
+                this.show = !this.show;
+            },
+            save() {
+                const state = serialiseState(this.$store.state);
+                const hashes = {
+                    ...this.baselineHashes,
+                    ...this.surveyAndProgramHashes
+                };
+                const file = new Blob([JSON.stringify({
+                    state, hashes
```
EmmaLRussell: Why do the hashes need to be added separately? Are they not just part of the state?

hillalex:
> How about red or dark red for the 'Naomi' background? It is a little harder to read as white on light grey...
> 
> Screenshot from 2019-10-11 16-22-19
> 
> Could the File menu be hidden when it loses focus?

I don't think red is a good idea because we are using red for links/buttons. Could up the contrast though.

src/app/static/src/app/components/UserHeader.vue
hillalex: They're not part of the state that gets serialised to localStorage at the moment, which is what `serialiseState` returns, because the files are requested from the server on page load.But even if they were in the serialised state it's going to be more useful to have them in a separate dictionary because that's the schema that will be POSTed to the server on "load" to persists to the db.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintApplicationTests.kt
hillalex: The logout link lives in the vue component, not the raw html, so won't be detectable here!

hillalex:
> > How about red or dark red for the 'Naomi' background? It is a little harder to read as white on light grey...
> > Screenshot from 2019-10-11 16-22-19
> > Could the File menu be hidden when it loses focus?
> 
> I don't think red is a good idea because we are using red for links/buttons. Could up the contrast though.

Ok, have gone with red after all, you're right that it looks better.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintApplicationTests.kt
EmmaLRussell: Fair point!

hillalex:
> Weirdly, the onblur event seems to stop the file from getting saved (in Chrome)! (I took it out locally and it worked again.) Is it blurring as soon as the click happens and the save event doesn't ever happen?
> 
> I think the style change has caused the Logout link to go dark red on hover, which looks a bit strange on the dark grey background. I think it was just staying white before (but gaining underline) which is probably better.

Ugh, sorry for not noticing this, and thanks for re-testing and catching this!

hillalex:
3rd time lucky? https://github.com/mrc-ide/hint/pull/95/commits/0330767f71ec421378f1e9d439f42a3cdc5ef554

PR #96, opened 2019-10-10
wip

hillalex:
Just playing around with the schema - it seems to me like it's gotten a bit tangled, with a mix of hard-coded and inferred info.

My idea is to simplify the metadata schema to avoid hard-coded keys, see https://github.com/mrc-ide/hint/pull/96/files#diff-24c81d0cd2b03472ab3329e20c7dd2fdR175

Then the code itself becomes agnostic, e.g.
https://github.com/mrc-ide/hint/pull/96/files#diff-963d30ddf5ac85423ec0c01d65e1a4b4R114
and 
https://github.com/mrc-ide/hint/pull/96/files#diff-963d30ddf5ac85423ec0c01d65e1a4b4R92

This branch obv won't compile, but just trying to illustrate a way forward. What do you thinK?

### DIFF
--- a/src/app/static/src/app/store/filteredData/getters.ts
+++ b/src/app/static/src/app/store/filteredData/getters.ts
@@ -1,8 +1,9 @@
 import {RootState} from "../../root";
-import {DataType, FilteredDataState, SelectedChoroplethFilters, SelectedFilters} from "./filteredData";
-import {IndicatorRange, Indicators, IndicatorValues} from "../../types";
-import {interpolateCool, interpolateWarm} from "d3-scale-chromatic";
-import {FilterOption, NestedFilterOption} from "../../generated";
+import {DataType, FilteredDataState, SelectedChoroplethFilters} from "./filteredData";
+import {IndicatorRange, IndicatorValues} from "../../types";
+import * as d3 from "d3-scale-chromatic";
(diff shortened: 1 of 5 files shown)


src/app/static/src/app/store/filteredData/getters.ts
```diff
@@ -87,119 +84,52 @@ export const getters = {
...
                 result[areaId] = {};
             }
 
-            const indicators = result[areaId];
-            switch (indicator) {
-                case("prev"):
-                    indicators.prev = {value: value, color: ""};
+            const indicatorsForArea = result[areaId];
+            indicatorMetadata.map((i: Indicator) => {
+                const value = row[i.value_column];
```
hillalex: I'm sure this logic isn't right because I still don't understand what value_column, indicator_column etc mean, but you get the idea!

hillalex: Ok, great to hear we're converging! I think making the schema fully agnostic would be nice because the whole motivation for this approach was that the plots should be able to change without our code ever changing - of course if we just treat it as a dictionary without knowledge of the keys that would also work, although then it feels like the schema is unnecessarily opinionated/complicated.

hillalex:
Closing now as this was just for demo purposes

PR #98, opened 2019-10-14
dynamic form prototype

hillalex:


src/app/static/src/app/components/forms/CustomForm.vue
```diff
@@ -0,0 +1,46 @@
...
+<script lang="ts">
+
+    import Vue from "vue";
+    import {BForm} from "bootstrap-vue";
+    import {formMeta} from "./fakeFormMeta";
+    import ControlGroup from "./ControlGroup.vue";
+    import {Dictionary} from "vuex";
+
+    export default Vue.extend({
+        name: "CustomForm",
```
hillalex: One slightly annoying thing about using real html tags and not jsx is that we can't name our custom components things like "form" because <form> will always be interpreted as the html tag. Any suggestions for a good name for this component?

src/app/static/src/app/components/forms/CustomForm.vue
```diff
@@ -0,0 +1,46 @@
...
+            submit(e: Event) {
+                e.preventDefault();
+                const form = this.$refs[this.id] as HTMLFormElement;
+                const formData = new FormData(form);
+                const data: Dictionary<any> = {};
+                formData.forEach(function (value, key) {
+                    data[key] = value;
+                });
+                const json = JSON.stringify(data);
+                this.$emit("submit", json);
```
hillalex: The idea is that this component can be pulled out into a standalone js package at some point. The parent component that instantiates it (in our case the ModelRun component) can do what it wants with the json string that's returned. Not sure how this will work if we want to trigger the submit on "continue" though :thinking:

hillalex: In fact maybe this should emit JSON rather than a string, more flexible for the caller?

hillalex: Ok in fact this is easy enough: https://github.com/mrc-ide/hint/pull/98/commits/443fd1fbbed0056f4680b8684637b5fbc28d1ea9

src/app/static/src/app/components/forms/CustomForm.vue
EmmaLRussell: Would MetaForm be too meta? DataForm? DynamicForm? Don't know, can't really think of anything that isn't either potentially misleading or too long!

src/app/static/src/app/components/forms/FormControl.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+    <b-col :md="colWidth">
+        <label v-if="control.label">{{control.label}}</label>
```
EmmaLRussell: I wonder if there's some way to avoid having to put this v-if in for every type of control component. Using dynamic components somehow? https://alligator.io/vuejs/dynamic-components/
Could the control.type map exactly to the tag of the control component?
`<component is:="control.type">`?

src/app/static/src/app/components/forms/CustomForm.vue
```diff
@@ -0,0 +1,60 @@
...
+            return {
+                form: formMeta
+            }
+        },
+        components: {
+            BForm,
+            ControlGroup
+        },
+        methods: {
+            submit(e: Event) {
```
EmmaLRussell: So if includeFormButton = false (and so e is falsy), that means that the submit method will be called from outside the form (as is done when integrated into the ModelComponent)? So in that case, it might make more sense to return the data from this method rather than to emit the event?

hillalex: Ooh yes good idea. And no reason for it not to do both, so the user can choose how to consume it.

src/app/static/src/app/components/forms/CustomForm.vue
hillalex: Maybe DynamicForm is the most natural.

hillalex:
> Looks like a good approach to me.
> I guess arguably you might not have completely separate components for select/multi-select but could just parameterise one - but this is a good illustration of the approach.
> 
> (This is currently failing to show choropleth with the master of hintr as the plotting metadata has changed, but that doesn't affect the model run stuff.)

Yeah I sorta did that on purpose because I wanted to use a simple bootstrap select for the single select, rather than the fancier treeselect. But actually it's a bit of a faff to style them exactly the same, so maybe easier to just use treeselect for both.

src/app/static/src/app/components/forms/FormControl.vue
hillalex: Much nicer, thanks for the prompt: https://github.com/mrc-ide/hint/pull/98/files#diff-88e4ce74e07a86408ea587fc3edef44eR4

src/app/static/package.json
```diff
@@ -29,14 +29,15 @@
...
+    "bootstrap-vue": "^2.0.3",
     "d3-scale-chromatic": "^1.5.0",
     "form-data": "^2.5.1",
     "gulp": "^4.0.0",
     "gulp-clean-css": "^4.0.0",
     "gulp-rename": "^1.4.0",
     "gulp-sass": "^4.0.2",
     "ify-loader": "^1.1.0",
-    "jest": "^24.8.0",
+    "jest": "^25.0.0",
```
hillalex: I had to updated jest to a beta version because of a bug with jsdom that only just got fixed. Seems to be working ok.

src/app/static/src/app/components/forms/DynamicFormSelect.vue
```diff
@@ -0,0 +1,28 @@
+<template>
```
hillalex: I kept this separate from the multiselect because it seemed like overkill to add the treeselect widget when a normal html5 select would do. I have done some work to style them the same though.

src/app/static/src/app/components/forms/types.ts
```diff
@@ -0,0 +1,120 @@
...
+    min?: number
+    max?: number
+    default?: number
+}
+
+export interface DynamicForm {
+    controlSections: DynamicControlSection[]
+}
+
+export const formMeta: DynamicForm = {
```
hillalex: This is here for testing, can be removed once we're fetching real data.

src/app/static/src/scss/partials/form.scss
```diff
@@ -0,0 +1,88 @@
+.dynamic-form {
```
hillalex: These style rules are mostly targeted at the select/multiselect, to make their styles consistent (and consistently Bootstrap-y). I've defined all the styles here to apply only within the dynamic form but actually probably make sense to extend them to the whole app though?

EmmaLRussell: Almost certainly! Probably worth doing a style audit at some point to make sure that things like the forgot password form etc aren't getting left behind.

hillalex: Good idea, added a ticket https://vimc.myjetbrains.com/youtrack/issue/mrc-603

src/app/static/src/app/components/forms/DynamicForm.vue
```diff
@@ -0,0 +1,101 @@
...
+
+                return requiredDict;
+            }
+        },
+        methods: {
+            submit(e: Event) {
+                if (e) {
+                    e.preventDefault();
+                }
+                const form = this.$refs[this.id] as HTMLFormElement;
```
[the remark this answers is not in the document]
hillalex: I thought it would be nice to actually return the list of inputs which are missing, and I couldn't think of a better way of doing that with built in APIs?

src/app/static/src/app/components/forms/DynamicFormMultiSelect.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+    <div :class="cssClass">
+        <tree-select :multiple="true"
+                     :clearable="!formControl.required"
```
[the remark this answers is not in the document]
hillalex: Good point. I've decoupled them and set clearable to false because the cross icon gets in the way of the dropdown icon. There seems to be a bug with the treeselect hidden input - it gets updated to the most recent selected value rather than a comma separated list of all values - hence managing a hidden input manually, so the required attribute won't do us any good here.

src/app/static/src/app/components/forms/DynamicForm.vue
```diff
@@ -0,0 +1,101 @@
...
+        name: "DynamicForm",
+        data() {
+            return {
+                validated: false
+            }
+        },
+        props: {
+            id: {
+                type: String,
+                default: Math.random().toString(36).slice(-5)
```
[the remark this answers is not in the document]
hillalex: Ah I just assumed it corresponded to an actual html5 id but I think you're right.

PR #101, opened 2019-10-15
mrc-514 New endpoint to set all session files on load

EmmaLRussell:


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/SessionTests.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.http.HttpEntity
+import org.springframework.http.HttpHeaders
+import org.springframework.http.HttpStatus
+import org.springframework.http.MediaType
+
+class SessionTests : SecureIntegrationTests() {
+    @ParameterizedTest
+    @EnumSource(IsAuthorized::class)
+    fun `can set session files`(isAuthorized: IsAuthorized) {
+        //Get a valid hash
```
hillalex: So at the moment an invalid hash will throw a foreign key error? Might be a good idea to throw a custom 400 error with a user friendly message in case this does happen, not quite sure what, but indicating to the user that they've uploaded a corrupted file?

PR #103, opened 2019-10-16
Mrc 597 Include filenames in saved state file

EmmaLRussell:


src/app/static/src/tests/components/userHeader.test.ts
```diff
@@ -80,25 +81,33 @@ describe("user header", () => {
...
-            state: {modelRun: mockModelRunState()}, hashes: {
-                pjnz: "2.csv",
-                population: "1.csv",
-                shape: "3.csv",
-                survey: "4.csv",
-                program: "5.csv",
-                anc: "6.csv"
-            }
-        })], {type: "application/json"});
-        expect((window.URL.createObjectURL as jest.Mock).mock.calls[0][0]).toStrictEqual(expectedBlob);
```
EmmaLRussell: This test continued to pass after i'd changed the output, so was obviously  not really comparing the output with expected json. I think because reading from a Blob is async, if you try to look at two Blobs without async, it thinks they're both empty, so the comparison was passing.

PR #104, opened 2019-10-16
mrc-609 add model options step

hillalex:
Small PR to replace the current "Review inputs" step with a "Model options" step. At the moment it uses the fake data to create the form, and just sets a "valid" flag on the state when the form is successfully submitted.


hillalex:
> Looks good to me! Nice to get rid of the 'Coming soon..' gap...
> You're going to do the actual saving of options to the state in another PR? At the moment they disappear when you step back/forward.
> Would it be worth putting in a required signifier (*) for required fields?

Yeah, there's another ticket for actually persisting the options. Wanted to keep PRs small. required signifier makes sense too. I'll open a separate ticket since that's conceptually different from this.

PR #105, opened 2019-10-16
Mrc 515 Load: Set session files from saved state

EmmaLRussell:
This PR adds a load module and setsFiles in the backend when user selects a file to load.
Getting file contents and overwriting store state will be in a separate PR.


src/app/static/src/app/store/load/load.ts
```diff
@@ -0,0 +1,25 @@
+import {actions} from "./actions";
+import {mutations} from "./mutations";
+import {Module} from "vuex";
+import {RootState} from "../../root";
+
+export enum LoadingState { NotLoading, SettingFiles, UpdatingState, LoadFailed }
```
EmmaLRussell: We might not actually need to keep track of all these states for the purposes of the front end, but it's been helpful for me to keep track of what's happening!

src/app/static/src/tests/components/userHeader.test.ts
```diff
@@ -108,4 +124,108 @@ describe("user header", () => {
...
+                })
+            });
+
+        const input = wrapper.find("input");
+
+        const vm = wrapper.vm;
+        const testFile = new File(["test file contents"], "filename");
+        //Can't programmatically construct a FileList to give to the read rendered input element, so we need to trick
+        //the component with a mocked ref
+        (vm.$refs as any).loadFile = {
```
hillalex: Ooh this is clever! I wonder if we can use this in the FileUpload tests to fix this: https://github.com/mrc-ide/hint/blob/master/src/app/static/src/tests/components/fileUpload.test.ts#L99

src/app/static/src/tests/components/userHeader.test.ts
```diff
@@ -108,4 +124,108 @@ describe("user header", () => {
...
+        const wrapper = shallowMount(UserHeader,
+            {
+                store: createStore()
+            });
+
+        expect(wrapper.find(Modal).attributes("open")).toBeFalsy();
+    });
+
+    it("opens modal if load error", () => {
+        const wrapper = shallowMount(UserHeader,
```
hillalex: Can pull the wrapper creation out into a shared function to be re-used between this and the test above?

src/app/static/src/app/store/load/load.ts
hillalex: might be useful for showing progress/loading spinners or something

EmmaLRussell:
> Small suggestion for test readability. This all looks good, is there a separate ticket for adding user feedback? Probably want a loading spinner similar to the one seen on page load?

Good idea, I'll get the spinner to show when the load state changes.

src/app/static/src/tests/components/userHeader.test.ts
EmmaLRussell: There's already a createStore method, I'll use that and just poke in the load module bits I need, should be much terser.

src/app/static/src/tests/components/fileUpload.test.ts
```diff
@@ -93,22 +93,24 @@ describe("File upload component", () => {
...
 
         const uploader = jest.fn();
         const wrapper = createSut({
             upload: uploader
         });
 
-        (wrapper.find("input").element as HTMLInputElement).files = mockFileList("TEST");
+        const testFile = mockFile("TEST FILE NAME", "TEST CONTENTS");
+        const vm = wrapper.vm;
+        (vm.$refs as any).pjnz = {
```
hillalex: Ah, thanks for updating this! :tada:

PR #107, opened 2019-10-17
mrc-606 hash content and add to downloaded file

hillalex:
Adds 2 helper methods, one to add a hash to downloadable content and one to verify such content.


src/app/static/src/app/utils.ts
```diff
@@ -29,4 +30,19 @@ export const mapGettersByNames = <K extends string>(namespace: string, names: st
...
+    return JSON.stringify([hash.toString(CryptoJS.enc.Base64), data]);
+};
+
+export const verifyDownloadableContent = (content: string): false | any => {
+    const result = JSON.parse(content);
+    const hash = result[0];
+    const data = result[1];
+    const valid = CryptoJS.MD5(data).toString(CryptoJS.enc.Base64) === hash;
+
+    return valid && JSON.parse(data);
```
hillalex: Made this so it returns false if invalid, otherwise returns the original object. Seem like the best behaviour?

Also I don't think either of these function names are great, better ideas?

EmmaLRussell: Agreed re return values. 

addChecksum / verifyChecksum ?

PR #109, opened 2019-10-18
accept map instead of typed params

hillalex:
* changes the model run interface over to pass a map instead of a strongly typed object to hint
* updates the dynamic form so that multiselect values are returned as an array rather than a comma separated string
* regenerated types from the hintr schema


src/app/static/src/tests/components/forms/dynamicForm.test.ts
```diff
@@ -32,12 +34,30 @@ describe('Dynamic form component', function () {
...
+        } as SelectControl);
+
     it("renders form with id", () => {
         const rendered = mount(DynamicFormComponent, {
             propsData: {
                 formMeta: fakeFormMeta,
                 id: "test-id"
-            }
+            },
+            sync: false
```
hillalex: This `sync: false` everywhere is due to a bug in vue test-utils. It seems to be triggered when doing a full mount where a child component has a watcher/computed property, so only surfaced in this test suite once I added the multiselect. Bizarely you have to add "sync: false" to every mount or shallow mount in the suite to stop it happening, even if it only affects one test in the suite. https://github.com/vuejs/vue-test-utils/issues/1130

src/app/static/src/tests/filteredData/getters.test.ts
```diff
@@ -137,7 +132,7 @@ describe("FilteredData getters", () => {
         const testFilters = {
             age: [{id: "age1", name: "0-4"}, {id: "age2", name: "5-9"}],
             quarter: [{id: "1", name: "2019 Q1"}],
-            indicator: []
+            indicators: []
```
EmmaLRussell: We probably ought to decide on a plural/singular policy for filter options, and stick to it! If it's 'indicators' then shouldn't it also be 'quarters' and 'ages'...?

EmmaLRussell: We probably ought to decide on a plural/singular policy for filter options, and stick to it! If it's 'indicators' then shouldn't it also be 'quarters' and 'ages'...?

src/app/static/src/tests/components/forms/dynamicForm.test.ts
EmmaLRussell: Weird! Candidate for pulling that shallowMount out into a common method, maybe with partialPropsData as a parameter (and include `formMeta: fakeFormMeta` in the props in the method, as I think that's in all the instances?

src/app/static/src/tests/filteredData/getters.test.ts
hillalex: Yep, agreed. Probably singular if these names end up getting reflected in the filter headers. Want to raise a ticket for this? Should be done in hintr.

EmmaLRussell: https://vimc.myjetbrains.com/youtrack/issue/mrc-626

PR #113, opened 2019-10-22
mrc-610 rewrite form to use 2 way data binding

hillalex:


src/app/static/src/app/components/forms/DynamicForm.vue
```diff
@@ -24,100 +30,85 @@
...
+                return controls;
+            },
+            disabled() {
+                return this.controls
+                    .filter(c => c.required && (c.value == null || c.value == ""))
+                    .length > 0
             }
         },
         methods: {
-            validate(formData: FormData) {
```
hillalex: Rather than "validating" on submit, which was actually just checking that required values as present, we now simply disable the submit button until required values are present. When the button is enabled and clicked, the form data is returned.

src/app/static/src/app/components/forms/DynamicFormControl.vue
```diff
@@ -1,8 +1,10 @@
 <template>
     <b-col :md="colWidth">
-        <label v-if="formControl.label">{{formControl.label}}</label>
+        <label v-if="formControl.label">{{formControl.label}}
+            <span v-if="formControl.required">*</span>
```
hillalex: I also added these required asterisks.

src/app/static/src/app/components/forms/types.ts
```diff
@@ -47,7 +57,7 @@ export const formMeta: DynamicFormMeta = {
                         name: "area_scope",
                         type: "multiselect",
                         options: [{id: "MWI", label: "Malawi"}, {id: "MWI.1", label: "Central"}],
-                        default: "MWI",
```
hillalex: Have replaced "default" with "value", which is always set after the form is rendered, but can optionally be passed in the original metadata if a default is wanted.

src/app/static/src/app/components/modelOptions/ModelOptions.vue
```diff
@@ -1,28 +1,34 @@
...
+                    return this.$store.state.modelOptions.optionsFormMeta
+                },
+                set (value: DynamicFormMeta) {
+                    this.update(value);
+                }
+            }
         },
         methods: {
-            ...mapActionsByNames<keyof Methods>("modelOptions", ["validated"])
+            ...mapMutationsByNames<keyof Methods>("modelOptions", ["validate", "update"])
```
hillalex: Mapping the mutations directly here, but once we're validating from the server will want to introduce a "validate" action.

src/app/static/src/tests/components/forms/dynamicForm.test.ts
```diff
@@ -70,85 +76,80 @@ describe('Dynamic form component', function () {
...
+                            name: "id_1",
+                            type: "number",
+                            required: false
+                        } as NumberControl
+                    ]
+                }]
+            }]
+        };
+        const rendered = getWrapper(formMeta, mount);
+        expect(rendered.find("button").attributes("disabled")).toBeUndefined();
```
hillalex: This is driving me crazy - why is this test failing??? Have run up the app and confirmed that the disabled/enabled state works as expected. But this test and the next one are failing. Any ideas?

hillalex:
Ok, absolutely no mutating of props now :tada: 

I've changed the styling for required inputs to make it more explicit and I've styled the button so it's grey while disabled, making the change to enabled more obvious. I've not added a tick or moved it near the continue link, just because the continue link is probably going to move and more broadly we have an open ticket for styling around the stepper and how to clearly indicate when a step becomes enabled.

src/app/static/src/app/components/forms/DynamicFormControlSection.vue
```diff
@@ -4,22 +4,36 @@
...
 </template>
 
 <script lang="ts">
 
     import Vue from "vue";
     import DynamicFormControlGroup from "./DynamicFormControlGroup.vue";
+    import {DynamicControlGroup, DynamicControlSection} from "./types";
 
-    export default Vue.extend({
+    export default Vue.extend<{}, { change: (newVal: DynamicControlGroup, index: number) => void },
```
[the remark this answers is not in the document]
hillalex: Yeah just laziness on my part! Will change.

src/app/static/src/app/components/forms/DynamicFormMultiSelect.vue
```diff
@@ -1,37 +1,42 @@
 <template>
-    <div :class="cssClass">
+    <div>
         <tree-select :multiple="true"
                      :clearable="false"
                      v-model="value"
                      :options="formControl.options"></tree-select>
-        <input type="hidden" :value="value" :name="formControl.name"/>
+        <input type="hidden" :value="formControl.value" :name="formControl.name"/>
```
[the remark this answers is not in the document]
hillalex: Tbh it's not needed now. Previously it was because the html form was the source of truth for the data, so we needed to keep track of the value in an input. I left it in because I thought maybe it's nice that you could still inspect the form element and see all the values there if you wanted to, or indeed create a FormData object from it.

hillalex:
Ready for review now

PR #115, opened 2019-10-23
mrc-641 Add tabs to ModelOutput component

EmmaLRussell:


hillalex:
> I've made the tab go across the full width of the container as each output visualisation will have its own 'local' filters unlike in the SurveyAndProgram view, where filters are shared between data types when you change tabs.

Yep, that makes sense. It's missing some padding around the edges though, just need to add a `<div class="col">` between the `<div class="row">` and the `<ul>` https://github.com/mrc-ide/hint/pull/115/files#diff-c28a53e46b9489acf62eed0941f347ccR4

Once that's done feel free to merge.

PR #116, opened 2019-10-24
mrc-647 Add modelOutput module with getter for barchart plotting metadata

EmmaLRussell:
This adds a modelOutput module in order to access the plotting metadata for the barchart (currently hardcoded, but will come from the modelRunResult response). This module doesn't actually have any state of its own, so maybe it doesn't deserve to be a separate module, but it seemed like it belonged separately from the model run stuff...

### DIFF
--- a/src/app/static/src/app/types.ts
+++ b/src/app/static/src/app/types.ts
@@ -1,4 +1,5 @@
 import {Payload} from "vuex";
+import {FilterOption} from "./generated";
 
 export interface PayloadWithType<T> extends Payload {
     payload: T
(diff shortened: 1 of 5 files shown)


src/app/static/src/app/types.ts
```diff
@@ -27,4 +28,21 @@ export type IndicatorValuesDict= Dict<IndicatorValues>;
...
+    indicator_column: string,
+    indicator_value: string,
+    name: string,
+    error_low_column: string,
+    error_high_column: string
+}
+
+export interface Filter {
+    id: string,
+    column_id: string,
```
EmmaLRussell: column_id is here because, as things stand with the filters returned in modelRunResults, their identifiers (e.g. 'age') don't actually correspond to the columns in which data is found, so we'll need both id (to find the filter options) and column_id (to find the data). This could be changed to avoid need for both, but would be a breaking change for the choropleth filters.

PR #118, opened 2019-10-25
Mrc 642 Barchart component

EmmaLRussell:
Contains changes from https://github.com/mrc-ide/hint/pull/123 which should be merged first

I've left the utils file with the getProcessedOutputData in this PR as it makes sense to look at that together with the Barchart. That does make it looks like a very large diff, but a lot of that is color configuration!

### DIFF
--- a/src/app/static/src/app/components/modelOutput/ModelOutput.vue
+++ b/src/app/static/src/app/components/modelOutput/ModelOutput.vue
@@ -17,7 +17,7 @@
                 <choropleth></choropleth>
             </div>
 
-            <div class="col-md-12" :display="selectedTab==='Bar' ? '' : 'none'">
+            <div id="barchart-container" :class="selectedTab==='Bar' ? 'col-md-12' : 'd-none'">
                 <barchart :chartdata="chartdata" :filters="barchartFilters" :indicators="barchartIndicators"></barchart>
             </div>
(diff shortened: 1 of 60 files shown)


hillalex:
Do you want to merge master in here and resolve the conflicts, then I'll review?

EmmaLRussell:
> Do you want to merge master in here and resolve the conflicts, then I'll review?

Should be ready to review now..

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -17,8 +17,8 @@
                 <choropleth></choropleth>
             </div>
 
-            <div class="col" v-if="selectedTab==='Bar'">
-                Bar chart coming soon!
+            <div id="barchart-container" :class="selectedTab==='Bar' ? 'col-md-12' : 'd-none'">
```
hillalex: So this was because we don't want the filters to reset when the user tabs away right? Seems fine as a first pass, we might want to think about a way of passing selected filter options as props instead so we can persist them.

EmmaLRussell: Yeah, we talked about doing something like that. I'll add a ticket to persist barchart filters.

PR #119, opened 2019-10-25
mrc-602 performance

hillalex:
First PR:

* freezes big objects coming from the API
* moves region options and flattened region options into the state

Also less importantly but makes some minor difference:
* storing selected filters as simple strings or arrays of strings, so we just pass around ids rather than large nested  objects

General principles at work here:
- making large immutable objects immutable so vue doesn't bloat them out by making everything observable
- not using computed properties for things that are not actually dynamic, i.e. will not be re-computed, as computeds have performance overhead associated with them

### DIFF
--- a/src/app/static/src/app/store/baseline/mutations.ts
+++ b/src/app/static/src/app/store/baseline/mutations.ts
@@ -27,7 +27,7 @@ export const mutations: MutationTree<BaselineState> & BaselineMutations = {
         const data = action.payload;
         if (data) {
             state.country = data.data.country;
-            state.pjnz = Object.freeze(data);
+            state.pjnz = data;
         } else {
             state.country = "";
             state.pjnz = null;
@@ -38,7 +38,7 @@ export const mutations: MutationTree<BaselineState> & BaselineMutations = {
     ShapeUpdated(state: BaselineState, action: PayloadWithType<ShapeResponse>) {
         state.shape = Object.freeze(action.payload);
         if (action.payload && action.payload.filters.regions){
-            state.regionFilters = Object.freeze(action.payload.filters.regions.children) as NestedFilterOption[];
+            state.regionFilters = action.payload.filters.regions.children as NestedFilterOption[];
             state.flattenedRegionFilters = Object.freeze(flattenOptions(state.regionFilters));
         }
         state.shapeError = "";
@@ -54,7 +54,7 @@ export const mutations: MutationTree<BaselineState> & BaselineMutations = {
     },
 
     PopulationUploadError(state: BaselineState, action: PayloadWithType<string>) {
-        state.populationError = Object.freeze(action.payload);
+        state.populationError = action.payload;
     },
 
     ...readyStateMutations
--- a/src/app/static/src/app/store/baseline/mutations.ts
+++ b/src/app/static/src/app/store/baseline/mutations.ts
@@ -1,8 +1,9 @@
 import {Mutation, MutationTree} from 'vuex';
 import {BaselineState} from "./baseline";
-import {PjnzResponse, PopulationResponse, ShapeResponse} from "../../generated";
+import {NestedFilterOption, PjnzResponse, PopulationResponse, ShapeResponse} from "../../generated";
 import {PayloadWithType} from "../../types";
 import {readyStateMutations} from "../shared/readyStateMutations";
+import {flattenOptions} from "../filteredData/utils";
 
 type BaselineMutation = Mutation<BaselineState>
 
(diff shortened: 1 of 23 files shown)


hillalex:
(Not this branch is `mrc-602_freeze`)

src/app/static/src/app/store/baseline/mutations.ts
```diff
@@ -26,7 +27,7 @@ export const mutations: MutationTree<BaselineState> & BaselineMutations = {
         const data = action.payload;
         if (data) {
             state.country = data.data.country;
-            state.pjnz = data;
+            state.pjnz = Object.freeze(data);
```
hillalex: A more radical re-architecture would be to keep this big objects in our own local cache, outside of vue, and just keep an index over them in the store. I feel like that might be slightly nicer, but it would be a more sizable refactor, so I think just something to consider at a later date.

src/app/static/src/app/store/filteredData/utils.ts
```diff
@@ -85,3 +86,27 @@ const flattenOption = (filterOption: NestedFilterOption): NestedFilterOption =>
     }
     return result;
 };
+
+export const getIdsAndChildrenIds = (ids: string[], lookup: Dict<NestedFilterOption>): Set<string> => {
```
hillalex: These 2 functions aren't very well named, any better ideas?

src/app/static/src/app/store/baseline/mutations.ts
EmmaLRussell: Agreed. It's probably worth using this frozen object technique for a bit to see what the use cases are, to find a reusable pattern for using frozen state with dynamic state

EmmaLRussell:
>storing selected filters as simple strings or arrays of strings, so we just pass around ids rather than large nested objects

Was that actually having a performance impact? Would just be passing by reference. I suppose if you're only ever going to need the ids then you might as well pass around as strings, but if the labels are needed to you might want to keep them together.

src/app/static/src/app/store/baseline/mutations.ts
```diff
@@ -49,7 +54,7 @@ export const mutations: MutationTree<BaselineState> & BaselineMutations = {
     },
 
     PopulationUploadError(state: BaselineState, action: PayloadWithType<string>) {
-        state.populationError = action.payload;
+        state.populationError = Object.freeze(action.payload);
```
EmmaLRussell: Is it the case that we actually want to freeze anything coming back from the API (it's not going to change until we re-request). Could the apiService return objects pre-frozen so that all the mutations don't have to freeze the payloads? 

Would we still need to remember to freeze parts of object (like filters.regions.children..)? Freeze just freezes the top layer of the object? I guess we could implement a recursive freeze?

src/app/static/src/app/store/filteredData/utils.ts
EmmaLRussell: flattenToIdSet, flattenToIdArray?

src/app/static/src/app/store/baseline/mutations.ts
hillalex: Yeah, also not totally sure how this will translate when making components reusable as this relies on the object being passed to the component being frozen already.

hillalex:
> > storing selected filters as simple strings or arrays of strings, so we just pass around ids rather than large nested objects
> 
> Was that actually having a performance impact? Would just be passing by reference. I suppose if you're only ever going to need the ids then you might as well pass around as strings, but if the labels are needed to you might want to keep them together.

No that wasn't, though it did make the code more verbose than it needed to be and involved unneccessary array iterations (albeit small ones) to look up the labels even though they aren't needed.

src/app/static/src/app/store/baseline/mutations.ts
hillalex: Good idea; yes it would have to be recursive. Will add.

hillalex:
I called this part 1 but actually there may not be a part 2. This is certainly the lion's share of the optimisation, need to run some more checks to see if the others things I tried actually made a difference or not.

src/app/static/src/app/utils.ts
```diff
@@ -54,3 +54,21 @@ export const verifyCheckSum = (content: string): false | any => {
 
     return valid && JSON.parse(data);
 };
+
+export const freezer = {
+
+    deepFreeze : (data: any): any => {
+        if (Array.isArray(data)) {
```
EmmaLRussell: Will this not do a shallow freeze on the array items, when we probably need to deep freeze those too? (e.g. gejson features, I think).

hillalex: Oh yeah, you're right, I misunderstood the documentation on this.

PR #120, opened 2019-10-25
mrc-602 performance part 2

hillalex:
This PR moves the rest of the filtering logic out of the getters. It means we lose a bit of the "magic" in the choropleth and have to manually call an update function every time a filter changes, but we make a big gain in performance by removing expensive watchers on large objects.

Sorry this is such a massive changeset, but it's mostly just moving code around! The important files to look at are `Choropleth.vue` and `choropleth.test.ts`, as those are the places where logic has changed.


src/app/static/src/app/components/plots/Choropleth.vue
```diff
@@ -90,14 +86,9 @@
             ),
             ...mapStateProps<FilteredDataState, keyof FilteredDataComputed>("filteredData", {
                     selectedDataType: state => state.selectedDataType,
-                    selectedRegions: state => state.selectedChoroplethFilters.regions || []
+                    selectedFilters: state => state.selectedChoroplethFilters
                 }
             ),
-            ...mapGettersByNames<keyof FilteredDataGetters>("filteredData", [
-                    "regionIndicators",
-                    "colorFunctions"
```
hillalex: The computed `regionIndicators`  has gone, we now manually update region colours when anything changes. This `colorFunctions` wasn't being used anywhere.

src/app/static/src/app/components/plots/Choropleth.vue
```diff
@@ -120,10 +110,11 @@
                 });
 
                 this.features.forEach((feature: Feature) => {
-                    const areas = feature.properties!!["area_id"].split(".");
+                    const id = feature.properties!!["area_id"];
+                    const areas = id.split(".");
                     const adminLevel = areas.length - 1;  //Country (e.g. "MWI") is level 0
                     if (result[adminLevel]) {
-                        result[adminLevel].push(feature);
+                        result[adminLevel].push({id, feature});
```
hillalex: This small change actually shaved quite a few milliseconds off.

src/app/static/src/app/components/plots/Choropleth.vue
```diff
@@ -132,83 +123,40 @@
...
+                if (this.selectedFilters.regions.length > 0) {
+                    return this.selectedFilters.regions.map((id: string) => this.getFeatureFromAreaId(id)!!);
                 }
-                return [];
+                return [this.countryFeature];
             }
         },
         data(): Data {
             return {
-                style: {
```
hillalex: Before I figured the proper bottlenecks out and was still just shaving off milliseconds here and there I moved the static bits of the style into the css. I don't think it made any difference to performance but it's a bit neater so I left it that way.

src/app/static/src/app/components/plots/choroplethData.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    if (selectedRegions.size > 0 && !selectedRegions.has(row.area_id)) {
+        return true
+    }
+
+    return false;
+};
+
+export const getRegionIndicators = (state: RootState, metadata: IndicatorMetadata): IndicatorValuesDict => {
```
hillalex: The one thing that has changed here is from iterating over all indicators each time and returning a map of objects like
`{ "area1": { "art": {value: 0.2, color: ""}, "prev": { value: 0.1, color: "" }}`
to taking a single indicator as an argument and returning a simpler dictionary
`{ "area1": {value: 0.2, color: ""}`

src/app/static/src/app/store/filteredData/getters.ts
```diff
@@ -45,90 +45,5 @@ export const getters = {
             default:
                 return null;
         }
-    },
```
hillalex: All the logic here has moved into `choroplethData.ts`, almost unchanged, see one comment there.

src/app/static/src/app/store/filteredData/utils.ts
```diff
@@ -46,22 +44,6 @@ export const getColor = (value: number, metadata: IndicatorMetadata) => {
     return colorFunction(colorValue);
 };
 
-export const getUnfilteredData = (state: FilteredDataState, rootState: RootState) => {
```
hillalex: This has moved into `choroplethData.ts`

src/app/static/src/tests/components/plots/choroplethData/ancIndicators.test.ts
```diff
@@ -0,0 +1,104 @@
+import {
```
hillalex: This file and all of the files in this folder contain test logic that was previously in `getters.test.ts` and `getter_regionIndicators.test.ts`. I broke the latter up into separate files so that I could get my head around it a bit better, and the setup has been simplified due to the functions no longer being getters, but nothing really new here.

PR #121, opened 2019-10-28
mrc-646 pass model options to model run

hillalex:
Just takes the model options saved in the previous step and passes them to the model run.

### DIFF
--- a/src/app/static/src/app/components/modelRun/ModelRun.vue
+++ b/src/app/static/src/app/components/modelRun/ModelRun.vue
@@ -1,7 +1,7 @@
 <template>
     <div>
         <button class="btn btn-red btn-lg"
-                v-on:click="runModelWithParams"
+                v-on:click="run"
                 :disabled="running">Run model
         </button>
         <h4 v-if="success" class="mt-3" id="model-run-complete">Model run complete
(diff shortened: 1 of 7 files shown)


src/app/static/src/app/components/modelRun/ModelRun.vue
```diff
@@ -64,7 +63,7 @@
             }
         },
         created() {
-            if (this.runId) {
+            if (this.runId && this.pollId == -1) {
```
hillalex: Just noticed a bug here, was creating new polling interval even if one was already running.

src/app/static/src/app/store/modelRun/mutations.ts
```diff
@@ -19,6 +19,7 @@ export const mutations: MutationTree<ModelRunState> & ModelRunMutations = {
     ModelRunStarted(state: ModelRunState, action: PayloadWithType<ModelSubmitResponse>) {
         state.modelRunId = action.payload.id;
         state.status = {id: action.payload.id} as ModelStatusResponse;
+        state.errors = [];
```
hillalex: another unrelated bug, now under test

PR #122, opened 2019-10-28
mrc-650 Use iso3 to fetch plotting metadata instead of country name

EmmaLRussell:
Front end changes to get iso3 as well as country name from PJNZ response and send that to the plotting metadata endpoint. 

The backend behaviour hasn't changed, but I renamed country to iso3 so as not to be misleading.

### DIFF
--- a/src/app/static/src/tests/baseline/mutations.test.ts
+++ b/src/app/static/src/tests/baseline/mutations.test.ts
@@ -57,6 +57,16 @@ describe("Baseline mutations", () => {
         expect(testState.pjnzError).toBe("Some error");
     });
 
+    it("sets country and filename and clears error if present on PJNZUpdated", () => {
+        const testState = {...initialBaselineState, pjnzError: "test"};
+        mutations.PJNZUpdated(testState, {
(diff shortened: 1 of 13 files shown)


src/app/static/src/tests/baseline/mutations.test.ts
```diff
@@ -56,23 +57,13 @@ describe("Baseline mutations", () => {
         expect(testState.pjnzError).toBe("Some error");
     });
 
-    it("sets country and filename and clears error if present on PJNZUpdated", () => {
-
-        const testState = {...initialBaselineState, pjnzError: "test"};
```
hillalex: I'm guessing you deleted this because it's a duplicate of https://github.com/mrc-ide/hint/pull/122/files#diff-5fa3e0f1539bd0a346ad87323763551fL9 ? But looks like this one tests that the `pjnzError` gets cleared, whereas the other one doesn't.

EmmaLRussell: Ah yeah, I just saw that it was testing the same properties, but of course this is with an error already present...

PR #123, opened 2019-10-28
mrc-642_prep Barchart integration prep

EmmaLRussell:
This PR contains the preliminary work for barchart integration. 

It does not contain the Barchart component itself, or changes to the ModelOutput component, but does contain the FilterSelect and chartJsBar components. So no actual apploication behaviour is changed.

### DIFF
--- a/src/app/static/src/tests/components/plots/barchart/chartjsBar.test.ts
+++ b/src/app/static/src/tests/components/plots/barchart/chartjsBar.test.ts
@@ -27,7 +27,6 @@ describe("chartjsBar component", () => {
     it("renders as expected", () => {
         const wrapper = shallowMount(ChartjsBar, {propsData, localVue});
         expect(wrapper.findAll("canvas").length).toEqual(1);
-        console.log(wrapper.html());
     });
 
     it("updates render when props chartdata changes", () => {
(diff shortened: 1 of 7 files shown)


src/app/static/src/tests/components/plots/barchart/chartjsBar.test.ts
```diff
@@ -0,0 +1,52 @@
...
+                    }
+                }
+            ]
+        }
+    };
+
+    it("renders as expected", () => {
+        const wrapper = shallowMount(ChartjsBar, {propsData, localVue});
+        expect(wrapper.findAll("canvas").length).toEqual(1);
+        console.log(wrapper.html());
```
hillalex: rogue console log?

src/app/static/src/tests/components/plots/barchart/chartjsBar.test.ts
```diff
@@ -0,0 +1,52 @@
...
+        const newChartData = {
+            ...propsData.chartdata,
+            labels: ["group1"]
+        };
+        wrapper.setProps({
+            ...propsData,
+            chartdata: newChartData
+        });
+
+        expect(mockUpdate.mock.calls.length).toBe(1);
```
hillalex: Might be good to test that this is called with the right arguments as well (including the axis labels that come from the props)

src/app/static/src/tests/components/plots/barchart/filterSelect.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        expect(vm.isXAxisOrDisagg).toEqual(false);
+    });
+
+    it("computes selectedValues", () => {
+        const wrapper = getWrapper();
+        const vm = (wrapper as any).vm;
+        expect(vm.selectedValues).toStrictEqual(["fo2"]);
+    });
+
+    it("computes badge", () => {
```
hillalex: What about if it's both?

EmmaLRussell: So at the moment, the x-axis overrides disagg, so it will just say 'xaxis' and it won't show any disagg column. I think that's probably a reasonable behavior, since you wouldn't see any disagg groupings if you'd selected the same column to be both. Does that sound ok?
Will add a test for that too.

src/app/static/src/tests/components/plots/barchart/chartjsBar.test.ts
EmmaLRussell: updateRender doesn't have any parameters, but I'll add a test mocking the inherited method (renderChart) and check that gets called with expected data by updateRender

src/app/static/src/tests/components/plots/barchart/filterSelect.test.ts
hillalex: cool, that makes sense to me :+1:

PR #125, opened 2019-10-28
mrc-615 persist model options in local storage

hillalex:


src/app/static/src/app/store/modelRun/modelRun.ts
```diff
@@ -13,8 +13,6 @@ export interface ModelRunState extends ReadyState {
     result: ModelResultResponse | null
 }
 
-export const localStorageKey = "modelRun";
```
hillalex: Just noticed this superfluous const.

EmmaLRussell: Tests are failing now, looks like mutations is looking for that const.

hillalex: Sorry, should have waited for build to pass before submitting. Passing now.

PR #127, opened 2019-10-29
mrc-653 Use barchart metadata types generated from hintr. 

EmmaLRussell:
This just changes hint to use BarchartIndicator and Filter classes as defined in hintr (https://github.com/mrc-ide/hintr/pull/58) rather than defining them itself. Barchart Metadata still not actually coming from hintr yet.

### DIFF
--- a/src/app/static/src/app/components/modelOutput/ModelOutput.vue
+++ b/src/app/static/src/app/components/modelOutput/ModelOutput.vue
@@ -31,11 +31,9 @@
     import ChoroplethFilters from "../plots/ChoroplethFilters.vue";
     import Barchart from "../plots/barchart/Barchart.vue";
     import {DataType} from "../../store/filteredData/filteredData";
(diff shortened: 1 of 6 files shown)


hillalex:
I've gone with the opposite apporach with the form - sticking with locally defined types, with the idea that when we want to factor out these components into a library they can't be tied to hintr. We can check that hintr responses conform to the types using integration tests.

EmmaLRussell:
> I've gone with the opposite apporach with the form - sticking with locally defined types, with the idea that when we want to factor out these components into a library they can't be tied to hintr. We can check that hintr responses conform to the types using integration tests.

Ok ok, probably just don't need this ticket then?

hillalex:
> > I've gone with the opposite apporach with the form - sticking with locally defined types, with the idea that when we want to factor out these components into a library they can't be tied to hintr. We can check that hintr responses conform to the types using integration tests.
> 
> Ok ok, probably just don't need this ticket then?

If you agree that this makes sense then no, I would just close this without merging.

PR #128, opened 2019-10-29
mrc-598 Add download endpoints to hint

EmmaLRussell:


src/config/hintr_version
```diff
@@ -1 +1 @@
-master
+f3b6d65
```
EmmaLRussell: Targeting previous master commit until latest master is fixed

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/ExtensionTests.kt
```diff
@@ -33,4 +39,31 @@ class ExtensionTests {
         res = Response(URL("http://whatever"), 500)
         assertThat(res.asResponseEntity().statusCode).isEqualTo(HttpStatus.INTERNAL_SERVER_ERROR)
     }
+
+    @Test
+    fun `asStreamingResponseEntity copies response to output stream`() {
+        val mockInputStream = mock<InputStream>{
```
hillalex: Could create a real stream here like 
` val inputStream = "test".byteInputStream()` and then at the end wrote back to a string to test it matches like 
```
  ByteArrayOutputStream().use {
            result.body?.writeTo(it)
            assertThat(it.toString()).isEqualTo("test")
        }
```
I'd find that easier to reason about than ` verify(mockOutputStream).write(any(), eq(0), eq(1))` - I'm not too sure what the latter really means!

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintrAPIClientTests.kt
```diff
@@ -67,4 +67,20 @@ class HintrApiClientTests {
...
+
+        assertThat(result.statusCodeValue).isEqualTo(400)
+    }
+
+    @Test
+    fun `can download summary data`() {
+        val sut = HintrAPIClient(ConfiguredAppProperties(), ObjectMapper())
+        val result = sut.downloadSummary("ef5b621dfd9beb8851fcd7de2d6c1e08");
+
+        assertThat(result.statusCodeValue).isEqualTo(400)
```
hillalex: What are these 2 tests testing?

src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
```diff
@@ -16,9 +21,35 @@ fun Response.asResponseEntity(): ResponseEntity<String> {
...
     return ResponseEntity.status(httpStatus)
             .contentType(MediaType.APPLICATION_JSON)
             .body(body)
 }
+
+fun Response.asStreamingResponseEntity(): ResponseEntity<StreamingResponseBody> {
+    val httpStatus = httpStatusFromCode(this.statusCode)
+    val headers = headersToMultiMap(this.headers)
+
+    val inputStream = this.body().toStream()
```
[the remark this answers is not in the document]
EmmaLRussell: Hmm, good point. I'll try to find out!

src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
```diff
@@ -16,9 +21,35 @@ fun Response.asResponseEntity(): ResponseEntity<String> {
         404 -> HttpStatus.NOT_FOUND
         else -> HttpStatus.INTERNAL_SERVER_ERROR
     }
+}
+
+fun headersToMultiMap(headers: Headers): MultiValueMap<String, String> {
+    val result = LinkedMultiValueMap<String, String>()
```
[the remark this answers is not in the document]
EmmaLRussell: That's the idea, though I don't think it's actually including the Content-Disposition header at the moment. We could add it if it's not present?

src/app/src/main/kotlin/org/imperial/mrc/hint/AppStart.kt
```diff
@@ -35,4 +33,15 @@ class MvcConfig(val config: Config) : WebMvcConfigurer {
                 .addPathPatterns("/**")
                 .excludePathPatterns("/login", "/login/", "/password/**", "/callback", "/callback/", "/public/**")
     }
+
```
EmmaLRussell: Spring requires this to be configured for StreamingResponseBody. Settings might need tweaking.

src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
```diff
@@ -16,9 +25,39 @@ fun Response.asResponseEntity(): ResponseEntity<String> {
...
 
     val body = this.body().asString("application/json")
     return ResponseEntity.status(httpStatus)
             .contentType(MediaType.APPLICATION_JSON)
             .body(body)
 }
+
+fun DownloadRequest.asStreamingResponseEntity(): ResponseEntity<StreamingResponseBody> {
+
+    val responseBody = StreamingResponseBody { outputStream: OutputStream ->
```
EmmaLRussell: Having read the Fuel docs, this seems more what needs to be done. 
https://github.com/kittinunf/fuel/tree/master/fuel#download-response-to-output-file-or-outputstream

What's not ideal is that I'm calling .response() on the DownloadRequest twice, once in order to get the headers and httpStatus to put into the ResponseEntity, once inside the StreamingResponseBody lambda in order to specify the output stream as the destination for the download. So I guess this is going to make two separate requests. Am I missing something obvious? Is there some clever async way to avoid this..?

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/ExtensionTests.kt
EmmaLRussell: The code this is testing has all changed so test is now very different. I ended up putting the extension method on Request rather than DownloadRequest where it really belongs to make it possible to test with mockito (since DownloadRequest is a final class...)

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintrAPIClientTests.kt
EmmaLRussell: I guess not a great deal, since unlike the other tests it isn't checking JSON schema, but only that you're getting a 400 back from hintr rather than a 500 or 404. Could change it so that it waits for a run to complete and gets actual successful download data? That's already happening in DownloadTests though.

hillalex: Getting a 400 here is consistent with it being the wrong hintr url, so I don't think these add anything unless there's some way of confirming that the correct url has been hit (maybe the error message has something in it that identifies the endpoint?)

src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
hillalex: Making an extra HEAD request would be inexpensive seeing as it's just between hint and hintr which are on the same network, but this does seem like a pretty unintuitive interface. Had a little fiddle and this seems to do the trick and be somewhat easier to reason about:
```
        val url = URL("$baseUrl/download/spectrum/${id}")
        val conn = url.openConnection() as HttpURLConnection

        val responseEntity = ResponseEntity.status(conn.responseCode)
        responseEntity.headers(HttpHeaders(conn.headerFields as MultiValueMap<String, String>))
        responseEntity.body(StreamingResponseBody { outputStream: OutputStream ->
            conn.inputStream.use {
                it.copyTo(outputStream)
            }
        })

        return responseEntity.build()
```

hillalex: Having said this the test seems to be taking way longer to run with this code, I can't see why? :thinking:

hillalex: Ok, scratch this, this doesn't seem to be working at all.

Btw another useful check to add in the integration test is
```assertThat(response.body!!.size.toString()).isEqualTo(contentLength)```

Comment (hillalex) [reply src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt]: Ok, so the following works, not wrapping the `treamingResponseBody` in a response entity but just writing directly to the `HttpServletResponse`. But now I've come back around and thinking maybe the `Fuel` syntax isn't that bad, it's very terse after all, and making an extra request for headers should be inexpensive :woman_shrugging: 
```
    @GetMapping("/spectrum/{id}")
    @ResponseBody
    fun getSpectrum(@PathVariable("id")id: String, response: HttpServletResponse): StreamingResponseBody {

        return StreamingResponseBody { outputStream: OutputStream ->
            val url = URL("http://localhost:8888/download/spectrum/${id}")
            val conn = url.openConnection() as HttpURLConnection
            response.setContentLength(conn.contentLength)
            response.contentType = conn.contentType
            response.status = conn.responseCode
            conn.inputStream.use {
                it.copyTo(outputStream)
            }
        }
    }
```

Comment (EmmaLRussell) [reply src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt]: Wow, thanks for going on that journey! I quite like the Fuel syntax too, it's much terser as you say. I'll try putting in the HEAD request.

Comment (EmmaLRussell) [reply src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt]: Ah so of course, hintr doesn't currently support HEAD requests (getting 404s) - I'll ask Rob to put that in.

Comment (hillalex) [reply src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt]: Ah of course. Want to ask him to add the content disposition header too?

File: src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt [hunk 8]
```diff
@@ -66,7 +67,7 @@ fun Response.asResponseEntity(): ResponseEntity<String> {
 
 }
 
-fun Request.getStreamingResponseEntity(): ResponseEntity<StreamingResponseBody> {
+fun Request.getStreamingResponseEntity(headRequest: (url: String, parameters: Parameters?) -> Request): ResponseEntity<StreamingResponseBody> {
```
EmmaLRussell: This function parameter for fetching the head response is a bit weird, but makes it much easier to unit test..

hillalex: Ok, yeah it's pretty weird but I can see the issue with testing.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/DownloadTests.kt
```diff
@@ -0,0 +1,73 @@
...
+                val entity = getModelRunJsonEntity()
+                val runResult = testRestTemplate.postForEntity<String>("/model/run/", entity)
+                ObjectMapper().readValue<JsonNode>(runResult.body!!)["data"]["id"].textValue()
+            }
+            IsAuthorized.FALSE -> "nonsense"
+        }
+
+        do {
+            Thread.sleep(500)
+            val statusResponse = testRestTemplate.getForEntity<String>("/model/status/$id")
```
hillalex: I do wonder if this is going to re-introduce the non-deterministic test failure we just removed, since this is the same logic that was failing and I didn't actually get to the bottom of why, I just ditched it. Can we just stick some logging in here to get ahead of it? Just logging this statusResponse would be good.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/DownloadTests.kt
```diff
@@ -0,0 +1,73 @@
...
+            assertResponseHasExpectedDownloadHeaders(responseEntity)
+        }
+    }
+
+    fun assertResponseHasExpectedDownloadHeaders(response: ResponseEntity<ByteArray>) {
+        val headers = response.headers
+        assertThat(headers["Content-Type"]?.first()).isEqualTo("application/octet-stream")
+        val contentLength = headers["Content-Length"]?.first()!!
+        assertThat(contentLength.toInt()).isGreaterThan(0)
+        assertThat(headers["Connection"]?.first()).isEqualTo("close")
```
hillalex: Can we also assert that the body length matches the Content-Length header?

PR #129, opened 2019-10-29
mrc-614 Reset if incomplete step before any complete step

EmmaLRussell:
Consider steps invalid if an incomplete step comes before active step or any complete step

### DIFF
--- a/src/app/static/src/tests/root/actions.test.ts
+++ b/src/app/static/src/tests/root/actions.test.ts
@@ -44,5 +44,51 @@ describe("root actions", () => {
 
         actions.validate(mockContext as any);
         expect(mockContext.commit).toHaveBeenCalled();
-    })
+    });
+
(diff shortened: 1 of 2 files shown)


src/app/static/src/tests/root/actions.test.ts
```diff
@@ -44,5 +44,51 @@ describe("root actions", () => {
 
         actions.validate(mockContext as any);
         expect(mockContext.commit).toHaveBeenCalled();
-    })
+    });
+
+    it("resets state if a state following current step is not valid", () => {
```
hillalex: Took me a minute to understand the name of this test (and the others in the class, some of which I wrote!). Could we just add a comment explaining that "valid" means "either complete or not succeeded by an incomplete step"?

hillalex: It'd be great if you could add that and then merge

PR #130, opened 2019-10-29
mrc-603 style tweaks

hillalex:
* make select styling consistent across forms and filters
* makes the file uploads in step 2 the same width as those in step 1 initially, shrinking once the map is loaded


hillalex:
> LGTM. Did you mean to re-add package-lock? It got deleted a while ago I think. Should go into gitignore if it isn't?

It must have gotten deleted by accident? Should be committed.

PR #131, opened 2019-10-30
mrc-575

hillalex:
When survey or programme data is uploaded, request model run options from the backend and save in the store ready to be rendered in the next step.


src/app/static/src/app/apiService.ts
```diff
@@ -74,19 +74,20 @@ export class APIService<S extends string, E extends string> implements API<S, E>
...
     private _handleAxiosResponse(promise: Promise<AxiosResponse>) {
         return promise.then((response: AxiosResponse) => {
             const success = response && response.data;
+            const data = success.data;
             if (this._onSuccess) {
-                this._onSuccess(success);
+                this._onSuccess(data);
             }
-            return success;
+            return data;
```
hillalex: Figured it's actually much more useful to return the response data rather than the whole response object. We weren't actually using this anywhere yet so this change hasn't affected any other code.

src/app/static/src/app/components/forms/DynamicFormMultiSelect.vue
```diff
@@ -35,7 +35,13 @@
         computed: {
             value: {
                 get() {
-                    return this.formControl.value ? this.formControl.value as string[] : []
+                    if (Array.isArray(this.formControl.value)) {
```
hillalex: Support default values provided either as a single string, or an array of strings.

src/app/static/src/app/store/modelOptions/mutations.ts
```diff
@@ -17,6 +17,11 @@ export const mutations: MutationTree<ModelOptionsState> & MetadataMutations = {
     },
 
     update(state: ModelOptionsState, payload: DynamicFormMeta) {
-        state.optionsFormMeta = payload;
+        const allControlSectionLabels = payload.controlSections.map(c => c.label);
```
hillalex: We don't want to fully overwrite the existing state with the api response, in case some values have been filled out already. So if a control section already exists we don't overwrite it, but simply add any missing control sections.

src/app/static/src/app/store/modelOptions/mutations.ts
```diff
@@ -17,6 +17,12 @@ export const mutations: MutationTree<ModelOptionsState> & MetadataMutations = {
...
```
hillalex: The tests should make clear what's going on here: https://github.com/mrc-ide/hint/pull/131/files#diff-97445ba69387f1718e7c5f330133e49d

src/app/static/src/tests/integration/surveyAndProgram.itest.ts
```diff
@@ -41,7 +45,7 @@ describe("Survey and programme actions", () => {
         const formData = new FormData();
         formData.append('file', file);
 
-        await actions.uploadProgram({commit} as any, formData);
+        await actions.uploadProgram({commit, state : {}} as any, formData);
```
hillalex: The weird thing is that this test and the one below are *also* now timing out, even though they are not making any extra requests. This makes me think there is something weird going on.

PR #133, opened 2019-11-01
fix model run tests and other housekeeping

hillalex:
Some housekeeping:
* increase integration test default timeout from 5 to 6s
* add missing assertion in model run integration test
* clear polling interval in model run integration test
* upgrade all js packages
* point cli to master in testing script


src/app/static/src/tests/integration/modelRun.itest.ts
```diff
@@ -36,23 +36,34 @@ describe("Model run actions", () => {
...
     });
 
     it("can get model run result", async () => {
         const commit = jest.fn();
         const mockState = {
-            modelRunId: runId,
+            modelRunId: "1234",
             status: {
                 done: true,
                 id: runId
```
EmmaLRussell: I guess for tidiness, this should be the same invalid run Id, rather than 'runId' taken from the run submit in beforeAll

PR #134, opened 2019-11-04
Timeout bug

hillalex:
* increase fuel timeouts to 60s
* return informative message on an empty fuel response


hillalex:
Oh wow, good catch, looks like I was not setting up the tests correctly. Fixed now. 

To get the fixed hintr image you have to delete the existing one from your machine and then pull a fresh copy:

*  `docker image rm mrcide/hintr:master`
*  `docker pull mrcide/hintr:master`

I think that will fix the timeouts (I'm just doing the same on my laptop at home and it's taking ages to pull the new image so haven't confirmed this yet)

hillalex:
Can this be merged @EmmaLRussell ?

PR #136, opened 2019-11-06
mrc-490 Validate combined baseline files

EmmaLRussell:


src/app/static/src/app/store/baseline/baseline.ts
```diff
@@ -29,12 +32,16 @@ export const initialBaselineState: BaselineState = {
...
+    validatedComplete: false,
+    validatedConsistent: false,
+    baselineError: ""
 };
 
 export const baselineGetters = {
   complete: (state: BaselineState) => {
-      return !!state.country && !!state.iso3 && !!state.shape && !!state.population
+      return state.validatedComplete && state.validatedConsistent &&
+                state.country!! && state.iso3!! && state.shape!! && state.population!!
```
EmmaLRussell: Still need to check the individual file states as well, as the combine validate endpoint will return complete and consistent as true if there is an error with an individual file (but no inconsistent / missing files)

src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -24,26 +25,33 @@ export const actions: ActionTree<BaselineState, RootState> & BaselineActions = {
             .postAndReturn<PjnzResponse>("/baseline/pjnz/", formData)
             .then(() => {
                 dispatch('metadata/getPlottingMetadata', state.iso3, {root: true});
+                dispatch('validate');
```
EmmaLRussell: Calling baseline validate after every file upload so can see inconsistencies immediately

src/app/static/src/app/store/baseline/baseline.ts
```diff
@@ -41,7 +41,7 @@ export const initialBaselineState: BaselineState = {
 export const baselineGetters = {
   complete: (state: BaselineState) => {
       return state.validatedComplete && state.validatedConsistent &&
-                state.country!! && state.iso3!! && state.shape!! && state.population!!
+                !!state.country && !!state.iso3 && !!state.shape && !!state.population
```
EmmaLRussell: Still need to check the individual file states as well, as the combine validate endpoint will return complete and consistent as true if there is an error with an individual file (but no inconsistent / missing files)

src/app/static/src/app/store/baseline/baseline.ts
```diff
@@ -29,12 +32,16 @@ export const initialBaselineState: BaselineState = {
...
+    validatedComplete: false,
+    validatedConsistent: false,
+    baselineError: ""
 };
 
 export const baselineGetters = {
   complete: (state: BaselineState) => {
-      return !!state.country && !!state.iso3 && !!state.shape && !!state.population
+      return state.validatedComplete && state.validatedConsistent &&
+                !!state.country && !!state.iso3 && !!state.shape && !!state.population
```
EmmaLRussell: Still need to check the individual file states as well, as the combine validate endpoint will return complete and consistent as true if there is an error with an individual file (but no inconsistent / missing files)

src/app/static/src/tests/integration/baseline.itest.ts
```diff
@@ -54,15 +55,26 @@ describe("Baseline actions", () => {
...
         expect(commit.mock.calls[1][0]["payload"]["filename"])
             .toBe("population.csv");
     });
 
+    it ("can validate baseline data", async () => {
+        const commit = jest.fn();
+        const dispatch = jest.fn();
+
+        await actions.validate({commit, dispatch} as any);
+        console.log(JSON.stringify(commit.mock.calls[1][0]));
```
hillalex: Debugging statement?

src/app/static/src/app/store/baseline/baseline.ts
hillalex: So what does `complete` actually mean?

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/BaselineController.kt
```diff
@@ -48,4 +48,14 @@ class BaselineController(private val fileManager: FileManager,
         return getAndValidate(FileType.Population)
     }
 
+    @GetMapping("/validate/")
+    @ResponseBody
+    fun validate(): ResponseEntity<String> {
+        return apiClient.validateBaselineCombined(
+                fileManager.getFile(FileType.PJNZ),
```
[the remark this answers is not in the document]
EmmaLRussell: Ah ok - yeah the reason I didn't use getFiles was because I'd then have to extract the values from the returned map by string key to construct the json to pass to hintr, which seemed more of a faff. But if we can just serialize it straight through that'll save some code.

EmmaLRussell:
> So what does complete actually mean?
Complete means that 'validate' has confirmed that all three files are present and they don't conflict with each other, *and* there weren't any errors returned by the individual file uploads. This would be simpler if the validate endpoint wouldn't validate sessions as complete and consistent if they have one duff file, but it seemed that it does. Could raise that as a ticket to fix in hintr.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/BaselineController.kt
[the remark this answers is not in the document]
hillalex: There is also `mapValues` kotlin function which transforms one map into another with the same keys just by transforming the values: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map-values.html

hillalex:
> Complete means that 'validate' has confirmed that all three files are present and they don't conflict with each other, _and_ there weren't any errors returned by the individual file uploads. 

If this is the intended meaning then it sounds like there's a bug in the current behaviour. So I guess we keep our logic as is for now and can simplify once the bug is fixed in hintr. Want to raise a ticket?

EmmaLRussell:
> > Complete means that 'validate' has confirmed that all three files are present and they don't conflict with each other, _and_ there weren't any errors returned by the individual file uploads.
> 
> If this is the intended meaning then it sounds like there's a bug in the current behaviour. So I guess we keep our logic as is for now and can simplify once the bug is fixed in hintr. Want to raise a ticket?

https://vimc.myjetbrains.com/youtrack/issue/mrc-682

EmmaLRussell:
Given the Slack coversation with Rob on this, I'll remove the use of 'complete' from the validate response, as this is going to be removed.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/BaselineController.kt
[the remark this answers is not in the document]
EmmaLRussell: Actually, it turned out that another issue is that getFiles misses out the entry for missing files, while the validate baseline endpoint required all three properties to be present, so I have modified this approach slightly.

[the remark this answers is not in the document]
hillalex: Ah of course

PR #137, opened 2019-11-07
 mrc-672 line wrap in filter treeselects should expand component

EmmaLRussell:


src/app/static/src/scss/partials/form.scss
```diff
@@ -17,6 +17,7 @@
     .vue-treeselect__control {
       @extend .form-control;
       font-size: inherit;
+      display: table;
```
hillalex: There's just something slightly weird going on with the height/padding now which I think is fixed by adding 
`padding-bottom: 0;`

PR #138, opened 2019-11-08
mrc-575 tool for testing that a js object conforms to the dynamic form types

hillalex:
First PR. This utility will be used for checking that the form options object returned from hintr conforms to the expected type.

### DIFF
--- a/src/app/static/src/app/components/forms/types.ts
+++ b/src/app/static/src/app/components/forms/types.ts
@@ -1,20 +1,26 @@
 import {Dict} from "../../types";
 
-export interface DynamicControlSection {
+export type DynamicControlSection = {
     label: string
     description?: string
     controlGroups: DynamicControlGroup[]
 }
 
-export interface DynamicControlGroup {
+export type DynamicControlGroup = {
     label?: string
     controls: Control[]
+}
+
+export type Option = {
+    id: string,
+    label: string,
+    children?: Option[]
 }
 
 export type DynamicControlType = "multiselect" | "select" | "number"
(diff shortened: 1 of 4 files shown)


src/app/static/src/app/components/forms/types.ts
```diff
@@ -1,20 +1,26 @@
 import {Dict} from "../../types";
 
-export interface DynamicControlSection {
+export type DynamicControlSection = {
```
hillalex: There's practically no difference between interfaces and types in typescript these days. Making these types felt a bit tidier.

src/app/static/src/tests/components/forms/dynamicFormChecker.test.ts
```diff
@@ -0,0 +1,309 @@
...
+    it("should recognise valid number input with value", () => {
+        expect(isDynamicFormMeta({
+            controlSections: [{
+                label: "l1",
+                controlGroups: [{
+                    controls: [{
+                        name: "i1",
+                        required: true,
+                        type: "number",
+                        value: 10
```
EmmaLRussell: Maybe add test that "10" is recognised as invalid?

hillalex: Done. Although maybe it would be better to accept string values that can be parsed as numbers :thinking:

hillalex: In fact the html input will accept "10", so we should as well. Updated accordingly: https://github.com/mrc-ide/hint/pull/138/commits/76a54a7bd35ea286a5c90da113522fe3087e366c

PR #139, opened 2019-11-08
mrc-575 actually fetch model options from hintr

hillalex:
Depends on https://github.com/mrc-ide/hint/pull/138 so that should be merged first.

Note branch name here is `mrc-575_form` (`mrc-575` is an outdated branch)


src/app/static/src/app/store/modelOptions/mutations.ts
```diff
@@ -18,5 +20,16 @@ export const mutations: MutationTree<ModelOptionsState> & MetadataMutations = {
 
     update(state: ModelOptionsState, payload: DynamicFormMeta) {
         state.optionsFormMeta = payload;
+    },
+
+    FormMetaUpdated(state: ModelOptionsState, action: PayloadWithType<DynamicFormMeta>) {
```
hillalex: There's a possible workflow that looks like:
1. user uploads survey data
2. user proceeds to model options step and selects some options
3. user returns to upload step and uploads program data

In this case we want to append the new section returned from the model options endpoint (some sections are conditional on program data being present) but we don't want to overwrite the already selected options. Hence this mutation which appends or removes sections (once https://vimc.myjetbrains.com/youtrack/issue/mrc-657 is in users will also be able to remove uploaded files, so could go from program + survey -> survey only), but does not overwrite selected values.

See tests for this here: https://github.com/mrc-ide/hint/pull/139/files#diff-97445ba69387f1718e7c5f330133e49dR20

src/app/static/src/app/components/forms/DynamicFormMultiSelect.vue
```diff
@@ -35,7 +35,13 @@
         computed: {
             value: {
                 get() {
-                    return this.formControl.value ? this.formControl.value as string[] : []
+                    if (Array.isArray(this.formControl.value)) {
```
hillalex: Now supporting either single string or array of strings as a default value.

src/app/static/src/app/apiService.ts
```diff
@@ -74,19 +74,20 @@ export class APIService<S extends string, E extends string> implements API<S, E>
...
     private _handleAxiosResponse(promise: Promise<AxiosResponse>) {
         return promise.then((response: AxiosResponse) => {
             const success = response && response.data;
+            const data = success.data;
             if (this._onSuccess) {
-                this._onSuccess(success);
+                this._onSuccess(data);
             }
-            return success;
+            return data;
```
hillalex: Realised it made more sense to just return the data from the response here, rather than the whole wrapped success response. This wasn't actually being used anywhere yet, [this](https://github.com/mrc-ide/hint/pull/139/files#diff-cbdcdbb269d6697eb9210ac538053e44R39) is the first instance of it.

src/app/static/src/app/store/modelOptions/mutations.ts
```diff
@@ -1,12 +1,14 @@
...
 import {DynamicFormData, DynamicFormMeta} from "../../components/forms/types";
+import {PayloadWithType} from "../../types";
 
 type ModelOptionsMutation = Mutation<ModelOptionsState>
 
 export interface MetadataMutations {
     validate: ModelOptionsMutation,
-    update: ModelOptionsMutation
+    update: ModelOptionsMutation,
+    FormMetaUpdated: ModelOptionsMutation
```
EmmaLRussell: Should these Mutations all start with upper-case to be consistent with the other Mutations, or is there a reason why validate and update are lower-case?

src/app/static/src/app/store/modelOptions/mutations.ts
EmmaLRussell: And the control sections themselves shouldn't change structure in the course of mucking about with different files? They'll always have the same controls and options if they're present at all?

hillalex: This is a good question. At the moment yes, the only thing that changes is whether or not the ART and ANC sections are included or not. But I'll check with Rob if that could change. If so, we'd have to just accept that options will get overwritten if you change your input files.

src/app/static/src/app/store/modelOptions/mutations.ts
hillalex: Yeah this is a weird one. `validate` and `update` get called as functions directly from the component via `mapMutations`, rather than committed from actions. 

We could fire off actions instead of mapping mutations directly into the component, or we could just stick to the TitleCase and have some weird function names in the component.

They also differ in signature - they take the payload directly rather than an action payload with type. I'm actually not clear on why we're usually passing an action with type and payload, and whether we should switch all mutations to this simpler signature.

Overall this is the bit I really don't like about Vuex - the matching of actions to mutations based on strings. I much prefer the Redux "reducer" pattern. Anyway, maybe we can chat about this next week and decide on a convention/pattern that we like.

src/app/static/src/app/store/modelOptions/mutations.ts
hillalex: "Yeah internal structure of each section is fixed. Only expecting ART and ANC to be possibly missing.
Whether it is likely to change I'm not sure, there isn't a mechanism for that in the R side atm. Only able to exclude/include at the section level.
I'd have to check with Jeff though for sure"

src/app/static/src/app/store/modelOptions/mutations.ts
hillalex: Hmm, in fact, in the docs they use `pascalCase` for all mutation names. But they also have a suggested pattern for mitigating the thing I don't like, the string matching, by using constants https://vuex.vuejs.org/guide/mutations.html#using-constants-for-mutation-types :thinking:

src/app/static/src/app/store/modelOptions/mutations.ts
hillalex: Following discussion today it seems that controls within sections *will* change. And we'll have to re-request model options every time the user lands on the model options page. Which means we probably don't want to request them when files are uploaded, as I've done on this branch. So if you hold off re-reviewing this, I'll make those changes on this branch.

PR #144, opened 2019-11-12
mrc-696 remove springboot context from cli

hillalex:
2 motivations for these changes:
1. It has been pointed out that the cli is really slow, which will be tedious when adding many real users. This is because it's booting up the whole app, which is slow to start up.
2. To get production db password into the cli would currently have to mount something at `/etc/hint/config` (as we do in Montagu). Passing extra bit of config as environment variables is more lightweight.

### DIFF
--- a/src/app/build.gradle
+++ b/src/app/build.gradle
@@ -6,18 +6,6 @@ apply plugin: 'application'
 
 mainClassName = "org.imperial.mrc.hint.AppStartKt"
 
-jacoco {
-	toolVersion = "0.8.3"
-	reportsDir = file("$projectDir/coverage")
-}
-
-jacocoTestReport {
-	reports {
-		xml.enabled = true
-		csv.enabled = false
-	}
-}
-
 group = 'org.imperial.mrc'
 sourceCompatibility = '1.8'
 
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/db/DbConfig.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/db/DbConfig.kt
@@ -18,4 +18,13 @@ class DbConfig {
         dataSourceBuilder.password(appProperties.dbPassword)
         return dataSourceBuilder.build()
     }
+
+    fun dataSource(url: String, user: String, password: String): DataSource {
+        val dataSourceBuilder = DataSourceBuilder.create()
+        dataSourceBuilder.driverClassName("org.postgresql.Driver")
+        dataSourceBuilder.url(url)
+        dataSourceBuilder.username(user)
+        dataSourceBuilder.password(password)
+        return dataSourceBuilder.build()
+    }
 }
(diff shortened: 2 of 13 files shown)


src/userCLI/src/test/kotlin/org/imperial/mrc/hint/userCLI/AppTests.kt
```diff
@@ -1,53 +1,58 @@
...
+        @JvmStatic
+        fun cleanup() {
+            dataSource.connection.close()
+        }
     }
 
-    @Autowired
-    private lateinit var context: ApplicationContext
+    @BeforeEach
+    fun `remove user if exists`() {
```
hillalex: we no longer have spring magic "transactional" unfortunately, so just manually resetting the db each time

src/userCLI/src/main/kotlin/org/imperial/mrc/hint/userCLI/DatabaseProperties.kt
```diff
@@ -0,0 +1,33 @@
...
+    private fun fromEnvOrDefault(propName: String): String {
+        return environment.get(propName) ?: properties[propName].toString()
+    }
+
+    val user: String get() {
+        return fromEnvOrDefault("db_user")
+    }
+
+    val password: String get() {
+        return fromEnvOrDefault("db_password")
```
hillalex: these props are all implicitly under test in here: https://github.com/mrc-ide/hint/pull/144/files#diff-85c6cbe4768ca3720255b3c06ef36a41R17

src/app/build.gradle
```diff
@@ -6,18 +6,6 @@ apply plugin: 'application'
 
 mainClassName = "org.imperial.mrc.hint.AppStartKt"
 
-jacoco {
```
hillalex: Noticed that this is already declared in the parent project file for all subprojects, so didn't need to be here.

src/app/src/main/kotlin/org/imperial/mrc/hint/db/DbConfig.kt
```diff
@@ -18,4 +18,13 @@ class DbConfig {
         dataSourceBuilder.password(appProperties.dbPassword)
         return dataSourceBuilder.build()
     }
+
+    fun dataSource(url: String, user: String, password: String): DataSource {
```
hillalex: Codecov thinks this isn't under test. It is actually in the `beforeAll` method of a test suite, so as far as I'm concerned it is under test!

src/userCLI/src/main/kotlin/org/imperial/mrc/hint/userCLI/App.kt
```diff
@@ -15,54 +16,49 @@ Usage:
...
     }
 }
 
-class UserCLI(private val appContext: ApplicationContext)
-{
-    fun addUser(options: Map<String, Any>, userRepository: UserRepository=userRepository()): String
-    {
+class UserCLI(dataSource: DataSource) {
+
+    private val profileService = DbProfileService(dataSource, SecurePasswordEncoder())
```
EmmaLRussell: I suppose we could use a HintDbProfileService here, rather than the base class, though at the moment that wouldn't make any practical difference as the only override relates to login validation.

PR #148, opened 2019-11-13
mrc-688 send email when user is added without pw

hillalex:
Should be merged after https://github.com/mrc-ide/hint/pull/149

Sorry, this is a big PR because it contains a refactor of the email classes. Kept it in one PR because I thought it might be easier to see the motivation for the refactor alongside its use, but can pull it out into its own PR if you'd rather?


src/app/src/main/kotlin/org/imperial/mrc/hint/db/UserRepository.kt
```diff
@@ -1,50 +1,68 @@
...
+    override fun addUser(email: String, password: String?) {
+        if (getUser(email) != null) {
             throw UserException("User already exists")
         }
 
         val profile = DbProfile()
         profile.build(email, mapOf("username" to email))
 
-        profileService.create(profile, password)
+        val pw = if (password.isNullOrEmpty()) {
```
EmmaLRussell: It feels a little bit weird for it to be the user repo that is responsible for sending emails - outside its jurisdiction as a repo. Would maybe feel more natural for there to be a logic class doing that - or put the logic in the cli itself?

src/app/src/main/kotlin/org/imperial/mrc/hint/emails/EmailManager.kt
```diff
@@ -0,0 +1,27 @@
...
+interface EmailManager {
+    fun sendEmail(data: EmailData, emailAddress: String)
+    fun sendPasswordEmail(emailAddress: String, username: String, emailTemplate: PasswordEmailTemplate)
+}
+
+abstract class BaseEmailManager(private val appProperties: AppProperties,
+                                private val oneTimeTokenManager: OneTimeTokenManager) : EmailManager {
+
+    override fun sendPasswordEmail(emailAddress: String, username: String, emailTemplate: PasswordEmailTemplate) {
+        val token = oneTimeTokenManager.generateOnetimeSetPasswordToken(username)
```
EmmaLRussell: This also seems a little bit odd from a responsibilities perspective - for the emailManager to be responsible for generating oneTimeTokens rather than some coordinating logic class.

hillalex:
Agree with your comments and have opened this ticket: https://vimc.myjetbrains.com/youtrack/issue/mrc-707

PR #149, opened 2019-11-13
mrc-696 remove spring boot from cli (take 2)

hillalex:
* moves all db related config into config.properties 
* removes spring boot app from user cli

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/security/tokens/OneTimeTokenManager.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/security/tokens/OneTimeTokenManager.kt
@@ -6,6 +6,7 @@ import org.pac4j.core.profile.CommonProfile
 import org.pac4j.jwt.profile.JwtGenerator
 import org.springframework.stereotype.Component
 import org.pac4j.jwt.config.signature.SignatureConfiguration
+import org.pac4j.jwt.credentials.authenticator.JwtAuthenticator
 import java.time.Duration
 import java.security.SecureRandom
 import java.time.Instant
@@ -16,7 +17,7 @@ class OneTimeTokenManager(
         appProperties: AppProperties,
         private val tokenRepository: TokenRepository,
         signatureConfiguration: SignatureConfiguration,
-        private val authenticator: OneTimeTokenAuthenticator
+        private val authenticator: JwtAuthenticator
 )
 {
     private val generator = JwtGenerator<CommonProfile>(signatureConfiguration)
--- a/scripts/add-test-user.sh
+++ b/scripts/add-test-user.sh
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 NETWORK=hint_nw
+HERE=$(readlink -f "$(dirname $0)")
+TEST_CONFIG=$HERE/../src/app/static/scripts/test.properties
 
 image=mrcide/hint-user-cli:master
 docker pull $image
-docker run --network=hint_nw $image add-user test.user@example.com password
+docker run --network=hint_nw \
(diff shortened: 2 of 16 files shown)


hillalex:
Branch failed with same intermittent model run test failure. Just re-rerunning that build.

src/app/src/main/kotlin/org/imperial/mrc/hint/security/tokens/OneTimeTokenManager.kt
```diff
@@ -16,7 +17,7 @@ class OneTimeTokenManager(
         appProperties: AppProperties,
         private val tokenRepository: TokenRepository,
         signatureConfiguration: SignatureConfiguration,
-        private val authenticator: OneTimeTokenAuthenticator
```
EmmaLRussell: Why did this change?

hillalex: Ah this isn't actually relevant until the next PR. `OneTimeTokenAuthenticator` has a more complicated constructor than the class it inherits from. The UserCLI doesn't actually do any authentication, so no need for any custom logic there, we may as well just pass the base class which is easier to instantiate. https://github.com/mrc-ide/hint/pull/148/files#diff-7083312459829a5d01f15b4f2e7dd398R109

hillalex:
> Address already in use

Looks like you have something else already running on port 8080!

EmmaLRussell:
> > Address already in use
> 
> Looks like you have something else already running on port 8080!

Yep, thought I'd killed everything but had a rogue debug process :woman_facepalming:

src/app/src/main/kotlin/org/imperial/mrc/hint/security/tokens/OneTimeTokenManager.kt
EmmaLRussell: Oh, so can we get rid of OneTimeTokenAuthenticator then? Doesn't get used anywhere else, except  as a mock in OneTimeTokenManagerTests (which I guess should mock a JwtAuthenticator instead now).

hillalex: Isn't it essential for verifying one time tokens?? The main app still has to do that!

hillalex: It should still be being wired up as a bean and injected by spring, just not used in the UserCLI where we're doing the dep injection manually.

hillalex: Actually the more correct solution here would be to split out the `OnetimeTokenManager` into a `OnetimeTokenIssuer` and `OnetimeTokenAuthenticator`, then the `EmailManager` could just take the former as a dependency. Although I'm somewhat concerned that we need to get these new users added asap, so I'm minded to leave as is for now and do that refactor on a separate branch.

EmmaLRussell: >It should still be being wired up as a bean and injected by spring
Ok, because it's the only implementation of JwtAuthenticator that's got @Configuration on it, so that's the one which will be instantiated when it creates the parameters for the OneTimeTokenManager.

hillalex: Exactly :+1: But I will add a ticket for refactoring into separate Issuer and Authenticator

PR #150, opened 2019-11-13
mrc-700 add logging for debugging tests

hillalex:
To help debug intermittent Travis test failures

### DIFF
--- a/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/IntegrationTests.kt
+++ b/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/IntegrationTests.kt
@@ -13,7 +13,6 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.test.context.SpringBootTest
 import org.springframework.boot.test.web.client.TestRestTemplate
 import org.springframework.boot.test.web.client.getForEntity
-import org.springframework.boot.test.web.client.postForEntity
 import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.test.context.ActiveProfiles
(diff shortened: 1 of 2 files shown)


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/IntegrationTests.kt
```diff
@@ -51,10 +50,17 @@ abstract class SecureIntegrationTests: CleanDatabaseTests() {
...
             IsAuthorized.TRUE -> {
-                Assertions.assertThat(responseEntity.headers.contentType!!.toString()).isEqualTo("application/json")
-                Assertions.assertThat(responseEntity.statusCode).isEqualTo(HttpStatus.OK)
-                if (schemaName != null) {
-                    JSONValidator().validateSuccess(responseEntity.body!!, schemaName)
+                Assertions.assertThat(responseEntity.headers.contentType!!.toString())
+                        .isEqualTo("application/json")
+
+                if (responseEntity.statusCode == HttpStatus.OK) {
+                    Assertions.assertThat(responseEntity.statusCode).isEqualTo(HttpStatus.OK)
```
EmmaLRussell: Not sure why this bit has changed - seems to be asserting the condition of the preceding 'if'?

hillalex: If it's not true, we want to log the response and manually fail the test. If true, just assert true. So yes, this assertion is superfluous, I just thought it was clearer this way that there's always an assertion. But obviously not :laughing:

hillalex: The alternative would just be to remove that assertion and have
```
 if (responseEntity.statusCode == HttpStatus.OK) {
                    if (schemaName != null) {
                        JSONValidator().validateSuccess(responseEntity.body!!, schemaName)
                    }
                } else {
                    Assertions.fail("Expected success response but got failure: ${responseEntity.body}")
                }
```

Would that be clearer?

hillalex: https://github.com/mrc-ide/hint/pull/150/commits/1e474bcfab381299aceaddc20ee55f51ee9d2835

PR #151, opened 2019-11-14
mrc-575 add util for updating form and update multiselect

hillalex:
* adds util for updating a form without overwriting existing control values
* allow vue multiselect to accept default value of a single string as well as an array

### DIFF
--- a/src/app/static/src/app/apiService.ts
+++ b/src/app/static/src/app/apiService.ts
@@ -74,8 +74,8 @@ export class APIService<S extends string, E extends string> implements API<S, E>
     };
 
     withSuccess = (type: S) => {
-        this._onSuccess = (success: Response) => {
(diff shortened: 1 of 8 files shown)


src/app/static/src/app/apiService.ts
```diff
@@ -74,19 +74,20 @@ export class APIService<S extends string, E extends string> implements API<S, E>
...
     private _handleAxiosResponse(promise: Promise<AxiosResponse>) {
         return promise.then((response: AxiosResponse) => {
             const success = response && response.data;
+            const data = success.data;
             if (this._onSuccess) {
-                this._onSuccess(success);
+                this._onSuccess(data);
             }
-            return success;
+            return data;
```
hillalex: Also this minor change to return just the response data, rather than the larger response object from the API.

PR #152, opened 2019-11-14
mrc-575 fetch model run options

hillalex:
contains commits from https://github.com/mrc-ide/hint/pull/151

* fetch options every time the user navigates to the relevant step
* render loading spinner while fetching options
* render tick and message when options are valid


src/app/static/src/app/apiService.ts
```diff
@@ -61,6 +68,11 @@ export class APIService<S extends string, E extends string> implements API<S, E>
 
     private _onSuccess: ((success: Response) => void) | null = null;
 
+    freezeResponse = () => {
```
hillalex: We were freezing everything by default, but actually this is unnecessary and made updating the form options a pain (it's much easier to mutate the incoming object by just adding existing values, than to mutate the old object or create a new one.)

src/app/static/src/tests/components/modelOutput/modelOutput.test.ts
```diff
@@ -30,6 +30,14 @@ function getStore() {
                 state: mockModelRunState({
                     result: mockModelResultResponse({data: ["TEST DATA"] as any})
                 })
+            },
+            modelOutput: {
```
hillalex: Just added this to make lots of test warnings go away

PR #153, opened 2019-11-15
mrc-623 delete endpoints for all input files

hillalex:
Note that when we "delete" a file we just remove it from the session in the db, we don't delete the physical file. This is because it might belong to a downloaded state that will later need to be rehydrated.

Incidentally, seems like it would pretty easy to track which files can be deleted by incrementing some count in the db whenever a file is part of a downloaded state. Then when count is 0 and file does not belong to a session, can remove. 

Contains commits from https://github.com/mrc-ide/hint/pull/154/files to fix failing tests on master.


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/IntegrationTests.kt
```diff
@@ -5,28 +5,13 @@ import com.fasterxml.jackson.databind.ObjectMapper
...
 import org.springframework.http.ResponseEntity
-import org.springframework.test.context.ActiveProfiles
-import java.io.File
-
-@ActiveProfiles(profiles = ["test"])
-@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
-abstract class IntegrationTests {
-
-    @AfterEach
-    fun tearDown() {
```
hillalex: Just noticed this class wasn't being used anywhere

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/DiseaseTests.kt
```diff
@@ -82,4 +85,25 @@ class DiseaseTests : SecureIntegrationTests() {
         val responseEntity = testRestTemplate.getForEntity<String>("/disease/anc/")
         assertSecureWithSuccess(isAuthorized, responseEntity, "ValidateInputResponse")
     }
+
+    @ParameterizedTest
+    @EnumSource(IsAuthorized::class)
+    fun `can delete survey data`(isAuthorized: IsAuthorized) {
```
hillalex: For these tests I've not done the more complete thing of adding a file and then confirming it gets deleted as in the [baseline tests](https://github.com/mrc-ide/hint/pull/153/files#diff-d0e7b6eef5bc7c7f8d5e2a6e82f4fd13R101). That's because to do so would involve uploading the shape file as well as the survey/program file each time. Since all units are under tests, and these use the same logic as the baseline methods, I figured it wasn't worth making the build even slower than it is. So these tests are just checking that the url is valid and secured. But let me know if you think this is not robust enough.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/BaselineController.kt
```diff
@@ -24,6 +31,13 @@ class BaselineController(fileManager: FileManager,
         return getAndValidate(FileType.PJNZ)
     }
 
+    @DeleteMapping("/pjnz/{hash}/")
```
EmmaLRussell: Are we requiring the hash here (unlike for the get endpoint) as a security measure (to check the front end is really deleting the file it intends to)? *Could* just look up the hash as we do for the GET.

src/app/src/main/kotlin/org/imperial/mrc/hint/db/SessionRepository.kt
```diff
@@ -76,6 +76,14 @@ class JooqSessionRepository(private val dsl: DSLContext) : SessionRepository {
 
     }
 
+    override fun removeSessionFile(sessionId: String, type: FileType, hash: String) {
+        dsl.deleteFrom(SESSION_FILE)
```
EmmaLRussell: This won't throw an error if no rows found? We should probably return a 404 if the hash doesn't exist for the session and file type?

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/DiseaseTests.kt
EmmaLRussell: Should be fine - there's pretty comprehensive unit tests of all the bits..

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/BaselineController.kt
hillalex: You're right, the hash isn't needed. Could remove?

src/app/src/main/kotlin/org/imperial/mrc/hint/db/SessionRepository.kt
hillalex: This was deliberate. I think it's better if idempotent. If it already doesn't exist, great! Means e.g. the front-end can call a "delete all" method without worrying about whether some exist and others don't.

PR #141, opened 2019-11-08
mrc-649 add missing integration tests

hillalex:
Does what it says on the tin :)

PR #154, opened 2019-11-15
mrc-705 fix flaky model run results test

hillalex:
We don't need to test hintr functionality, just that we're hitting the correct endpoint. So have simplified the test to just pass a non-existent run id and check that the error message is as expected for that endpoint.

This also fixes a front-end integration test failure. master had changed since the checks for https://github.com/mrc-ide/hint/pull/141 had been run, and one of the tests had become out of date.

### DIFF
--- a/src/app/static/src/tests/integration/load.itest.ts
+++ b/src/app/static/src/tests/integration/load.itest.ts
@@ -18,7 +18,7 @@ describe("load actions", () => {
         formData.append('file', file);
 
         await baselineActions.uploadShape({commit} as any, formData);
-        shape = (commit.mock.calls[1][0]["payload"] as ShapeResponse);
+        shape = (commit.mock.calls[2][0]["payload"] as ShapeResponse);
     });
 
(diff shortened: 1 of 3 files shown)


src/app/static/src/tests/integration/load.itest.ts
```diff
@@ -16,9 +16,8 @@ describe("load actions", () => {
         const file = fs.createReadStream("../testdata/malawi.geojson");
         const formData = new FormData();
         formData.append('file', file);
-
-        await baselineActions.uploadShape({commit} as any, formData);
-        shape = (commit.mock.calls[1][0]["payload"] as ShapeResponse);
+        await baselineActions.uploadShape({commit, dispatch: jest.fn()} as any, formData);
+        shape = (commit.mock.calls[2][0]["payload"] as ShapeResponse);
```
hillalex: There is now an extra commit happening in this method (to invalidate survey/program when shape is updated), the actual shape response is the 3rd commit.

PR #156, opened 2019-11-15
Mrc 656 Persist Barchart selections

EmmaLRussell:
This adds a plottingSelections module to the store to remember selections (selected filter options and axis columns) for the barchart (and we can also use it for other any output plots we add). Will need to remember to reset this state when prior steps become invalidated. 

I tried to use a similar model to the model run options form for the Barchart component  - it emits a changed event when any new selections are made.

I've left the responsibility for setting sensible initial values for the selections in the Barchart for now - but this will be fetched from hintr response metadata in a future ticket.


src/app/static/src/tests/components/plots/barchart/filterSelect.test.ts
```diff
@@ -3,17 +3,19 @@ import {createLocalVue, shallowMount} from "@vue/test-utils";
...
+        isXAxis: true,
+        isDisaggregateBy: false,
+        value: [{id: "fo2", label: "option 2"}],
+        label: "Test Filter"
+    }
 };
 
 const getWrapper = (props: any = {}) => {
-    return shallowMount(FilterSelect, {propsData: {...propsData, ...props}, localVue});
+    return shallowMount(FilterSelect, {propsData: {...defaultProps(), ...props}, localVue});
```
EmmaLRussell: This was a bit weird - propsData seemed to get mutated somewhere in the course of these tests running, even though they were only being used with ...   Changing to a function seemed to sort it, but it's bizarre...

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -18,7 +18,8 @@
             </div>
 
             <div id="barchart-container" :class="selectedTab==='Bar' ? 'col-md-12' : 'd-none'">
-                <barchart :chartdata="chartdata" :filters="barchartFilters" :indicators="barchartIndicators"></barchart>
+                <barchart :chartdata="chartdata" :filters="barchartFilters" :indicators="barchartIndicators"
+                          :selections="barchartSelections" v-on:change-selections="updateBarchartSelections"></barchart>
```
hillalex: I see why you've gone for a custom event name over "change", but "change-selections" is quite verbose - how about "select"?

src/app/static/src/app/components/plots/barchart/Barchart.vue
```diff
@@ -1,42 +1,46 @@
...
+                                :value=selections.xAxisId
+                                @input="changeXAxisId"></treeselect>
                 </div>
                 <div id="disagg-fg" class="form-group">
                     <label class="font-weight-bold">Disaggregate by</label>
                     <treeselect :multiple=false
                                 :clearable="false"
                                 :options="filtersAsOptions"
-                                v-model=disaggregateById></treeselect>
+                                :value=selections.disaggregateById
```
hillalex: What do you think about sticking with `v-models` for these controls but just using computed props with getters/setters? 

E.g.
```
            disaggregateById: {
                get() {
                    return this.selections.disaggregateById
                },
                set(value: string) {
                    this.$emit("change-selections", {disaggregateById: value});
                }
            }
```

The `created` handler would become a bit simpler e.g.

```
if (!this.indicatorId) {
    this.indicatorId = this.indicators[0].indicator;
}

if (!this.xAxisId) {
    this.xAxisId = this.filters[0].id;
}
...etc.
```

hillalex:
> This adds a plottingSelections module to the store to remember selections (selected filter options and axis columns) for the barchart (and we can also use it for other any output plots we add). Will need to remember to reset this state when prior steps become invalidated.
> 
> I tried to use a similar model to the model run options form for the Barchart component - it emits a changed event when any new selections are made.
> 
> I've left the responsibility for setting sensible initial values for the selections in the Barchart for now - but this will be fetched from hintr response metadata in a future ticket.

Oh, and I think having the `BarChart` component set some defaults if selected values are missing is a good idea anyway, I'd leave that functionality in even when values are fed in from the api :+1:

src/app/static/src/tests/components/plots/barchart/filterSelect.test.ts
hillalex: Was it the `value` or `options` properties that were mutated? spread operator will only give a shallow clone, could be why.

EmmaLRussell: Ah, yeah I think you're right. It was `value` that was causing the problems - I guess the clone just got a pointer to the same array.

src/app/static/src/app/components/modelOutput/ModelOutput.vue
EmmaLRussell: Not that keen on `select` as it's quite associated with selecting from a drop-down. I could go with `update`?

src/app/static/src/app/components/plots/barchart/Barchart.vue
EmmaLRussell: OK, sounds good - the only thing with that is getting a flurry of emitted changes on initialise - but that will largely go away once we're getting defaults from the API.

src/app/static/src/app/components/modelOutput/ModelOutput.vue
hillalex: Ah yeah fair point. "update" seems good :+1:

hillalex:
Had another thought about this approach which is that it isn't ideal for when we want to factor this out into a re-usable component because it makes it essential to manage a selections object externally to the component (which presumably many use cases don't warrant). But I figure lets cross that bridge when we come to it.

EmmaLRussell:
> Had another thought about this approach which is that it isn't ideal for when we want to factor this out into a re-usable component because it makes it essential to manage a selections object externally to the component (which presumably many use cases don't warrant). But I figure lets cross that bridge when we come to it.

Maybe we could make it handle an 'internal selections' mode - if you don't give it a selections property, it assumes it needs to create an manage its own one - the computed setters either emit the event or update the internal object depending on what mode it's in.. Could limit than change to the changeSelections method.

PR #159, opened 2019-11-15
mrc-679 Barchart metadata from hintr

EmmaLRussell:
Use barchartIndicators and barchartFilters from plottingMetadata from modelRunResult


EmmaLRussell:
This has got more stuff in it that will eventually come out than will stay in! But have highlighted that in the code...

src/app/static/src/app/store/modelOutput/modelOutput.ts
```diff
@@ -9,87 +9,34 @@ export interface ModelOutputState {}
...
+                label: "Quarter"
+            };
+            newFilters = filters.filter((f: any) => f.id != "quarter");
+            newFilters.push(newQuarter);
+
+            filters = newFilters;
+        }
+        const sex = filters.find((f: any) => f.id == "sex");
+        if (!sex){
+            filters.push({
```
hillalex: So is `sex` sometimes not present, sometimes present? I know I was the one who originally suggested hard-coding filters we already knew about but now seems more clear that was the wrong call, can we just rely on it being returned by hintr?

PR #161, opened 2019-11-18
fix cli

hillalex:
Can be confirmed working by:
* uncommenting https://github.com/mrc-ide/hint/pull/161/files#diff-7083312459829a5d01f15b4f2e7dd398R65
* run `./src/userCLI/scripts/build` to build local image
* run `./scripts/test-cli.sh add-user test@exmaple.com password` and `./scripts/test-cli.sh add-user test@exmaple.com` - the former should print "password" and the latter "null"


hillalex:
How exactly could this changeset cause a decrease in codecov :thinking: :roll_eyes:

PR #162, opened 2019-11-18
mrc-586

hillalex:
* moves baseline, surveyAndProgram and modelOptions mutations over to using enums for method names
* extends the plugin that runs after every mutation to reset later steps if any input data, including model options, has been changed

Note that this resetting of data happens without warning; there is a further ticket for prompting the user to confirm before data is reset: https://vimc.myjetbrains.com/youtrack/issue/mrc-721


hillalex:
closing in favour of separate PRs

PR #163, opened 2019-11-19
mrc-706 also delete files when baseline inputs are changed

hillalex:
Sorry, noticed one omission - as well as deleting survey and program files when baseline files are deleted, we should also delete all survey and program files when baseline inputs are *changed*.


hillalex:
> Ah yes, of course.
> 
> Is there now an argument for adding some checks into surveyAndProgram deleteAll to see if the files are actually populated before sending API requests to delete them, just to minimise unnecessary API traffic? Since the most common usage of this is going to be to set all the baseline files first, you're mostly not going to have any surveyAndProgram files to delete yet..

Do you think those extra calls are a big deal? It's all async after all (and it's not like the server is going to be overloaded with traffic ;))

hillalex:
But you're right it'd be dead easy to check, I'll add that in.

PR #164, opened 2019-11-19
mrc-586 preliminary PR

hillalex:
Changes baseline, surveyAndProgram and modelOptions over to using consts for mutations.

This PR should contain no logical changes!

### DIFF
--- a/src/app/static/src/app/components/modelOptions/ModelOptions.vue
+++ b/src/app/static/src/app/components/modelOptions/ModelOptions.vue
@@ -7,7 +7,7 @@
         <dynamic-form v-if="!loading"
                       v-model="modelOptions"
                       submit-text="Validate"
-                      @submit="Validate"></dynamic-form>
+                      @submit="validate"></dynamic-form>
         <h4 v-if="valid" class="mt-3">Options are valid
             <tick color="#e31837" width="20px"></tick>
         </h4>
@@ -21,14 +21,14 @@
     import LoadingSpinner from "../LoadingSpinner.vue";
     import Tick from "../Tick.vue";
 
-    import {mapActionByName, mapMutationsByNames, mapStateProps} from "../../utils";
+    import {mapActionByName, mapMutationByName, mapStateProps} from "../../utils";
     import {ModelOptionsMutation} from "../../store/modelOptions/mutations";
     import {ModelOptionsState} from "../../store/modelOptions/modelOptions";
 
     interface Methods {
         fetchOptions: () => void
-        [ModelOptionsMutation.Validate]: (data: DynamicFormData) => void
-        [ModelOptionsMutation.Update]: (data: DynamicFormMeta) => void
+        validate: (data: DynamicFormData) => void
+        update: (data: DynamicFormMeta) => void
     }
 
     interface Computed {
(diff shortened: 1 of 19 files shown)


src/app/static/src/app/store/baseline/mutations.ts
```diff
@@ -46,46 +50,48 @@ export const mutations: MutationTree<BaselineState> & BaselineMutations = {
...
 
-    BaselineError(state: BaselineState, action: PayloadWithType<string>) {
+    [BaselineMutation.BaselineError](state: BaselineState, action: PayloadWithType<string>) {
         state.validating = false;
         state.validatedConsistent = false;
         state.baselineError = action.payload;
     },
 
-    ...readyStateMutations
+    [BaselineMutation.Ready](state: ReadyState) {
```
EmmaLRussell: It's a bit of a shame if this is going to have to be reimplemented in every module. Could we can a ReadyStateMutations enum and continue to pull in ...readyStateMutations here?

src/app/static/src/app/store/surveyAndProgram/actions.ts
```diff
@@ -74,25 +74,22 @@ export const actions: ActionTree<SurveyAndProgramDataState, RootState> & SurveyA
             .delete("/disease/survey/")
             .then(() => {
                 commit({type: "SurveyUpdated", payload: null});
-                commit({type: "filteredData/Reset", payload: null}, {root: true});
```
EmmaLRussell: Why have these resets been removed? Is it because filteredData updates itself if a file gets deleted, and you actually want to still see the files that remain?

src/app/static/src/app/components/modelOptions/ModelOptions.vue
```diff
@@ -7,25 +7,28 @@
         <dynamic-form v-if="!loading"
                       v-model="modelOptions"
                       submit-text="Validate"
-                      @submit="validate"></dynamic-form>
+                      @submit="Validate"></dynamic-form>
```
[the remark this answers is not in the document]
hillalex: Yeah true, could just name it something else internally. Ok will do

src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -3,13 +3,7 @@ import {BaselineState} from "./baseline";
 import {RootState} from "../../root";
 import {api} from "../../apiService";
 import {PjnzResponse, ValidateBaselineResponse} from "../../generated";
-
-export type BaselineActionTypes = "PJNZUpdated" | "ShapeUpdated" | "PopulationUpdated" | "Validated"
```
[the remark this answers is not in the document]
hillalex: Because we now have the enum values (which correspond exactly to these). It's true could pass "keyof BaselineMutation" as a type to the api service for type checking there, could add that.

src/app/static/src/app/store/surveyAndProgram/actions.ts
hillalex: This was accidental! The next PR does remove them because moves all resetting logic to the plugin, must have crept into the wrong PR. I'll revert here.

src/app/static/src/app/store/baseline/mutations.ts
hillalex: I actually think this is better. It only gets repeated 3 times, it's only a couple of lines of code, and abstracting it into a shared class obscures what's happening a bit.

hillalex:
Would be good to get this and the next PR looked at, as there's more work (showing the confirmation modal) that builds off of these...

PR #166, opened 2019-11-19
mrc-586 resetting state plugin

hillalex:
Contains commits from https://github.com/mrc-ide/hint/pull/164

Moves resetting responsibility into a plugin, adds logic to reset options and outputs, depending on which mutation is called.


src/app/static/src/app/root.ts
```diff
@@ -42,6 +47,34 @@ const persistState = (store: Store<RootState>) => {
...
+    store.subscribe((mutation: MutationPayload, state: RootState) => {
+
+        if (state.baseline.ready
+            && state.surveyAndProgram.ready
+            && state.modelRun.ready) {
+
+            const type = stripNamespace(mutation.type);
+
+            if (BaselineUpdates.includes(type as BaselineMutation)) {
+                store.commit("ResetInputs");
```
hillalex: In fact this `ResetInputs` is just resetting the `filteredData` state (the actions handle all the deletion of input files.) So perhaps this should be renamed.

src/app/static/src/app/store/modelOutput/modelOutput.ts
```diff
@@ -5,7 +5,9 @@ import {FilterOption} from "../../generated";
 
 const namespaced: boolean = true;
 
-export interface ModelOutputState {}
+export interface ModelOutputState {
+    dummyProperty: boolean // unused except for testing
```
hillalex: Why do we have an empty module here, and should the bar chart plotting selections stuff rightly live in this module?

EmmaLRussell: Hmm.. I think you could argue that the selections are different to the model output since they're modified by the user rather than coming from the model output. But maybe it would make sense for the modelRun.result to inhabit modelOutput rather than modelRun - I've always found that a bit confusing, though I can see why it happened.

EmmaLRussell:
Sorry, ended up with a bunch of duplicate comments there as i thought I'd lost them with butterfingers navigate-away, and retyped them, but apparently they were still in there!

src/app/static/src/app/store/modelOutput/modelOutput.ts
hillalex: > One thing I noticed is that if I go back to the options after being on a subsequent step and edit them, the output steps become invalidated as expected, but I don't need to re-validate the options before continuing. Probably should have to though?

Thanks for picking this up, yes this was an oversight. But we're actually going to ditch the validation altogether though I think worth just leaving on this branch as is. Ticket for removing the validation logic - https://vimc.myjetbrains.com/youtrack/issue/mrc-792

> Is there a separate ticket for the issue where the choropleth disappears when you remove survey, but have other surveyandProgram files present? 

Ha, no, this was supposed to be working here! Clearly my tests are not doing their job :woman_facepalming: 

> Oh, and one other little thing I've just noticed, not directly related to this ticket, is that if you have an error with anc or programme, you can still continue to next step, because survey is valid. Maybe this is ok because I can't actually 'remove' the errored file, and maybe I don't actually have a valid one and just hit the wrong upload button. It just seems a little odd to be able to continue with errors. What do you think?

Good question, does seem like we should block continuing if in error state. Added a ticket: https://vimc.myjetbrains.com/youtrack/issue/mrc-791

hillalex:
> One thing I noticed is that if I go back to the options after being on a subsequent step and edit them, the output steps become invalidated as expected, but I don't need to re-validate the options before continuing. Probably should have to though?

Thanks for picking this up, yes this was an oversight. But we're actually going to ditch the validation altogether though I think worth just leaving on this branch as is. Ticket for removing the validation logic - https://vimc.myjetbrains.com/youtrack/issue/mrc-792

> Is there a separate ticket for the issue where the choropleth disappears when you remove survey, but have other surveyandProgram files present? 

Ha, no, this was supposed to be working here! Clearly my tests are not doing their job :woman_facepalming: 

> Oh, and one other little thing I've just noticed, not directly related to this ticket, is that if you have an error with anc or programme, you can still continue to next step, because survey is valid. Maybe this is ok because I can't actually 'remove' the errored file, and maybe I don't actually have a valid one and just hit the wrong upload button. It just seems a little odd to be able to continue with errors. What do you think?

Good question, does seem like we should block continuing if in error state. Added a ticket: https://vimc.myjetbrains.com/youtrack/issue/mrc-791

src/app/static/src/app/store/filteredData/filteredData.ts
```diff
@@ -19,7 +20,6 @@ export interface SelectedChoroplethFilters {
 export interface FilteredDataState {
     selectedDataType: DataType | null
     selectedChoroplethFilters: SelectedChoroplethFilters
-    regionIndicators: {[k: string]: any};
```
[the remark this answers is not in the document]
hillalex: This was a ghost interface property as far as I could tell, never actually referenced or updated - the list of all available region indicators is a property on the baseline module, and selected regions are a property on `selectedChoroplethFilters`. Unless I'm missing something?

src/app/static/src/app/root.ts
```diff
@@ -42,6 +47,34 @@ const persistState = (store: Store<RootState>) => {
...
+const resetState = (store: Store<RootState>) => {
+    store.subscribe((mutation: MutationPayload, state: RootState) => {
+
+        if (state.baseline.ready
+            && state.surveyAndProgram.ready
+            && state.modelRun.ready) {
+
+            const type = stripNamespace(mutation.type);
+
+            if (BaselineUpdates.includes(type as BaselineMutation)) {
```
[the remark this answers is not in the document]
hillalex: Actually I think this is a good idea!

src/app/static/src/app/store/modelOutput/modelOutput.ts
hillalex: I was thinking of the modules as roughly corresponding to steps of the app, not sure there's a reason to distinguish user mutated from api mutated content. Anyway, just something to consider refactoring when we get a chance! I agree, result should also be in modelOutput, it's confusing as is!

src/app/static/src/app/root.ts
hillalex: Ah so no, I've only moved control for committing __mutations__ to the plugin, not for dispatching __actions__. It's not 100% clear to me from the Vuex docs, but I was under the impression plugins should only commit mutations, not dispatch actions. They don't make this very explicit, but they describe a plugin as:

"The plugin simply receives the store as the only argument and can either listen to mutations (for outbound data persistence, logging, or debugging) or dispatch mutations (for inbound data e.g. websockets or observables)." (https://vuex.vuejs.org/api/#plugins)

hillalex: Just noticed that confusingly there they say "dispatch" mutations. Lol.

hillalex:
Thanks for reviewing this so carefully! Have made some updates, could you confirm the resetting filters/selected tab logic is working as expected now?

hillalex:
>  But we're actually going to ditch the validation altogether though I think worth just leaving on this branch as is. Ticket for removing the validation logic - https://vimc.myjetbrains.com/youtrack/issue/mrc-792

Changed my mind about how to handle this - probably makes sense to leave the front-end as is, even though it's not wired up to anything in the back end. So have added resetting logic: https://github.com/mrc-ide/hint/pull/166/commits/4bc7a08958ad4ebf276a721a6d7ddfebab31abcb

src/app/static/src/app/store/filteredData/filteredData.ts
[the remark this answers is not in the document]
EmmaLRussell: Can't remember, probably got used once upon a time then superseded

hillalex:
> There is still an issue with removing files, which is that you can't remove a file and then re-add the same file. Happy for that to go into a separate bug ticket though!

ugh :woman_facepalming: added https://vimc.myjetbrains.com/youtrack/issue/mrc-1124

PR #167, opened 2019-11-20
mrc-737 passing by reference bug

hillalex:
This moves over all intial data states to be functions rather than objects, so that we always get a fresh copy rather than shallow cloning.

Also adds some logic to the root mutations tests to make them a bit more robust, as that's where I first discovered this bug.

Many of the modules aren't currently vulnerable to this bug because their states are shallow objects, but I've moved over everything to this new pattern for consistency.

### DIFF
--- a/src/app/static/src/app/store/plottingSelections/plottingSelections.ts
+++ b/src/app/static/src/app/store/plottingSelections/plottingSelections.ts
@@ -15,16 +15,18 @@ export interface BarchartSelections {
     selectedFilterOptions:  { [key: string]: FilterOption[] }
 }
 
-export const initialBarchartSelections = {
-    indicatorId: "",
-    xAxisId: "",
-    disaggregateById: "",
-    selectedFilterOptions: {}
+export const initialBarchartSelections = (): BarchartSelections => {
+    return {
+        indicatorId: "",
+        xAxisId: "",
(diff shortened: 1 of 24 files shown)


hillalex:
Sorry hold off reviewing for 5 mins!

hillalex:
Ready for review now

src/app/static/src/app/store/plottingSelections/plottingSelections.ts
```diff
@@ -22,16 +22,18 @@ export const initialBarchartSelections = {
     selectedFilterOptions: {}
 };
 
-export const initialPlottingSelectionsState = {
-  barchart: {...initialBarchartSelections}
+export const initialPlottingSelectionsState = (): PlottingSelectionsState => {
+    return {
+        barchart: {...initialBarchartSelections}
```
EmmaLRussell: I think technically we'd get away with this because the selectedFilterOptions are replaced rather than modified when they're initialised by the Barchart component. But might be safer to make initiaBarchartSelections a function too, same as initialSelectedChoroplethFilters.

hillalex: Thanks, agree good to be complete, it's future proofing if nothing else.

PR #168, opened 2019-11-20
mrc-718 Logging errors

EmmaLRussell:
The easiest way to manually test the logging in this is to upload a duff file, and see the error logged in the bootRun console.


src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
```diff
@@ -18,21 +19,22 @@ import javax.validation.ConstraintViolationException
...
-class HintExceptionHandler : ResponseEntityExceptionHandler() {
+class HintExceptionHandler(private val errorCodeGenerator: ErrorCodeGenerator,
+                           private val appProperties: AppProperties) : ResponseEntityExceptionHandler() {
 
     override fun handleExceptionInternal(e: java.lang.Exception,
                                          @Nullable body: Any?,
                                          headers: HttpHeaders,
                                          status: HttpStatus,
                                          request: WebRequest): ResponseEntity<Any> {
         logger.error(e.message)
```
EmmaLRussell: So this ExceptionHandler is using its inherited logger to log out any exception messages anyway. Maybe this is redundant now? Maybe the Filter should be using a configured logger rather than defaulting to system log?

src/app/src/main/resources/config.properties
```diff
@@ -12,3 +12,4 @@ hintr_url=http://localhost:8888
 db_user=hintuser
 db_password=changeme
 db_url=jdbc:postgresql://localhost/hint
+support_email=naomi-support@imperial.ac.uk
```
EmmaLRussell: We need to decide what this is actually going to be!

EmmaLRussell:
Codecov is failing because the PSQL exception handling isn't being tested. Can't really think of a way to force a SQL error, can you?

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
```diff
@@ -18,21 +19,22 @@ import javax.validation.ConstraintViolationException
...
-        return ErrorDetail(status, e.message ?: "Something went wrong").toResponseEntity()
+        return ErrorDetail(status, (e.message ?: "Something went wrong").appendErrorCodeInstructions()).toResponseEntity()
     }
 
     @ExceptionHandler(HintException::class)
     protected fun handleHintException(e: HintException): ResponseEntity<Any> {
         logger.error(e.message)
-        return ErrorDetail(e.httpStatus, e.message!!).toResponseEntity()
+        return ErrorDetail(e.httpStatus, e.message!!.appendErrorCodeInstructions()).toResponseEntity()
```
hillalex: I don't think these want the error code instructions appended as these are known errors that contain their own instructions - e.g. "You must upload a shape file before uploading survey or programme data".

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
hillalex: I guess a log factory logger might be better for the filter, so that the error messages are formatted the same throughout the app. Not too worried about redundancy and in some instances this is logging something different to the filter: https://github.com/mrc-ide/hint/pull/168/files#diff-2a06c226eeb55efc9a48ced0a230f53bL46

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
```diff
@@ -44,7 +46,14 @@ class HintExceptionHandler : ResponseEntityExceptionHandler() {
...
         logger.error(e.message)
-        return ErrorDetail(HttpStatus.INTERNAL_SERVER_ERROR, "An unexpected error occurred.").toResponseEntity()
+        return ErrorDetail(HttpStatus.INTERNAL_SERVER_ERROR, "An unexpected error occurred.".appendErrorCodeInstructions())
+                .toResponseEntity()
+    }
+
+    private fun String.appendErrorCodeInstructions(): String {
+        val code = errorCodeGenerator.newCode()
+        val fullStopIfNeeded =  if (this.endsWith('.')) ""  else "."
+        return "$this$fullStopIfNeeded Please contact support at ${appProperties.supportEmail} and quote this code: $code"
```
hillalex: Shall we change this to: "If you see this message while you are using naomi a workshop, please contact your workshop technical support and show them this code. Otherwise <same as previous>"

src/app/src/main/kotlin/org/imperial/mrc/hint/logging/ErrorLoggingFilter.kt
```diff
@@ -0,0 +1,44 @@
...
+import javax.servlet.http.HttpServletResponse
+import org.springframework.web.util.ContentCachingResponseWrapper
+import java.nio.charset.Charset
+
+@Component
+class ErrorLoggingFilter(private val logMethod: (msg: String) -> Unit = ::println): Filter {
+    override fun doFilter(request: ServletRequest, response: ServletResponse, chain: FilterChain) {
+        //ContentCachingResponseWrapper doesn't play nicely with streamed data, so don't use wrapper for a download.
+        //We'll still be able to log any status errors
+        val isDownload = (request as HttpServletRequest).servletPath.startsWith("/download")
```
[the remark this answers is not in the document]
hillalex: I see, the headers aren't set til after the chain is called. Ok.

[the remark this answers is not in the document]
EmmaLRussell: Yep!

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
EmmaLRussell: I think that might be the only case where that's true! (And it shouldn't be possible to hit it in the app anyway.) The messages for TokenException and SessionException are fairly cryptic. None of these exceptions should really happen (it's not like uploading a bad file, which is a 400), so I guess the point is more the error code than the instructions, so we can find the place in the log where they happened.

hillalex: But these are all known exceptions - i.e. we throw them manually. If they're cryptic, we should improve the error messages.

PR #170, opened 2019-11-21
mrc-713 updated data schemas

hillalex:
* regenerates types
* updates test data
* splits up the large `getters_regionIndicators` test file into separate test suites for each data type, to make the tests easier to reason about. also we now have complete typings for the return data, so am using that in those tests for type checking.
* updates code and tests to be consistent with new types

The latter is mostly just renaming "quarter" to "year" and "age_group_id" to "age_group". The one logical change is from deriving the region level from the id, to reading it explicitly from the "area_level" property in the geojson.

~~Contains commits from https://github.com/mrc-ide/hint/pull/171 so please merge that first.~~


src/app/static/src/app/apiService.ts
```diff
@@ -7,26 +7,14 @@ import {
...
+    Response, SessionFile,
     ShapeResponse,
-    SurveyResponse,
+    SurveyResponse, ValidateBaselineResponse,
     ValidateInputResponse
 } from "./generated";
 import {Commit} from "vuex";
 import {freezer} from "./utils";
-
-type ResponseData =
```
hillalex: This was pointless and incorrect, since we get all sorts of response back from the API.

src/app/static/src/app/store/filteredData/getters.ts
```diff
@@ -34,14 +34,12 @@ export const getters = {
                         quarter: undefined
                     } : null;
             case (DataType.Output):
-                return rootState.modelRun.result ?
-                    {
-                        ...rootState.modelRun.result.filters,
```
hillalex: There's been a regression in hintr and these filters are not actually present (on master as well as this branch). We should re-add once they are back in hintr, thought they will likely be under a different property name (something like `modelRun.result.choropleth.filters`)

src/app/static/src/tests/filteredData/getters.test.ts
```diff
@@ -16,24 +16,8 @@ import {
...
-        choroplethRanges: {
-            prev: {min: 0, max: 1},
-            art: {min: 0, max: 1}
-        }
-    } as any;
-
-    self.excludeRow = getters.excludeRow(state, self, rootState);
-    return self;
+export function testGetters(state: FilteredDataState) {
+    return {excludeRow: getters.excludeRow(state)};
```
hillalex: The other functions being set up (colorFunctions and choroplethRanges) were out of date/obsolete.

src/app/static/src/tests/filteredData/getters.test.ts
```diff
@@ -175,105 +158,6 @@ describe("FilteredData getters", () => {
         expect(filters).toBeNull();
     });
 
-    it("gets unfilteredData when selectedDataType is Survey", () => {
```
hillalex: These tests moved to own file https://github.com/mrc-ide/hint/pull/170/files#diff-62db21a33d4682b884efe0101fc4ef1e

src/app/static/src/app/components/plots/ChoroplethFilters.vue
```diff
@@ -28,12 +28,12 @@
...
 
         <div class="py-2" v-if="!isOutput">
-            <filter-select label="Quarter"
+            <filter-select label="Year"
                            :multiple="false"
-                           :options="quarterFilters.available"
-                           :value="quarterFilters.selected"
-                           :disabled="quarterFilters.disabled"
-                           @select="selectQuarter"></filter-select>
+                           :options="yearFilters.available"
```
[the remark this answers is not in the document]
hillalex: Ah so we'll have both year and quarter? At the moment no filters are coming back for output, I believe @r-ash has an open ticket for fixing this, at which point we should add whatever is missing.

hillalex:
> Looks good. Age doesn't seem to be coming back from hintr as a filter for ANC, is that expected? (Age_group is in the ANC data.)
> 
> Age also shows as 'Not Used' in the choro for Output - I think that's because as per your comment, those filters aren't currently being returned from the modelRun result.
> 
> Thanks for splitting up those tests, they seem much more manageable now..

Yeah that's why the age filter is disabled for output, it's still rendered in the view but missing in the state. Not sure about ANC - is that a bug @r-ash ?

src/app/static/src/app/components/plots/ChoroplethFilters.vue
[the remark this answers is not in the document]
r-ash: https://vimc.myjetbrains.com/youtrack/issue/mrc-746 is the back end ticket where I have updated the filters. That fixes the bug with the missing choropleth filters - I think if any issues with the filters aren't blocking you then let's handle those in a separate ticket?https://vimc.myjetbrains.com/youtrack/issue/mrc-781

If there are problems with missing filters which are blocking merging then let me know and I can try and move those fixes from mrc-746 to here.

r-ash:
> > Looks good. Age doesn't seem to be coming back from hintr as a filter for ANC, is that expected? (Age_group is in the ANC data.)
> > Age also shows as 'Not Used' in the choro for Output - I think that's because as per your comment, those filters aren't currently being returned from the modelRun result.
> > Thanks for splitting up those tests, they seem much more manageable now..
> 
> Yeah that's why the age filter is disabled for output, it's still rendered in the view but missing in the state. Not sure about ANC - is that a bug @r-ash ?

Yep probably a bug if those are missing - see note in https://github.com/mrc-ide/hint/pull/170#discussion_r349560978 it should be fixed in https://vimc.myjetbrains.com/youtrack/issue/mrc-746

src/app/static/src/app/store/filteredData/getters.ts
hillalex: No, separate.

hillalex:
Ok, thanks @r-ash. The missing filters are not a blocker, they can be fixed iteratively in further branches.

EmmaLRussell:
@hillalex  Do you want to add the Quarter filter for Choro output in this branch or shall we put it in a separate one? If separate, I think that was the only outstanding issue, so this branch can be merged along with the hintr branch.

hillalex:
> @hillalex Do you want to add the Quarter filter for Choro output in this branch or shall we put it in a separate one? If separate, I think that was the only outstanding issue, so this branch can be merged along with the hintr branch.

I def think separate, this PR is big enough as it is and that is new logic rather than just making things work with the latest hintr.

PR #172, opened 2019-11-22
mrc-773 allow deleting of invalid files

hillalex:
Show the remove button when a file is valid or invalid.


hillalex:
> This change works fine for showing remove button, still seeing file names being left behind in the input box after removal though. Did you get to the bottom of why that was happening?

Yes there's a separate PR open for that bug: https://github.com/mrc-ide/hint/pull/169

PR #173, opened 2019-11-22
mrc-695 save version info with model options and pass back to hintr

hillalex:
* updates backend to expect and pass through version info to the model run
* saves version info when model options are requested
* passes version info to backend on model run


src/app/static/src/app/apiService.ts
```diff
@@ -95,13 +76,13 @@ export class APIService<S extends string, E extends string> implements API<S, E>
...
-        return promise.then((response: AxiosResponse) => {
-            const success = response && response.data;
+        return promise.then((axioxResponse: AxiosResponse) => {
+            const success = axioxResponse && axioxResponse.data;
             const data = success.data;
             if (this._onSuccess) {
                 this._onSuccess(data);
             }
-            return data;
+            return axioxResponse.data;
```
hillalex: I can't remember which PR (it was quite recent) that I changed this to return only the data part of 
```{status: "success", errors: [], data: whatever}```
but I now couldn't see it being used anywhere. Anyway we now need to revert it back to returning the whole object so that we can pull out the version info.

File: src/app/static/src/app/generated.d.ts [hunk 2]
```diff
@@ -28,6 +28,50 @@ export type AncResponseData = {
   art_coverage?: number;
   [k: string]: any;
 }[];
+export interface BarchartIndicator {
```
hillalex: It seems the typings were actually quite out of date, so lots of unrelated changes here!

src/app/static/src/app/store/modelRun/mutations.ts
```diff
@@ -1,46 +1,55 @@
...
+    [ModelRunMutation.RunResultError](state: ModelRunState, action: PayloadWithType<string>) {
         state.errors.push(action.payload);
     },
 
-    ...readyStateMutations
+    [ModelRunMutation.Ready](state: ModelRunState) {
+        state.ready = true;
+    },
+
+    [ModelRunMutation.ModelRunError](state: ModelRunState, action: PayloadWithType<string>) {
```
hillalex: I noticed that we were telling the api to commit these mutations on error, but they didn't actually exist! So I've migrated this mutation tree over to using the enum pattern, and I added a helper test method that checks that all the enum values correspond to methods: https://github.com/mrc-ide/hint/pull/173/files#diff-4f54fb77cd41d6dd4f4bed5bdac03397

src/app/static/src/tests/mocks.ts
```diff
@@ -277,28 +282,3 @@ export const mockPlottingMetadataResponse = (props: Partial<PlottingMetadataResp
         ...props
     }
 };
-
-export interface Indicators {
```
hillalex: Just noticed these were unused

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintrAPIClientTests.kt
```diff
@@ -47,28 +46,21 @@ class HintrApiClientTests {
     @Test
     fun `can submit model run`() {
         val sut = HintrAPIClient(ConfiguredAppProperties(), ObjectMapper())
-        val result = sut.submit(mapOf(), mapOf())
-        assertThat(result.statusCodeValue).isEqualTo(200)
-        JSONValidator().validateSuccess(result.body!!, "ModelSubmitResponse")
+        val result = sut.submit(emptyMap(), ModelRunOptions(emptyMap(), emptyMap()))
+        assertThat(result.statusCodeValue).isEqualTo(400)
+        JSONValidator().validateError(result.body!!, "VERSION_OUT_OF_DATE")
```
hillalex: Here just confirming that we're hitting hte right endpoint. In ModelRunTests we are more complete and test the happy path.

src/config/hintr_version
```diff
@@ -1 +1 @@
-mrc-577
+mrc-693
```
hillalex: After this is merged we can return hintr to master.

src/app/static/src/app/store/modelOptions/modelOptions.ts
```diff
@@ -4,20 +4,23 @@ import {DynamicFormData, DynamicFormMeta} from "../../components/forms/types";
...
 }
 
 export const initialModelOptionsState = (): ModelOptionsState => {
     return {
         optionsFormMeta: {controlSections: []},
         options: {},
         valid: false,
-        fetching: false
+        fetching: false,
+        version: {hintr: "unknown", naomi: "unknown", rrq: "unknown"}
```
EmmaLRussell: If we're expecting particular property names in VersionInfo in the front end, should there be a correspondingly strong type in the back end, rather than just Map<String, String>? I guess the backend doesn't care much what the properties are and will just let hintr error if the format is wrong.

src/app/static/src/app/store/modelRun/mutations.ts
```diff
@@ -1,46 +1,55 @@
...
+        state.errors.push(action.payload);
+    },
+
+    [ModelRunMutation.RunStatusError](state: ModelRunState, action: PayloadWithType<string>) {
+        state.errors.push(action.payload);
```
EmmaLRussell: Maybe we need to be careful about this - if something goes wrong and we start getting an error on every poll request, this array is going to get huge with no obvious way to stop it. Should we stop polling if we get an error? I don't think we do at the moment.

EmmaLRussell: As thing stand, if you got one error during polling then the run successfully completed sometime later, the user wouldn't be able to proceed to subsequent steps anyway, as there would be an error in the list.

hillalex: Good point, not sure how to handle this. If the status endpoint errors, it might actually go on indefinitely. But we also don't want to halt the polling on a single failure. Maybe we need to do something clever like if we get 5 consecutive errors, halt and report the last one? I might just revert this here and leave for the subsequent ticket about reporting model run errors if that's ok?

src/app/static/src/app/store/modelOptions/modelOptions.ts
hillalex: Well the front-end is the bit that is actually hooked up to the schema - i.e. we don't have a way of auto-generating kotlin models from the schema. So this way is more flexible - if the schema changes, just have to regenerate the types and the backend will still work.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintrAPIClientTests.kt
```diff
@@ -47,28 +46,21 @@ class HintrApiClientTests {
...
     }
 
     @Test
     fun `can get model run status`() {
         val sut = HintrAPIClient(ConfiguredAppProperties(), ObjectMapper())
-        val submitResult = sut.submit(mapOf(), mapOf())
-
-        val id = ObjectMapper().readValue<JsonNode>(submitResult.body!!)["data"]["id"].textValue()
-        val result = sut.getStatus(id)
+        val result = sut.getStatus("1234")
```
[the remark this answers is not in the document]
hillalex: Apparently not. Maybe it should?

src/app/static/src/app/store/modelRun/mutations.ts
EmmaLRussell: That's fine, I'm doing that ticket now and there's not much else to it. Will put in 'five strikes and you're out' into that. And I guess if you get polling errors but then it recovers, it will clear the errors so that the stepper can proceed.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintrAPIClientTests.kt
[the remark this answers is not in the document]
EmmaLRussell: @r-ash Any thoughts on this? Run Status polling doesn't return an error response if you pass it a nonexistent polling id.

EmmaLRussell:
I merged with master and this branch has now failed because master now requires hintr mrc-746 (data format changes) whereas this branch required mrc-693 (model run versions). I've chased Rob about getting at least one of these branches merged...

hillalex:
I *think* mrc-693 of hintr has the data format changes merged into it, so pointing this back to mrc-693 should result in tests passing.

PR #174, opened 2019-11-22
mrc-717 errors module

EmmaLRussell:


src/app/static/src/app/apiService.ts
```diff
@@ -116,13 +127,17 @@ export class APIService<S extends string, E extends string> implements API<S, E>
         if (!failure || !failure.status) {
             throw new Error("Could not parse API response");
         }
-        if (this._onError) {
+        else if (this._onError) {
             this._onError(failure);
         } else {
-            throw new Error(APIService.getFirstErrorFromFailure(failure));
+            this._addErrorToStore(APIService.getFirstErrorFromFailure(failure));
```
EmmaLRussell: I'm including a commit to errors here so we can see the module in action - any remaining unhandled throws will be added to the module in mrc-719. You can force this error condition by killing the hintr container then attempting to upload a file. The files coming out of the backend are pretty useless at the moment, but should include an error code to quote once https://github.com/mrc-ide/hint/pull/168 is merged.

EmmaLRussell: We could also include the support email address in the header message when the backend PR is merged.

src/app/static/src/app/components/Errors.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+    <div v-if="hasErrors" class="container">
+        <div class="content">
+            <div class="alert alert-danger alert-dismissible fade-show" role="alert">
+                <p>The following errors occurred. Please contact {{title}} support if this problem persists.</p>
```
hillalex: Duplicating instructions here? i.e. "Please contact... etc" will already be encompassed in the API error message? I think just remove this bit.

src/app/static/src/app/apiService.ts
```diff
@@ -49,8 +51,18 @@ export class APIService<S extends string, E extends string> implements API<S, E>
...
     static getFirstErrorFromFailure = (failure: Response) => {
-        const firstError = failure.errors[0];
-        return firstError.detail ? firstError.detail : firstError.error;
+        const firstError = failure.errors && failure.errors[0];
+        let result;
+        if (firstError) {
+            result =  firstError.detail ? firstError.detail : firstError.error;
+        }
+
+        if (!result)
```
hillalex: The idea is that we have already checked that the API response is correctly formatted before calling this method, and thrown an error saying "Could not parse API response" if not.  So no need to do all these null checks here. That check happens [here](https://github.com/mrc-ide/hint/pull/174/files#diff-0f0781cfcf5e78654409fce90c2c466fR116). Could make it more complete though I agree, by checking that `errors` exists and is non-empty/correctly formatted.

src/app/static/src/app/components/Errors.vue
EmmaLRussell: The 'Please contact...' bit would be there if the error had originated in an exception in the backend, I guess there are other error cases that might need to be displayed in the front end though?

src/app/static/src/app/apiService.ts
EmmaLRussell: Yeah, I put that in because errors was undefined in my test case (which I think I'd forced by killing the hintr) container. Good idea to move that check to the format checking part.

src/app/static/src/app/components/Errors.vue
hillalex: Yeah, I've just made the front-end errors a bit more descriptive: https://github.com/mrc-ide/hint/pull/175/files#diff-0f0781cfcf5e78654409fce90c2c466fL55

They don't have the app name or email in there though - but we could get that in by having them as variables on the page rendered by freemarker?

PR #182, opened 2019-11-27
mrc-781 Re-order choropleth filters, added area to start of barchart filters

EmmaLRussell:
As specified in ticket https://vimc.myjetbrains.com/youtrack/issue/mrc-781 (see comments too)


src/app/static/src/tests/components/plots/choroplethFilter.test.ts
```diff
@@ -476,14 +476,25 @@ describe("ChoroplethFilters component", () => {
             .toStrictEqual([FilterType.Quarter, "1"]);
     });
 
-    it("shows quarter filter if data type is output", () => {
```
EmmaLRussell: didn't need these tests any more as now we're testing that all the filters are there in the right order

PR #186, opened 2019-11-27
mrc-721 confirm editing files

hillalex:
This PR adds the confirmation modal for the file uploads. See what you think of the modal content itself, as well as the code. Could format the steps being reset as a bullet pointed list perhaps.

 I've just got 2 call to actions - cancel or continue, rather than adding any save or download buttons into the modal. My thinking was that:
a) would be too many calls to action
b) context is important for meaning and it might be confusing/ambiguous what the various save/download buttons mean if they appear outside of their normal context

Contains commits from https://github.com/mrc-ide/hint/pull/185 so please merge that one first.

### DIFF
--- a/src/app/static/src/app/components/FileUpload.vue
+++ b/src/app/static/src/app/components/FileUpload.vue
@@ -2,7 +2,7 @@
     <div class="form-group">
         <label class="font-weight-bold">{{label}}</label>
         <tick color="#e31837" v-if="valid" width="20px"></tick>
-        <a v-if="showRemove" class="small float-right" href="#" v-on:click="remove">remove</a>
+        <a v-if="showRemove" class="small float-right" href="#" v-on:click="handleFileDelete">remove</a>
         <loading-spinner v-if="uploading" size="xs"></loading-spinner>
         <slot></slot>
         <div class="custom-file">
@@ -20,6 +20,12 @@
             </label>
         </div>
         <error-alert v-if="hasError" :message="error"></error-alert>
+        <reset-confirmation :continue-editing="uploadSelectedFile"
+                            :cancel-editing="cancelEdit"
+                            :open="showUploadConfirmation"></reset-confirmation>
+        <reset-confirmation :continue-editing="deleteSelectedFile"
+                            :cancel-editing="cancelEdit"
+                            :open="showDeleteConfirmation"></reset-confirmation>
     </div>
 </template>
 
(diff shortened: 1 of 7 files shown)


src/app/static/src/app/components/FileUpload.vue
```diff
@@ -80,32 +96,47 @@
...
             },
-            showRemove: function() {
+            showRemove: function () {
                 return this.hasError || this.valid
-            }
+            },
+            editsRequireConfirmation: mapGetterByName("stepper", "editsRequireConfirmation")
         },
         methods: {
-            handleFileSelect(_: Event, files: FileList | null) {
```
hillalex: I've removed this slightly weird thing we were doing of optionally passing a file list to this method. It was used only for testing, but since you innovated a much better way of testing file inputs it was actually obsolete. I updated all the existing tests to use your technique.

EmmaLRussell: Are we not going to have a future situation where people might want to upload multiple files of a type? Or are they always going to be zipped? I've kind of lost track of that whole debate...

src/app/static/src/app/store/stepper/getters.ts
```diff
@@ -20,5 +22,12 @@ export const getters: StepperGetters & GetterTree<StepperState, RootState> = {
             5: rootGetters['modelRun/complete'],
             6: false
         }
+    },
+    laterCompleteSteps: (state: StepperState, getters: any) => {
```
EmmaLRussell: I suppose a limitation of this is that it won't prevent you from clearing a later partially completed step. Not sure there's an obvious way to prevent this. Could mark a state as partially complete when it gets saved to local storage and isn't identical to its initial state? Might not be worth it. But user might be surprised at being warned about some changes being lost but not others..

hillalex: Yeah I did consider this. We could have a property "edited" or something like that, but overall I think not worth it. The only step this applies to currently is the model options - if you have filled out some model options but not validated them, they will be reset without warning. But I think this is a minor inconvenience.

hillalex:
> Works great. I like the descriptive dialog. Might modify to "Changing this file.." to make it really clear what it means.
> 
> "Being reset" is maybe not very meaningful, and if you have a lot of subsequent steps comes slightly awkwardly at the end of quite a long sentence. Could maybe change the message to something like: "If you change this file, the following steps will be cleared of all settings: ..."
> 
> Or, since the button says 'Discard', could be 'the following steps will be discarded'.
> 
> I suppose if you wanted to be super-helpful, you could add a little 'How do I save my work?' link which expands to instructions on saving a file or downloading results. Particularly since there isn't a 'save' button to be seen anywhere. Maybe another ticket, or maybe not worth it..?

Good idea on the language, I'll have a go at making that clearer :+1: (I think for now leave the save instructions, but add a ticket as a could have/see what user feedback we get about this?)

src/app/static/src/app/components/FileUpload.vue
hillalex: No we've gone with zip of zips!

hillalex: (But in any case more than this would have to change to support multiple)

PR #187, opened 2019-11-27
mrc-1147 auto-version css and js

hillalex:
So that every time we deploy, users get the fresh js and css.

* move templates to static folder
* use webpack to version bundles and inject them into templates
* use gulp to append versions to css and inject them into template heads


hillalex:
> I don't quite understand why the templates themselves need to go into the static folder, if the bundles are going to get versioned and the version included in the generated html..? Oh, is that where they need to be for webpack to inject the versioned js bundle name?
> 
> Was the favicon removed in error?



> I don't quite understand why the templates themselves need to go into the static folder, if the bundles are going to get versioned and the version included in the generated html..? Oh, is that where they need to be for webpack to inject the versioned js bundle name?
> 
> Was the favicon removed in error?

Yes favicon was a mistake, thanks! Pre-compiled templates could stay where they were, but now that they're more intimately involved with the front-end builders seemed clearer to have them all relative to the same directory root. Not sure whether possible to get the "compiled" (i.e. with js injected) templates to end up in a different location to the bundles, certainly this setup requires the least faffing with webpack, although could always have a further gulp task to move them wherever we wanted. Is there any reason they should live in their own directory though?

PR #193, opened 2019-12-03
mrc-725 Get barchart default selections from model run result

EmmaLRussell:


src/app/static/src/app/store/modelRun/actions.ts
```diff
@@ -45,6 +46,21 @@ export const actions: ActionTree<ModelRunState, RootState> & ModelRunActions = {
                 .withError(ModelRunMutation.RunResultError)
                 .freezeResponse()
                 .get<ModelResultResponse>(`/model/result/${state.modelRunId}`)
+                .then(() => {
+                    if (state.result && state.result.plottingMetadata.barchart.defaults) {
+                        const defaults = state.result.plottingMetadata.barchart.defaults;
+                        commit({
+                                type: "plottingSelections/updateBarchartSelections",
+                                payload: {
```
EmmaLRussell: Hintr's new BarchartDefaults schema doesn't quite match our existing BarchartSelections type, so you get this mapping. They could be more integrated, though it feels like defaults and selections (which can subsequently change) are slightly different things, though they have the same structure..

src/config/hintr_version
```diff
@@ -1 +1 @@
-master
+mrc-1160
```
hillalex: Looks like this is merged so can return this to master.

src/app/static/src/app/store/modelRun/actions.ts
hillalex: Ultimately I would like all the hintr responses to be camelCased as is standard for JSON, but not a big deal.

PR #195, opened 2019-12-03
mrc-1162 Do not fetch plotting metadata if error fetching pjnz

EmmaLRussell:


src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -39,7 +41,7 @@ export const actions: ActionTree<BaselineState, RootState> & BaselineActions = {
             .withSuccess(BaselineMutation.ShapeUpdated)
             .withError(BaselineMutation.ShapeUploadError)
             .freezeResponse()
-            .postAndReturn<PjnzResponse>("/baseline/shape/", formData)
+            .postAndReturn<ShapeResponse>("/baseline/shape/", formData)
```
EmmaLRussell: Just noticed these had the wrong type

src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -27,7 +27,9 @@ export const actions: ActionTree<BaselineState, RootState> & BaselineActions = {
             .freezeResponse()
             .postAndReturn<PjnzResponse>("/baseline/pjnz/", formData)
             .then(() => {
-                dispatch('metadata/getPlottingMetadata', state.iso3, {root: true});
+                if (!state.baselineError && state.iso3) {
+                    dispatch('metadata/getPlottingMetadata', state.iso3, {root: true});
+                }
                 dispatch('validate');
```
EmmaLRussell: I guess we don't really need to validate either since we know there's nothing to validate against - have left as is but could change this.

PR #196, opened 2019-12-03
mrc-583: Add partner logos to login screen

r-ash:
This has been a long time coming

This PR will

* Add logos with links to web pages for partners

Still a question of whether we should add Fjelltopp & Avenir here - I can ask Jeff & Mary and make that change.

The test is bit odd I feel - let me know if there is a better way to check that.


EmmaLRussell:
Do you what's going om with the Travis build failure?

EmmaLRussell:
Need to commit the images too? :)

hillalex:
Looks like one more typo in the tests @r-ash

hillalex:
Sorry @r-ash still need to fix that test

PR #197, opened 2019-12-03
mrc-617 progress bars

hillalex:
It takes a second or 2 for the progress indicators to be returned with the model run status, so this also adds a loading spinner until then with the text "initialising model run".


src/app/static/gulpfile.js
```diff
@@ -39,4 +40,8 @@ gulp.task('copy-templates', function () {
         .pipe(gulp.dest('public'));
 });
 
-gulp.task('styles', gulp.series(gulp.parallel('sass', 'cssmap'), 'inject'));
+gulp.task('clean-css', function () {
+    return del('public/css/**', {force: true});
```
hillalex: clean the css folder before compiling the new css to avoid weird clashes and files being included twice! only happens in dev because in docker it's always compiled fresh.

PR #198, opened 2019-12-04
Add report bug link to header

r-ash:
This adds a link which opens in another tab

I put some margin in the vue file on the a tag - I can move that to a css class if that is better practice? Not sure what should be done here.

One question - I am seeing some styling issues on the stepper see
image

Not sure whether that is related to my change or it is some local development environment issue.

### DIFF
--- a/src/app/static/src/app/components/UserHeader.vue
+++ b/src/app/static/src/app/components/UserHeader.vue
@@ -23,7 +23,7 @@
                         <input type="file" style="display: none;" ref="loadFile" v-on:change="load" accept=".json">
                     </div>
                 </div>
-                <a href="https://forms.gle/QxCT1b4ScLqKPg6a7" target="_blank" style="margin-right: 10px">Report a bug</a>
+                <a href="https://forms.gle/QxCT1b4ScLqKPg6a7" target="_blank" class="pr-2 mr-2 border-right">Report a bug</a>
                 <a href="/logout">Logout</a>
             </div>
         </nav>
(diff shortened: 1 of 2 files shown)


hillalex:
> This adds a link which opens in another tab
> 
> I put some margin in the vue file on the a tag - I can move that to a css class if that is better practice? Not sure what should be done here.
> 
> One question - I am seeing some styling issues on the stepper see
> image
> 
> Not sure whether that is related to my change or it is some local development environment issue.

Those look like old style rules, can you hard refresh?

src/app/static/src/app/components/UserHeader.vue
```diff
@@ -23,6 +23,7 @@
                         <input type="file" style="display: none;" ref="loadFile" v-on:change="load" accept=".json">
                     </div>
                 </div>
+                <a href="https://forms.gle/QxCT1b4ScLqKPg6a7" target="_blank" style="margin-right: 10px">Report a bug</a>
```
hillalex: Bootstrap has a bunch of extremely useful spacing utilities that can be used here instead of the inline style. And how about putting a right border also so there's a clear separation between this and the logout link? So the classes would be "pr-2 mr-2 border-right"

For reference: https://getbootstrap.com/docs/4.4/utilities/spacing/

hillalex:
Oh, re the styles, if you're running locally try deleting the contents of `public/css` before re-building.

r-ash:
> Oh, re the styles, if you're running locally try deleting the contents of `public/css` before re-building.

That fixed it - thanks

r-ash:
Fixed the margin and spacing. That is a nice divider!

PR #199, opened 2019-12-04
Mrc 1172 Show trace for errors which have it

EmmaLRussell:
This is just required for the modelRun just now, but trace is in the Error schema, so I've taken the approach of changing all errors from strings to Errors, and getting the ErrorAlert to show trace if it's available in the Error. 

We might want to limit this in future e.g. explicitly don't show trace in password components. But it's only coming back from the server for model run responses at the moment anyway!

The easiest way to test manually is to comment out `-e USE_MOCK_MODEL=true \` in run-dependencies.sh


src/app/static/src/app/apiService.ts
```diff
@@ -41,12 +41,18 @@ export class APIService<S extends string, E extends string> implements API<S, E>
 
     static getFirstErrorFromFailure = (failure: Response) => {
         if (failure.errors.length == 0) {
-            return "API response failed but did not contain any error information. Please contact support."
+            return APIService.createError("API response failed but did not contain any error information. Please contact support.");
```
hillalex: This feels more like an error detail than an error code (and similarly below).

src/app/static/src/app/components/ErrorAlert.vue
```diff
@@ -1,11 +1,63 @@
...
-        props: ["message"]
+    import Vue from "vue";
+    import {Error} from "../generated.d";
+
+    interface Data {
+        showTrace: Boolean
+    }
+
+    interface Props {
+        error: Error | null
```
hillalex: This seems a bit odd, so the component might exist even though the error is null? The only place I can spot this happening is https://github.com/mrc-ide/hint/pull/199/files#diff-4c86fe55240bb0dc3f3348916613943dR19 - if we add a `v-if` there then can remove null checking logic from this component. 

Ideally the component would just not render if the error is null. It would be easy with `redux/recompose`! I think there is a functional syntax for vue components, maybe once this project time pressure is over we should look into it.

src/app/static/src/app/components/ErrorAlert.vue
```diff
@@ -1,11 +1,63 @@
...
+            message: function() {
+                return this.error ?
+                    (this.error.detail ? this.error.detail : this.error.error) :
+                    "";
+            },
+            hasTrace: function() {
+                return !!this.error && !!this.error.trace && this.error.trace.length > 0
+            },
+            traceLinkText: function() {
+                return (this.showTrace ? "hide" : "show") + " details";
```
hillalex: Maybe "show stack trace"? Show details sounds kind of friendly, and this is not a friendly thing to look at :laughing:

src/app/static/src/app/components/ErrorAlert.vue
```diff
@@ -1,11 +1,63 @@
 <template>
     <div class="pt-1 text-danger">
-        {{message}}
+        <div class="error-message">{{message}}</div>
+        <div v-if="hasTrace" >
+            <a href="#" @click="toggleTrace"><strong>{{traceLinkText}}</strong></a>
```
hillalex: Could style this with up/down carets by conditionally adding "up"/"down" classes to this link and the style rules:
```

a.up::after {
  content: '\25b2';
}

a.down::after {
  content: '\25bc';
}
```

PR #200, opened 2019-12-04
mrc-758 tooltips on form

hillalex:
See what you think of the look and feel as well as the code.

### DIFF
--- a/src/app/static/src/app/components/forms/DynamicFormControlGroup.vue
+++ b/src/app/static/src/app/components/forms/DynamicFormControlGroup.vue
@@ -18,7 +18,7 @@
     import {BCol, BRow} from "bootstrap-vue";
     import {Control, DynamicControlGroup} from "./types";
     import DynamicFormControl from "./DynamicFormControl.vue";
-    import {VTooltip} from 'v-tooltip'
+    import {VTooltip} from 'v-tooltip';
     import {InfoIcon} from "vue-feather-icons";
 
     Vue.directive('tooltip', VTooltip);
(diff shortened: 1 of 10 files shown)


src/app/static/src/app/components/forms/DynamicFormControlGroup.vue
```diff
@@ -64,6 +71,10 @@
             required() {
                 return this.controlGroup.controls.length == 1
                     && this.controlGroup.controls[0].required
+            },
+            helpText() {
+                return this.controlGroup.controls.length == 1 ?
+                    this.controlGroup.controls[0].helpText : ""
```
EmmaLRussell: Doesn't this control show its own help text, in DynamicFormControl? Will this show it twice?

hillalex: Appreciate this is slightly confusing. A "control group" corresponds to a single row in the form, so if you want a control to appear in its own row, you create a control group with one control, and you give the control group a label, but not the control itself. The help text won't show up on the control because it only appears if the control has a label. 

Here for example, "Include ART data" is a group with 2 controls, whereas "Estimate neighbouring district ART attendance" and "Logit prior" are groups with 1 control each.

Screenshot from 2019-12-04 17-00-02

Arguably the help text should then be a property on the group rather than the control, but I'm a bit loathe to change the interface as would create more work for Rob.

hillalex:
> Looks good. Wonder if a circled ? would be more obvious than the i - I've seen both but I think the question mark might be more common, and that icon is quite subtle.

Thanks, you're right, have changed to a ?

PR #204, opened 2019-12-04
mrc-794  bug fix - email domains should be case insensitive

hillalex:
"To reproduce:
1. Ask for password reset email to <user>@IMPERIAL.ac.uk
2. No password reset email is received

I would expect us to convert IMPERIAL.ac.uk to lowercase and use that. This has bitten us with avenir health accounts which have been created as AvenirHealth but requesting pw reset as avenirhealth."

Have turned the `UserRepository` into `UserLogic` (which is a more apt name, as previously noted, as performs lots of logic e.g. sending emails.) Created a new `UserRepository` which just looks up usernames from the db, allowing `UserLogic` to do the case insensitive matching against stored usernames.

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt
@@ -55,7 +55,7 @@ class DbProfileServiceUserLogic(private val userRepository: UserRepository,
 
         val emailArray = email.split("@")
         if (emailArray.count() == 1) {
-            return profileService.findById(email)
+            throw UserException("Please provide a valid email address")
         }
 
         val caseInsensitiveDomainRegex = Regex("(?-i)${emailArray[0]}@(?i)${emailArray[1]}")
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt
@@ -59,6 +59,7 @@ class DbProfileServiceUserLogic(private val userRepository: UserRepository,
         }
 
         val caseInsensitiveDomainRegex = Regex("(?-i)${emailArray[0]}@(?i)${emailArray[1]}")
+
         val username = userRepository.getAllUserNames()
                 .find { caseInsensitiveDomainRegex.matches(it) }
                 ?: return null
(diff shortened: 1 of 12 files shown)


src/userCLI/src/main/kotlin/org/imperial/mrc/hint/userCLI/App.kt
```diff
@@ -49,58 +48,52 @@ fun main(args: Array<String>) {
...
-    private fun Any?.getStringValue(): String
-    {
-        return this.toString().replace("[", "").replace("]","")
+    private fun Any?.getStringValue(): String {
+        return this.toString().replace("[", "").replace("]", "")
     }
 }
 
-fun getUserRepository(dataSource: DataSource): UserRepository {
+fun getUserRepository(dataSource: DataSource): UserLogic {
```
EmmaLRussell: Might as well rename the function I guess

src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    override fun getUser(email: String): CommonProfile? {
+
+        val emailArray = email.split("@")
+        if (emailArray.count() == 1) {
+            throw UserException("Please provide a valid email address")
+        }
+
+        val caseInsensitiveDomainRegex = Regex("(?-i)${emailArray[0]}@(?i)${emailArray[1]}")
```
EmmaLRussell: This means we'll stay case sensitive for the name part of the email address - is that what we want?

hillalex: Yep, emails are case sensitive in general.

hillalex: Well, in theory they could be case sensitive. In practice, unlikely. https://stackoverflow.com/questions/9807909/are-email-addresses-case-sensitive

EmmaLRussell: Yeah you're right, they're meant to be case-sensitive, but most providers let you get away with variants. So users might be used to treating them as case insensitive and being a bit cavalier with using variants. I think it's fine to keep it case sensitive for our purposes and see if anyone objects.

PR #207, opened 2019-12-05
mrc-582 cancel model run

EmmaLRussell:
This should be reviewable though it's maybe not quite ready to merge yet, since the hintr endpoint isn't there (Rich says he'll put that in next week). It's taking a fire-and-forget approach anyway - it sends a request to cancel the run, but immediately clears the front end modelRun state regardless of the result of the request (ignoring errors - currently getting a 404 of course).


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintrAPIClientTests.kt
```diff
@@ -72,4 +72,11 @@ class HintrApiClientTests {
 
         JSONValidator().validateSuccess(metadataResult.body!!, "PlottingMetadataResponse")
     }
+
+    @Test
+    fun `can can model run`() {
```
hillalex: can can :dancer:

src/app/static/src/app/apiService.ts
```diff
@@ -155,6 +156,11 @@ export class APIService<S extends string, E extends string> implements API<S, E>
         return this._handleAxiosResponse(axios.post(fullUrl, data, config));
     }
 
+    async post(url: string) {
+        const fullUrl = this._buildFullUrl(url);
```
hillalex: We already have a `post` method:  https://github.com/mrc-ide/hint/pull/207/files#diff-0f0781cfcf5e78654409fce90c2c466fR146 Should just be able to change `data: any` to `data?: any`

src/app/static/src/app/store/modelRun/mutations.ts
```diff
@@ -61,6 +62,14 @@ export const mutations: MutationTree<ModelRunState> = {
             });
         }
     },
+
+    [ModelRunMutation.RunCancelled](state: ModelRunState) {
+        stopPolling(state);
+        state.modelRunId = "";
+        state.status = {} as ModelStatusResponse;
+        state.errors = [];
+        state.result = null;
```
hillalex: Could use `initialModelRunState()` here perhaps?
```
   Object.assign(state.modelRun, initialModelRunState());
   state.modelRun.ready = true;
```

EmmaLRussell:
Yeah, I didn't put the backend integration test in, since that will always get a 404 just now. So maybe I'll add it and let it fail and not merge until it passes when the hintr endpoint is there.
 
I'm not sure the front end integration test is going to tell us anything beyond what the unit test does, since it ignores any result from the api call and just does the commit come what may. Could add it for the sake for completeness though!

hillalex:
> Yeah, I didn't put the backend integration test in, since that will always get a 404 just now. So maybe I'll add it and let it fail and not merge until it passes when the hintr endpoint is there.
> 
> I'm not sure the front end integration test is going to tell us anything beyond what the unit test does, since it ignores any result from the api call and just does the commit come what may. Could add it for the sake for completeness though!

So the idea for the integration tests is to confirm that the HINT endpoint exists, is secured, and that the front-end is calling the correct endpoint. Looks like the back-end integration test does the job, i.e. confirms that our cancel endpoint is correct even if the hintr one isn't there. But as you say the front-end one doesn't actually confirm that we're hitting the correct endpoint. Might have to factor our the API call into it's own method that can be awaited and the response checked?

src/app/static/src/tests/integration/modelRun.itest.ts
```diff
@@ -32,6 +34,9 @@ describe("Model run actions", () => {
             }
         } as RootState;
         await actions.run({commit, rootState: mockState} as any);
+
+        console.log("CALLS: " + JSON.stringify(commit.mock.calls));
```
hillalex: Can remove this console log

src/app/static/src/tests/integration/modelRun.itest.ts
```diff
@@ -73,4 +78,26 @@ describe("Model run actions", () => {
         expect(commit.mock.calls[0][0]["payload"].detail).toBe("Missing some results");
     });
 
+   it("can cancel model run", async () => {
```
hillalex: I don't think this test adds anything over and above the unit tests - may as well delete to save a few build seconds!

PR #211, opened 2019-12-09
mrc-1191 translations

hillalex:
This creates a new component `Translated.vue`, and a dropdown in the header to toggle the language. It wraps 2 pieces of text in the new component as a demonstration.

It's quite a lot of lines changed (although 130 of them are just adding the translations) so let me know if you'd rather I broke it up more. But I thought would be good to see the component in action with the dropdown, so you can see the language being changed and the text being translated in real time.

Nb this branch is called `mrc-1191_header`


src/app/static/src/app/components/header/DropDown.vue
```diff
@@ -0,0 +1,47 @@
+<template>
```
hillalex: I factored out this component to be re-used by the file menu and the language menu.

src/app/static/src/app/components/Translated.vue
```diff
@@ -0,0 +1,29 @@
...
+
+<script lang="ts">
+    import Vue from "vue";
+    import i18next from "i18next";
+    import {mapState} from "vuex";
+    import {RootState} from "../root";
+
+    export default Vue.extend({
+        name: "Translated",
+        props: ["value"],
```
hillalex: I don't love "value" as the prop name here - it's really the key of the translations dictionary, but "key" is a reserved word. Any ideas?

src/app/static/src/tests/components/baseline/baseline.test.ts
```diff
@@ -11,7 +11,6 @@ import ErrorAlert from "../../../app/components/ErrorAlert.vue";
 import LoadingSpinner from "../../../app/components/LoadingSpinner.vue";
 
 const localVue = createLocalVue();
-Vue.use(Vuex);
```
hillalex: Have just moved this shared setup code into `setup.ts`

src/app/static/src/tests/testHelpers.ts
```diff
@@ -35,3 +38,23 @@ export function testUploadErrorCommitted(url: string,
         });
     });
 }
+
+export function expectAllMutationsDefined(mutationDefinitions: any, mutationTree: MutationTree<any>) {
```
hillalex: Consolidated the test helper classes into one file, this used to live in it's own file.

src/app/static/src/app/components/Translated.vue
EmmaLRussell: Yeah, that's a bit confusing. I think some variant of 'key' would be better than 'value'. 'stringKey'?

hillalex:
> Seems like a really nice approach, very clear.
> 
> Language menu is rendering weirdly for me - always seems to be open. And changing language doesn't have any effect on the strings which should be translated - though I can see the ChangeLanguage mutations coming through in the console. All the stuff on the right of the UserHeader seems to have migrated westwards.
> 
> Screenshot from 2019-12-10 11-33-00
> 
> Seem to have lost the fjelltop and avenir logos!

is this def on the mrc-1191_header branch? there's another old mrc-1191 branch which wasn't working

EmmaLRussell:
> > Seems like a really nice approach, very clear.
> > Language menu is rendering weirdly for me - always seems to be open. And changing language doesn't have any effect on the strings which should be translated - though I can see the ChangeLanguage mutations coming through in the console. All the stuff on the right of the UserHeader seems to have migrated westwards.
> > Screenshot from 2019-12-10 11-33-00
> > Seem to have lost the fjelltop and avenir logos!
> 
> is this def on the mrc-1191_header branch? there's another old mrc-1191 branch which wasn't working

Ha, I've done it again :woman_facepalming:

EmmaLRussell:
Yes, sorry that's all fine, I'm an idiot. Just the 'value' thing to change, if you want to.

src/app/static/src/app/components/Translated.vue
hillalex: Cool, I'll go with textKey

PR #212, opened 2019-12-10
mrc-1203 Show modal to user when state is reset on load

EmmaLRussell:


src/app/static/src/app/store/root/actions.ts
```diff
@@ -22,6 +22,7 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
                 dispatch("surveyAndProgram/deleteAll")
             ]);
             commit({type: "Reset"});
+            commit({type: "load/LoadFailed", payload: {detail: "State is invalid. Resetting all steps."}});
```
hillalex: "State" and "resetting" are more our words than user facing ones. How about just "This file could not be loaded. Taking you back to step 1."

Maybe also "Please contact support..."?

I reckon we should also ditch the "Failed to load state." message in the modal, we already have "Load Error" as a title (and again I don't think "state" is meaningful to the user.)

src/app/static/src/app/components/UserHeader.vue
```diff
@@ -113,7 +113,7 @@
         computed: {
             ...mapStateProps<LoadState, keyof LoadComputed>("load", {
                 hasError: state => state.loadingState==LoadingState.LoadFailed,
-                loadError: state => state.loadError
+                loadError: state => state.loadError && state.loadError.detail
```
hillalex: One of the other ways this error gets populated is if the checksum doesn't match, in which case looks like a string gets committed as the `loadError` at the moment so won't have a `detail` property, should be changed to an error object: https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/load/actions.ts#L34

src/app/static/src/app/store/root/actions.ts
EmmaLRussell: Since the other ticket I'm working on will potentially rescue some valid steps, I might make it something even more generic, "There was a problem loading your data. Some data may have been invalid. Please contact support etc.."

PR #213, opened 2019-12-10
mrc-1190 pass language accept header to API

hillalex:
This won't do anything yet because the back-end isn't there, but pass the `Accept-Language` header on every POST and GET request.

### DIFF
--- a/src/app/static/src/app/store/password/password.ts
+++ b/src/app/static/src/app/store/password/password.ts
@@ -1,6 +1,8 @@
 import {Error} from "../../generated";
+import {TranslatableState} from "../../root";
+import {Language} from "../translations/locales";
 
-export interface PasswordState {
+export interface PasswordState extends TranslatableState {
     resetLinkRequested: boolean
     requestResetLinkError: Error | null
     passwordWasReset: boolean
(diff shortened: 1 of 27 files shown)


src/app/static/src/app/store/password/password.ts
```diff
@@ -1,13 +1,16 @@
...
-export interface PasswordState {
+export interface PasswordState extends TranslatableState {
     resetLinkRequested: boolean
     requestResetLinkError: Error | null
     passwordWasReset: boolean
     resetPasswordError: Error | null
 }
 
 export const initialPasswordState: PasswordState = {
+    language: Language.en,
```
hillalex: So that the password app can use the same api client as the main app, it needs to have this language property as well. There's currently no way to change this of course for the password pages, but we could make the `LanguageMenu` component re-usable and have it on the password pages as well?

src/app/static/src/tests/apiService.test.ts
```diff
@@ -279,9 +282,42 @@ describe("ApiService", () => {
         expect(warnings[1][0]).toBe("No success handler registered for request /baseline/.");
     });
 
+    it("passes language header on get", async () => {
```
hillalex: Testing that we're passing the `Accept-Language` header but can't write an integration test yet because the backend isn't there.

src/app/static/src/app/store/password/password.ts
EmmaLRussell: Makes sense..

PR #215, opened 2019-12-11
mrc-1234 HINT translations

hillalex:
Adds mechanism for translating all HINT originating errors. All `HintException` instances should now have a key rather than a message, that gets translated in the `ExceptionHandler`. All SpringBoot MVC errors will be returned with the unexpected error message (translated) but their original message (untranslated, because we aren't in control of these) will be added to the error object's `trace` field.


src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/PasswordController.kt
```diff
@@ -51,9 +51,13 @@ class PasswordController(private val userLogic: UserLogic,
     @ResponseBody
     @Throws(TokenException::class)
     fun postResetPassword(@RequestParam("token") token: String,
-                          @RequestParam("password") @Size(min = 6, message = "Password must be at least 6 characters long")
+                          @RequestParam("password")
                           password: String): String {
-        val user = oneTimeTokenManager.validateToken(token) ?: throw TokenException("Token is not valid")
+        if (password.length < 6) {
+            throw HintException("invalidPasswordLength", HttpStatus.BAD_REQUEST)
```
hillalex: I changed this over as it makes the `ExceptionHandler` simpler (https://github.com/mrc-ide/hint/pull/215/files#diff-2a06c226eeb55efc9a48ced0a230f53bL41) and also removes that `postResetPassword.password` prefix on the error message https://github.com/mrc-ide/hint/pull/215/files#diff-fdb106bec0ed7f070cacea437a573925L108

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
```diff
@@ -14,48 +13,70 @@ import org.springframework.web.bind.annotation.ControllerAdvice
...
+                appProperties.supportEmail
+        )
+        message = formatter.format(messageArguments)
+
+        val originalMessage = if (!e.message.isNullOrEmpty()) {
+            listOf(e.message!!)
+        } else {
+            null
+        }
+        return ErrorDetail(status, message, originalMessage)
```
hillalex: Using the `trace` field here to return the original error message, e.g. "no handler found" etc. It's not actually the trace of course...

src/app/src/main/kotlin/org/imperial/mrc/hint/models/Response.kt
```diff
@@ -21,4 +22,6 @@ data class ErrorResponse(val errors: List<ErrorDetail>) {
     val data = mapOf<Any, Any>()
     val status = "failure"
 }
-fun ErrorResponse.toJsonString() = ObjectMapper().writeValueAsString(this)
+fun ErrorResponse.toJsonString() = ObjectMapper().apply {
+    setSerializationInclusion(JsonInclude.Include.NON_NULL)
```
hillalex: So that `trace` doesn't get serilalised if its null

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
```diff
@@ -14,48 +13,77 @@ import org.springframework.web.bind.annotation.ControllerAdvice
...
+        // so do not pass the original error message here
+        return unexpectedError(HttpStatus.INTERNAL_SERVER_ERROR, request)
+    }
+    
     @ExceptionHandler(HintException::class)
-    protected fun handleHintException(e: HintException): ResponseEntity<Any> {
+    protected fun handleHintException(e: HintException, request: WebRequest): ResponseEntity<Any> {
         logger.error(e.message)
-        return ErrorDetail(e.httpStatus, e.message!!).toResponseEntity()
+        return translatedError(e.message!!, e.httpStatus, request)
```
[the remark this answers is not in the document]
hillalex: Yeah good shout. "message" could just be null, or it could be something like "error with key: {key}".

hillalex:
> Could the keys be more strongly typed, like they are in the front end? Probably not easily since it's tied into the ResourceBundle which is loaded at runtime. Could inherit from ResourceBundle with properties linked to the string keys in the class. Again, not sure it's worth it, particularly since all the translations are well tested...
> 
> Code coverage is moaning, I think about the UnexpectedError path where there's no default originalMessage, and where translation falls back to returning the key if it isn't the bundle.

Ah thanks, I couldn't spot what codecov was complaining about! Yeah I thought about having an enum or something for keys, but given we'd still have to keep it manually in sync with the ResourceBundle didn't seem like it would add much - having every error message under test (as I believe they are) would still be needed.

PR #218, opened 2019-12-12
mrc-1243 pass language accept header to hintr

hillalex:
Pass the language header from the front-end through to hintr calls.

Contains commits from https://github.com/mrc-ide/hint/pull/215


hillalex:
Hmm, I  merged master in here and now the build is failing because the cancel run tests are failing, I presume because the hintr changes aren't merged to master? Can they be?

EmmaLRussell:
I think Rich still had some work to do on the cancel run branch, and needs PR review, but thought it was pretty much there.

hillalex:
> I think Rich still had some work to do on the cancel run branch, and needs PR review, but thought it was pretty much there.

Unfortunately because the translations don't exist on that hintr branch, this is now blocked until that gets merged.

EmmaLRussell:
Could reset the hintr branch to master then. I think there's just a backend integration test which could be changed to expect a 404 for now, and a front end test which could also expect a 404 rather than a 400. The actual cancel behaviour in the front end shouldn't be affected since it ignores errors, and in the backend the run would continue unaffected.

hillalex:
This is now unblocked @EmmaLRussell

PR #223, opened 2019-12-13
mrc-1235 a bunch of tranlsations

hillalex:
Components:
* baseline
* downloadResults
* header
* modelOptions
* modelOutput
* modelRun
* resetPassword

Thought this was a big enough changeset, so I'll do the rest on a separate branch.

Contains commits from https://github.com/mrc-ide/hint/pull/222

PR #224, opened 2019-12-16
mrc-1235 final batch of translations

hillalex:
Components:
* barchart
* choropleth
* surveyAndProgram
* fileUpload
* resetConfirmation
* stepper/step

Contains commits from https://github.com/mrc-ide/hint/pull/223


src/app/static/src/app/components/plots/ChoroplethFilters.vue
```diff
@@ -1,6 +1,6 @@
 <template>
     <div v-if="hasFilters">
-        <h4>Filters</h4>
+        <h4 v-translate="'Filters'"></h4>
```
EmmaLRussell: I think the key is lower case, 'filters'..?

hillalex:
> Nice one, this must have been boring!
> 
> So the only other rogue strings I can find in the front end are:
> /store/root/actions line 46 (loading data error)
> various error messages in apiService
> 
> Are these going into a separate branch?

Ah thanks, no I had missed those. I'll add them in here :+1:

src/app/static/src/app/apiService.ts
```diff
@@ -46,15 +47,15 @@ export class APIService<S extends string, E extends string> implements API<S, E>
...
+            return APIService.createError("apiMissingError");
         }
         return failure.errors[0];
     };
 
     static createError(detail: string) {
         return {
             error: "MALFORMED_RESPONSE",
-            detail
+            detail: i18next.t(detail)
```
hillalex: So this of course means that once the error is displayed, if the language is toggled it remains in the language it was generated with. But this is the same situation as with the errors coming back from the API.

EmmaLRussell: I think that's fair enough!

PR #226, opened 2019-12-16
Mrc 1261 Bubble plot filters

EmmaLRussell:
Sorry, this is quite massive. Do you want me to try to split it up?


src/app/static/src/app/components/plots/FilterSelect.vue
```diff
@@ -7,17 +7,21 @@
...
 
 <script lang="ts">
     import Vue from "vue";
     import Treeselect from '@riophae/vue-treeselect';
-    import {NestedFilterOption} from "../../generated";
+    import {FilterOption, NestedFilterOption} from "../../generated";
 
     interface Methods {
-        select: (value: string[]) => void
+        input: (value: string[]) => void
```
EmmaLRussell: We've ended up with two FilterSelect components, one living with the Barchart and one at the plots level. I've modified the latter so that it can deal with both types of event emitting (string ids or full selected options - Choro uses former, Barchart and Bubble Plot the latter). I've left the Barchart FilterSelect where it is as it has extra functionality around isXAxis/isDisagg etc.

EmmaLRussell:
How can code coverage be failing having succeeded in previous builds. All this commit did was remove code!

hillalex:
> Sorry, this is quite massive. Do you want me to try to split it up?

If there is a sensible way or carving this up into small PRs that would be super helpful :pray:

PR #227, opened 2019-12-16
mrc-1248 change language on password pages

hillalex:
Pulls out the changing language action/mutation into their own files to be shared between root and password app, and adds a header to the password related pages.

Also figured out a way to test that the .ftl templates themselves are configured correctly (without resorting to selenium :grin:) e.g. https://github.com/mrc-ide/hint/pull/227/files#diff-6ab8de68b3bdfe3735aca230f988b32a


src/app/static/src/app/components/header/LoggedOutHeader.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+    <header class="mb-5">
+        <nav class="navbar navbar-dark bg-secondary">
+            <div class="container-fluid">
+                <div class="navbar-header">
+                    {{title}}
```
EmmaLRussell: I suppose one day we might have different titles in French and English, but I'm assuming French for 'Naomi' is just Naomi with a French accent!

src/app/static/src/tests/components/password/forgotPasswordPage.test.ts
```diff
@@ -0,0 +1,18 @@
+const fs = require('fs');
+const path = require('path');
+const html = fs.readFileSync(path.resolve(__dirname, '../../../../templates/forgot-password.ftl'), 'utf8');
+
+document.documentElement.innerHTML = html.toString();
```
EmmaLRussell: Very cunning! Just wonder if this should go into a different root test folders than 'components' since it's not actually testing a component. 'pages'  or 'templates'. No doubt we'll make other tests like this, and they should probably be kept together.

hillalex: Good shout

src/app/static/src/app/components/header/LoggedOutHeader.vue
hillalex: I mean, I think Naomi makes as much sense in French as it does in English ;)

PR #228, opened 2020-01-07
add translations

hillalex:


src/app/src/main/resources/ErrorMessageBundle_fr.properties
```diff
@@ -1,3 +1,10 @@
-unexpectedError = Une erreur inattendue s''est produite. If you see this message while you are using {0} at a workshop, \
-  please contact your workshop technical support and show them this code: {1}. \
-  Otherwise please contact support at {2} and quote this code: {1}
+invalidEmail = Veuillez fournir une adresse e-mail valide.
+invalidToken = Le jeton n'est pas valide.
+invalidPasswordLength = Veuillez entrer un nouveau mot de passe d'au moins 6 caractères.
+loadFailed = UImpossible de charger des fichiers pour la session. Les fichiers indiqués n'existent pas sur le serveur.
```
EmmaLRussell: Is this a stray U?

EmmaLRussell:
Are we sure that 'PJNZ'  in French is definitely the same? Is it actually an acronym, so potentially changes by language?

src/app/src/main/resources/ErrorMessageBundle_fr.properties
hillalex: well spotted

hillalex:
> Are we sure that 'PJNZ' in French is definitely the same? Is it actually an acronym, so potentially changes by language?

it's a file extension :)

hillalex:
> Looking good so far as I can tell! There are some missing translations, as shown. These should maybe go in separate ticket though? I'm going to need to add some for the Bubble plot too...
> 
> Step 1: 'File' menu title and 'Shape file'
> Screenshot from 2020-01-08 09-18-30
> 
> Step 2:
> Filter titles (these are hardcoded in the Choro rather than coming from metadata, so we'll need to translate them ourselves)
> 'Survey'
> 'Select'
> 'Not used'
> Tooltips on the file selects say 'No file chosen'
> As shown, the French text for select a file runs over the length of the control, but not sure that's something we'd want to change.
> Screenshot from 2020-01-08 09-22-42
> 
> Step 3
> 'required' marker for parameters
> 
> Step 5
> Same issues in Choropleth as in Step 2
> 'x axis' and 'disaggregate by' in the bar chart
> Screenshot from 2020-01-08 09-31-56

Cool, thanks for spotting these. I'll fix any that are due to missing the directive here, it's possible some of the translations are actually missing, if so we can make a short list to request further translations (including bubble plot related ones)

hillalex:
Ok so I've kept the barchart and form decoupled from the state, passing all translated strings down to them as props. The other coupled bits I've left as is for now (the choropleth filters basically.)

PR #232, opened 2020-01-07
mrc-1261_components

EmmaLRussell:
Adding filters to BubblePlot component, and associated component changes

Contains changes from https://github.com/mrc-ide/hint/pull/231

### DIFF
--- a/src/app/static/src/app/components/plots/FilterSelect.vue
+++ b/src/app/static/src/app/components/plots/FilterSelect.vue
@@ -7,17 +7,22 @@
                     :value=treeselectValue
                     :disabled=disabled
                     :placeholder=placeholder
-                    @input="select"></treeselect>
+                    @input="input"
+                    @select="select"
+                    @deselect="deselect"></treeselect>
     </div>
 </template>
 
 <script lang="ts">
     import Vue from "vue";
     import Treeselect from '@riophae/vue-treeselect';
-    import {NestedFilterOption} from "../../generated";
+    import {FilterOption, NestedFilterOption} from "../../generated";
+    import {flattenOptions} from "../../utils";
 
     interface Methods {
-        select: (value: string[]) => void
+        input: (value: string[]) => void
+        select: (node: FilterOption) => void
+        deselect: (node: FilterOption) => void
     }
 
     interface Computed {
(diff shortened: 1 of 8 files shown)


src/app/static/src/app/components/plots/FilterSelect.vue
```diff
@@ -7,17 +7,22 @@
...
 <script lang="ts">
     import Vue from "vue";
     import Treeselect from '@riophae/vue-treeselect';
-    import {NestedFilterOption} from "../../generated";
+    import {FilterOption, NestedFilterOption} from "../../generated";
+    import {flattenOptions} from "../../utils";
 
     interface Methods {
-        select: (value: string[]) => void
+        input: (value: string[]) => void
```
EmmaLRussell: We've ended up with two FilterSelect components, one living with the Barchart and one at the plots level. I've modified the latter (this one) so that it can deal with both types of event emitting (string ids or full selected options - Choro uses former, Barchart and Bubble Plot the latter). I've left the Barchart FilterSelect where it is as it has extra functionality around isXAxis/isDisagg etc. Choro and Bubble Plot both using this more generic one.

src/app/static/src/app/components/plots/bubble/BubblePlot.vue
```diff
@@ -140,20 +195,60 @@
...
+                if (selectedOptions && selectedOptions.length > 0) {
+                    return selectedOptions
+                }
+                return this.countryFilterOption ? [this.countryFilterOption] : [];
+            },
+            flattenedAreas() {
+                return flattenOptions(this.areaFilter.options);
+            },
+            allSelectedAreaIds() {
+                //get intersection of selected areas and current Features(ie current level)
```
hillalex: I don't think it's neccessary to take this intersection since only current features are rendered on the map - you can just directly pass the flattened selected area ids into the `getFeatureIndicators` function.

EmmaLRussell: Yeah, that's right - I think I put that in there to keep the featureIndicators dictionary small. I guess it's a trade off between handling more data than you need and recalculating it more often. I don't think I saw a performance hit either way really - for Malawi at least... It would keep the bubblePlot code a little simpler to do away with allSelectedAreaIds so that's reason enough!

hillalex: I wondered if it was a deliberate choice, but yeah I think if we can simplify this a little bit that's preferable!

PR #234, opened 2020-01-07
mrc-1254_new Indicator selectors in Bubble plot

EmmaLRussell:
Contains changes from https://github.com/mrc-ide/hint/pull/232

Still need to add French translations for 'Bubble plot', 'Colour Indicator' and 'Size Indicator'. Rob is adding those to the spreadsheet, and we're going to ask Anne for a sensible initial translation.


src/app/static/src/app/store/translations/locales.ts
```diff
@@ -95,6 +97,7 @@ const en: Translations = {
     cancelRun: "Cancel run",
     chooseFile: "Choose a file",
     clickHere: "Click here",
+    colorIndicator: "Colour Indicator",
```
EmmaLRussell: I'm assuming here that we're using UK English in the UI, but the code uses 'color' elsewhere so would be better for the key to be consistent with that, hence this slight oddity

PR #235, opened 2020-01-08
mrc-1251 Color legend in bubble plot

EmmaLRussell:
Contains changes from https://github.com/mrc-ide/hint/pull/234


EmmaLRussell:
Hang fire on this for now, and I'll add in using actual values for min and max

EmmaLRussell:
> Hang fire on this for now, and I'll add in using actual values for min and max

Actually, I've changed my mind about his after the discussions yesterday afternoon. I'm putting in some preparatory work for the MapLegend to potentially accept a range separate to the metadata object so it can show arbitrary ranges, since it seems we're going to have to add some variation on that (use defined and/or data defined). 

But for now let's keep the colour ranges in the Choropleth and Bubble Plot consistent with each other. The Size range in the Bubble plot will remain data-defined (I think - will check with Jeff when we meet with him today).

src/app/static/src/app/components/plots/utils.ts
```diff
@@ -0,0 +1,33 @@
+import * as d3ScaleChromatic from "d3-scale-chromatic";
```
EmmaLRussell: Moved the getColor function here rather than it living in filteredData utils

PR #236, opened 2020-01-08
Mrc 1252 Size legend in bubble plot

EmmaLRussell:
Contains changes from https://github.com/mrc-ide/hint/pull/235


src/app/static/src/app/main.ts
```diff
@@ -8,4 +8,4 @@ Vue.use(Vuex);
 export const store = new Vuex.Store<RootState>(storeOptions);
 registerTranslations(store);
 
-export const inactiveFeatures: string[] = ["BubblePlot"];
+export const inactiveFeatures: string[] = [];
```
EmmaLRussell: Time to throw the switch?

src/app/static/src/app/components/plots/bubble/BubblePlot.vue
```diff
@@ -161,15 +166,18 @@
...
         },
         props: props,
         data(): Data {
             return {
                 style: {
-                    className: "geojson-grey"
-                }
+                    className: "geojson-grey",
+                },
+                maxRadius: 70,
```
EmmaLRussell: Bumped this max radius down from 100 which just massive.

src/app/static/src/app/components/plots/bubble/SizeLegend.vue
```diff
@@ -0,0 +1,93 @@
...
+        computed: {
+            width: function() {
+                return (this.maxRadius * 2) + 2; //leave room for stroke width
+            },
+            height: function() {
+                return (this.maxRadius * 2) + 10; //leave room for the top text
+            },
+            circles: function() {
+                const x = this.width / 2;
+                const circleFromRadius = (r: number, value: number, under: boolean = false) => {
```
hillalex: Can `circleFromRadius` be a component method rather than defined here? Just to make this function shorter/easier to read.

src/app/static/src/app/components/plots/bubble/SizeLegend.vue
```diff
@@ -0,0 +1,93 @@
...
+                //We treat the minimum circle differently, since the smallest radius is actually likely to cover quite
+                //a wide range of low outliers, so we show the value for the next pixel up and prefix with '<'
+                const nextMinRadius = this.minRadius + 1;
+                const valueScalePoint = (Math.pow(nextMinRadius,2) - Math.pow(this.minRadius,2))/
+                                            (Math.pow(this.maxRadius, 2) - Math.pow(this.minRadius, 2));
+                const nextValue = (valueScalePoint * (this.indicatorRange.max - this.indicatorRange.min))
+                                            + this.indicatorRange.min;
+                const minCircle = circleFromRadius(this.minRadius, nextValue, true);
+
+                const steps = [0.25, 0.5, 0.75, 1];
```
hillalex: Can steps be a data property of the component?

Also do you think it gives a better sense of scale to use [0.125, 0.25, 0.5, 1], so each bubble is twice the size of the previous? Or [0.1, 0.25, 0.5, 1] even - starting at 0.25 means there's a very big jump from the negligible bubble to the first size, but even 0.125 is quite a big jump.

To compare:

(this is [0.1, 0.25, 0.5, 1])
Screenshot from 2020-01-10 13-13-19

vs

Screenshot from 2020-01-10 13-14-35

src/app/static/src/app/components/plots/bubble/SizeLegend.vue
```diff
@@ -0,0 +1,93 @@
...
+                        text  = "<" + text;
+                    }
+
+                    return {x: x,  y: y, radius: r, text: text, textX: x, textY: y-r}
+                };
+
+                //We treat the minimum circle differently, since the smallest radius is actually likely to cover quite
+                //a wide range of low outliers, so we show the value for the next pixel up and prefix with '<'
+                const nextMinRadius = this.minRadius + 1;
+                const valueScalePoint = (Math.pow(nextMinRadius,2) - Math.pow(this.minRadius,2))/
```
hillalex: Can this line and the next be pulled out into their own util methods to make this function a bit easier to read?

src/app/static/src/app/components/plots/bubble/SizeLegend.vue
```diff
@@ -0,0 +1,93 @@
...
+            },
+            height: function() {
+                return (this.maxRadius * 2) + 10; //leave room for the top text
+            },
+            circles: function() {
+                const x = this.width / 2;
+                const circleFromRadius = (r: number, value: number, under: boolean = false) => {
+                    const y = this.height - r;
+
+                    let text = value > 1000 ? numeral(value).format("0a") : (+value.toFixed(2)).toString();
```
hillalex: I think we could do with more precision here - probably 3 dp for small numbers and `0.0a` format for large?
E.g. makes the smallest circle here <0.004 rather than literally 0:

prevalencecircles
 vs

Screenshot from 2020-01-10 13-23-57

and < 5k becomes < 4.6k here:

Screenshot from 2020-01-10 13-26-35

The tooltips actually have 4 dp but that might be too cluttered for the legend. At the moment the big numbers in the tooltips aren't formatted, we should open a ticket to format them to 0.0a also.

src/app/static/src/app/main.ts
hillalex: yeah!

src/app/static/src/app/components/plots/bubble/SizeLegend.vue
EmmaLRussell: Yeah you're right, particularly since in pratice people are mostly going to be looking at the smaller bubbles (those at district level rather than country/region level)

src/app/static/src/app/components/plots/bubble/SizeLegend.vue
EmmaLRussell: OK, that sounds good. Do you think we should also have a couple of levels of 'smallness'? Small currently means anything under 100, so we'd potentially get '8647.1522' etc 
Maybe make it 4 sig figs rather than 4dps?

hillalex: 4 sig fig sounds good to me, want to change that over?

PR #237, opened 2020-01-14
Add link to help document

r-ash:
This PR will

* Add link in user header to help document

At the moment the French version is the same pdf as the English one - UNAIDS are currently working on a French version so we should update this when that is available.

Let me know if this testing is not extensive enough?


src/app/static/src/tests/components/header/userHeader.test.ts
```diff
@@ -34,4 +37,23 @@ describe("user header", () => {
...
+    it("computes language", () => {
+        const wrapper = shallowMount(UserHeader, {localVue, store});
+        const vm = (wrapper as any).vm;
+        const result = vm.currentLanguage;
+        expect(result).toStrictEqual("EN");
+    });
+
+    it("contains help document link", () => {
+        const store = new Vuex.Store({
+            state: {language: Language.en}
```
EmmaLRussell: Language defaults to English anyway so shouldn't really need this, but I guess you're using a different store to the one used by all the other tests because you're changing its value. 

I might be inclined to use the file-level store to test the default (English) link, and make a local frenchStore to test another wrapper with the French version. 

Or at least to rename the local variable 'store' here to languageStore or something, so it's not confusable with the file level one (would then need to shallow mount as eg {store: languageStore}).

r-ash: I've updated this now, it requires setting the translations though after switching to French which isn't super nice. 

Is there a junit `@Before` equivalent for this test framework? I could use that to make sure `registerTranslations` is called with English `store` before every test?

EmmaLRussell: Right, yes, sorry, I had forgotten that registerTranslations affects the translation module state, so of course you need to re-register. You can use beforeEach: https://jestjs.io/docs/en/setup-teardown

src/app/static/src/app/components/header/UserHeader.vue
```diff
@@ -38,13 +39,21 @@
...
+        helpFilename: string
     }
 
     export default Vue.extend<{}, {}, Computed, Props>({
         computed: {
-            currentLanguage: mapStateProp<RootState, string>(null,
-                (state: RootState) => state.language.toUpperCase())
+            helpFilename: mapStateProp<RootState, string>(null,
+                (state: RootState) => {
+                    var filename = "";
```
EmmaLRussell: Could be let rather than var..

src/app/static/src/app/components/header/UserHeader.vue
```diff
@@ -38,13 +39,21 @@
...
+                    if (state.language == Language.en) {
+                        filename = "Naomi-basic-instructions.pdf";
+                    } else if (state.language == Language.fr) {
+                        filename = "Naomi-instructions-de-base.pdf";
+                    }
+                    return filename;
```
EmmaLRussell: This would default to linking to nothing if we introduce another language and forget to add a new helpfile. Which shouldn't happen of course... but since we're falling back to EN everywhere else, maybe it would be better to do that here too.

PR #240, opened 2020-01-23
MRC-1313: add a link to a troubleshooting page

richfitz:
This PR adds an extra "Troubleshooting" link to the toolbar.

As discussed on slack it would be better if the page appeared to be hosted on the same domain, but this approach will let us modify the page fairly frequently without redeploying the front end.

I have copied the approach of the Help/Aide button which changes url based on the language. Translations have been added too - all basically just duplicating code that was there.


src/app/static/src/tests/components/header/userHeader.test.ts
```diff
@@ -50,11 +50,13 @@ describe("user header", () => {
...
         const vm = (wrapper as any).vm;
         expect(vm.helpFilename).toStrictEqual("Naomi-basic-instructions.pdf");
+        expect(vm.troubleFilename).toStrictEqual("index-en.html");
 
         const frStore = createFrenchStore();
         const frWrapper = shallowMount(UserHeader, {localVue, store: frStore});
         const frVm = (frWrapper as any).vm;
-        expect(frVm.helpFilename).toStrictEqual("Naomi-instructions-de-base.pdf")
+        expect(frVm.helpFilename).toStrictEqual("Naomi-instructions-de-base.pdf");
+        expect(frVm.troubleFilename).toStrictEqual("index-fr.html");
```
EmmaLRussell: Could also test that the rest of the link looks legit, as "contains help document link" does

PR #242, opened 2020-02-21
mrc-1363 Make Choropleth re-usable

EmmaLRussell:
This is a preparatory PR to pulling the Choro out as a library component. Refactors the Choropoleth to be like the Bubble plot and Barchart, with no knowledge of HINT store, but dealing with generic Filter objects etc. 

This means that the filteredData module in the store was redundant, so I've removed that and moved any bits that the new Choropleth needs into surveyAndProgram and modelOutput state. The Choropleth filter state for the surveyAndProgram and for the modelOutput are now independent from each other. 

There's still a lot of shared code with the Bubble plot, might refactor those in a separate branch. 

Sorry this is so massive, but on the bright side, it looks like there were about twice as many deletions as additions!


src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -0,0 +1,379 @@
+<template>
+    <div class="row">
+        <div :class="hideControls ? 'col-sm-6 col-md-8' : 'col-md-3'">
+            <slot></slot>
```
EmmaLRussell: This bit of annoyance is here so we can keep the file uploads in the SAP view within the footprint of the choro (so the Filters can tuck underneath them). Hence the fiddly css and needing to 'hide controls' before we're ready to display the map and filters.

EmmaLRussell: Hm, I've just realised that this isn't a great solution as keeping the tabs in a row about the choropleth means that they stay there with the filters and map wrap on smaller screens, rather than jumping down to be above the map... Not really sure what to do about that!

EmmaLRussell: Maybe we should keep the option to keep Choro and Choro-filters as two separate controls, and lose the <slot> in the choro. Choro could be configured to either include its own choro-filters or not (it could in case of Output). They'd just need to both be wired up to the same plottingSelections state. Actually, if we did that, it would have the advantage that we could only show both controls once a file was loaded, so we could avoid a lot of that horrible intitialising/hideControls logic. What do you think? Put that in here or in a separate ticket?

EmmaLRussell:
Not sure what code cov is complaining about, the diff is all green!

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
hillalex: I agree this is a bit weird; filters as a separate component sounds good, it will also have the benefit of breaking up this megalithic component. Happy for that to be on a fresh ticket though.

PR #243, opened 2020-02-26
mrc-1376 Separate out filters into separate component from the Choropleth

EmmaLRussell:


src/app/static/src/app/components/plots/Filters.vue
```diff
@@ -0,0 +1,64 @@
...
+<script lang="ts">
+    import Vue from "vue";
+    import FilterSelect from "./FilterSelect.vue";
+    import {Dict, Filter} from "../../types";
+    import {FilterOption} from "../../generated";
+
+    interface Props {
+        filters: Filter[],
+        selectedFilterOptions: Dict<FilterOption[]>,
+        selectMultipleFilterIds: string[]
```
hillalex: Could we just add `allowMultiple` as a field on `Filter`?

src/app/static/src/app/components/plots/utils.ts
```diff
@@ -118,4 +118,19 @@ export const roundToContext = function (value: number, context: number) {
     const roundingNum = Math.pow(10, maxDecPl + 1);
 
     return Math.round(value * roundingNum) / roundingNum;
+};
+
+export const replaceAreaFilterOptionsWithCountryChildren = (filters: Filter[], areaFilterId: string) => {
```
hillalex: I don't understand what's going on here! The area filter gets passed to the component with options including the top level country but that's not actually what we want, so this method changes it to remove the top level? If I'm understanding that correctly, shouldn't the filter just be created with the desired options to begin with, i.e. in the getter?

src/app/static/src/app/components/plots/Filters.vue
EmmaLRussell: Yep. Whether multiple values should be selectable for a filter depends on the (output) plot - you can select multiple values for other filters in the barchart. So I was sort of seeing Filters as immutable things which come from the backend and can be presented with different functionality in different plots. However, we do actually have different filters getters for each plot type as there was initially the idea that you might want to have different filters per plot. So I could make a subclass of Filter which includes allowMultiple and return that in the getter for the choropleth filters, allowing multiple for area but not the others.

src/app/static/src/app/components/plots/utils.ts
EmmaLRussell: As for selectMultiple it depends on the plot - we do include the top level region for the Barchart. But again, I could put it in the getter for the maps but not the barchart..

EmmaLRussell: Ha, except this doesn't quite for the SAP getter, since the Choropleth needs the full option set so it can 'select' the country option if no options selected in the Filters component. Otherwise its default selection if the first region..

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -200,29 +195,20 @@
...
                 return flattenOptions(this.areaFilter.options);
             },
             selectedAreaFeatures(): Feature[] {
-                if (this.initialised && this.selectedAreaFilterOptions && this.selectedAreaFilterOptions.length > 0) {
+                if (this.initialised) {
                     return this.selectedAreaFilterOptions.map(o => this.getFeatureFromAreaId(o.id)!!);
                 }
                 return [];
             },
-            countryFilterOption(): FilterOption {
```
hillalex: Nice! Wow this simplified things a fair bit :smile:

PR #245, opened 2020-02-28
mrc-1374 replace bar chart with npm package

hillalex:


src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -25,13 +25,12 @@
             </div>
 
             <div id="barchart-container" :class="selectedTab==='bar' ? 'col-md-12' : 'd-none'">
-                <barchart :chartdata="chartdata" :filters="barchartFilters" :indicators="barchartIndicators"
-                          :selections="barchartSelections"
-                          :x-axis-text="xAxisText"
```
hillalex: So looks like we had already replaced the v-translate with passing in the labels here. But I think the filter config object is a little neater anyway.

PR #246, opened 2020-03-18
mrc-1415 Set custom colour scale for Survey and Program choropleth

EmmaLRussell:
Sorry this a bit large, should probably have been a couple of tickets. 

- This adds a new colourScales module to the store. The colour scale preferences could have gone into plottingSelections, but I though that would bloat that module, and also there's a structural distinction in that when we come to do outputs, bubble plot has its own plotting selections but will share colour scales with the choropleth.

- The dictionary of colourScales is lazy-created as needed rather than initialised for every indicator when the sap state arrives. This means they would be retained if sap state was reloaded with different data.. which I think is probably good from the user's pov but maybe it would be more consistent with other modules if it was wiped? 

- See what you think of the 'not a modal' UI - the map updates immediately when you make changes, which is pretty cool, but also has slightly odd effects when you temporarily have an empty (NaN) value, or max < min. Nothing actually break though.. It should probably at least warn if max <= min!

- I haven't done anything with the up/down controls in the number inputs - these default to delta of 1 which isn't always appropriate. Could try to something cleverer with those in a separate ticket.

- When you switch from custom back to default it retains your custom min and max, which seemed like it might be good if you wanted to toggle, but then I found myself forgetting what the defaults had been, so maybe those custom values should always show the actual min and max.


src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -146,14 +158,24 @@
             },
             featureIndicators() {
                 const selectedAreaIdSet = flattenToIdSet(this.selectedAreaFilterOptions.map(o => o.id), this.flattenedAreas);
+
+                let customMin = null;
+                let customMax = null;
+                if (this.indicatorColourScale.type == ColourScaleType.Custom) {
```
EmmaLRussell: I'm moving this logic into the utils for adding dynamic range - let utils rather than choro work out what it the actual min and mac for a given trio of colour scale, metadata and data.

EmmaLRussell:
Actually, maybe hold fire on this, I may have broken the output view...

src/app/static/src/app/components/plots/MapAdjustScale.vue
```diff
@@ -0,0 +1,71 @@
...
+    interface Props {
+        show: Boolean,
+        colourScale: ColourScaleSettings
+    }
+
+    interface Computed {
+        disableCustom: Boolean
+    }
+
+    export default Vue.extend<{}, {}, {}, Props>({
```
hillalex: I'm suprised it didn't complain about this type declaration? Missing computed, methods and data!

src/app/static/src/app/store/colourScales/actions.ts
```diff
@@ -0,0 +1,15 @@
...
+import {ColourScaleSelections, ColourScalesState} from "./colourScales";
+
+export interface ColourScalesActions {
+    updateSelectedSAPColourScales: (store: ActionContext<ColourScalesState, RootState>, colourScales: ColourScaleSelections) => void
+}
+
+
+export const actions: ActionTree<ColourScalesState, RootState> & ColourScalesActions = {
+    updateSelectedSAPColourScales(context, colourScales) {
+        const {commit, rootState} = context;
```
hillalex: Since there's nothing async going on here and it just calls straight through to the commit, can we just dispense with this action altogether and directly call the commit from the component using the `mapCommit`helper?

src/app/static/src/app/store/colourScales/colourScales.ts
```diff
@@ -0,0 +1,52 @@
+import {Dict} from "../../types";
+import {localStorageManager} from "../../localStorageManager";
+import {Module} from "vuex";
+import {RootState} from "../../root";
+import {getters} from "./getters";
+import {actions} from "./actions";
+import {mutations} from "./mutations";
+
+export interface ColourScalesState {
```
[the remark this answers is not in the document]
EmmaLRussell: It could do, though as I said above wouldn't really be consistent with the other structure of plottingSelections (where bubble and choro have their own selections), so these would need to be top level properties on plottingSelections anyway, so I thought it might be cleaner as a new module. Happy to combine if you'd prefer.

EmmaLRussell:
I'll maybe prevent it from emitting the event if min and max are the same..

> Whenever I change the values I see this error in the console:
> 
> ```
> vue.esm.js?a026:628 [Vue warn]: Duplicate keys detected: '0'. This may cause an update error.
>

src/app/static/src/app/store/colourScales/actions.ts
hillalex: I see about the rootState - but this is called from a component that already know what the selected data type is - could parameterise the mutation with `selectedDataType` as well as `colourScales`?

src/app/static/src/app/store/colourScales/colourScales.ts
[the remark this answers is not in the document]
hillalex: My feeling is it's cleaner as one module, but since we're split on this and you've already written the code I'll leave it up to you!

src/app/static/src/app/store/colourScales/actions.ts
hillalex: (also yeah, mutations clearly can't access rootState so no idea what we were both remembering!)

EmmaLRussell:
Thanks for the style tweaks, looks a lot better. I'm actually going to make it a little taller than the legend, to accommodate an error message (unless you can think of a better way of presenting that..), and it'll get taller still when the dynamic options are in.

src/app/static/src/app/store/colourScales/colourScales.ts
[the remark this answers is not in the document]
EmmaLRussell: Fair enough, you've talked me into it! :)

EmmaLRussell:
Ach! Having spotted a little bit in your PR that needs translating, I've realised this all needs translating too. I'll open another ticket for that if that's OK...

hillalex:
Ah yeah, the design challenges of red as primary theme colour!

On a very minor point we're using Bootstrap's `text-danger` and `alert-danger` classes rather than literally using our theme red to style errors (although the difference is perhaps not visible to the naked eye :laughing:) 

The only other thing I can think is could add a red outline to the inputs themselves, the way a Bootstrap form-control has when it's in an error state (and then you could maybe even have the error message in grey, or leave as red). I think it's pretty obvious to the user what's going on though, so I'm not too worried about the colour clash.

Might be nice to add  `v-bind:min="colourScaleToAdjust.customMin"` to the max input and the opposite to the min so that when using the up/down arrows it automatically stops you choosing outside that range? The step="any" doesn't seem to do anything for me - it's still navigating up/down in intervals of 1. Maybe step="0.1" is a reasonable heuristic? Or could try and be a bit cleverer and take the default scale range and divide by 10 or something to get the step?

I forgot about translations also! Def a new ticket :+1:

src/app/static/src/app/components/plots/MapAdjustScale.vue
```diff
@@ -17,22 +17,26 @@ import {ColourScaleType} from "../../store/colourScales/colourScales";
         </div>
 
         <div class="mt-2 ml-2">
-            <div class="row p-0 mb-2">
-                <label for="custom-min-input" class="col col-form-label col-2">Min</label>
-                <div class="col pt-1">
-                    <input id="custom-min-input" type="number" step="any" v-model.number="colourScaleToAdjust.customMin"
-                           @change="update" @keyup="update" :disabled="disableCustom">
+            <form novalidate>
```
EmmaLRussell: Need this to stop the inputs complaining with warning message when you manually enter a value into a numeric input which isn't 'legal' according to the step value e.g. 0.5 if step is 1

hillalex:
This is looking great. Sorry to be a pain but I'm still seeing some errors as well as some weird behaviour if you enter negative values:

Screenshot from 2020-03-20 17-10-58

Changing line 9 in MapLegend.vue to 

```v-for="(level, index) in levels" v-bind:key="index"```

 fixes the errors, but you still get suprising colouring when you put negative values in!

Screenshot from 2020-03-20 17-13-01

I realise it's stupid for someone to enter negative values so this is a bit of an edge case, but it shouldn't be much extra work to disallow negatives in the same way that min=max is disallowed (and set min=0 on the input with id = custom-min-input)

EmmaLRussell:
Interesting, I thought negative values were actually behaving themselves, but I think I may have been looking at a wider range. I think there might be something a bit odd going on with the rounding in the legend. Will check.

Will negative values definitely never be relevant? Don't seem to be for any of the current indicators, but could imagine negative values that might make sense. I'll ask Jeff.

hillalex:
> Interesting, I thought negative values were actually behaving themselves, but I think I may have been looking at a wider range. I think there might be something a bit odd going on with the rounding in the legend. Will check.
> 
> Will negative values definitely never be relevant? Don't seem to be for any of the current indicators, but could imagine negative values that might make sense. I'll ask Jeff.

It kinds looks like the colour scale values are being treated as absolute values - like values that are 0.2 are getting coloured by the -0.2 colour? Maybe? I haven't confirmed this, just eyeballing the map seems plausible.

EmmaLRussell:
Ha ok, so the duplicate keys were to do with the rounding rather than negatives, in that it was making the decision on how to round based only on the max value and not considering min. We should  make the change to consider both regardless of whether negatives are allowed, now that users can put in any pair of numbers they fancy. Still need to work out what's going on with the colours though...

EmmaLRussell:
There was a bug in getColor which meant that it was setting a default range of 1 if max was falsy. Which was probably ok before we contemplated having ranges including negatives. Leaving allowing negatives in, since Jeff thinks might be useful in future. Think it should be behaving now though...

hillalex:
Working great, nice one :tada:

PR #248, opened 2020-03-19
mrc-1419 allow unauthenticated use

hillalex:
So this just opens up the whole app. Most of the code changes are just updating the tests. I have mostly left the security testing framework code in place, in case we end up up having some endpoints that are for logged in users only (I guess we will as soon as we implement the session view where users can view past sessions.)

Have added some indication in the header whether you are logged in or not - if you are logged in it shows your email and a "log out" link; if you're not logged in it shows a "Log In" link.

To test this locally you have to manually add the guest user to the db (we will also have to do this on deploy for the first time):

* build the cli for the local branch by running `./src/userCLI/build`
* run the locally built cli to add the guest using `./scripts/test-cli.sh add-user guest`


src/app/static/src/app/components/header/UserHeader.vue
```diff
@@ -6,6 +6,7 @@
                     {{title}}
                 </div>
                 <file-menu :title="title"></file-menu>
+                <span v-if="!isGuest" class="pr-2 mr-2 border-right">Logged in as {{user}}</span>
```
EmmaLRussell: This should be translated

hillalex:
> Looks good! I wondered if it would be more consistent, since you see the 'Login as guest' option and login page when you explicitly log out, whether you should also see that when you first arrive unauthenticated. So you'd always explicitly choose to continue as guest or login as yourself. This might be less confusing with users who do want to log in. But since this would mean changing the code quite a bit, and since there's currently no reason for anyone to log in, I think we should put off worrying about that until we do have stuff for users to log in for.
> 
> One weird thing - somehow seem to have lost the text from the file menu:
> Screenshot from 2020-03-20 10-43-06
> This is a css issue - the text is there, just white.

Oops, fucked up that css. Fixed now. Good point about user flow, that does sound like it might be preferable but as you say, a bit trickier to implement. I'll raise a ticket for it.

PR #250, opened 2020-03-20
mrc-1417 Dyamic colour scales for Survey and Program

EmmaLRussell:


src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -174,11 +169,16 @@
             },
             selectedAreaIds() {
                 const selectedAreaIdSet = flattenToIdSet(this.selectedAreaFilterOptions.map(o => o.id), this.flattenedAreas);
+
+                //Should also ensure include top level (country) included if no filters selected
+                const selectedOptions = this.selections.selectedFilterOptions[this.areaFilterId];
+                if (!selectedOptions || selectedOptions.length == 0) {
```
EmmaLRussell: This fixes a previously undiagnosed bug which meant that data at level 0 (country) was not shown when selected because it wasn't included in the selectedAreaIds set

src/app/static/src/app/components/plots/utils.ts
```diff
@@ -107,6 +114,22 @@ export const getColourRanges = function(data: any,
   return result;
 };
 
+export const roundRange = function(unrounded: NumericRange) {
```
hillalex: This is really clever and works really nicely for most cases, but I randomly put in 1.1234 and 1.1235 and it rounded things to 1 and 1. I put in a debug statement and it's because 

`((1.1235-1.1234)/100).toString() = 9.999999999998898e-7`

Maybe you could get around this by just checking if there's an "e" in the string and if so don't round them? I'm guessing it's an edge case for values to be so super close together. 


Could also just use the difference e.g. `1.1235-1.1234` as the magnitude and then just prepend "00" to the `fractionString` rather than actually dividing the number by 100? Of course that still runs into the same bug for even tinier numbers like 1.1234567 - 1.1234566 (but things are getting a bit ridiculous at that point!)

EmmaLRussell: Ha, well spotted. I've taken out reliance on strings..

src/app/static/src/app/components/plots/utils.ts
```diff
@@ -118,13 +118,10 @@ export const roundRange = function(unrounded: NumericRange) {
...
     let magnitude = unrounded.max == unrounded.min ? unrounded.min : (unrounded.max - unrounded.min);
-    magnitude = magnitude / 100;
 
+    magnitude = magnitude / 100;
     if (magnitude < 1 && magnitude > 0) {
-        let fractionString = magnitude.toString().split('.')[1];
-        while((fractionString.length > decPl) && (fractionString[decPl] =='0')) {
-            decPl++;
-        }
+        decPl = Math.trunc(Math.abs(Math.log10(magnitude)));
```
hillalex: Much neater, nice!

PR #253, opened 2020-03-26
mrc-1444 performance improvement

hillalex:
I'm actually quite surprised by what a difference this one change makes - it takes "excludeRow" from taking up around 50% of the total scripting time to around 9%, and at least with the test data takes the filtering back to feeling snappy and responsive.

It still takes about a second to toggle between map and bubble tabs so will continue to look into that to see if anything can be done.

Edited: ok, yeah so if the colour range is set to "whole data set" then it's still slow (and the bubble plot is slow because it works out the size range by iterating the whole data set) so reducing those iterations should help. I'll do that on a separate branch.

### DIFF
--- a/src/app/static/src/app/components/plots/utils.ts
+++ b/src/app/static/src/app/components/plots/utils.ts
@@ -2,7 +2,6 @@ import * as d3ScaleChromatic from "d3-scale-chromatic";
 import {ChoroplethIndicatorMetadata, FilterOption} from "../../generated";
 import {Dict, Filter, NumericRange} from "../../types";
 import {ColourScaleSelections, ColourScaleType} from "../../store/plottingSelections/plottingSelections";
-import {mockAncResponse} from "../../../tests/mocks";
 
 export const getColor = (value: number, metadata: ChoroplethIndicatorMetadata,
                          colourRange: NumericRange) => {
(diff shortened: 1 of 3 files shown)


src/app/static/src/app/components/plots/utils.ts
```diff
@@ -136,8 +135,17 @@ export const iterateDataValues = function(
     func: (areaId: string,
            indicatorMeta: ChoroplethIndicatorMetadata, value: number) => void) {
 
+    const selectedFilterValueIds: Dict<string[]> = {};
+    const validFilters = filters && filters.filter(f => f.options && f.options.length > 0);
+
+    if (validFilters && selectedFilterValues) {
+        for (const f of validFilters) {
+            selectedFilterValueIds[f.id] = selectedFilterValues[f.id].map(n => n.id)
```
EmmaLRussell: I get an error on this line: `TypeError: Cannot read property 'map' of undefined` when, having just arrived in the Output step, I set the colour scale for the default indicator to 'Filtered dataset' then switch to the Bubble plot. I guess the selectedFilterValues in the BubblePlot haven't yet been initialised when that  code is first called.

hillalex:
Ok, handling this case now inside the `iterateDataValues` method but also stopped the bubble plot calling that method before it's initialised.

hillalex:
GitHub being a bit sketchy again, the Travis builds have actually passed.

PR #254, opened 2020-03-26
mrc-1444 further performance improvement

hillalex:
The output file is >20MB and there are 10 indicators, so iterating all the rows x 10 indicators is just too much. So this PR changes things from fetching colors and values for all indicators at once and storing them in a big dict, to just requesting colors and values (and radius in bubble case) for the currently selected indicator. 

This branch inherits from https://github.com/mrc-ide/hint/pull/253 so that should be merged first.

It's a lot of lines of code changed, sorry, but a lot of it is tests and hopefully since you're so familiar with this code it should still be parseable for review.

### DIFF
--- a/src/app/static/src/app/components/plots/bubble/BubblePlot.vue
+++ b/src/app/static/src/app/components/plots/bubble/BubblePlot.vue
@@ -256,9 +256,6 @@
                         };
                     case ColourScaleType.Default:
                     default:
-                        if (!this.initialised) {
-                            return {max: 1, min: 0}
-                        }
                         return {max: this.colorIndicator.max, min: this.colorIndicator.min}
                 }
             },
--- a/src/app/static/src/app/components/plots/bubble/BubblePlot.vue
+++ b/src/app/static/src/app/components/plots/bubble/BubblePlot.vue
@@ -60,7 +60,7 @@
                              :level-labels="featureLevels"
                              @detail-changed="onDetailChange"></map-control>
                 <map-legend :metadata="colorIndicator"
-                            :colour-range="colourIndicatorRange"
+                            :colour-range="colourRange"
                             :colour-scale="colourIndicatorScale"
                             @update="updateColourScale"
                 ></map-legend>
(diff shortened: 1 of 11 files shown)


src/app/static/src/app/components/plots/bubble/BubblePlot.vue
```diff
@@ -207,23 +210,54 @@
...
+                const sizeId = this.selections.sizeIndicatorId;
+                if (!this.fullIndicatorRanges.hasOwnProperty(sizeId)) {
+                    // cache the result in the fullIndicatorRanges object for future lookups
+                    this.fullIndicatorRanges[sizeId] =
+                        getIndicatorRange(this.chartdata, this.sizeIndicator)
+                }
+                return this.fullIndicatorRanges[sizeId];
+            },
+            colourRange() {
+                const colorId = this.selections.colorIndicatorId;
```
hillalex: This logic needs to live in the component itself so that the result is cached and it doesn't get re-calculated unnecessarily. To tidy up these 2 components which are massive and share a lot of code, we should refactor to make use of mixins. I'll open a ticket to do this.

EmmaLRussell: Good idea, wanted to do that anyway, even more desirable now!

src/app/static/src/app/components/plots/utils.ts
```diff
@@ -44,73 +44,35 @@ export const colorFunctionFromName = function (name: string) {
...
-                                            selectedFilterValues: Dict<FilterOption[]> | null = null,
-                                            selectedAreaIds: string[] | null = null): Dict<NumericRange> {
-    const result = {} as Dict<NumericRange>;
-    iterateDataValues(data, indicatorsMeta, selectedAreaIds, filters, selectedFilterValues,
+export const getIndicatorRange = function (data: any,
+                                           indicatorMeta: ChoroplethIndicatorMetadata,
+                                           filters: Filter[] | null = null,
+                                           selectedFilterValues: Dict<FilterOption[]> | null = null,
+                                           selectedAreaIds: string[] | null = null): NumericRange {
+    let result = {} as NumericRange;
```
EmmaLRussell: I think we want to roundRange the full dynamic as well as filtered dynamic ranges. It's getting applied in getDyanamicFilteredColourRange, which calls this method, but maybe it should be applied to the result here before returning, so it applies to full ranges too.

src/app/static/src/app/components/plots/bubble/BubblePlot.vue
hillalex: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1448

src/app/static/src/app/components/plots/utils.ts
hillalex: Ah cool, I've just moved the rounding logic into `getIndicatorRange` and removed the duplicate `getDyanamicFilteredColourRange` method - does that seem right to you? There's no reason not to always round the former right?

PR #255, opened 2020-03-27
mrc-1434 Translations for custom colour scales

EmmaLRussell:


src/app/static/src/tests/components/plots/mapLegend.test.ts
```diff
@@ -25,7 +34,7 @@ describe("Map legend component", () => {
                 customMax: 2.5
             },
             colourRange
-        }
+        }, localVue
```
hillalex, line 37: why add this localVue here? and actually this wrapper is only used in 1 test, so probably makes sense to declare it within the scope of that test to keep things clear

hillalex: oops, merged this (code and functionality all looks great) but meant to raise this comment about the test code @EmmaLRussell

EmmaLRussell: I think that might have been a relic from an earlier version of the test and isn't actually needed, will remove in separate branch. .

hillalex: nice, thanks

PR #259, opened 2020-06-19
mrc-1641 Create version

EmmaLRussell:
This adds a front page into the application flow, where logged in user must create a new version. Creating a new version initialises a first snapshot for the version and sets that as the current snapshot. 
Version management is not available to guest users. 
Logged in users will see the version page when they first log in, and are required to create a new version to continue. 
Users can also return to this page using the 'Versions' link at any time, and create a fresh version.
The page shows the name of the current version, if any, and a link to return to the current version. 

Future tickets will cover saving front end state for the version snapshot, spawning new snapshots for invalidating changes ,showing and loading previous versions etc.


src/app/src/test/kotlin/org/imperial/mrc/hint/database/SnapshotRepositoryTests.kt
```diff
@@ -30,25 +31,32 @@ class SnapshotRepositoryTests {
...
     fun `can save snapshot without version id`() {
         sut.saveSnapshot(snapshotId, null)
 
         val snapshot = dsl.selectFrom(VERSION_SNAPSHOT)
                 .fetchOne()
-
         assertThat(snapshot[VERSION_SNAPSHOT.ID]).isEqualTo(snapshotId)
-        assertThat(snapshot[VERSION_SNAPSHOT.VERSION_ID]).isNull()
+
+        val versionId: Int? = snapshot[VERSION_SNAPSHOT.VERSION_ID]
```
EmmaLRussell, line 51: Oddly, version_id has type Int rather than Int? in the generated db interface - it's a foreign key but a nullable foreign key, and it's quite happy to have nulls in the table, but there was a runtime type exception when comparing to null without this cast..

src/app/static/src/app/apiService.ts
```diff
@@ -15,7 +15,7 @@ export interface ResponseWithType<T> extends Response {
 export interface API<S, E> {
 
     withError: (type: E) => API<S, E>
-    withSuccess: (type: S) => API<S, E>
+    withSuccess: (type: S, root: boolean) => API<S, E>
```
EmmaLRussell, line 18: Added this option so I could use a root mutation as the success mutation for a new version, to reset the entire root state (without it, Vuex assumes the named mutation is relative to the home module, versions).

src/app/static/src/scss/partials/versions.scss
```diff
@@ -0,0 +1,3 @@
+.versions-header {
```
hillalex, line 1: Suggest dispensing with this custom class and just using Bootstrap's built in `lead` class.

src/app/static/src/app/components/versions/Versions.vue
```diff
@@ -1,21 +1,41 @@
...
+        <div v-if="!loading" id="versions-content" >
+            <div class="versions-header">
+                <span v-translate="'versionsHeaderCreate'"></span>
+                <span v-if="currentVersion">
+                    <span v-translate="'or'"></span>
+                    <a v-translate="'versionsHeaderReturn'"
+                       href="#" @click="setManageVersions(false)"></a> ({{currentVersion.name}})
+                </span>
+            </div>
+            <div class="my-3 col-6 clearfix">
```
hillalex, line 12: There is some weird padding going on here and it's because this `col-6` is not wrapped in a parent `row`. I would suggest adding the `row` class to the div with id `versions-content`, then giving `versions-header` a `col-12` class.

EmmaLRussell:
Agreed, will do this in another ticket: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1710

> I did also realise that there is 1 big advantage to using proper routing - back button usage. I without thinking tried to use the back button to navigate between the screens which of course does not work. This might be a reason to switch the routing.

PR #263, opened 2020-07-01
mrc-1641_schema Use new schema tables

EmmaLRussell:
Use the Version_snapshot table rather than user_session, update all related classes and methods accordingly, and use a snapshot id variable in the session, rather than session id.


src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
```diff
@@ -38,14 +39,31 @@ class Pac4jConfig {
...
 
     fun getUserProfile(): CommonProfile {
         val manager = ProfileManager<CommonProfile>(webContext)
         val profiles = manager.getAll(true)
         return profiles.singleOrNull() ?: CommonProfile().apply {
             id = GUEST_USER
         }
+    }
+
+    fun getSnapshotId() :String {
```
hillalex, line 54: So I *think* this is all implicitly under test because otherwise the integration tests that upload files that depend on the shape file wouldn't work (e.g. if it was generating a new shapshot id every time rather than correctly reading it from the session). Could add some unit tests for these methods but would be a faff to mock the session store, so I'm happy with implicit tests if you are!

hillalex: (Go ahead and merge if so)

EmmaLRussell: Yeah I noticed the same thing and came to the same conclusio!

PR #265, opened 2020-07-03
mrc-1645 Save version snapshot state

EmmaLRussell:
Save the current version snapshot state to the backend whenever a mutation occurs. Because mutations sometimes come in a cascade of several close together we wait a couple of seconds before saving any pending changes.


src/app/static/src/app/store/versions/mutations.ts
```diff
@@ -0,0 +1,31 @@
...
+    },
+    [VersionsMutations.SetSnapshotUploadPending](state: VersionsState, action: PayloadWithType<boolean>) {
+        state.snapshotUploadPending = action.payload;
+    },
+    [VersionsMutations.VersionError](state: VersionsState, action: PayloadWithType<Error>) {
+        state.error = action.payload;
+        state.loading = false;
+    },
+    [VersionsMutations.SnapshotUploadError](state: VersionsState, action: PayloadWithType<Error>) {
+        console.error(action.payload);
```
EmmaLRussell, line 29: Wasn't really sure what to do here. This could happen anywhere in the stepper flow, so not really sure how to flag to user, if at all.

src/app/src/main/kotlin/org/imperial/mrc/hint/db/SnapshotRepository.kt
```diff
@@ -144,6 +144,32 @@ class JooqSnapshotRepository(private val dsl: DSLContext) : SnapshotRepository {
...
+    private fun checkSnapshotExists(snapshotId: String, versionId: Int, userId: String)
+    {
+        val count = dsl.select(count(VERSION_SNAPSHOT.ID))
+                .from(VERSION_SNAPSHOT)
+                .join(VERSION)
+                .on(VERSION_SNAPSHOT.VERSION_ID.eq(VERSION.ID))
+                .where(VERSION_SNAPSHOT.ID.eq(snapshotId))
+                .and(VERSION.ID.eq(versionId))
+                .and(VERSION.USER_ID.eq(userId))
+                .fetchOne().value1()
```
hillalex, line 165: could do `fetchAny()?: throw ...`

src/app/static/src/app/store/versions/mutations.ts
hillalex: We already console log all errors via the `apiService` so may as well just remove this method and have `.ignoreErrors()` set. 

Since there's no action a user can take if an auto-save fails, I guess there's not much point in advertising it but maybe we could have a little icon somewhere like "last saved: timestamp"? Which could also turn into an error icon if it fails to save?

src/app/static/src/app/store/versions/actions.ts
```diff
@@ -1,22 +1,56 @@
...
+                commit({type: VersionsMutations.SetSnapshotUploadPending, payload: true});
+                setTimeout(() => {
+                    immediateUploadSnapshotState(context);
+                }, 2000);
+            }
+        }
     }
 };
+
+const immediateUploadSnapshotState = (context: ActionContext<VersionsState, RootState>) => {
```
hillalex, line 44: This also needs an integration test I think?

src/app/static/src/app/store/versions/mutations.ts
EmmaLRussell: Good idea - I've had a stab at doing a little component for this in another branch, see what you think: https://github.com/mrc-ide/hint/pull/280

src/app/static/src/app/root.ts
```diff
@@ -57,7 +57,7 @@ const persistState = (store: Store<RootState>) => {
 
         const {dispatch} = store;
         const type = stripNamespace(mutation.type);
-        if (type[0] !== "versions") {
+        if (type[0] !== "versions" && type[0] !== "errors") {
```
EmmaLRussell, line 60: Don't attempt snapshot save on error, get infinite death spiral if connection goes down.

PR #266, opened 2020-07-17
prototype table implementation from f2f323b

Nicolas-Dolan:
f2f323b


src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -0,0 +1,183 @@
...
+    selections: {
+        type: Object
+    }
+}
+export default Vue.extend<{}, {}, Computed, Props>({
+    name: "table-view",
+    props: props,
+    data() {
+      return {
+          styleObject: {
```
EmmaLRussell, line 80: Could be a class in the scss?

src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -0,0 +1,183 @@
...
+
+            if (result.length > 0){
+
+                let filterObject: any = {...this.selections.selectedFilterOptions}
+                delete filterObject['area']
+                Object.keys(filterObject).map(function(key, index) {
+                    if (filterObject[key].length < 1 || key === 'area') {
+                        delete filterObject[key]
+                    } else if (key === 'survey'){
+                        filterObject[key] = filterObject[key][0]['label']
```
EmmaLRussell, line 129: This won't work when you've got multiple values for a filter as in the barchart I think. I'd be inclined not to worry about the filter labels from the seleted filters object until after you've constructed the complete set of rows based on values. It feels safer to me to compare values and just apply labels when you're ready to display.

EmmaLRussell:
I had a look at the country level bug, and it's actually quite a nasty issue caused by the fact that when we're setting up the filters we deliberately remove the top level area filter option because we don't want it appearing in any filter drop downs. So it isn't actually available to the Table to access, annoyingly. The Choropleth doesn't have this issue because it uses features (the geo shape data) to find that top level area id. There's a couple of ways we could address this but I'd suggest leaving it for another ticket.

src/app/static/src/tests/components/plots/table/table.test.ts
```diff
@@ -0,0 +1,247 @@
...
+registerTranslations(store);
+
+const propsData = {
+    selections: {
+        indicatorId: "prevalence",
+        detail: 4,
+        selectedFilterOptions: {
+            age: [{id: "0:15", label: "0-15"}],
+            sex: [{id: "female", label: "Female"}],
+            area: []
```
EmmaLRussell, line 20: Could we also have a test where an area filter is set to include one or more values, and check that this works with the detail level?

src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
```diff
@@ -107,6 +121,9 @@
             };
         },
         computed: {
+            filterTableIndicators(){
+                return this.sapIndicatorsMetadata.filter((val: any) => val.indicator === this.plottingSelections.indicatorId)
```
EmmaLRussell, line 125: This is the line that's making the tests fail, but I think it's because the Computed sapIndicatorsMetadata prop above in the Computed interface is mistyped. It runs ok, but the test fails to compile that new line. Should be `sapIndicatorsMetadata: ChoroplethIndicatorMetadata` I think?

PR #268, opened 2020-07-27
mrc-1641_nav Navigation logic and simple Versions component

EmmaLRussell:
This adds the navigation logic where a logged in user should always be working within a current version and snapshot. 

Logged in users see a Versions screen on log in, and a menu item which will take them back there at any time. 

The Versions screen is not available to the guest user. 

In this ticket, the Versions screen consists of a Create Version button, which sets current version to a hardcoded string value. The next ticket will create the new version and snapshot by calling endpoint, and update all state.

Future tickets will provide the option to load previous versions.


EmmaLRussell:
Yes I wondered about using Vuex routing as well, and iirc decided that it would be more straightforward as you say to just to manage the logic (of the user types and requiring logged in users to have a current version) through the state. I'd want to reconsider if that logic changes though. 

> This all looks fine, I'm just wondering whether this sort of bespoke navigation is the way to go vs using Vuex router (so the versions page would be viewable at `/versions`.) What do you think are the pros/cons here? Tricky to accommodate the difference between guest/logged in user perhaps?

PR #270, opened 2020-07-27
mrc-1706 db methods for managing API keys

hillalex:
~~Contains commits from https://github.com/mrc-ide/hint/pull/269~~

Stuck these methods in the `UserRepo` rather than proliferate repo classes (seeing as there's only ever likely to be these 2 methods pertaining to this table.)


hillalex:
Just realised my comment is of course wrong, there will be 1 more method - `getADRKey()`

hillalex:
Replacing this PR with an updated one based on the updated db interface

PR #272, opened 2020-07-28
mrc-1678 endpoints for managing ADR keys

hillalex:
~~Contains commits from https://github.com/mrc-ide/hint/pull/274 so that should be merged first and then the base here switched back to master.~~

This PR adds GET, POST and DELETE endpoints for managing ADR keys.


hillalex:
> This all looks good but it's just occurred to me to wonder what should happen with these ADR keys for a guest user. They shouldn't have the option to set ADR keys (?), so I guess the widget should be hidden in the front end? Or should they be able to save a key just for the session?
> That could impact these endpoints too.

Yeah it should all be hidden for the guest user. I'm imagining the select widget will have the ADR option greyed out with a tooltip saying "you have to log in to use this". But would be prudent to lock down these endpoints as well, I'll do that :+1:

hillalex:
I've had a very confusing couple of hours that I attribute to something updating in our un-versioned dependencies (all the spring boot ones) but I can't pin down what exactly has changed. 

The `TestRestTemplate` is throwing an error when a 401 response is returned, instead of just returning a nice entity with a 401 status. I found a fix for this here (this q/a dates back to 2017, which makes it even more baffling as we didn't see this prior to March 2020 - until https://github.com/mrc-ide/hint/pull/248 we were using this testing pattern everywhere) : https://stackoverflow.com/questions/27341604/exception-when-using-testresttemplate

However introducing the new apache http client dependency breaks 2 other tests, the ones that check content is being gzipped. This is because the apache client automatically decompresses content and removes the content encoding header (bizarre). The solution for this I found here: https://stackoverflow.com/questions/30216375/extracting-gzip-data-from-apache-httpclient-without-decompression

So this is all fixed, but I am very confused about what has caused this in the first place.

EmmaLRussell:
How bizarre, no idea!

> I've had a very confusing couple of hours that I attribute to something updating in our un-versioned dependencies (all the spring boot ones) but I can't pin down what exactly has changed.
> 
> The `TestRestTemplate` is throwing an error when a 401 response is returned, instead of just returning a nice entity with a 401 status. I found a fix for this here (this q/a dates back to 2017, which makes it even more baffling as we didn't see this prior to March 2020 - until #248 we were using this testing pattern everywhere) : https://stackoverflow.com/questions/27341604/exception-when-using-testresttemplate
> 
> However introducing the new apache http client dependency breaks 2 other tests, the ones that check content is being gzipped. This is because the apache client automatically decompresses content and removes the content encoding header (bizarre). The solution for this I found here: https://stackoverflow.com/questions/30216375/extracting-gzip-data-from-apache-httpclient-without-decompression
> 
> So this is all fixed, but I am very confused about what has caused this in the first place.

PR #273, opened 2020-07-28
mrc-1676 add ADR client

hillalex:
~~Contains commits from https://github.com/mrc-ide/hint/pull/274 which should be merged first and the base here pointed back to master.~~

This PR:
* factors out an abstract `FuelClient` for both the hintr client and the ADR client to inherit from
* adds an `ADRClient` 
* adds a helper for instantiating an `ADRClient` using the session user's api key


src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
```diff
@@ -51,16 +53,22 @@ fun Response.asResponseEntity(): ResponseEntity<String> {
...
+                    .body(body)
+        } else {
+            // this is an ADR response, so convert to our response schema
+            SuccessResponse(json["result"])
+                    .asResponseEntity()
+        }
 
     } catch (e: IOException) {
-        ErrorDetail(httpStatus, "Could not parse response.")
+        ErrorDetail(httpStatusFromCode(500), "Could not parse response.")
```
hillalex, line 71: Not related to this changeset but if we're returning an error response, we always want this to be a 500, even if the original request returned a different status code. See updated test of this here: https://github.com/mrc-ide/hint/pull/273/files#diff-5db22b6b5a666b3836c7f617a4378d0eR76

src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
```diff
@@ -51,16 +53,22 @@ fun Response.asResponseEntity(): ResponseEntity<String> {
...
-                .contentType(MediaType.APPLICATION_JSON)
-                .body(body)
+        return if (json.has("status")) {
+            // this is a hintr response, so already conforms to our response schema
+            ResponseEntity.status(httpStatus)
+                    .contentType(MediaType.APPLICATION_JSON)
+                    .body(body)
+        } else {
+            // this is an ADR response, so convert to our response schema
+            SuccessResponse(json["result"])
```
EmmaLRussell, line 66: Is this going to behave correctly if ADR gives back something other than a 200?

hillalex: Er, no it won't! Thanks for spotting. Actually the ckan API always returns a 200, even if the request errors, but we should be looking at the `success` property to work out whether it succeeded or not: https://github.com/mrc-ide/hint/pull/273/commits/bd51356bca226ea6d422f0f94f9c527219aaf53c

PR #274, opened 2020-07-28
mrc-1709 methods for managing api keys in db

hillalex:
This PR:
* updates the db interface to store api keys as byte arrays
* updates the encryption methods to deal with encrypted keys as byte arrays
* adds save/delete/get methods for interfacing with keys in the db

Stuck these methods in the UserRepo rather than proliferate repo classes (seeing as there's only ever likely to be these 3 methods pertaining to this table.)


hillalex:
Ok, I'm going to merge this as quite a lot of other stuff depends on it.

PR #276, opened 2020-07-28
mrc-1679 add ADR API key widget

hillalex:
This PR adds a widget for adding/editing/removing an ADR API key. It just stores the key in the Vuex store state; a further PR will wire this up to the backend.


src/app/static/src/app/components/adr/ADRKey.vue
```diff
@@ -0,0 +1,80 @@
+<template>
+    <div class="row mb-3">
+        <div class="col-8">
+            <div class="d-flex">
+                <label for="key"
+                       class="font-weight-bold align-self-stretch"
+                       >ADR API Key: </label>
```
EmmaLRussell, line 7: I guess this and the edit links need a translation..

src/app/static/src/app/components/adr/ADRKey.vue
```diff
@@ -0,0 +1,80 @@
...
+                if (this.key) {
+                    let str = ""
+                    let count = this.key.length
+                    while (count) {
+                        str += "*";
+                        count--;
+                    }
+                    return str;
+                } else {
+                    return "none provided"
```
EmmaLRussell, line 57: Translation needed for this too

src/app/static/src/app/components/adr/ADRKey.vue
```diff
@@ -0,0 +1,80 @@
...
+    <div class="row mb-3">
+        <div class="col-8">
+            <div class="d-flex">
+                <label for="key"
+                       class="font-weight-bold align-self-stretch"
+                       >ADR API Key: </label>
+                <div class="align-self-stretch pl-2">
+                    <div v-if="!editing">
+                        <span class="pr-2">{{keyText}}</span>
+                        <a href="#" v-if="!key" @click="edit">add</a>
```
EmmaLRussell, line 11: Using falsiness here is nice because I guess if the user hadn't spotted 'remove' they might just delete the whole key and I guess end up with empty string rather than null, which should behave as if 'none provided'. Worth a unit test?

src/app/static/src/app/components/adr/ADRKey.vue
```diff
@@ -0,0 +1,80 @@
...
+                        <a href="#" v-if="key" @click="edit">edit</a>
+                        <span v-if="key">/</span>
+                        <a href="#" v-if="key" @click="remove">remove</a>
+                    </div>
+                    <div class="input-group"
+                         style="margin-top: -11px; min-width: 390px"
```
EmmaLRussell, line 17: Is there a reason for bespoke width rather than bootstrap class? Is this the right size for showing real ADR key?

src/app/static/src/app/components/Stepper.vue
```diff
@@ -37,6 +37,7 @@
         </div>
         <div v-if="!loading" class="content">
             <div class="pt-4">
+                <adr-key v-if="isActive(1) || isActive(2)"></adr-key>
                 <baseline v-if="isActive(1)"></baseline>
                 <survey-and-program v-if="isActive(2)"></survey-and-program>
```
Nicolas-Dolan, line 42: I had been wondering what the parent component of the sap and output were.

src/app/static/src/app/components/adr/ADRKey.vue
```diff
@@ -0,0 +1,80 @@
...
+                         v-if="editing">
+                        <input id="key"
+                               class="form-control"
+                               v-model="editableKey"
+                               type="text"
+                               placeholder="Enter key"/>
+                        <div class="input-group-append">
+                            <button class="btn btn-red" type="button" @click="save">Save</button>
```
Nicolas-Dolan, line 25: Presumably you would need a translation for Save.

hillalex: Great catch, thanks

src/app/static/src/app/components/adr/ADRKey.vue
hillalex: The margin is to bottom align the input with the label. The width is the min width needed for a real ADR key.

src/app/static/src/app/components/adr/ADRKey.vue
hillalex: I've actually added a check to stop them from saving an empty string:
https://github.com/mrc-ide/hint/pull/276/commits/e43ff8d8007ff9ef54c4bf3f2ba56c7e1f67106f

Nb, we'll probably expand this at some point to get some validation of the key by making a request with it to ADR.

hillalex:
Added translations (ticket open to add the French: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1714.) Also added in a check so that the widget only renders if the user is logged in. 

We might want to think about changing the way we handle the current user variable? At the moment it's a global var on the page that gets passed to each top level component as a prop, which is fine for those top level components but doesn't work for something like this. Would probably be better if we could get it in the store state so any component can access it in the normal way.

hillalex:
> Looks good! Sorry, I just thought of another thing, which is that it should probably focus the input when you start editing. But happy for this to be merged first.

Good idea, it's those little details that make the user experience better :+1:

hillalex:
Added in the auto-focusing here:  https://github.com/mrc-ide/hint/pull/276/commits/cdb8fabde1a4ee5e9c4a9718606b34cb4057f805

EmmaLRussell:
You need to log in maybe @Nicolas-Dolan  ?

> I'm having trouble viewing the widget locally. I suppose there is some condition I need to meet for it to appear.

PR #278, opened 2020-07-29
mrc-1635_frontend Show version history

EmmaLRussell:
Display version history in Versions page - no controls to load or delete yet.

Snapshots are nested under versions. Currently there is only ever one snapshot per version since they are not automatically spawned at invalidation points yet. Once this is up and running we could maybe not nest snapshots where there only is one in the version, as the latest will be shown at the version level anyway. Maybe we also shouldn't show the current snapshot in the history list.


src/app/static/src/app/components/versions/VersionHistory.vue
```diff
@@ -0,0 +1,69 @@
...
+
+    interface Props {
+        versions: Version[];
+    }
+
+    interface Methods {
+        format: (date: string) => string
+    }
+
+    export default Vue.extend<{}, {}, {}, Props>({
```
hillalex, line 51: `Methods` interface isn't actually used!

src/app/static/src/app/components/versions/VersionHistory.vue
```diff
@@ -0,0 +1,69 @@
...
+
+<script lang="ts">
+    import Vue from "vue";
+    import {Version} from "../../types";
+    import {BCollapse} from "bootstrap-vue";
+    import { VBToggle } from 'bootstrap-vue';
+    import {ChevronDownIcon, ChevronRightIcon} from "vue-feather-icons";
+    import {formatDateTime} from "../../utils"
+
+    Vue.directive('b-toggle', VBToggle);
```
hillalex, line 41: If this directive is just for use here you can register directives per component in the same way you register other components:
`directives: {'b-toggle': VBToggle}`

If you think it'll be used throughout the app then maybe stick the registration in `main.ts` where we register the `v-translate` directive.

src/app/static/src/app/components/versions/VersionHistory.vue
```diff
@@ -0,0 +1,69 @@
...
+                        <chevron-down-icon size="20" class="icon when-open"></chevron-down-icon>
+                    </button>
+                </div>
+                <div class="col-md-3 version-cell">
+                    {{v.name}}
+                </div>
+                <div class="col-md-3 version-cell">{{format(v.snapshots[0].updated)}}</div>
+            </div>
+            <b-collapse :id="`snapshots-${v.id}`">
+                <div v-for="s in v.snapshots" :id="`s-${s.id}`" class="row font-italic bg-light pb-3">
```
hillalex, line 24: Generally I think this looks and works great. There's some slight padding weirdness - this should probably be `py-2` rather than `pb-3` so it's even top and bottom, the same `py-2` added to each `version-cell` (line 12), and then `py-0` on the button toggle (line 13) so that it doesn't add extra vertical padding

src/app/static/src/app/components/versions/VersionHistory.vue
EmmaLRussell: OK, just need it here for now.

hillalex:
I think each `version-cell` should also have `py-2` added (or add a css rule  `.version-cell {@extend .py-2 }`).

But also something weird going on with the toggle buttons (not sure how I failed to notice this before sorry :disappointed: )

Screenshot from 2020-08-04 17-17-05

EmmaLRussell:
Should just need to refresh npm install to get the right version of bootstrp-vue

> Screenshot from 2020-08-04 17-17-05

hillalex:
> Should just need to refresh npm install to get the right version of bootstrap-vue - older versions didin't add the `while-closed` class iirc
> 

Ah that explains why I only just noticed it! Yes looks good after a fresh install :+1: Just the `version-cell` padding then, feel free to merge after that

PR #279, opened 2020-07-29
mrc-1713 wire up front-end ADR key widget

hillalex:
Contains commits from https://github.com/mrc-ide/hint/pull/276 which should be merged first and the base of this PR put back to master.

This PR wires up the ADR API key widget so it actually fetches/saves/removes the key from the backend.

### DIFF
--- a/src/app/static/src/app/store/root/actions.ts
+++ b/src/app/static/src/app/store/root/actions.ts
@@ -5,9 +5,14 @@ import {RootMutation} from "./mutations";
 import {LanguageActions} from "../language/language";
 import {changeLanguage} from "../language/actions";
 import i18next from "i18next";
+import {api} from "../../apiService";
+import qs from "qs";
 
 export interface RootActions extends LanguageActions<RootState> {
     validate: (store: ActionContext<RootState, RootState>) => void;
+    fetchADRKey: (store: ActionContext<RootState, RootState>) => void;
+    saveADRKey: (store: ActionContext<RootState, RootState>, key: string) => void;
+    deleteADRKey: (store: ActionContext<RootState, RootState>) => void;
 }
 
 export const actions: ActionTree<RootState, RootState> & RootActions = {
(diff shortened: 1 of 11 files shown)


src/app/static/src/app/store/root/actions.ts
```diff
@@ -52,5 +57,27 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
 
     async changeLanguage(context, payload) {
         await changeLanguage<RootState>(context, payload)
+    },
+
+    async fetchADRKey(context) {
+        await api<RootMutation, RootMutation>(context)
+            .withSuccess(RootMutation.UpdateADRKey)
```
EmmaLRussell, line 64: Should include ignoreError to suppress console warning? Or do something on error?

src/app/static/src/app/store/root/actions.ts
```diff
@@ -52,5 +57,27 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
...
+    async saveADRKey(context, key) {
+        await api<RootMutation, RootMutation>(context)
+            .withSuccess(RootMutation.UpdateADRKey)
+            .postAndReturn("/adr/key/", qs.stringify({key}));
+    },
+
+    async deleteADRKey(context) {
+        await api<RootMutation, RootMutation>(context)
+            .delete("/adr/key/")
+            .then((response) => {
```
EmmaLRussell, line 77: Why not withSuccess as for the others?

src/app/static/src/app/store/root/actions.ts
hillalex: Probably should handle this error actually and show back to the user :+1:

src/app/static/src/app/store/root/actions.ts
hillalex: I was thinking the payload back from the API would be "OK", wile the payload we want to commit is `null`. But actually looking at the controller the payload back from the API will be null, so this _should_ work as `withSuccess`, will change that over.

hillalex:
Codecov at the project level complaining, no idea why tbh. I'll turn that off as we've done on other projects, as it's really the diff coverage we should be considering.

src/app/static/src/app/store/root/actions.ts
```diff
@@ -52,5 +57,28 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
 
     async changeLanguage(context, payload) {
         await changeLanguage<RootState>(context, payload)
+    },
+
+    async fetchADRKey(context) {
+        await api<RootMutation, RootMutation>(context)
+            .ignoreErrors()
```
EmmaLRussell, line 64: Does this still need to be changed to withError?

hillalex: I think if fetching the API key fails there's not much point in complaining about it as the user can't really take another action to rectify it? Unlike saving/deleting where they can try again.

EmmaLRussell: Fair enough!

PR #281, opened 2020-08-03
mrc-1710 Use vue routing for versions page

EmmaLRussell:
This branch replaces the management of which view to show using a manageVersions store property with the use of vue-router. We push the router programmatically when required: to ensure a logged in user with no current version is sent to the /versions view, and to navigate to the root stepper view when a version is successfully created. 

Back and forward browser buttons should not behave as expected.


src/app/static/src/app/index.ts
```diff
@@ -3,12 +3,19 @@ import {store} from "./main"
...
 import Errors from "./components/Errors.vue";
+import Stepper from "./components/Stepper.vue";
+import Versions from "./components/versions/Versions.vue";
 import {mapActions, mapState} from "vuex";
 import {RootState} from "./root";
 import VueRouter from "vue-router";
 
 Vue.use(VueRouter);
 
+router.addRoutes( [
```
EmmaLRussell, line 14: The routes were originally being added in router.ts when router is created, but I had to move them out of that module as the merge with the ADR widget caused a circular dependency: Stepper -> ADR component -> Root Mutations -> router -> Stepper... 
Other ways of doing that, but this seemed easiest while not being too hacky.

PR #282, opened 2020-08-03
Mrc 1705 fixes top-level country data in table

Nicolas-Dolan:
Resolves the issue regarding the top-level country data not being shown correctly in the tables. The issue was being caused upstream by the getters for the sap and model output removing top-level country data. This update adds a one-off getter for each called countryAreaFilterOption, which allows the tables to reference the country-level data.


Nicolas-Dolan:
> This isn't how I expected you to do this, but it does work!
> 
> You aren't actually returning a 'countryAreaFilterOption' from your getter but rather,I think, a copy of all the filters, where the area filter still has the country level option in place).
> In a way that potentially simplifies the code in the Table since it doesn't need to do anything special when detail == 0. However, to me, it seems a bit confusing that the areaFilter you're handling in the Table is a different one from the areaFilter the other plots deal with.
> 
> I'd suggest there's two possible approaches you could take:
> 
> 1. revert to the original plan to have a 'countryAreaFilterOption', actually return that from the getter, and return it from selectedAreaFilterOptions when detail == 0
> 2. accept the confusion of a slightly different area filter in Table. Definitely rename the getter to something more accurate and just return the area Filter there (you don't need copies of the others) - call it something like areaFilterWithCountry perhaps. Then in Table, nonAreaFilters can filter out the (countryless) areaFilter, and you can use areaFilterWithCountry in place of the current area filter (I think).
> 
> One slightly odd behaviour with the current implementation - if you have areas selected and detail = Country, the first selected area is shown in the table. The table usually filters selected data by detail, so showing something other than country seems wrong here. Since you can't actually select Country other than by selecting nothing, maybe the table could consider the country to always be selected if detail ==0. This should simplify the logic in the selected are computed props too. Though it would mean making detail == 0 a special case anyway, so maybe that's an argument for option 1. above...

I'm currently adapting the code to match the first approach you mentioned.

src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -91,24 +91,27 @@ export default Vue.extend<{}, {}, Computed, Props>({
...
+            return this.areaFilter ? flattenOptions([this.countryAreaFilterOption]) : {};
+          }
+          return this.areaFilter ? flattenOptions(this.areaFilter.options) : {};
         },
         selectedAreaIds() {
             const selectedAreaIdSet = flattenToIdSet(this.selectedAreaFilterOptions.map(o => o.id), this.flattenedAreas);
             const areaArray = Array.from(selectedAreaIdSet)
             if (this.selections.detail === 0 || !this.selections.detail){
-                return [areaArray[0]]
+                return areaArray[0].length === 3 ? [areaArray[0]] : []
```
EmmaLRussell, line 106: This seems a bit suspect, why `===3`? Because that's the number of regions at level 1? What if it was dealing with a country with 4 top level regions? 

I think this can be simplified further. Really, we just want selectedAreaFilterOptions to always give back the right options, and then selectedAreaIds can just map those to the right Ids (allowing for country level ids having different format). 

How about changing this to:
```
 selectedAreaIds() {
            const areaArray = Array.from(selectedAreaIdSet);
            if (this.selections.detail === 0){
                return [areaArray[0]]
            } else return areaArray.filter(val => parseInt(val[4]) === this.selections.detail);
        },
        selectedAreaFilterOptions() {
            const selectedOptions = this.selections.selectedFilterOptions[this.areaFilterId];
            if(this.selections.detail === 0){
                return this.countryAreaFilterOption ? [this.countryAreaFilterOption] : []
            }
            else if (selectedOptions && selectedOptions.length > 0) {
                return selectedOptions
            } else return this.areaFilter ? this.areaFilter.options : []; //consider all top level areas to be selected if none are
        },
```

Nicolas-Dolan: > This seems a bit suspect, why `===3`? Because that's the number of regions at level 1? What if it was dealing with a country with 4 top level regions?
> 
> I think this can be simplified further. Really, we just want selectedAreaFilterOptions to always give back the right options, and then selectedAreaIds can just map those to the right Ids (allowing for country level ids having different format).
> 
> How about changing this to:
> 
> ```
>  selectedAreaIds() {
>             const selectedAreaIdSet = flattenToIdSet(this.selectedAreaFilterOptions.map(o => o.id), this.flattenedAreas);
>             const areaArray = Array.from(selectedAreaIdSet);
>             if (this.selections.detail === 0){
>                 return [areaArray[0]]
>             } else return areaArray.filter(val => parseInt(val[4]) === this.selections.detail);
>         },
>         selectedAreaFilterOptions() {
>             const selectedOptions = this.selections.selectedFilterOptions[this.areaFilterId];
>             if(this.selections.detail === 0){
>                 return this.countryAreaFilterOption ? [this.countryAreaFilterOption] : []
>             }
>             else if (selectedOptions && selectedOptions.length > 0) {
>                 return selectedOptions
>             } else return this.areaFilter ? this.areaFilter.options : []; //consider all top level areas to be selected if none are
>         },
> ```

The idea of the ```===3``` was to prevent the table simply showing the first value in the area array when detail === 0 and some of the sub areas have been selected. For example, if Northern and Central were selected but detail was set to 0, the table would return the Northern row (the first value in the array). The country level area id string is always 3 characters long (I believe), eg, Malawi is MWI. So for the above example, the code I wrote return no data available are for these selections instead, which is more accurate.

EmmaLRussell: Oh I see - ok, but it's probably even simpler just to use the countryLevelFilterOption you get back from selectedAreFilterOption if detail === 0 - that's what it's there for.

EmmaLRussell: Actually simplest of all would be to combine the two methods since nothing else uses them and since they're both needing to distinguish detail == 0 from other cases it's probably less confusing to combine them. But that can be a refactor for another day.

Nicolas-Dolan: I'll merge this in. This will have the effect of always returning the country level row if detail is set 0, no matter what is selected in the area filter, rather than no data are available for these selections.

PR #283, opened 2020-08-04
mrc-1707 get datasets from ADR

hillalex:
This PR adds an endpoint to retrieve all datasets matching the appropriate ADR schema, and that the currently logged in user can access resources for. This is for use in a typeahead in the front-end allowing the user to select a dataset.

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/clients/FuelClient.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/clients/FuelClient.kt
@@ -6,11 +6,15 @@ import com.github.kittinunf.fuel.httpPost
 import org.imperial.mrc.hint.asResponseEntity
 import org.springframework.http.ResponseEntity
 
-abstract class FuelClient(protected val baseUrl: String) {
+interface HttpClient {
+    fun get(url: String): ResponseEntity<String>
+}
+
+abstract class FuelClient(protected val baseUrl: String): HttpClient {
 
     abstract fun standardHeaders(): Map<String, Any>
 
-    fun get(url: String): ResponseEntity<String> {
+    override fun get(url: String): ResponseEntity<String> {
         return "$baseUrl/$url".httpGet()
                 .header(standardHeaders())
                 .addTimeouts()
--- a/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ADRTests.kt
+++ b/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ADRTests.kt
@@ -56,6 +56,31 @@ class ADRTests : SecureIntegrationTests() {
         }
     }
 
+    @ParameterizedTest
+    @EnumSource(IsAuthorized::class)
+    fun `can get ADR datasets`(isAuthorized: IsAuthorized) {
+        testRestTemplate.postForEntity<String>("/adr/key", getPostEntity())
+        val result = testRestTemplate.getForEntity<String>("/adr/datasets")
+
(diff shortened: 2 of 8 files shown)


src/app/src/main/kotlin/org/imperial/mrc/hint/clients/FuelClient.kt
```diff
@@ -6,11 +6,15 @@ import com.github.kittinunf.fuel.httpPost
 import org.imperial.mrc.hint.asResponseEntity
 import org.springframework.http.ResponseEntity
 
-abstract class FuelClient(protected val baseUrl: String) {
+interface HttpClient {
```
hillalex, line 9: for mocking

hillalex:
Sorry hold off reviewing, realised I missed something

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ADRTests.kt
```diff
@@ -56,6 +56,31 @@ class ADRTests : SecureIntegrationTests() {
...
+    fun `can get ADR datasets`(isAuthorized: IsAuthorized) {
+        testRestTemplate.postForEntity<String>("/adr/key", getPostEntity())
+        val result = testRestTemplate.getForEntity<String>("/adr/datasets")
+
+        assertSecureWithSuccess(isAuthorized, result, null)
+
+        if (isAuthorized == IsAuthorized.TRUE) {
+            val data = ObjectMapper().readTree(result.body!!)["data"]
+            assertThat(data.isArray).isTrue()
+            // we used a fake key so no results will come back
```
EmmaLRussell, line 70: Does this mean that if a user enters their key wrongly they won't see any sort of error, but will just see no results returned? Maybe we should handle that in the front end (or with a displayed error message originating in the backend) - "No datasets found - this can be caused by an incorrect ADR key."  One for another ticket?

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ADRTests.kt
```diff
@@ -56,6 +56,31 @@ class ADRTests : SecureIntegrationTests() {
...
+            assertThat(data.count()).isEqualTo(0)
+        }
+
+        val resultWithResources = testRestTemplate.getForEntity<String>("/adr/datasets?showInaccessible=true")
+        if (isAuthorized == IsAuthorized.TRUE) {
+            val data = ObjectMapper().readTree(resultWithResources.body!!)["data"]
+            assertThat(data.isArray).isTrue()
+            // with the showInaccessible flag set, we can see packages
+            // that aren't actually accessible to our fake user
+            assertThat(data.count()).isEqualTo(3)
```
EmmaLRussell, line 80: So these are all the datasets in the unaids schema? Can we rely on that to stay as 3, or might some new stuff get added causing this test to fail?

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ADRTests.kt
hillalex: In the mockups we've now added a validation step when the key is first added. Not implemented yet but it should prevent saving if the key isn't valid (by hitting some ADR endpoint to check validity.)

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ADRTests.kt
hillalex: On the ADR dev server. But in fact it's now 1 lol, I'll change this to > 0.

PR #284, opened 2020-08-04
mrc-1637_backend Endpoint to create new snapshot in existing version

EmmaLRussell:
This will be called by the front end to spawn new snapshots in existing versions when invalidation points are reached in the UI. 

Requires a parent snapshot id to be provided in query string. Copies the snapshot's state to a new snapshot, and also copies snapshot files. Includes check that parent snapshot and version exist and belongs to user.

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/VersionsController.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/VersionsController.kt
@@ -35,9 +35,12 @@ class VersionsController(private val session: Session,
     fun newSnapshot(@PathVariable("versionId") versionId: Int,
                     @RequestBody state: String): ResponseEntity<String>
     {
+        //TODO: take 'parent' snapshot id as qs parameter, and copy over all snapshot file rows
         val newSnapshotId = session.generateNewSnapshotId()
-        val snapshot = snapshotRepository.saveUserSnapshot(newSnapshotId, versionId, userId())
+        snapshotRepository.saveUserSnapshot(newSnapshotId, versionId, userId())
         snapshotRepository.saveSnapshotState(newSnapshotId, versionId, userId(), state)
+
+        val snapshot = snapshotRepository.getSnapshot(newSnapshotId)
         return SuccessResponse(snapshot).asResponseEntity();
     }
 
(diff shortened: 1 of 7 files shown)


EmmaLRussell:
Snapshot-of-a-version seems even more peculiar..  :camera_flash: 

So what I was imagining:
1. User wants to make invalidating change
2. Store immediately saves current snapshot to ensure we've persisted its latest state
3. Store requests copy of the current snapshot, gets a new snapshot id back and assigns it as the current snapshot, with the current store state. 
4. Make invaliding change and carry on as usual

I started off imagining just a post for making a new snapshot with a state, but then I thought, since we have to make sure we save the final state of the old snapshot anyway, there didn't seem to be any point in sending state twice. 

Could make it a bit less chatty with a single request, to 'saveStateAndMakeCopy'. Maybe that would be better?

Since there will be a bit of network delay (even if we cut it down to one request) you've probably also got:
1.5. Stick up a spinner
3.5 Take down the spinner

> "snapshot" sounds funnier the more I read it laughing
> 
> Code looks fine but could you just spell out the front-end flow for me? I'm a bit confused about why we're copying the current snapshot state rather than just creating a new one with it's own, modified, state.

hillalex:
> Snapshot-of-a-version seems even more peculiar.. 
> 
> So what I was imagining:
> 
> 1. User wants to make invalidating change
> 2. Store immediately saves current snapshot to ensure we've persisted its latest state
> 3. Store requests copy of the current snapshot, gets a new snapshot id back and assigns it as the current snapshot, with the current store state.
> 4. Make invaliding change and carry on as usual
> 
> I started off imagining just a post for making a new snapshot with a state, but then I thought, since we have to make sure we save the final state of the old snapshot anyway, there didn't seem to be any point in sending state twice.
> 
> Could make it a bit less chatty with a single request, to 'saveStateAndMakeCopy'. Maybe that would be better?
> 
> Since there will be a bit of network delay (even if we cut it down to one request) you've probably also got:
> 1.5. Stick up a spinner
> 3.5 Take down the spinner
> 
> > "snapshot" sounds funnier the more I read it laughing
> > Code looks fine but could you just spell out the front-end flow for me? I'm a bit confused about why we're copying the current snapshot state rather than just creating a new one with it's own, modified, state.

Thanks. Ok, I think this makes sense to me!

PR #286, opened 2020-08-05
mrc-717 fix warnings in js tests

hillalex:
An excessive number of warnings in the jest tests have been making it very difficult to read the test output and work out what's going on when a build fails (see e.g. https://travis-ci.com/github/mrc-ide/hint/jobs/368343813#L1021) This PR fixes all the warnings by either fixing the cause, or in the case where Vue is warning about incorrect props that are being deliberately used for test purposes (`surveyAndProgram.test.ts`), temporarily silences warnings.

### DIFF
--- a/src/app/static/src/tests/components/surveyAndProgram/surveyAndProgram.test.ts
+++ b/src/app/static/src/tests/components/surveyAndProgram/surveyAndProgram.test.ts
@@ -5,7 +5,8 @@ import Vuex from 'vuex';
 import SurveyAndProgram from "../../../app/components/surveyAndProgram/SurveyAndProgram.vue";
 import {
     mockAncResponse,
-    mockBaselineState, mockPlottingSelections,
+    mockBaselineState,
+    mockPlottingSelections,
     mockProgramResponse,
     mockSurveyAndProgramState,
     mockSurveyResponse
@@ -17,11 +18,19 @@ import {actions} from "../../../app/store/surveyAndProgram/actions";
 import {mutations} from "../../../app/store/surveyAndProgram/mutations";
 import {getters} from "../../../app/store/surveyAndProgram/getters";
 import {mutations as selectionsMutations} from "../../../app/store/plottingSelections/mutations";
-import {ColourScaleType} from "../../../app/store/plottingSelections/plottingSelections";
+import {ColourScaleSelections, ColourScaleType} from "../../../app/store/plottingSelections/plottingSelections";
 
 const localVue = createLocalVue();
 
 describe("Survey and programme component", () => {
+
+    beforeAll(() => {
+        Vue.config.silent = true;
+    });
+
+    afterAll(() => {
+        Vue.config.silent = false;
+    })
 
     testUploadComponent("surveys", 0);
     testUploadComponent("program", 1);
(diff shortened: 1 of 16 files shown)


src/app/static/src/tests/components/surveyAndProgram/surveyAndProgram.test.ts
```diff
@@ -115,6 +138,11 @@ describe("Survey and programme component", () => {
         expect(choro.props().featureLevels).toBe("TEST LEVEL LABELS");
         expect(choro.props().indicators).toStrictEqual(["TEST INDICATORS"]);
         expect(choro.props().selections).toStrictEqual({selectedFilterOptions: "TEST SELECTIONS"});
+        expect(choro.props().colourScales).toEqual({prevalence: {
```
hillalex, line 141: In this case the warning was that the getter "selectedSAPColourScales" did not exist, which led me to notice that this prop was not being tested.

hillalex:
> This all looks good. I am still seeing a few warnings e.g.
> 
> ```
> PASS  src/tests/modelRun/actions.test.ts (11.189s)
>   ● Console
> 
>     console.info node_modules/vue/dist/vue.runtime.common.dev.js:8397
>       Download the Vue Devtools extension for a better development experience:
>       https://github.com/vuejs/vue-devtools
>     console.warn src/app/apiService.ts:6032
>       No success handler registered for request /model/cancel/123.
> 
>  PASS  src/tests/components/header/loggedOutHeader.test.ts
>  PASS  src/tests/modelOutput/modelOutput.test.ts
>  PASS  src/tests/metadata/metadata.test.ts
>  PASS  src/tests/modelRun/getters.test.ts
>  PASS  src/tests/versions/mutations.test.ts
>   ● Console
> 
>     console.error node_modules/jest-environment-jsdom/node_modules/jest-mock/build/index.js:847
>       test error
> 
>  PASS  src/tests/plottingSelections/getters.test.ts
>  PASS  src/tests/root/mutations.test.ts
>  PASS  src/tests/components/header/dropDown.test.ts
>  PASS  src/tests/components/plots/utils.test.ts
>   ● Console
> 
>     console.warn src/app/components/plots/utils.ts:2999
>       Unknown color function: not-a-color-function
> ```
> 
> Is that expected?
> 
> Do you think we should set up jest to fail on warnings to stop this issue creeping back in? or just to prevent it from happening by stopping any console logging? Some ideas here - overriding console might cause other problems though.
> https://stackoverflow.com/questions/28615293/is-there-a-jest-config-that-will-fail-tests-on-console-warn

Must have missed those console.infos, will mock out console.info everywhere cos they just clutter  the output. Since in many test classes we expect console.logs as part of the normal running of the app, they've also been mocked out so they don't clutter the output: https://github.com/mrc-ide/hint/pull/286/files#diff-0f5aaf01a1b0a24681c565f5fd5a8e51R35 I sort of left in the console.warns in case we want to be aware of them, like not having a success handler registered (although haven't you now added an `ignoreSuccess` option? that might be a better solution there.)

hillalex:
We can't make tests fail on console.warn because in some cases, a console.warn is the expected outcome of a test! E.g. here: https://github.com/mrc-ide/hint/blob/master/src/app/static/src/tests/components/plots/utils.test.ts#L16

We could mock `console.warn` in that case and explicitly test that it's called though, which would be a better test really.

hillalex:
Also, in general I think failing on `console.warn` is too strong but maybe making tests fail on `console.error` is a good shout. I'll try add that.

PR #287, opened 2020-08-05
mrc-1726 select ADR dataset front-end

hillalex:
This PR adds the component for selecting a dataset, but does not yet do anything with that selection. Partly to keep PR small, partly to run by you a couple of architecture choices - see comments.

This also adds a parent component that wraps both this new component and the API key widget.

### DIFF
--- a/src/app/static/src/app/components/Stepper.vue
+++ b/src/app/static/src/app/components/Stepper.vue
@@ -37,7 +37,7 @@
         </div>
         <div v-if="!loading" class="content">
             <div class="pt-4">
-                <adr-key v-if="isActive(1)"></adr-key>
+                <adr-integration v-if="isActive(1)"></adr-integration>
                 <baseline v-if="isActive(1)"></baseline>
                 <survey-and-program v-if="isActive(2)"></survey-and-program>
                 <model-options v-if="isActive(3)"></model-options>
@@ -53,7 +53,7 @@
 
     import Vue from "vue";
     import {mapActions} from "vuex";
-    import AdrKey from "./adr/ADRKey.vue";
+    import AdrIntegration from "./adr/ADRIntegration.vue";
     import Step from "./Step.vue";
     import Baseline from "./baseline/Baseline.vue";
     import SurveyAndProgram from "./surveyAndProgram/SurveyAndProgram.vue";
@@ -128,7 +128,7 @@
             }
         },
         components: {
-            AdrKey,
+            AdrIntegration,
             Step,
             Baseline,
             SurveyAndProgram,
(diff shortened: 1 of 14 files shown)


src/app/static/src/app/components/Stepper.vue
```diff
@@ -37,7 +37,7 @@
         </div>
         <div v-if="!loading" class="content">
             <div class="pt-4">
-                <adr-key v-if="isActive(1) || isActive(2)"></adr-key>
+                <adr-integration v-if="isActive(1)"></adr-integration>
```
hillalex, line 40: I think it only makes sense to show this on step 1, since it's supposed to be sort of step 0 (and editing it on step 2 would potentially invalidate the step you were on.)

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -0,0 +1,88 @@
...
+                            <span class="font-weight-bold">${d.organization.title}</span>
+                        </div>`
+                }))
+            }
+        },
+        methods: {
+            getDatasets: mapActionByName(null, 'getADRDatasets'),
+            importDataset() {
+                // set loading spinner
+                // set selected dataset in store
```
hillalex, line 71: I'm imagining `selectedDataset` makes sense on `BaselineState`, as it is really a part of this first step. I considered introducing a new module to keep ADR stuff together, but it adds some complication with respect to saving/reloading/resetting state, whereas that is more easily handled if this is part of the baseline state.

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -0,0 +1,88 @@
...
+                // import each file
```
hillalex, line 72: My idea is to simply fire off 6 calls to import each of the 6 files from the ADR at this point. So basically 6 new actions, 3 baseline actions, 3 surveyAndProgram actions. No new mutations needed as the new "import from ADR" endpoints will return data in exactly the same format as the current upload ones do.

This will be async by default, but we'll wait until all 6 promises resolve before closing this modal.

src/app/static/src/app/components/adr/SelectDataset.vue
hillalex: If you agree, perhaps `adrKey` and `adrDatasets` should also live in `BaselineState` (they're currently in `RootState`)

src/app/static/src/app/components/adr/SelectDataset.vue
hillalex: Good q about multiple files of the same type - will ask ADR

PR #289, opened 2020-08-05
mrc-1725 preliminary refactor

hillalex:
* removes the `HttpClient` interface and replaces with an `ADRClient` interface, as this will end up having other methods on it
* while testing I noticed that we were not handling arbitrary exceptions - unhandled exception types were not returning in our API format but in the SprintBoot default format, so I've added this in to the `ExceptionHandler`

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
@@ -14,6 +14,7 @@ import org.springframework.web.bind.annotation.ControllerAdvice
 import org.springframework.web.bind.annotation.ExceptionHandler
 import org.springframework.web.context.request.WebRequest
 import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler
+import java.lang.reflect.UndeclaredThrowableException
 import java.text.MessageFormat
 import java.util.*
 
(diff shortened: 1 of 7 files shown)


src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
```diff
@@ -34,10 +35,18 @@ class HintExceptionHandler(private val errorCodeGenerator: ErrorCodeGenerator,
         return unexpectedError(status, request, e.message)
     }
 
-    @ExceptionHandler(PSQLException::class)
-    fun handlePSQLException(e: PSQLException, request: WebRequest): ResponseEntity<Any> {
-        logger.error(e.message)
-        // for security reasons we should not return arbitrary db errors to the frontend
+    @ExceptionHandler(Exception::class)
+    fun handleArbitraryException(e: Exception, request: WebRequest): ResponseEntity<Any> {
+        val error = if (e is UndeclaredThrowableException) {
```
hillalex, line 40: When you handle *all* exceptions, you start catching these - they happen when a method isn't marked with the `@Throws` annotation, and wrap the actual error that was thrown.

hillalex:
PR is failing due to broken master build, should be fixed once that is!

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/ExceptionHandler.kt
EmmaLRussell: Could this be unit tested?

PR #290, opened 2020-08-05
mrc-1725 endpoint to GET ADR schema names

hillalex:
When the user selects a dataset, one of the dataset fields is an array resources that look something like:
`[{"name": "Spectrum File", "url": "https://dev.adr.fjelltopp.org/dataset/4ffa1536-5be5-4444-8f32-b1a183b0de1f/resource/26efad44-a2ed-475d-bd30-d76e394040aa/download/spectrum.pjnz", "resource_type": "inputs-unaids-spectrum-file", ...}... ]`

The `resource_type` field of each resource identifies which of the 6 input files it is. Once a dataset is selected, the front-end can just pick out each file and make a POST request to `/import/pjnz` or whatever, but it needs to know what the 6 schema names are so it can select the right resource. Rather than hard-code these in the front-end I thought it was best to make them configurable. So this PR adds an endpoint to return a dictionary of configured schema names that the front-end can use to work out which resources to import.

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/AppProperties.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/AppProperties.kt
@@ -7,13 +7,13 @@ import java.net.URL
 import java.util.*
 
 interface AppProperties {
-    val adrANC: String
-    val adrART: String
-    val adrPJNZ: String
-    val adrPop: String
-    val adrShape: String
-    val adrSurvey: String
-    val adrSchema: String
+    val adrANCSchema: String
+    val adrARTSchema: String
+    val adrPJNZSchema: String
+    val adrPopSchema: String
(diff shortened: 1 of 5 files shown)


src/app/src/main/kotlin/org/imperial/mrc/hint/AppProperties.kt
```diff
@@ -7,6 +7,12 @@ import java.net.URL
 import java.util.*
 
 interface AppProperties {
+    val adrANC: String
+    val adrART: String
+    val adrPJNZ: String
+    val adrPop: String
+    val adrShape: String
+    val adrSurvey: String
```
EmmaLRussell, line 15: Since all of these are adrSchema, could we maybe rename adrSchema to make it clear what that is? (It's the dataset schema??)

PR #292, opened 2020-08-05
mrc-1724 add border separator between Versions | File

hillalex:
The top menu should have a visual separator between the "Versions" and "File" options.


hillalex:
> Looks good but need to hide the separator when not logged in.

doh, nice spot.

PR #295, opened 2020-08-06
mrc-1637_frontend Create new snapshots when invalidating changes made by logged in users

EmmaLRussell:
Invalidating changes cause new snapshots to be created if user is logged in. We're assuming that we still want to warn users that subsequent  steps will be deleted, but that these can be reloaded in future. Tricky to get the wording right for this, see what you think, will add all translations in a subsequent ticket. 

To create a new snapshot, first the latest state is immediately saved for the current snapshot, then we ask the backend to make a new snapshot from the old one, then finally set the current snapshot as the new one returned from the backend, We don't need to set any state or files at that point because our starting point for the new snapshot is exactly the same as the finish point for the old one.

The component waits until the new snapshot is set before invoking the continueEditing event, when changes actually start being made.

As usual, I wasn't entirely sure what to do in the error case. I've gone with logging a global error which will show up in the global errors alert and cancelling the edit, but maybe it would be better to continue editing.

Sorry this is quite a large changeset, it's mostly tests!


src/app/static/src/app/components/ResetConfirmation.vue
```diff
@@ -1,36 +1,58 @@
...
     import Vue from "vue";
     import Modal from "./Modal.vue";
-    import {mapGetterByName} from "../utils";
+    import {mapActionByName, mapGetterByName, mapStateProp} from "../utils";
     import {StepDescription} from "../store/stepper/stepper";
+    import LoadingSpinner from "./LoadingSpinner.vue";
+    import {VersionsState} from "../store/versions/versions";
+    import {ErrorsState} from "../store/errors/errors";
+
+    declare const currentUser: string;
```
hillalex, line 49: I feel like this hack is in enough places now that we should sort out getting it into the store state - have written a ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1743

src/app/static/src/app/components/ResetConfirmation.vue
```diff
@@ -1,36 +1,58 @@
...
+            <p v-translate="'discardWarning'"></p>
+            <ul>
+                <li v-for="step in laterCompleteSteps">
+                    Step {{step.number}}: <span v-translate="step.textKey"></span>
+                </li>
+            </ul>
+
+            <p v-if="guestUser"  v-translate="'savePrompt'"></p>
+            <p v-if="!guestUser">
+                These steps will automatically be saved in a snapshot. You will be able to reload this snapshot from the Versions page.
```
hillalex, line 16: The logic all seems fine to me here, re the wording I think we discussed ditching "snapshot" as a user facing term?

hillalex (commented):
Looks good, just some conflicts to fix and maybe wording to update following your chat with UNAIDs about terminology?

src/app/static/src/app/components/ResetConfirmation.vue
EmmaLRussell: Good call

src/app/static/src/app/components/ResetConfirmation.vue
EmmaLRussell: Pending with Jeff atm. Ticket for changing user-facing terms as decided: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1745

PR #297, opened 2020-08-10
Mrc 1696 adds a region hierarchy column to the table

Nicolas-Dolan:
Added a function to iterateDataValues which uses the countryAreaFilterOption object to find the regional hierarchy of each row in the table. It finds the current rows areaId within the countryAreaFilterOption object and iterates through the path to that id giving the hierarchy.

Perhaps this would be better as a tooltip, when someone mouses over the value in the area column, rather than its own dedicated column?


src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -6,11 +6,13 @@
                 <div>
                     <tr>
                         <th v-translate="'area'"></th>
+                        <th>Area Hierarchy</th>
```
Nicolas-Dolan, line 9: I'll need to add a v-translate for this.

src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -138,13 +152,15 @@ export default Vue.extend<{}, {}, Computed, Props>({
                     const key = [current.areaId, ...this.nonAreaFilters.map(f => current.filterValues[f.id])].join("_");
                     if (!(key in displayRows)) {
                         const areaLabel =  this.flattenedAreas[current.areaId].label;
+                        const areaHierarchy = findPath(current.areaId, this.countryAreaFilterOption).replace(/undefined\//g, "")
```
hillalex, line 155: These `undefined`s are coming in because on this line: https://github.com/mrc-ide/hint/pull/297/files#diff-607cf6b7d106639467b8b36fabc00bfcR131 `obj.label` is not defined where `obj` is an array. I think replacing 

```
if (path) return obj.label + "/" + path;
```

with 

```
if (path) {
      return (obj.label ? obj.label + "/": "") + path;   
}
```

should give you want you need and then you can remove this `replace`. Side note - as a matter of convention we always use brackets for if/else statements, even if they are 1 liners. Keeps things readable and less error prone.

src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -119,6 +122,17 @@ export default Vue.extend<{}, {}, Computed, Props>({
             return this.nonAreaFilters.filter(f => f.options.length > 0)
         },
         filteredData() {
+            function findPath(id: string, obj: any): any {
```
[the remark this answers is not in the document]
EmmaLRussell: Agreed, it's a little opaque as is.

src/app/static/src/tests/components/plots/utils.test.ts
```diff
@@ -428,4 +428,42 @@ describe("plot utils", () => {
 
         expect(result).toStrictEqual([12, 0.5, 14, 0.6, 14]);
     });
+
+    it("findPath produces expected result", () => {
```
hillalex, line 432: It's a good idea with functions like this to unit test multiple cases, including edge/unusual cases if there are any - here I'd add a test where the first argument, the id, comes from each level. So test it brings back the right thing for `MWI` and `MWI11` say.

src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -152,7 +143,9 @@ export default Vue.extend<{}, {}, Computed, Props>({
                     const key = [current.areaId, ...this.nonAreaFilters.map(f => current.filterValues[f.id])].join("_");
                     if (!(key in displayRows)) {
                         const areaLabel =  this.flattenedAreas[current.areaId].label;
-                        const areaHierarchy = findPath(current.areaId, this.countryAreaFilterOption).replace(/undefined\//g, "")
+                        // The first replace removes the first layer of the path (ie, the country level)
```
hillalex, line 146: Would it make sense to start the recursion from `this.countryAreaFilterOption.children` in that case, so the country doesn't get included in the path in the first place? The country itself would have an undefined path, but that could easily be mapped to the empty string. 

Also seems like you could modify `findPath` so that the current area selected isn't included in the path, by returning an empty string when you find the matching area id: `if(id === obj[key]) return ""` (you'd just have to also change the `if (path)` check to `(if (path != undefined)` because the empty string is falsy.)

src/app/static/src/app/generated.d.ts
```diff
@@ -171,6 +171,7 @@ export interface Filter {
 export interface FilterOption {
   label: string;
   id: string;
+  children?: any[];
```
Nicolas-Dolan, line 174: I couldn't find a way to make the optional children prop more specific. I tried FilterOption[] and Array<FilterOption> but TS complained for both.

EmmaLRussell: Sorry, I'm not sure why you were trying to do this? These are the generated types (auto-generated from the hintr interface) so you shouldn't be changing them manually anyway. Children is on the NestedFilterOption type.

Nicolas-Dolan: I changed the countryAreaFilterOption type to NestedFilterOption and this has fixed the problem.

EmmaLRussell (approved):

PR #301, opened 2020-08-24
mrc-1731 set selected dataset

hillalex:
This PR just sets the selected ADR dataset in the store and reflects it back out to the user. Still to come - importing the individual files from ADR.

### DIFF
--- a/src/app/static/src/app/components/adr/SelectDataset.vue
+++ b/src/app/static/src/app/components/adr/SelectDataset.vue
@@ -50,7 +50,7 @@
     import {Dataset} from "../../types";
 
     interface Methods {
-        setDataset: (dataset: any) => void
+        setDataset: (dataset: Dataset) => void
         importDataset: () => void
         toggleModal: () => void
     }
@@ -117,9 +117,10 @@
             importDataset() {
                 this.loading = true;
                 this.setDataset(this.newDataset);
-                // import each file
-                // await all
+                // TODO import each file
+                // TODO await all
                 setTimeout(() => {
+                    // mock importing of files with a timeout
                     this.loading = false;
                     this.open = false;
                 }, 200)
--- a/src/app/static/src/app/components/adr/SelectDataset.vue
+++ b/src/app/static/src/app/components/adr/SelectDataset.vue
@@ -100,7 +100,7 @@
                     id: fullMetaData.id,
                     title: fullMetaData.title,
                     revision_id: fullMetaData.revision_id,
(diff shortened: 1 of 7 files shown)


EmmaLRussell (commented):
Works nicely, look good. The link to the test dataset on ADR gets a 404, is that expected if not actually logged in to ADR?
Wonder if that link should open a new tab? 
When navigated back from the 404 page, saw that the dataset had not been retained in the step. Is that because it's in the baseline module which isn't saved to local storage?

hillalex:
> Works nicely, look good. The link to the test dataset on ADR gets a 404, is that expected if not actually logged in to ADR?
> Wonder if that link should open a new tab?
> When navigated back from the 404 page, saw that the dataset had not been retained in the step. Is that because it's in the baseline module which isn't saved to local storage?

It's possible I've got the url wrong, will double check! New tab - good idea, will add that. And yeah, good point, need to persist the bits of state that relate to the datasets, created ticket here: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1746

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -61,15 +93,37 @@
...
                         </div>`
                 }))
+            },
+            newDataset() {
+                const fullMetaData = this.datasets.find(d => d.id = this.newDatasetId)
+                return fullMetaData && {
+                    id: fullMetaData.id,
+                    title: fullMetaData.title,
+                    revision_id: fullMetaData.revision_id,
+                    url: `https://adr.unaids.org/${fullMetaData.type}/${fullMetaData.name}`
```
hillalex, line 103: Ok, so this is the reason the url isn't working. I've hard-coded "https://adr.unaids.org" as the base url even though when running in development the back-end is talking to "https://dev.adr.fjelltopp.org/". I did think of adding a HINT endpoint GET `adr/url` which the front-end could call to get back the configured ADR location but it seemed like a faff and I thought hard-coding the url here would suffice for now.

hillalex: (nb, if you added your production ADR key, the reason it still worked is because the API keys on their dev server are exactly the same as on their prod server!)

hillalex: Actually just realised we already have an endpoint returning the individual resource schema names, should just add the base url into that endpoint. Will add that in on a separate branch.

EmmaLRussell (approved):

PR #302, opened 2020-08-25
mrc-1746 persist selectedDataset

hillalex:
~~Contains commits from https://github.com/mrc-ide/hint/pull/301~~

### DIFF
--- a/src/app/static/src/app/localStorageManager.ts
+++ b/src/app/static/src/app/localStorageManager.ts
@@ -4,6 +4,7 @@ const appStateKey = "appState";
 
 export const serialiseState = (rootState: RootState): Partial<RootState> => {
     return {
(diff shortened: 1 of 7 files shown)


EmmaLRussell (commented):
The code looks fine, but I'm struggling to test it manually - the test key doesn't seem to be showing any datasets any more. Are you seeing the same thing? Happy for this to be merged anyway.

hillalex:
> The code looks fine, but I'm struggling to test it manually - the test key doesn't seem to be showing any datasets any more. Are you seeing the same thing? Happy for this to be merged anyway.

The ADR dev server keeps changing the status of the test user from admin to member and they have this bug where only admins can use the API in the way we want. I've manually changed it back but will raise with them.

PR #303, opened 2020-08-25
mrc-1747 fetch ADR schemas from the back-end

hillalex:
This PR:
* adds `baseUrl` into the `GET /adr/schemas` endpoint
* adds a Vuex action and mutation for calling and storing the response from ^
* uses the retrieved `baseUrl` in the `SelectDataset` component


hillalex:
Hold off on this a min, just realised the /schemas endpoint being locked down to logged in users is a problem here.

src/app/static/src/app/root.ts
```diff
@@ -37,6 +38,7 @@ export interface RootState extends TranslatableState {
     adrDatasets: any[],
     adrKey: string | null,
     adrKeyError: Error | null,
+    adrSchemas: ADRSchemas | null,
```
EmmaLRussell, line 41: This is starting to feel like it might belong in an adr module rather than just on root state, what do you think? Will there be any more bits still to come on adr?

hillalex: Pretty sure this is the last one. Have considered same, but would it make the code simpler or more complicated?

EmmaLRussell (changes requested):
I'm actually still getting a 401 from the schemas endpoint when not logged in. I tried changing `excludePathPatterns("/adr/schemas*")` to `excludePathPatterns("/adr/schemas/")` which seemed to fix it so I guess it's just a wildcarding issue.

hillalex:
> I'm actually still getting a 401 from the schemas endpoint when not logged in. I tried changing `excludePathPatterns("/adr/schemas*")` to `excludePathPatterns("/adr/schemas/")` which seemed to fix it so I guess it's just a wildcarding issue.

That's a bit concerning because it means something very strange is going on with the integration test: https://github.com/mrc-ide/hint/pull/303/files#diff-771ff8aa668a23ae12ae8f75503e8e50R140

src/app/static/src/app/root.ts
EmmaLRussell: Yeah, it's a bit borderline. Fine as it is I think, but probably worth changing if the adr properties do expand any more..

hillalex:
Ah ok, I know exactly what's going on. The integration test is testing `/adr/schemas` and the front-end is calling `/adr/schemas/`. I thought it was a regex so would match both, but it must be a glob. Will update test and add both with and without trailing slash explicitly.

EmmaLRussell (approved):

PR #305, opened 2020-08-25
mrc-1636_frontend User can load snapshot from version history

EmmaLRussell:
This branch adds Load links to the snapshot rows in the Version History component. Clicking these invokes loading snapshot details from the backend, then loads these into the current state in the same way that file data is loaded.


src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/VersionsController.kt
```diff
@@ -53,10 +53,11 @@ class VersionsController(private val session: Session,
 
     @GetMapping("version/{versionId}/snapshot/{snapshotId}")
     @ResponseBody
-    fun getSnapshotDetails(@PathVariable("versionId") versionId: Int,
-                    @PathVariable("snapshotId") snapshotId: String): ResponseEntity<String>
+    fun loadSnapshotDetails(@PathVariable("versionId") versionId: Int,
+                            @PathVariable("snapshotId") snapshotId: String): ResponseEntity<String>
     {
         val snapshotDetails = snapshotRepository.getSnapshotDetails(snapshotId, versionId, userId())
+        session.setSnapshotId(snapshotId)
```
EmmaLRussell, line 60: Changed this from 'get' to 'load' as it needs to set snapshot id for the session, don't just imply that this is a fetch of data with no side effects.

src/app/static/src/app/store/load/actions.ts
```diff
@@ -42,15 +44,17 @@ export const actions: ActionTree<LoadState, RootState> & LoadActions = {
...
-                }
-            });
+        await getFilesAndLoad(context, files, savedState)
+    },
+
+    async loadFromSnapshot(context, snapshotContents) {
+        const {commit} = context;
+        commit({type: "SettingFiles", payload: null});
+
+        router.push("/", () => {
```
EmmaLRussell, line 54: We currently need to be on the root route by then time we reload the page or get a 404 from the backend. As discussed, it would be good at some point to make the backend recognise the /versions route so user can reload page from the versions screen without error.

hillalex (commented):
Code all looks fine (do you want to make naming changes here? or merge and do in a later branch?)

But there's something quite strange about the user experience.

Suppose I go into my version history and I have 2 snapshots:

Screenshot from 2020-08-27 11-56-18

I select the oldest one, the one dated 27/08/2020 11:51:52

It loads, then I go back to the versions page and this is what I now see:

Screenshot from 2020-08-27 11-57-10

The snapshot I just loaded has seemingly disappeared (in fact it's been updated and is now the most recent), and the oldest snapshot is now one that was first created later than the one I just loaded.

I think we need to go back to the drawing board a bit, because this is a very confusing user experience. As a user I naturally expect the timestamp to uniquely identify a snapshot which makes it very confusing when these change. The second way I try to identify a snapshot is by the order it appears in the list (e.g. I think "I want to go back to the oldest state of the app"), but this also changes! It's basically impossible to know which snapshot is which because they have no persistent id.

EmmaLRussell:
Created a ticket to improve user experience with persistent snapshot identifiers: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1757

PR #308, opened 2020-08-26
Mrc 1699 added sort and filter features to the table

Nicolas-Dolan:
Replaced the table with component from Bootstrap https://bootstrap-vue.org/docs/components/table


src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -2,23 +2,34 @@
...
+            >
+              <b-input-group size="sm">
+                <b-form-input
+                  v-model="filter"
+                  type="search"
+                  id="filterInput"
+                  placeholder="Type to Search"
+                ></b-form-input>
+                <b-input-group-append>
+                  <b-button :disabled="!filter" @click="filter = ''">Clear</b-button>
```
hillalex, line 16: Could do with a test for this "clear" functionality

hillalex (commented):
Something a bit off with the whitespace in the component and the test suite. Could you try auto-formating those files? Looks like in VS the shortcuts are <kbd>CMD ⌘</kbd> + <kbd>K</kbd>, <kbd>CMD ⌘</kbd> + <kbd>D</kbd>: https://stackoverflow.com/questions/5755942/how-do-you-auto-format-code-in-visual-studio

Side note - as we're using multiple IDEs now we should make sure we have the same style rules defined. Unfortunately there's no easy way to share style settings between IDEs. In general for front-end code we want tab size 4, indent 4, continuation indent 4, and children of the `<template>` and `<script>` tags in vue templates to be indented. Do you have any interest in switching to IntelliJ? We can get you a subscription.

src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -154,14 +176,47 @@ export default Vue.extend<{}, {}, Computed, Props>({
                         displayRows[key] = {
                             areaLabel,
                             areaHierarchy,
-                            filterLabels,
-                            indicatorValues: {}
+                            ...filterLabels,
                         }
                     }
-                    displayRows[key].indicatorValues[current.indicatorMeta.indicator] = current.value;
+                    displayRows[key][current.indicatorMeta.indicator] = current.value;
```
EmmaLRussell, line 182: We talked about prefixing the filter and indicator keys to avoid possible duplication (unlikely but possible). Did you decide against that?
Also maybe an issue when merge with your other ticket adding the upper and lower error columns..

src/app/static/src/app/components/plots/table/Table.vue
EmmaLRussell, line 16: I'm not sure you really need this Clear button, as you seem to get an 'x' which appears in the field when there's any filter text, and which has the same effect.
If we keep it, it'll need a translation too!

EmmaLRussell (changes requested):
Looks good, works really nicely.

I think we'd expect to see the 'No data available' message if there are no rows returned from the new text filter as well as the main filtered data method, but this isn't the case, you just see an empty table. Would that be easy to put in?

src/app/static/src/app/components/plots/table/Table.vue
Nicolas-Dolan: Because of the way the fields are generated for the table there isn't much of an issue of the indicators or filters being mislabelled -- unless you think there is a chance an indicator and a filter could have the same name? With regard to the error columns,  I took the liberty of adding them here. I prefixed them with the name of the indicator they belong to.

src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -2,23 +2,34 @@
...
-            </table>
+            <b-form-group
+              class="mb-0"
+            >
+              <b-input-group size="sm">
+                <b-form-input
+                  v-model="filter"
+                  type="search"
+                  id="filterInput"
+                  placeholder="Type to Search"
```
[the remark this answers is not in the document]
Nicolas-Dolan: These have been added now.

src/app/static/src/app/components/plots/table/Table.vue
Nicolas-Dolan: I think having a clear button is nice to have as it signals to the user what the filter tool is/does.

Nicolas-Dolan: Added

src/app/static/src/app/components/plots/table/Table.vue
Nicolas-Dolan: PS. the uncertainty range format I've chosen has been approved by Jeff

Nicolas-Dolan:
So I decided to just add in the uncertainty ranges along with this ticket as I was finding it too difficult trying to predict how they would play with everything else in this ticket and it was relatively simple just to add them in. Jeff has approved the format I've used for them. Weirdly despite not making any changes to the bubbleplot and choropleth indicators in the modelOutput store the ranges still came through (maybe this has something to do with changes Emma made?). I have left them as is for now but let me know if you still think I should change
```return rootGetters['metadata/outputIndicatorsMetadata'];```
to
```return rootState.modelRun.result!!.plottingMetadata.choropleth.indicators;```
for each

src/app/static/src/app/components/plots/table/Table.vue
hillalex: Cool, I've formatted it for you here but we should get you set up with IntelliJ or sort out code formatting rules for VS when you get the new laptop set up. We should also set up code style checking as part of CI rather than relying on PRs to pick up formatting issues, been meaning to do that for a while!

hillalex (approved):

Nicolas-Dolan:
Regarding the translations, it would be good if those could all be checked properly at some point. I noticed that Alex had created a ticket adding translations so perhaps these could be checked there too. Rich told me that everything will be checked by a professional at some point also.

EmmaLRussell (approved):

PR #312, opened 2020-08-28
mrc-1754 Rename snapshot to version

EmmaLRussell:
Rename snapshot to version in backend and frontend


richfitz (approved):
Nice - can't spot any missed ones

PR #313, opened 2020-08-28
mrc-1759 Add 'Load last updated' links to Projects view

EmmaLRussell:
Load the last updaed version for a project - quick links without needing to expand the project.


hillalex (approved):
"Load last updated" sounds a little funny to me but can't think of a better terminology, anyway this can easily be updated down the line if it comes up during user testing.

PR #314, opened 2020-09-01
mrc-1727 add link to ADR and tooltip text

hillalex:
Updates the ADR key widget to match the mockups. I must have accidentally deleted the first slide from the mockups, can be seen in this version though: https://docs.google.com/presentation/d/1IrToJoGxVd5pespysncY-2LacQgqDy6LH8nazzSf_ns/edit?ts=5f2a7a98#slide=id.g8ea8bb15f6_7_240


src/app/static/src/app/components/adr/ADRKey.vue
```diff
@@ -1,55 +1,63 @@
...
-                           v-translate="'cancel'"></a>
+                    <div class="input-group"
+                         style="margin-top: -8px; min-width: 390px"
+                         v-if="editing">
+                        <input id="key"
+                               ref="keyInput"
+                               class="form-control"
+                               v-model="editableKey"
+                               type="text"
+                               v-translate:placeholder="'enterKey'"/>
```
hillalex, line 40: Noticed this placeholder was hard-coded, so added it to the translations list

EmmaLRussell (commented):
This all looks good to me. The ADR url to get the key is just the root url for the ADR domain, is that right? It shouldn't be /profile or something like that?

hillalex:
> This all looks good to me. The ADR url to get the key is just the root url for the ADR domain, is that right? It shouldn't be /profile or something like that?

They're building a shortcut apparently that will take users to their key (at the moment there's no link like /profile, you have to know the username) so when they've implemented that we'll change this over to use it.

src/app/static/src/app/components/adr/ADRKey.vue
```diff
@@ -105,8 +118,11 @@
                     }
                     return str;
                 } else {
-                    return i18next.t("noneProvided", this.currentLanguage)
+                    return i18next.t("noneProvided", {lng: this.currentLanguage})
```
Nicolas-Dolan, line 121: good spot!

Nicolas-Dolan (commented):
The code is clear although I'm having trouble seeing the widget when I run it locally

hillalex:
> The code is clear although I'm having trouble seeing the widget when I run it locally

You have to be logged in to see it, that might be the issue?

Nicolas-Dolan:
> > The code is clear although I'm having trouble seeing the widget when I run it locally
> 
> You have to be logged in to see it, that might be the issue?

That'll be why. Is there a dummy login I can use for testing purposes?

hillalex:
> > > The code is clear although I'm having trouble seeing the widget when I run it locally
> > 
> > 
> > You have to be logged in to see it, that might be the issue?
> 
> That'll be why. Is there a dummy login I can use for testing purposes?

Yep, you can use `test.user@example.com` with password `password`, after running `./scripts/add-test-user.sh`

Nicolas-Dolan:
> > > > The code is clear although I'm having trouble seeing the widget when I run it locally
> > > 
> > > 
> > > You have to be logged in to see it, that might be the issue?
> > 
> > 
> > That'll be why. Is there a dummy login I can use for testing purposes?
> 
> Yep, you can use `test.user@example.com` with password `password`, after running `./scripts/add-test-user.sh`

Thanks, this worked except no validation occurred when I added an ADR access key

hillalex:
> > > > > The code is clear although I'm having trouble seeing the widget when I run it locally
> > > > 
> > > > 
> > > > You have to be logged in to see it, that might be the issue?
> > > 
> > > 
> > > That'll be why. Is there a dummy login I can use for testing purposes?
> > 
> > 
> > Yep, you can use `test.user@example.com` with password `password`, after running `./scripts/add-test-user.sh`
> 
> Thanks, this worked except no validation occurred when I added an ADR access key

Validation will be a separate ticket :+1:

EmmaLRussell (approved):

PR #321, opened 2020-09-02
mrc-1762 refresh files from the ADR 

hillalex:
Realised that I forgot to actually wire up the refreshing logic! This PR does so, the easiest way to see it in action is to import a dataset, then go to the ADR dataset and delete the PJNZ file, re-upload it, go back to HINT and refresh the page. Then you'll see the refresh button and can refresh the data.

NB the mechanism for detecting a change in the ADR isn't very good at the moment - it just makes a call to the API on component mount, meaning changes aren't detected unless you refresh the page, or navigate away and back. I want to change this to a poll for changes in a further branch.

Note you can login to the ADR with `username: testing123` and `password: password`


src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -158,15 +168,18 @@
                 return outOfDateResources
             },
             outOfDateMessage() {
+                const updatedNames = Object.keys(this.outOfDateResources)
+                    .map(r => names[r as keyof DatasetResourceSet]).join(", ")
                 if (Object.keys(this.outOfDateResources).length == 0) {
                     return ""
                 }
-                return "This dataset has been updated in the ADR. Use the refresh button to import the latest files."
+                return `The following files have been updated in the ADR: ${updatedNames}. Use the refresh button to import the latest files.`
```
hillalex, line 176: This more informative msg was requested by Jeff. One weird thing though is that the ADR gives false positives - sometimes changing 1 file updates the metadata for other files in the set, so it will report that they've all changed when they haven't. If this causes confusion we'll have to revert to the simpler message.

Nicolas-Dolan (commented):
The first time I attempted to download the dataset from ADR I ran into an error in which it failed to download the PJNZ file. "An unexpected error occurred..." I tried a few more times and would get errors seemingly at random with other files as well. I'll need to get more familiar with the ADR side of things to test this properly.

hillalex:
> The first time I attempted to download the dataset from ADR I ran into an error in which it failed to download the PJNZ file. "An unexpected error occurred..." I tried a few more times and would get errors seemingly at random with other files as well. I'll need to get more familiar with the ADR side of things to test this properly.

That's sad :cry: When you get errors like that, can you look at the HINT logs to see what the cause of the error was?

hillalex:
Emma was seeing intermittent failures with tests too. We think it might be their dev server restoring data, which happens every morning. Try again now!

src/app/static/src/app/components/adr/SelectDataset.vue
EmmaLRussell: Are ADR aware of this issue?

EmmaLRussell, line 176: This ought to come from translations file I guess. Not sure how we usually deal with string templates in the translations, have we avoided that issue up til now?

EmmaLRussell (changes requested):
Looks good, works for me. As you say, when I updated the spectrum file only, the tooltip indicates that all files have been updated. Just needs a translation for the message.

src/app/static/src/app/components/adr/SelectDataset.vue
hillalex: It's straight-forward with `i18next`, see https://github.com/mrc-ide/hint/pull/336/files#diff-305dda242c365630c350e025be821a5cR113 for example.

hillalex: Yep.

EmmaLRussell (approved):

PR #323, opened 2020-09-04
mrc-1757 Show persistent version numbers for versions

EmmaLRussell:
Associated with https://github.com/mrc-ide/hint-db/pull/17
Haven't put label translations in as they should be the same for English and French anyway.


src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -29,6 +31,7 @@
             <b-collapse :id="`versions-${p.id}`">
                 <div v-for="v in p.versions" :id="`v-${v.id}`" class="row font-italic bg-light py-2">
                     <div class="col-md-4 version-cell"></div>
+                    <div class="col-md-1 version-cell">{{`v${v.versionNumber}`}}</div>
```
Nicolas-Dolan, line 34: I'm getting 'vundefined' as the version number when I run the code myself here. I'll update my Docker and try again.

Nicolas-Dolan: Nevermind, updating my Docker fixed it.

scripts/run-dependencies.sh
```diff
@@ -12,7 +12,7 @@ HINTR_VERSION=$(<$HERE/../src/config/hintr_version)
 
 REGISTRY=mrcide
 DB_IMAGE=$REGISTRY/hint-db:master
-DB_MIGRATE_IMAGE=$REGISTRY/hint-db-migrate:master
+DB_MIGRATE_IMAGE=$REGISTRY/hint-db-migrate:mrc-1757
```
Nicolas-Dolan, line 15: Will this need to be changed back to master when you merge?

src/app/static/src/app/components/projects/ProjectHistory.vue
Nicolas-Dolan: It might be a good idea to make the version numbers also reflect the project id, which is also an integer that increments by one each time. Something like `v.${p.id}.${v.versionNumber}`. You wouldn't necessarily have to change anything in the back end as the 'true' version number could stay the same but the user is shown this adapted version number with the project id prefix.

src/app/static/src/tests/components/projects/projectHistory.test.ts
```diff
@@ -203,7 +207,7 @@ describe("Project history component", () => {
             }
         });
         const wrapper = getWrapper(testPropsData, mockStore);
-        const deleteLink = wrapper.find("#v-s11").findAll(".version-cell").at(3).find("a");
+        const deleteLink = wrapper.find("#v-s11").findAll(".version-cell").at(4).find("a");
```
Nicolas-Dolan, line 210: Maybe add a test to check that the version counter is updated when a version is deleted.

Nicolas-Dolan (changes requested):
Works well. I have some suggestions regarding the version number and testing.

src/app/static/src/app/components/projects/ProjectHistory.vue
EmmaLRussell: I do like that idea because you do get unique 'v' ids then for each version, but I think it might be confusing for the user because they've never seen the project id anywhere else. Just having v1, v2, for each project seems simple and intuitive to me. @hillalex any thoughts?

scripts/run-dependencies.sh
EmmaLRussell: Yes, will do that after merge.

src/app/static/src/tests/components/projects/projectHistory.test.ts
EmmaLRussell: Deleting the version doesn't actually happen within the component, as the component is reflecting the state which is being mocked here. We're checking that the delete action is invoked by clicking the link, and also separately checking that the version count is correct for the mocked state - that should be enough to unit test the component.

scripts/run-dependencies.sh
EmmaLRussell: Oh actually, that db branch is already merged, I'll make that change here.

src/app/static/src/app/components/projects/ProjectHistory.vue
hillalex: So projects also have a numeric id? For some reason I thought their names were primary keys. Agree adding more (seemingly to the user) random numbers in probably not worth it. I could image something like "project-name.v1" could be nice, but then again it's a lot of duplication. I reckon leave as `v1` etc for now and see what Jeff/users say. 

I have a feeling they might want to edit those version names themselves rather than just have them as auto-incremented numeric ids - v1, v2 etc is an improvement over no ids, but can imagine they still might be confusing (e.g. v1 ends up being the most recently updated, even though it was created first, could be a confusing user experience.)

hillalex (approved):
Code looks fine to me, some conflicts to fix. My biggest reservation is that numeric ids won't be enough to deconfuse the user, but makes sense as a first iteration to see how Jeff/users respond.

src/app/static/src/app/components/projects/ProjectHistory.vue
Nicolas-Dolan: Yeah, I suspect Jeff will want the user to be able to name the versions themselves to reflect what they are doing. As Alex points out, semantic non-ordinal names would be best, but we should leave as is for now and collect feedback.

Nicolas-Dolan (approved):
Looks good to me overall.

PR #325, opened 2020-09-10
Mrc 1743 moving currentUser to store and adding isGuest getter

Nicolas-Dolan:
Rather than declaring currentUser as a global const in each component, it is declared once in the root and then accessed from the store in all other components.


src/app/static/src/app/localStorageManager.ts
```diff
@@ -35,6 +35,8 @@ export class LocalStorageManager {
     };
 
     getState = (): Partial<RootState> | null => {
+        // console.log('storage manager currentUser', state.currentUser)
+        // console.log('storage manager ', this)
```
Nicolas-Dolan, line 39: I'm not sure if it is possible to access the root store from here so I have left this file as is (ie, declaring currentUser). Let me know if you think there is another way of doing this.

src/app/static/src/tests/components/header/userHeader.test.ts
```diff
@@ -80,6 +91,7 @@ describe("user header", () => {
     });
 
     it("contains help document links", () => {
+        const store = createStore()
         // Reset translations
         registerTranslations(store);
```
hillalex, line 96: now that you've refactored the test setup, this line registering translations is obsolete and can be deleted

src/app/static/src/app/localStorageManager.ts
hillalex: That seems correct to me :+1: Just need to delete the 2 lines of commented code.

src/app/static/src/app/components/header/UserHeader.vue
```diff
@@ -77,7 +77,7 @@
                     return filename;
                 }),
             isGuest() {
-                return this.user == "guest"
+                return this.$store.getters.isGuest
```
hillalex, line 80: This absolutely works, but isn't consistent with the codebase. We use the `mapGetters` helper (much like the `mapState`, `mapActions`, `mapMutations` helpers.) 

See https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue#L161 for an example. In this case, because it is a root getter the code will be even simpler than that example:

```
isGuest() {
     return this.$store.getters.isGuest         
}
```
becomes
```
...mapGetters(["isGuest"])
```

hillalex (commented):
Functionality is looking good, just need a change to syntax to keep the codebase consistent.

hillalex:
The Travis branch build is failing because of some breaking changes to hintr. 

To fix:

* `git checkout master && git pull`
* `git checkout mrc-1743 && git merge master`

Then push the merged changes.

src/app/static/src/app/components/header/UserHeader.vue
Nicolas-Dolan: I've applied this to every instance of isGuest except in the ADRIntegration.vue where TS seems to have issues with a computed property loggedIn returning a getter

src/app/static/src/app/components/adr/ADRIntegration.vue
```diff
@@ -11,13 +11,11 @@
...
 
-    declare const currentUser: string;
-
     export default Vue.extend({
         components: {adrKey, SelectDataset},
         computed: {
-            loggedIn() {
-                return currentUser != "guest"
+            loggedIn(): boolean {
+                return !this.$store.getters.isGuest
```
Nicolas-Dolan, line 18: TS seems to have issues with a computed property loggedIn returning a getter from mapGetters so I have left this as is. Let me know if there is a better way of doing this.

hillalex: We have a helper utility, `mapGetterByName` that can help here, although at the moment it's set up to only work with namespaced getters, so will need a small tweak to work with root getters.

This should become:
```
isGuest: mapGetterByName(null, "isGuest"),
loggedIn() {
       return !this.isGuest
}
```

You may also need to explicitly type the component, so:
```
    interface Methods {
        getDatasets: () => void
        fetchADRKey: () => void
    }

    interface Computed {
        isGuest: boolean
        loggedIn: boolean
        key: string | null
    }

    export default Vue.extend<{}, Methods, Computed, {}>({
       ...
```

And to make this work, in `app/utils`, the `mapGetterByName` function needs to become:
```
export const mapGetterByName = <T>(namespace: string | null, name: string): ComputedWithType<T> => {
    return (namespace && mapGetters(namespace, [name])[name]) || mapGetters([name])[name]
}
```

hillalex (commented):
There's also one failing test: https://travis-ci.com/github/mrc-ide/hint/jobs/385018311

hillalex (approved):
Brilliant, thanks :)

PR #326, opened 2020-09-10
mrc-1718 Bug fix: canceling a file edit then reselecting the same file

Nicolas-Dolan:
Ended up being a really simple fix despite taking a long time. The file upload method would reset the input value once it had finished uploading, allowing the input v-on:change to detect any subsequent files being uploaded (even if it was the same file).  However, this wouldn't happen if the cancel button was pressed instead so I had the cancel button do something similar, resetting the input value.


hillalex (approved):
Nice one, thanks!

PR #327, opened 2020-09-15
mrc-1639_modal

Nicolas-Dolan:
Users can copy a previous version into a new Project

AKA copy version. Effectively promotes that version to a new project. User is prompted to provide a name. Adds a row to the Projects table, and copies the selected version for the new project, and loads that new version

This PR just adds the basic UI for the feature.


src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -25,17 +25,23 @@
                 <div class="col-md-2 project-cell">
                     <a @click="deleteProject($event, p.id)" href="" v-translate="'delete'"></a>
                 </div>
+                <div class="col-md-2 project-cell">
+                    <a @click="copyProject($event, p.id)" href="" v-translate="'copyToNewProject'"></a>
```
hillalex, line 29: Coping whole projects will be a bit more involved than just copying a single version into a new project, and will almost certainly require some back-end work. Is it part of the spec, or is copying a single version into a new project enough to satisfy requirements at this point? I'd be tempted to leave whole project copying for now, and wait to see if it's needed.

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -153,6 +157,9 @@ const en: Translations = {
     clickHere: "Click here",
     colorIndicator: "Colour Indicator",
     continue: "continue",
+    copyProjectHeader: "Copying project to a new project",
+    copyToNewProject: "Copy to a new project",
+    copyVersionHeader: "Copying project-version to a new project",
```
hillalex, line 162: `project-version` looks a bit weird here, maybe just "Copying version to a new project"? Can be replaced with "Copying {versionId} to a new project" once we have version ids!

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -55,6 +61,30 @@
...
+            <h4 v-if="versionToCopy" v-translate="'copyVersionHeader'"></h4>
+            <h5 v-translate="'enterProjectName'"></h5>
+            <template v-slot:footer>
+                <div class="container">
+                    <div class="row">
+                        <input type="text" class="form-control" v-translate:placeholder="'projectName'" v-model="copiedProjectName">
+                    </div>
+                    <div class="row">
+                        <button type="button"
+                            class="btn btn-white mt-2 mr-1 col"
```
hillalex, line 75: I see that you've used `btn-white` for the OK action and `btn-red` for the cancel action because those are what's used in the delete modal. In fact this is backwards since `btn-red` is our primary call to action button (that one must have slipped through code review!), so both these modals should have the button colours swapped. Do you mind fixing that for both modals in this branch?

hillalex (changes requested):
Works really nicely :) See comments

hillalex (approved):

PR #328, opened 2020-09-18
mrc-1770 changed access ADR key url to UNAIDS site

Nicolas-Dolan:
Fjelltopp: "We have created a simpler way to get ADR api key for users. When you visit https://adr.unaids.org/me you'll be redirected to your own user details page and your api key will be displayed in a green box at the top.
If a user is not logged in he'll see a login page and will be redirected to the user details after a successful auth."


src/app/src/main/resources/config.properties
```diff
@@ -5,7 +5,7 @@ adr_pop_schema=inputs-unaids-population
 adr_shape_schema=inputs-unaids-geographic
 adr_survey_schema=inputs-unaids-survey
 adr_schema=inputs-unaids-estimates
-adr_url=https://dev.adr.fjelltopp.org/
```
hillalex, line 8: This should remain as `dev.adr.fjelltop.org` as it's the test ADR site we use for local testing. The real url gets injected into the config during deployment. All automated tests are run against the dev server where there is a test user and fake data, so changing this will cause them all to fail.

src/app/src/main/resources/config.properties
```diff
@@ -5,7 +5,7 @@ adr_pop_schema=inputs-unaids-population
...
+adr_url=https://adr.unaids.org/me/
```
hillalex, line 8: We also don't want to hardcode `/me` at the end of the url here because this url is used in various places, not just for the link to get the access code.  The link to the access code should just be updated in the `ADRKey.vue` component.

hillalex (changes requested):
There are a lot of failing tests, which should be a clue that this isn't quite right!

hillalex (approved):

PR #329, opened 2020-09-21
mrc-1658 frontend for renaming projects

Nicolas-Dolan:
Added a modal to front end to allow the user to rename a project.


src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -15,8 +15,9 @@
                         <chevron-down-icon size="20" class="icon when-open"></chevron-down-icon>
                     </button>
                 </div>
-                <div class="col-md-3 project-cell">
+                <div class="col-md-3 project-cell d-flex justify-content-between">
                     {{p.name}}
+                    <a @click="renameProject($event, p.id)" href="" v-translate="'renameProject'"></a>
```
Nicolas-Dolan, line 20: Could use an edit icon as a link here instead of text "Rename project"

EmmaLRussell: Yes, I think that would look neater and you'll probably find that you're starting to run out of columns when you merge this with the latest master. If it's an icon, should include text in tooltip. We're not leaving much room for the project name itself here, which might be quite long depending on how people use it - we should deal gracefully with overflow.

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -347,6 +351,8 @@ const fr: Partial<Translations> = {
     projectName: "Nom du projet",
     projects: "Projets",
     remove: "supprimer",
+    renameProject: "Rename project",
```
EmmaLRussell, line 354: I'd normally put in our rough French translations here, I don't think we should start putting any English into the French section. My rudimentary French knowledge + Google suggests 'Renommer le projet' and 'Entrez un nouveau nom pour le projet' (+ 's'il vous plaît'? - not sure if the 'please' seems weird in French). I usually run whatever I come up with past Rich.

EmmaLRussell (commented):
See comments, otherwise looks fine. You could take this ticket as far as the 'back of the front end' by adding and invoking the rename action, though of course you don't yet have the backend endpoint to call. That's why I usually do any backend  branch before the frontend.

hillalex:
>  You could take this ticket as far as the 'back of the front end' by adding and invoking the rename action, though of course you don't yet have the backend endpoint to call.

Just as a counterpoint to this, I like keeping the PRs small like this! Nice and easy to review :) And since we're batching features for release it doesn't matter that we're merging them incomplete.

hillalex (approved):
Edit icon plus tooltip sounds like a good shout - otherwise looks good!

Nicolas-Dolan:
I haven't mocked out the calls to the backend on the frontend here, which I'll do in the next ticket along with the backend endpoints. I've added a bunch more feature switch conditionals for the unit test for the project history as we are simply flipping the feature switches rather than removing them at this point. I've not mocked out every possible combination of features the front end could have, just a few key possibilities to make sure the rename project feature gets tested appropriately.

Nicolas-Dolan:
I haven't mocked out the calls to the backend on the frontend here, which I'll do in the next ticket along with the backend endpoints. I've added a bunch more feature switch conditionals for the unit test for the project history as we are simply flipping the feature switches rather than removing them at this point. I've not mocked out every possible combination of features the front end could have, just a few key possibilities to make sure the rename project feature gets tested appropriately.

src/app/static/src/app/featureSwitches.ts
```diff
@@ -1,7 +1,9 @@
 const urlParams = new URLSearchParams(window.location.search);
 const shareProject = !!urlParams.get('shareProject');
 const promoteProject = !urlParams.get('promoteProject');
+const renameProject = !urlParams.get('renameProject');
```
hillalex, line 4: this switch is turned on at the moment - just need one more `!` here to turn it off

hillalex (commented):
Looks great, thanks for fixing this up. Just need to flip the switch the other way for now, so it's off by default! Then go ahead and merge.

EmmaLRussell (approved):

PR #331, opened 2020-09-23
mrc-1661 enforce unique names for users

hillalex:
Some front and back end validation. One thing that's a bit weird - even if a project is deleted it hangs around in the db, meaning you still won't be able to create a new project with the same name. Could exclude any projects that have versions with the "deleted" flag not set, or could add a "deleted" flag to the project table.


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ADRTests.kt
```diff
@@ -126,6 +127,7 @@ class ADRTests : SecureIntegrationTests() {
     }
 
     @ParameterizedTest
+    @Ignore
```
hillalex, line 130: Sorry I broke this test by deleting a file from ADR and it won't let me re-upload the file for... reasons :shrug: . Will have to wait for 24 hours for their data to be refreshed before this passes again.

hillalex:
Hmm, just doing the clone project ticket and realised there's a clash here - if we allow cloning of projects from other users, we can't enforce name uniqueness. I think we should maybe just scrap this.

hillalex:
Closing in favour of https://github.com/mrc-ide/hint/pull/333

PR #332, opened 2020-09-23
mrc-1639 action - backend functionality for promoting a version into its own project

Nicolas-Dolan:
Renamed copyVersion action to promoteVersion for consistency and to avoid confusion with a different function in the versionRepository. Had a lot of conflicts with recent changes to master, which I believe I resolved without breaking anything.

Backend tests currently work and pass but might be a bit limited so comments regarding how they can be more robust would be appreciated.


src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -61,18 +61,19 @@
                 </button>
             </template>
         </modal>
-        <modal :open="projectToCopy || versionToCopy">
-            <h4 v-if="projectToCopy" v-translate="'copyProjectHeader'"></h4>
-            <h4 v-if="versionToCopy" v-translate="'copyVersionHeader'"></h4>
+        <modal :open="versionToCopy">
+            <h4 v-translate="'copyVersionHeader'"></h4>
```
Nicolas-Dolan, line 65: This header could be changed to give the actual version name once it is combined with Emma's ticket that is adding in version names. Something like `Copying ${versionName} to a new project`

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -26,7 +26,7 @@
                     <a @click="deleteProject($event, p.id)" href="" v-translate="'delete'"></a>
                 </div>
                 <div class="col-md-2 project-cell">
-                    <a @click="copyProject($event, p.id)" href="" v-translate="'copyToNewProject'"></a>
+                    <a @click="copyVersion($event, p.id, p.versions[0].id)" href="" v-translate="'copyLatestToNewProject'"></a>
```
Nicolas-Dolan, line 29: Once all of the tickets affecting the projectHistory have been combined (version name, rename project, and this one), we could replace all the links with icons that have tooltips.

src/app/static/src/app/components/projects/ProjectHistory.vue
Nicolas-Dolan: I have now merged in the latest master and added in this feature.

hillalex: Re translations, see this draft for an example of interpolating with `i18next`: https://github.com/mrc-ide/hint/pull/336/files#diff-305dda242c365630c350e025be821a5cR113

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -67,42 +67,64 @@
...
             <template v-slot:footer>
-                <button type="button"
-                        class="btn btn-red"
-                        v-translate="'createProject'">
-                </button>
-                <button type="button"
-                        class="btn btn-white"
-                        @click="cancelCopy"
-                        v-translate="'cancel'">
-                </button>
```
Nicolas-Dolan, line 85: I presume this button was ok to delete as it didn't do anything.

EmmaLRussell (changes requested):
Works really nicely :+1: 

There is one thing a bit weird going on - if I promote a version then refresh the page, I see this error. Not quite sure if it's just related to the code changes in this branch, but needs some investigation:
Screenshot from 2020-10-02 17-58-29


You should also add integration tests in the backend and front end. We can have a chat about those on Monday if you like.

src/app/static/src/tests/projects/actions.test.ts
```diff
@@ -416,4 +416,30 @@ describe("Projects actions", () => {
...
+    it("promoteVersion creates new project containing replicated version", async (done) => {
+        const commit = jest.fn();
+        const dispatch = jest.fn();
+        const state = mockProjectsState({
+            currentProject: mockProject,
+            currentVersion: mockProject.versions[0]
+        });
+
+        const stateUrl = "/project/1/version/testVersion/promote";
+        mockAxios.onPost(stateUrl, "newProject")
```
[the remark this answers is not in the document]
Nicolas-Dolan: I don't think so because qs.stringify({name}) converts the entered name into name=newProject

src/app/static/src/tests/projects/actions.test.ts
```diff
@@ -416,4 +416,30 @@ describe("Projects actions", () => {
...
+        const versionPayload = {
+            version: {projectId: 1, versionId: "testVersion"},
+            name: "newProject"
+        }
+        actions.promoteVersion({commit, state, rootState, dispatch} as any, versionPayload);
+
+        setTimeout(() => {
+
+            const posted = mockAxios.history.post[0].data;
+            expect(posted).toEqual("name=newProject");
```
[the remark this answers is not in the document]
Nicolas-Dolan: It's no longer making a mutation so not testing for that anymore.

Nicolas-Dolan:
Currently has the features `shareProject` and `promoteProject` switched off. This will mean the tests for these features will not assert. To switch on when running locally, go to `featureSwitches.ts` and remove a bang before each of the `urlParams`

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -114,90 +129,115 @@
...
-            confirmDelete() {
-                if (this.projectToDelete) {
-                    this.deleteProjectAction(this.projectToDelete);
-                    this.projectToDelete = null;
-                } else if (this.versionToDelete) {
-                    this.deleteVersionAction(this.versionToDelete);
-                    this.versionToDelete = null;
+            currentLanguage: mapStateProp<RootState, Language>(null,
+                (state: RootState) => state.language)
+        },
```
hillalex, line 167: Something a bit funny going on with spacing, can you auto-format this file? (sorry for nit-picking, we should hopefully soon have a linter for the front-end that will automatically pick these things up on Travis!)

src/app/static/src/tests/projects/actions.test.ts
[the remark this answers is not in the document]
EmmaLRussell: That's right, but that won't match the posted data anticipated in this line, which is just 'newProject'

src/app/static/src/tests/projects/actions.test.ts
```diff
@@ -436,9 +436,30 @@ describe("Projects actions", () => {
         actions.promoteVersion({commit, state, rootState, dispatch} as any, versionPayload);
 
         setTimeout(() => {
-
             const posted = mockAxios.history.post[0].data;
             expect(posted).toEqual("name=newProject");
+
+            expect(dispatch.mock.calls[0][0]).toBe("getProjects");
+            expect(commit.mock.calls.length).toBe(1);
```
EmmaLRussell, line 443: There shouldn't be any commit calls for the success state as we're ignoring success. So this should be checking for length 0. I think this case is actually going through the error route because of the post body matching failing - I tried fixing the mockAxios.onPost locallt to check for data name=newProject as discussed, and that gives a commit call count of 0.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ProjectTests.kt
```diff
@@ -282,8 +282,48 @@ class ProjectTests : VersionFileTests()
...
+        val headers = HttpHeaders()
+        headers.contentType = MediaType.APPLICATION_FORM_URLENCODED
+        val httpEntity = HttpEntity(map, headers)
+
+        testRestTemplate.postForEntity<String>("/project/$projectId/version/$versionId/promote", httpEntity)
+
+        val result = testRestTemplate.getForEntity<String>("/projects/")
+        val data = getResponseData(result) as ArrayNode
+        assertThat(data.count()).isEqualTo(2)
+        val versionsData = data[0]["versions"] as ArrayNode
```
EmmaLRussell, line 305: so versionsData is the versions belonging to your new project, not the old one? Should be, because that's been updated most recently, I think. Maybe worth checking the project name just to be sure.

src/app/static/src/tests/projects/actions.test.ts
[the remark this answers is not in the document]
hillalex: Looks like the `getProjects` action is dispatched whether the POST succeeds or fails (which is harmless enough) hence why this test passes even without the correct axios setup! You would expect there to be no commits though (length 0 not 1 here: https://github.com/mrc-ide/hint/pull/332/files#diff-06e69d219525b7c97c665bab17cb53a1R443)

hillalex (commented):
Great job on your first feature using Kotlin!

src/app/static/src/tests/integration/projects.itest.ts
```diff
@@ -146,4 +146,25 @@ describe("Projects actions", () => {
...
+        expect(commit.mock.calls.length).toBe(2);
+
+        const createdProject = commit.mock.calls[1][0]["payload"];
+        state.currentProject = createdProject;
+        state.currentVersion = createdProject.versions[0];
+        const versionPayload = {version: {projectId: state.currentProject!.id, versionId: state.currentVersion!.id}, name: 'newProject'};
+        
+        await actions.promoteVersion({commit, dispatch, state, rootState} as any, versionPayload);
+        setTimeout(() => {
+            expect(commit.mock.calls.length).toBe(2);
```
EmmaLRussell, line 165: I still think it would be less confusing to make a fresh commit for calling the action - expect no commits from the action, right?

EmmaLRussell (commented):
Nice, tests looks in much better shape. Code coverage is failing on the component but I guess that's because of the feature switches.

src/app/static/src/app/components/projects/ProjectHistory.vue
hillalex: Oh dear, the spacing here is still a bit off, but don't worry about it. We need to commit a style rules file and linter that enforces styles properly.

hillalex (approved):

PR #333, opened 2020-09-23
mrc-1661 front-end validation of project name

hillalex:
Ok, so this just prevents users from entering a project name that is already displayed in their project list. No actual restrictions in the db or the backend as to project name, because of 2 use cases:
1. user deletes a project (which leaves the project record still in the db) but then wants to create a project of the same name
1. user inherits a cloned project that has the same name as an existing one

Although on point 2 - perhaps we want to indicate where a project has been cloned from another user by appending some kind of suffix? E.g. "projectname - copy"


Nicolas-Dolan (commented):
I'm wondering whether this validation process should be moved to its own file where it can be accessed by other components. I am currently working on two tickets (rename project and promote version to project) in the projectHistory component, which both require the user to input a new name for a project. If they could access this validation it would mean not having to repeat code. Or perhaps it could just be exported from this component to make it available elsewhere.

hillalex:
> I'm wondering whether this validation process should be moved to its own file where it can be accessed by other components. I am currently working on two tickets (rename project and promote version to project) in the projectHistory component, which both require the user to input a new name for a project. If they could access this validation it would mean not having to repeat code. Or perhaps it could just be exported from this component to make it available elsewhere.

Great idea, I'll turn this into a mixin.

hillalex:
Ok, this is now a slightly bigger changeset. Both `Projects.vue` and `ProjectHistory.vue` inherit from a new mixin `ProjectsMixin` which exposes the list of projects as a prop called `projects`, a data property called `newProjectName` and a computed prop `invalidName` that tracks whether `newProjectName` is valid.

EmmaLRussell:
> 
> Although on point 2 - perhaps we want to indicate where a project has been cloned from another user by appending some kind of suffix? E.g. "projectname - copy"

Could do that, though of course if they clone it again, you'll get duplicates of  "project name - copy". We could get quite elaborate with this e.g. record when and where a project was cloned from in the db. But maybe we should just keep it simple to start with.

src/app/static/src/app/components/projects/Projects.vue
```diff
@@ -10,7 +10,12 @@
                 </span>
             </div>
             <div class="my-3 col-6 clearfix">
-                <input type="text" class="form-control" v-translate:placeholder="'projectName'" v-model="newProjectName">
+                <input type="text" class="form-control"
+                       v-translate:placeholder="'projectName'"
+                       v-model="newProjectName">
+                <div class="invalid-feedback d-block"
```
EmmaLRussell, line 16: If this is d-inline you can avoid the rest of the screen jumping around when the message shows, though I guess there's a danger of them colliding on smaller screens..

EmmaLRussell (commented):
Looks good to me

hillalex:
Some unrelated backend tests failing, a hintr change perhaps? https://travis-ci.com/github/mrc-ide/hint/jobs/390383854

EmmaLRussell (approved):

PR #335, opened 2020-09-24
mrc-1681 Load file to new project

EmmaLRussell:
This may not be the ideal interface because it presents the user with two dialogs, one to select the file to load and a second to provide the new project name but I figured this was the simplest thing to implement. What do you think?


src/app/static/src/app/store/load/actions.ts
```diff
@@ -78,8 +96,8 @@ async function getFilesAndLoad(context: ActionContext<LoadState, RootState>, fil
         .withError("LoadFailed")
         .postAndReturn<Dict<LocalSessionFile>>("/session/files/", files)
         .then(() => {
-            if (state.loadingState == LoadingState.UpdatingState) {
-                dispatch("updateStoreState", savedState);
+            if (state.loadingState != LoadingState.LoadFailed) {
+                dispatch("load/updateStoreState", savedState, {root: true});
```
EmmaLRussell, line 100: I changed this from a local dispatch because I wanted to write a meaningful integration test which tested the load action actually creating the project in the projects module via the dispatch, so I didn't want to just mock the dispatch. So I create a store in the integration test and take pass its dispatch to the action, but I couldn't work out how to make this a localised dispatch. Any ideas?

hillalex: 2 options:
1. use a fake local dispatch in the test:
```
const localDispatch = (name: string, payload: any, options: DispatchOptions) => {
            if (!options) {
                return store.dispatch("load/" + name, payload)
            } else {
                return store.dispatch(name, payload, options)
            }
}
```
2. use the actual module's dispatch:
```
const moduleDispatch = ((store as any)._modulesNamespaceMap["load/"] as any).context.dispatch
```

hillalex: I guess for an accurate integration test the latter is better, even though you have to access private methods and cast everything to `any`.

hillalex (commented):
The interface looks great to me!

src/app/static/src/app/store/load/actions.ts
EmmaLRussell: Agreed, thanks.

src/app/static/src/app/components/header/FileMenu.vue
```diff
@@ -24,6 +24,35 @@
                 </button>
             </template>
         </modal>
+        <modal id="load-project-name" :open="requestProjectName">
+            <h4 v-translate="'loadFileToProjectHeader'"></h4>
+            <h5 v-translate="'enterProjectName'"></h5>
+            <template v-slot:footer>
+                <div class="container">
+                    <div class="row">
```
hillalex, line 32: Sorry didn't notice this before but this is a bit weird - the buttons make sense in the footer, but the actual input should probably be in the modal body? Also the buttons being full width is inconsistent with our modal styling elsewhere, e.g. https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/components/projects/ProjectHistory.vue#L51

Although in that linked file I notice that the copy project modal is also styled in this idiosyncratic way. We should open a ticket to make that consistent also.

hillalex (changes requested):

src/app/static/src/app/components/header/FileMenu.vue
EmmaLRussell: Ha, yeah, I made this use the same styling as the copy modal as I figured they should be consistent with each other! But yes, will make this consistent with our other modals too.

EmmaLRussell: Added ticket to change the styling in copy modal: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1791

hillalex: Yeah I figured as much! My fault for not noticing the weird styling on the copy project modal when I reviewed that one! Thanks :)

hillalex (approved):

PR #336, opened 2020-09-24
mrc-1642 UI for sharing projects with other users

hillalex:
Adds a link for sharing projects, doesn't actually wire up an action yet. Also adds a test utility for testing translations.


src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -9,34 +9,37 @@
...
+                    {{ p.name }}
                 </div>
-                <div class="col-md-1 project-cell"><small class="text-muted">{{versionCountLabel(p)}}</small></div>
-                <div class="col-md-3 project-cell">{{format(p.versions[0].updated)}}</div>
-                <div class="col-md-2 project-cell">
-                    <a @click="loadVersion($event, p.id, p.versions[0].id)" href="" v-translate="'loadLastUpdated'"></a>
+                <div class="col-md-1 project-cell"><small class="text-muted">{{ versionCountLabel(p) }}</small></div>
+                <div class="col-md-3 project-cell">{{ format(p.versions[0].updated) }}</div>
+                <div class="col-md-1 project-cell">
+                    <a @click="loadVersion($event, p.id, p.versions[0].id)" href="" v-translate="'load'"></a>
```
hillalex, line 25: Changed "Load last updated" to just "Load" to allow this column to become narrower.

src/app/static/src/app/components/projects/ShareProject.vue
```diff
@@ -0,0 +1,122 @@
...
+                e.preventDefault();
+                this.open = true;
+            },
+            confirmShareProject() {
+                if (this.invalidEmail) {
+                    this.showValidationFeedback = true;
+                } else {
+                    this.showValidationFeedback = false;
+                    this.loading = true;
+                    const emails = this.emailsToShareWith.replace(/\s*/g, "").split(",")
```
hillalex, line 93: Just thinking about this interface, and how we report back to the user if say, they enter 1 correct email and 1 email that isn't registered with Naomi. 

Would it be better to have multiple inputs, one for each email, so that success/error can be reported per recipient? Probably right?

src/app/static/src/app/components/projects/ProjectHistory.vue
EmmaLRussell: I think that's ok - we've got the project/version indicator now to show what version people are editing and I guess it's intutive that you'd load the last edited version.

src/app/static/src/app/components/projects/ShareProject.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+    <div>
+        <a @click="shareProject" href=""
+           v-translate="'share'"></a>
+        <modal :open="open">
```
EmmaLRussell, line 5: Don't know if it's worth worrying about, but this will repeat the modal for every link. Could cut down the bloat with v-if="open" on the modal?

src/app/static/src/app/components/projects/ShareProject.vue
```diff
@@ -0,0 +1,122 @@
...
+            <h4 v-translate="'shareProject'"></h4>
+            <div v-if="!loading">
+                <div v-html="instructions" id="instructions"></div>
+                <div class="help-text text-muted small">E.g. someone@gmail.com, another@hotmail.com</div>
```
EmmaLRussell, line 9: I think this will need a French translation - don't think e.g. gets used in French, and could make the email addresses less English

src/app/static/src/tests/components/projects/shareProject.test.ts
```diff
@@ -0,0 +1,159 @@
...
+        link.trigger("click");
+        const expectedEnglish  = "This will create a copy of p1 for the given users." +
+        "Please enter the email address or comma separated list of email addresses you would like to share" +
+        "this project with. These email addresses must be already registered with Naomi."
+
+        const expectedFrench = "Cela créera une copie de p1 pour les utilisateurs désignés." +
+            "Veuillez entrer une adresse e-mail ou une liste d'adresses séparées par des virgules " +
+            "avec lesquelles vous souhaitez partager ce projet. Ces addresses e-mails doivent être déjà enregistrées dans Naomi."
+
+        expectTranslated(wrapper.find(Modal).find("#instructions"), expectedEnglish, expectedFrench, store);
```
EmmaLRussell, line 122: Ah nice one, I did add a ticket for putting his in everywhere. https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1789

src/app/static/src/app/components/projects/ShareProject.vue
hillalex: Ok, will leave the input styling as is. In terms of implementation, I could change the endpoint to accept a list of emails, but then if some fail and some succeed, what status code should it report? Alternatively we do a separate POST for each email, showing concatenated success/error messages from each call. Thoughts?

src/app/static/src/app/components/projects/ProjectHistory.vue
hillalex: Yeah that's what I thought.

src/app/static/src/tests/components/projects/shareProject.test.ts
hillalex: Great, thanks :) Once this is merged we can re-use this utility elsewhere

src/app/static/src/app/components/projects/ShareProject.vue
hillalex: My sense is that since its only once per project, and most people will likely only have 1 project (or a small number), it makes more sense not to use conditional rendering.

src/app/static/src/app/components/projects/ShareProject.vue
EmmaLRussell: Hm, good questions. I guess it would still be a 400 because it failed overall because of invalid input, but since it's affected the state by adding the valid emails that seems confusing. Maybe multiple calls would be cleaner.

EmmaLRussell (approved):
Looks good, just some conflicts to fix before merging.

src/app/static/src/app/components/projects/ShareProject.vue
hillalex: I guess another option is have a single endpoint but if any of the emails are invalid, the whole request fails, i.e. nothing persisted to the db?

PR #337, opened 2020-09-25
mrc-1780 Display project name and version label in VersionStatus component

EmmaLRussell:
Added a utils method which both ProjectHistory and VersionStatus use.
Also renamed the test file which was still named after 'snapshotStatus'.


hillalex (commented):
This is tripping me out - there's an issue with a `float-right` class on the parent container of the version status causing the wrong alignment of the following div which is the ADR access key (see extra padding to the left of the ADR key widget.) However the source code does *not* specify that this element has a `float-right` class?!?!? Am extremely confused - do you also see this `float-right` class being appended to the parent container when running the app?

Screenshot from 2020-09-28 10-33-51

EmmaLRussell:
I think it was because the component tag (<vesion-status> in the parent component (stepper) was given that class, so it was put on the top level div tag.

> This is tripping me out - there's an issue with a `float-right` class on the parent container of the version status causing the wrong alignment of the following div which is the ADR access key (see extra padding to the left of the ADR key widget.) However the source code does _not_ specify that this element has a `float-right` class (see [this line](https://github.com/mrc-ide/hint/pull/337/files#diff-177e28224a324021df7f1287832828d6R2))?!?!? Am extremely confused - do you also see this `float-right` class being appended to the parent container when running the app?
>

hillalex:
Ah, mystery solved!

PR #339, opened 2020-09-25
Update to testing instructions in readme

Nicolas-Dolan:
Updated readme with kotlin testing instructions.


EmmaLRussell (changes requested):
It's the test class that you're specifying here rather than the file name, so I'd probably refer to the 'fully qualified class name' here instead of the file. Might be useful to include this link which shows many variations on this theme - can specify at the package level, method level etc: https://docs.gradle.org/current/userguide/java_testing.html#full_qualified_name_pattern

EmmaLRussell (approved):

PR #340, opened 2020-09-25
mrc-1734 Add translations for resetting data for logged in user

EmmaLRussell:
We re-use the same dialog for guest and logged in user- translations already existed for guest, adding those for logged in user here.


hillalex (approved):

hillalex:
In general, laborious though it is, I wonder if we should be writing unit tests for translations. Otherwise we're relying on manual testing to pick up missing translations like these. We could write a test helper that would make this easy enough, something like `expectTranslatedText(element, englishText, frenchText, store)`.

EmmaLRussell:
> In general, laborious though it is, I wonder if we should be writing unit tests for translations. Otherwise we're relying on manual testing to pick up missing translations like these. We could write a test helper that would make this easy enough, something like `expectTranslatedText(element, englishText, frenchText, store)`.

Good idea, I'll add a ticket for it.

PR #341, opened 2020-09-25
mrc-1782 Add backend endpoint for /projects

EmmaLRussell:
Added '/projects' to the index mapping. This allows users to refresh the page when showing the projects view without getting an error.
Guest users are also be able to access this endpoint, but the front end routing now redirects them to the front end '/' route, as guests should never actually see the projects component.


hillalex (approved):

hillalex:
Just need to fix conflict, then go ahead and merge

PR #342, opened 2020-09-28
mrc-1788_fitmodel Rename "Run model" to "Fit model"

EmmaLRussell:
Preliminary branch for separating out the calibration step - this just renames 'run model' and associated strings to 'fit model' etc. I haven't updated the component names, action names etc as this terminology may not be nailed down..


hillalex (approved):
Good idea to separate this name change out into it's own PR!

PR #345, opened 2020-09-29
mrc-1788 Front end for model calibrate step

EmmaLRussell:
This step comes after model fitting (previously model running) and before review results. I'm assuming that once the hint endpoints have been updated, we won't get the model outputs until after the calibration step, but will leave that restructuring until then.

Includes placeholder action to get a dynamic form with single hardcoded control. 

I've not included a step reset confirmation dialog because the assumption is that users will go back and forward tweaking calibration, but will have to check if this means filter values in the review outputs step get wiped. 

**This branch should not be merged into master now.** Just looking for feedback - will continue to work on branches from here to wire up the backend endpoints until the feature as a whole works.


Nicolas-Dolan:
> This step comes after model fitting (previously model running) and before review results. I'm assuming that once the hint endpoints have been updated, we won't get the model outputs until after the calibration step, but will leave that restructuring until then.
> 
> Includes placeholder action to get a dynamic form with single hardcoded control.
> 
> I've not included a step reset confirmation dialog because the assumption is that users will go back and forward tweaking calibration, but will have to check if this means filter values in the review outputs step get wiped.
> 
> **This branch should not be merged into master now.** Just looking for feedback - will continue to work on branches from here to wire up the backend endpoints until the feature as a whole works.

I'm not sold on 'fit model' as a name. I realise this has probably been agreed and perhaps it is the correct term for what is happening, but my understanding was the interface was to be written in plain English to make it accessible. I don't think 'fit model' will mean much to many people.

EmmaLRussell:
I think that part's already merged in. Yeah, fair point though - I think Jeff thinks in terms of run = fit + calibrate. Hopefully in the workshops where the app will be mostly used, the distinction will be explained to people. 

> I'm not sold on 'fit model' as a name. I realise this has probably been agreed and perhaps it is the correct term for what is happening, but my understanding was the interface was to be written in plain English to make it accessible. I don't think 'fit model' will mean much to many people.

Nicolas-Dolan (approved):
This runs fine for me locally. Will be interesting to see its functionality when it is connected with the backend/model.

src/app/static/src/app/store/modelCalibrate/modelCalibrate.ts
```diff
@@ -0,0 +1,39 @@
...
+        optionsFormMeta: {controlSections: []},
+        options: {},
+        fetching: false,
+        complete: false
+    }
+};
+
+export const modelCalibrateGetters = {
+    complete: (state: ModelCalibrateState) => {
+        return state.complete;
```
hillalex, line 26: Why have a getter here when can just access the state prop directly?

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -0,0 +1,43 @@
...
+                       max: 10,
+                       value: 5
+                   }]
+               }]
+           }]
+       };
+       commit(ModelCalibrateMutation.ModelCalibrateOptionsFetched, {payload: placeholderFormMeta});
+    },
+
+    async calibrate(context) {
```
hillalex, line 38: calibrating is similar to fitting right? do we need to be queuing a calibration request, then polling for changes, showing progress etc, rather than expecting it to be synchronous like this?

hillalex (changes requested):
This sets up the calibration page as analogous to the model options page, which just sets options but doesn't trigger a run - isn't it more like the model fit page, where a run is queued and progress reported on that run? 

[Edited] - ok, so looking at hintr I see it's just one call that returns model result - is it quick?

EmmaLRussell:
Yes, calibration should be fast once the model has run, or so I'm told, so polling should not be necessary

> [Edited] - ok, so looking at hintr I see it's just one call that returns model result - is it quick?

src/app/static/src/app/store/modelCalibrate/modelCalibrate.ts
EmmaLRussell: I think I put this in when I wasn't sure what the API schema would be and thought this might end up being a more complex check on the returned data, but that isn't the case, so yes doesn't need to be a getter.

hillalex:
> Yes, calibration should be fast once the model has run, or so I'm told, so polling should not be necessary
> 
> > [Edited] - ok, so looking at hintr I see it's just one call that returns model result - is it quick?

Ok great, in which case this all looks good - want to fix up the branch wiring up the actions and create a PR into this branch? Then I can merge that in here, and then this into master.

PR #349, opened 2020-09-30
mrc-1642 validate that users exist before sharing a project with them

hillalex:
This PR restyles the share project modal and adds user email validation ahead of actually sharing the project.

This should also leave the UI in a nice place for Rob to take screenshots of the functinality, even though it's not really sharing the project yet.


EmmaLRussell (commented):
UI looks good. I'm not sure it's going to be obvious to users that you need to actually press the enter key after each email address, maybe need more explicit instructions.

It should remove the address if I hit the delete key while focussed on the input? That's not working for me. Again, without explicit instructions, my instinct was to try to press the red x that appears to delete, but that's just indicating invalidity, right?

hillalex:
> UI looks good. I'm not sure it's going to be obvious to users that you need to actually press the enter key after each email address, maybe need more explicit instructions.
> 
> It should remove the address if I hit the delete key while focussed on the input? That's not working for me. Again, without explicit instructions, my instinct was to try to press the red x that appears to delete, but that's just indicating invalidity, right?

Yeah that's the bootstrap invalid state marker, could remove that "x" maybe to make less confusing. Will delete the input when you delete the whole email address, not if you just delete one character, if that's what you mean? But maybe good to add an explicit "remove" button to the right of the input as well. You don't actually have to press enter, just unfocusing the input by e.g. trying to click something else should work? But will add "and press enter" to instructions :+1:

hillalex:
See if you think this is clear enough - haven't added a "remove" button, but have removed the input style that could be confusing. Simply deleting the email address is the way to remove it.

EmmaLRussell (approved):
Oh yeah, I just completely misinterpreted the code, that delete works fine. That is much clearer with the new text and without the misleading non-button.

PR #350, opened 2020-09-30
mrc-1781 enforce style rules

hillalex:
I found a great Kotlin linter, https://detekt.github.io/detekt/ which adds new gradle tasks for linting; have added one linter check to Travis and fixed most of the errors it threw up (the trivial ones like too long lines, missing end of file, wrongly named files etc.) To demonstrate, I committed a changeset with some errors left in, you can see it breaks the Travis build: https://travis-ci.com/github/mrc-ide/hint/jobs/393213805

I've just added linting of the `main` files, not the `test` files, to Travis because the latter have nearly 100 violations which will take a long time to fix! I've also added some fairly arbitrary customisations to the config to get the `main` test to pass, e.g. allowing up to 20 functions in a class (the default is 11) and allowing 10 issues without failing the build (the default is 0). The full list of possible config options is [here](https://github.com/detekt/detekt/blob/master/detekt-core/src/main/resources/default-detekt-config.yml) and explanations for each can be found in the [documentation](https://detekt.github.io/detekt/).

I recommend running the tasks `app:detektMain` and `app:detektTest` locally and looking at the output reports to get an idea of the sorts of errors it's reporting.

This PR also commits an IntelliJ style file and the whole project has been auto-formatted to conform with it. Unfortunately detekt 
doesn't seem to have a built in rule around brace alignment for functions so for now it won't actually detect those style violations that prompted this PR! It is [extensible](https://arturbosch.github.io/detekt/extensions.html) though so we will have to write our own rule to enforce that.

So still to do:
1. write a detekt rule that will enforce brace style
1. add a typescript linter as well for the front-end!


src/build.gradle
```diff
@@ -28,6 +29,18 @@ subprojects {
     apply plugin: 'org.jetbrains.kotlin.jvm'
     apply plugin: 'org.springframework.boot'
     apply plugin: "kotlin-spring"
+    apply plugin: "io.gitlab.arturbosch.detekt"
+
+    detekt {
```
hillalex, line 34: Just highlighting that this is where the task has been added and configured, amongst all the millions of re-formatted files in this PR!

src/config/detekt/detekt.yml
```diff
@@ -0,0 +1,33 @@
+build:
```
hillalex, line 1: And here's the actual config

hillalex:
Code coverage obviously hasn't changed, so I don't know why it's complaining :sob:

EmmaLRussell (approved):
Great stuff

PR #352, opened 2020-10-01
mrc-1642 final wiring up of share project widget

hillalex:
This PR adds the action and mutations for cloning a project, and wires this up to the UI. One slightly odd thing is that you can "share" the project with yourself, although this is useful for testing! (nb - we could fire off a "get all projects" action at the end of a successful share, in case you have shared it with yourself. I haven't included that here.)


EmmaLRussell (changes requested):
This seems to have the same issue that @Nicolas-Dolan 's copy branch has - when I load a copied/shared version, it loads ok but I see a 'Version does not exist' error - this happens when the version state is being saved.
https://github.com/mrc-ide/hint/pull/332#pullrequestreview-501260430
Looking at that screenshot from Nick's branch, the version number looks suspicious - an int rather than a guid. In this branch, it's a guid as expected, but still see the same error.
 
I wonder if the issue might be related to copying the state straight over from the source version without modifying the project state. - the new version when loaded maybe thinks its current version id is the source version id, because that comes from the projects module of the loaded state. I had to work around that when putting in the load file to project feature.

In support of this theory, there's no error when I load a version that the user shared with themselves, but if I create a second test user and share with them, the error shows when the second user loads the shared version. The Version does not exist error would be raised because the source version id in the state does not belong to the second user. 

This suggests that both the share and copy branches need to fix up the copied state in the backend - or we need to rethink project state..

hillalex:
> This seems to have the same issue that @Nicolas-Dolan 's copy branch has - when I load a copied/shared version, it loads ok but I see a 'Version does not exist' error - this happens when the version state is being saved.
> [#332 (review)](https://github.com/mrc-ide/hint/pull/332#pullrequestreview-501260430)
> Looking at that screenshot from Nick's branch, the version number looks suspicious - an int rather than a guid. In this branch, it's a guid as expected, but still see the same error.
> 
> I wonder if the issue might be related to copying the state straight over from the source version without modifying the project state. - the new version when loaded maybe thinks its current version id is the source version id, because that comes from the projects module of the loaded state. I had to work around that when putting in the load file to project feature.
> 
> In support of this theory, there's no error when I load a version that the user shared with themselves, but if I create a second test user and share with them, the error shows when the second user loads the shared version. The Version does not exist error would be raised because the source version id in the state does not belong to the second user.
> 
> This suggests that both the share and copy branches need to fix up the copied state in the backend - or we need to rethink project state..

Oh dear :( Yes that makes total sense about the state being copied over. We obviously need to have a bit of a think about how to architect this.

hillalex:
So, now that the projects back-end is fixed, this should just work. I haven't flipped the switch in this branch because I think we may as well leave that for now to allow for user acceptance testing before we release for good.

EmmaLRussell (approved):
Works for me :+1:

Nicolas-Dolan (approved):
This works well when I tested it locally. I also tested it in combination with the promote version feature and it worked as expected.

hillalex:
Thanks for the careful testing both :+1:

PR #353, opened 2020-10-02
mrc-1704 formatted outputs on tables and tooltips

Nicolas-Dolan:
Takes format, scale, and accuracy arguments from metadata for each indicator and uses a utility to apply them to the values displayed by the tables and tooltips of the choropleth and bubble plot. Barcharts to be included in a separate ticket.


src/app/static/src/app/components/plots/utils.ts
```diff
@@ -173,4 +174,17 @@ export const findPath = function (id: string, obj: any): any {
           }
       }
   }
-}
\ No newline at end of file
+};
+
+export const formatOutput = function (value: number | string, form: string) {
```
EmmaLRussell, line 179: I think this function should take the indicator metadata as a parameter rather than form. Once Rob's put in the backend then the scale and format values should be on the indicator metadata, and you can use those here to first scale the value, then pass the scaled value and format to numeral. 
Until the backend is available: rather than making a dictionary of formats, I was imagining just switching on the indicator id, to decide realistic scale and format for each.

src/app/static/src/app/components/plots/utils.ts
```diff
@@ -173,4 +174,17 @@ export const findPath = function (id: string, obj: any): any {
...
+    var numeral = require('numeral')
```
EmmaLRussell, line 180: This can be a const outside the function.

src/app/static/src/tests/components/plots/bubble/utils.test.ts
```diff
@@ -10,7 +10,10 @@ describe("Bubble plot utils", () => {
         min: 0,
         max: 1,
         colour: "interpolateGreys",
-        invert_scale: false
+        invert_scale: false,
```
Nicolas-Dolan, line 13: I'll look into ways to autoformat these.

src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -202,9 +201,11 @@
                             ...filterLabels,
                         }
                     }
-                    displayRows[key][current.indicatorMeta.indicator] = current.value;
-                    current.lower ? displayRows[key][`${current.indicatorMeta.indicator}_lower`] = current.lower : '';
-                    current.upper ? displayRows[key][`${current.indicatorMeta.indicator}_upper`] = current.upper : '';
+                    const {value, upper, lower} = current
+                    const { indicator, format, scale, accuracy } = current.indicatorMeta
+                    displayRows[key][indicator] = formatOutput(value, format, scale, accuracy);
+                    lower ? displayRows[key][`${indicator}_lower`] = formatOutput(lower, format, scale, accuracy): '';
```
hillalex, line 207: this is quite unconventional syntax, 

```
displayRows[key][`${indicator}_lower`] = lower ? formatOutput(lower, format, scale, accuracy): ''';
```

would be easier to read

src/app/static/src/app/components/plots/utils.ts
```diff
@@ -162,15 +163,45 @@ const roundToPlaces = function (value: number, decPl: number) {
...
+    }
+    return numeral(ans).format(format)
+};
+
+interface Indicator {
+    indicator: string, format: string, scale: number, accuracy: number | null
+}
+
+export const findMetaData = function(indicators: Indicator[], indicatorId: string){
+    return indicators.reduce((acc: Indicator, value: Indicator) => {
```
hillalex, line 201: This is a rather unconventional usage of `reduce` which is generally used when you want to aggregate all the elements of an array in some way; I think what you want here is `find`:
```
return indicators.find(i => i.indicator == indicator.id)
```

These two calls will be slightly different - the code here using `reduce` will return the last element of the array that matches the indicator id; the code using `find` will return the first. I presume we expect there to be only one? In which case `find` would be the clearer and more efficient way. 

Just to also give a bit of context about `reduce` - this code as is would be error prone in 2 ways.
1. it would return the first element of the array if nothing matched the indicator id (since it initialises `acc` to  be the first element of the array) 
1. it would error if the array was empty (because it can't initialise `acc`)

Both of these problems would be fixed by passing `null` as the second argument to `reduce`, which would initialise `acc` as `null`.

hillalex (changes requested):
This is looking great, good job!

hillalex (approved):
Works great, nice one

PR #354, opened 2020-10-04
mrc-1787 Endpoints for model calibration

EmmaLRussell:

PR #355, opened 2020-10-04
mrc-1788_callapi Wire up calibration front end to backend

EmmaLRussell:
As well as calling the api, the also introduces a 'calibrating' flag to the state, and shows any error in the component

Contains changes from https://github.com/mrc-ide/hint/pull/345 and https://github.com/mrc-ide/hint/pull/354

There is an outstanding bug, where the calibrate and review steps appear to be complete when the user goes back to earlier steps. I'm inclined to do this in a separate branch, and have added a ticket: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1875


EmmaLRussell:
Actually, I've realised that bug is worse than I thought it was because it actually prevents you getting past subsequent steps (you get stuck in an infinite loop of invalidation confirmation), so I should fix it in this branch after all!

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -34,12 +35,14 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
...
         const response = await api<ModelRunMutation, ModelCalibrateMutation>(context)
-            .withSuccess(ModelRunMutation.RunResultFetched)
+            .ignoreSuccess()
             .withError(ModelCalibrateMutation.SetError)
             .freezeResponse()
             .postAndReturn<ModelResultResponse>(`/model/calibrate/${modelRunId}`, {options, version});
 
         if (response) {
+            const data = freezer.deepFreeze(response.data);
+            commit({type: `modelRun/${ModelRunMutation.RunResultFetched}`, payload: data}, {root: true});
```
EmmaLRussell, line 45: I've ended up committing the main success route here, as this was failing going through the apiService withSuccess, which I don't think really supports mutation non-root modules which aren't the home module of the commit (because it doesn't construct the path for them)..?

EmmaLRussell:
OK, this should be ready for review now

hillalex (commented):
Thanks for this, testing suggests it all works as expected :tada: 

One thing that would be nice to have (certainly not essential for merging this but perhaps in a small further ticket) - disabling the "Calibrate" button if calibration is already in progress. I think this means stopping the dynamic form from rendering it's own submit button and handle submit outside the form.

One thing that I think should be fixed in this branch - because this changes the number of steps in the app, it is a breaking change as far as reloading state goes. If you have state persisted in localStorage with the old 6 steps, then booting up the app does this:

Screenshot from 2020-10-28 11-57-55

The simplest way to fix this is to change the local storage key (from `appState` to `hintAppStatev2` or something.) 

This will also affect any saved projects but since we haven't deployed that to prod I don't think we need to worry about that.

It will also affect any previously downloaded json files. Reloading them will cause the same behaviour. We should check with Jeff whether we need to support loading of old json files (I think we agreed this was ok, as we were planning on deleting old data anyway?) 

If we don't need to support loading of old files, we can increment `version` in RootState (we should do this anyway!) and when loading state, if the versions don't match, commit a `LoadFailed` mutation as we do when contents are corrupted, with a "this file is from an old version of Naomi" message. 

Alternatively, if we want these files to still be loadable, we could reset just stepper state when the number of steps don't match - this would result in the baseline/survey/options steps from the loaded file to be retained (though not the model run result.)

EmmaLRussell:
Good points! I've made a ticket for disabling the Calibrate button while calibrating: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1882

I think we've said, and Jeff has agreed, that we're not supporting old json files since so many things have changed, so I don't think we really need to worry about that specifically. Checking for the matching version seems like a good generic approach rather than failing in various weird ways - maybe we could allow minor versions to differ but if major version differs then it's a load fail? Happy for me to put that into this branch?

> Thanks for this, testing suggests it all works as expected tada
> 
> One thing that would be nice to have (certainly not essential for merging this but perhaps in a small further ticket) - disabling the "Calibrate" button if calibration is already in progress. I think this means stopping the dynamic form from rendering it's own submit button and handle submit outside the form.
> 
> One thing that I think should be fixed in this branch - because this changes the number of steps in the app, it is a breaking change as far as reloading state goes. If you have state persisted in localStorage with the old 6 steps, then booting up the app does this:
> 
> Screenshot from 2020-10-28 11-57-55
> 
> The simplest way to fix this is to change the local storage key (from `appState` to `hintAppStatev2` or something.)
> 
> This will also affect any saved projects but since we haven't deployed that to prod I don't think we need to worry about that.
> 
> It will also affect any previously downloaded json files. Reloading them will cause the same behaviour. We should check with Jeff whether we need to support loading of old json files (I think we agreed this was ok, as we were planning on deleting old data anyway?)
> 
> If we don't need to support loading of old files, we can increment `version` in RootState (we should do this anyway!) and when loading state, if the versions don't match, commit a `LoadFailed` mutation as we do when contents are corrupted, with a "this file is from an old version of Naomi" message.
> 
> Alternatively, if we want these files to still be loadable, we could reset just stepper state when the number of steps don't match - this would result in the baseline/survey/options steps from the loaded file to be retained (though not the model run result.)

src/app/static/src/app/localStorageManager.ts
```diff
@@ -4,6 +4,7 @@ const appStateKey = "hintAppState_v1.0.0";
 
 export const serialiseState = (rootState: RootState): Partial<RootState> => {
     return {
+        version: rootState.version,
```
EmmaLRussell, line 7: Turns out we weren't actually saving version yet..

hillalex:
> I think we've said, and Jeff has agreed, that we're not supporting old json files since so many things have changed, so I don't think we really need to worry about that specifically. Checking for the matching version seems like a good generic approach rather than failing in various weird ways - maybe we could allow minor versions to differ but if major version differs then it's a load fail? Happy for me to put that into this branch?

Cool, then changing local storage key and checking for inconsistent versions in the state when loading seems like way forward :+1:

src/app/static/src/app/hintVersion.ts
```diff
@@ -0,0 +1 @@
+export const currentHintVersion = "1.0.0";
```
EmmaLRussell, line 1: I originally had this in root.ts but that caused a circular dependency when using it from load actions.

src/app/static/src/app/store/load/actions.ts
```diff
@@ -56,6 +57,15 @@ export const actions: ActionTree<LoadState, RootState> & LoadActions = {
         const files = objectContents.files;
         const savedState = objectContents.state;
 
+        const majorVersion = (s: string) => s ? s.split(".")[0] : null;
+        if (majorVersion(savedState.version) != majorVersion(currentHintVersion)) {
```
hillalex, line 61: Considering we've never incremented this version number and have no protocol for doing so probably semantic versioning is overkill here and just v1, v2 etc might suffice! But fine to leave this as is, maybe we can have a bit more of a think about this longer term - I don't think we do any release versioning of the docker image either? Would be nice to figure out a strategy that ties in with releasing...

hillalex (commented):
Great. One tiny further suggestion, but feel free to merge after that, and then to merge the whole lot into master :tada:

PR #357, opened 2020-10-06
mrc-1728 renamed baseline components labels

carbowizzy:
I have just renamed "PJNZ" to "Spectrum file" and  "Shape file" to "Area boundary file" in baseline component. This changes will enable labels to match terminology from the ADR.


hillalex (commented):
Great job making your first PR so quickly! 

With respect to the code change, this isn't quite right. It's my fault for not sign-posting this on the YouTrack ticket.

If you look in the `ManageFile.vue` component, the `label` prop actually gets passed to the `v-translate` directive here:
https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/components/files/ManageFile.vue#L3

That means that these values should actually be translation keys. The translations can be found here: https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/translations/locales.ts#L273

So really we can leave the keys as they are and just update the value of the translations.

hillalex (approved):
Nice one :)

PR #358, opened 2020-10-08
mrc-1806 secured project endpoints and updated integration test

carbowizzy:
Added /project to secure paths for authentication and updated integration tests in ProjectTests to reflect the changes.

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/utils/SecurePaths.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/utils/SecurePaths.kt
@@ -1,0 +1,21 @@
+package org.imperial.mrc.hint.utils
+
+enum class SecurePaths{
+
+    ADD{
+        override fun pathList(): List<String>
+        {
+            return listOf("/adr/**","/user/**","/project/**")
+        }
+    },
+    EXCLUDE{
+        override fun pathList(): List<String>
+        {
+            return listOf("/adr/schemas", "/adr/schemas/")
+        }
+    };
+
+    abstract fun pathList():List<String>;
+
+
+}
--- a/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/IntegrationTests.kt
+++ b/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/IntegrationTests.kt
@@ -78,6 +78,35 @@ abstract class SecureIntegrationTests : CleanDatabaseTests()
 
     fun assertSecureWithSuccess(isAuthorized: IsAuthorized,
                                 responseEntity: ResponseEntity<String>,
+                                schemaName: String?, httpStatus: HttpStatus)
+    {
+
+        when (isAuthorized)
+        {
+            IsAuthorized.TRUE ->
+            {
+                Assertions.assertThat(responseEntity.headers.contentType!!.toString())
+                        .contains("application/json")
+
+                if (responseEntity.statusCode != httpStatus)
+                {
+                    Assertions.fail<String>("Expected BAD_REQUEST response but got error: ${responseEntity.body}")
+                }
+                if (schemaName != null)
+                {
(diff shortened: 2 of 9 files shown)


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/IntegrationTests.kt
```diff
@@ -76,6 +76,35 @@ abstract class SecureIntegrationTests : CleanDatabaseTests()
                 getTestEntity("survey.csv"))
     }
 
+    fun assertSecureWithSuccess(isAuthorized: IsAuthorized,
```
hillalex, line 79: Since this method is checking for a `HttpSstatus` other than 200/success, this name isn't quite right - how about `assertSecureWithHttpStatus`?

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/IntegrationTests.kt
```diff
@@ -76,6 +76,35 @@ abstract class SecureIntegrationTests : CleanDatabaseTests()
...
+        when (isAuthorized)
+        {
+            IsAuthorized.TRUE ->
+            {
+                Assertions.assertThat(responseEntity.headers.contentType!!.toString())
+                        .contains("application/json")
+
+                if (responseEntity.statusCode != httpStatus)
+                {
+                    Assertions.fail<String>("Expected BAD_REQUEST response but got error: ${responseEntity.body}")
```
hillalex, line 93: Since the function takes HttpStatus as an argument, here we want to pass in the expected status, not hard-code `BAD_REQUEST`.  The message should probably be a little different too, maybe 

`"Expected $httpStatus but got $responseEntity.status"`

hillalex (changes requested):
Great job on your first Kotlin ticket :)

src/app/src/main/kotlin/org/imperial/mrc/hint/utils/SecurePaths.kt
```diff
@@ -0,0 +1,21 @@
+package org.imperial.mrc.hint.utils
```
EmmaLRussell, line 1: Nice to pull these out - maybe it belongs in /security rather than /utils though?

src/app/src/main/kotlin/org/imperial/mrc/hint/utils/SecurePaths.kt
```diff
@@ -0,0 +1,21 @@
...
+
+enum class SecurePaths{
+
+    ADD{
+        override fun pathList(): List<String>
+        {
+            return listOf("/adr/**","/user/**","/project/**")
```
EmmaLRussell, line 8: Sorry, this probably wasn't in the ticket, but I think we should add /projects too, that's the route to fetch all projects for a user.

src/app/src/main/kotlin/org/imperial/mrc/hint/utils/SecurePaths.kt
carbowizzy: You are right, I can have a new package named Security under Utils package.

src/app/src/main/kotlin/org/imperial/mrc/hint/utils/SecurePaths.kt
hillalex: @EmmaLRussell I thought this was the route that just returns the Index and should be excluded here?

hillalex: In fact just looking at the code I'm now very confused - do we have 2 routes `projects/` and `projects` that return different things based only on the presence of a trailing slash? That seems... bad!

hillalex: I suggest we exclude from this ticket and open a new ticket to change this

hillalex (approved):

PR #360, opened 2020-10-08
mrc-1821 Clone project does not set session version id

EmmaLRussell:
The clone project controller method was using session.generateVersionId to create new version ids for the copied versions. However, generateVersionId also set the new version id as the session variable, which we don't want.

The branch the separates out the id generation from setting the session id.


src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
```diff
@@ -65,17 +65,22 @@ class Session(private val webContext: WebContext, private val pac4jConfig: Confi
     fun getVersionId(): String
     {
         //Generate a new id if none exists
-        return (pac4jConfig.sessionStore.get(webContext, VERSION_ID) ?: generateNewVersionId()) as String
+        return (pac4jConfig.sessionStore.get(webContext, VERSION_ID) ?: setNewVersionId()) as String
```
EmmaLRussell, line 68: Not sure why codecev is complaining about this line, as I'm explicitly testing both cases here...

Nicolas-Dolan: I'm not super clear on what this line of code does. It looks like it runs a conditional on getting the versionId for the session but then doesn't return anything if there is a versionId for the session? I'm not familiar with kotlin syntax but should it not return the session versionId between `?` and `:`?

Nicolas-Dolan (commented):
So the reasoning for changing the naming of generateNewVersionId to generateVersionId is to prevent it implying that it is replacing the existing versionId with a new one?

src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
EmmaLRussell: This is the Elvis operator - it evaluates the first clause, and returns that if it's not null. If it is null, it evaluates and returns the second clause. Useful for situations like these, where you want to fetch some data which may or may not exist, and do something else if it doesn't. 
https://kotlinlang.org/docs/reference/null-safety.html#elvis-operator
I think you're expecting it to behave like the tertiary operator in js, which it does look a bit like.

EmmaLRussell:
Yeah, plus 'new' seems a bit redundant when you're generating it anyway. 

> So the reasoning for changing the naming of generateNewVersionId to generateVersionId is to prevent it implying that it is replacing the existing versionId with a new one?

src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
```diff
@@ -65,17 +65,22 @@ class Session(private val webContext: WebContext, private val pac4jConfig: Confi
...
     }
 
-    fun generateNewVersionId(): String
+    fun generateVersionId(): String
     {
-        val newVersionId = UUID.randomUUID().toString()
+        return UUID.randomUUID().toString()
+    }
+
+    private fun setNewVersionId(): String
```
hillalex, line 81: Perhaps it would be even clearer if we do away with this method altogether?

hillalex (commented):
This makes a lot more sense, no hidden side-effects!

hillalex (approved):

PR #362, opened 2020-10-13
Mrc 1823 HINT endpoint that retrieves hintr and Naomi version info from hintr

carbowizzy:

PR #363, opened 2020-10-20
Mrc 1824 Show hintr & Naomi version info in the front end

carbowizzy:
added version info to the header menu


src/app/static/src/app/localStorageManager.ts
```diff
@@ -14,10 +14,11 @@ export const serialiseState = (rootState: RootState): Partial<RootState> => {
         modelOptions: rootState.modelOptions,
         modelOutput: rootState.modelOutput,
         stepper: rootState.stepper,
-        metadata: {...rootState.metadata, plottingMetadataError: null},
```
hillalex, line 17: was this line removed by mistake? doesn't seem like it has anything to do with the current ticket...

src/app/static/src/app/components/header/HintrVersionMenu.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <drop-down :text="`v${hintrVersions.naomi}`" :right="true" style="flex: none">
+    <a class="dropdown-item" href="#"> naomi: v{{ hintrVersions.naomi }} </a>
+    <a class="dropdown-item" href="#"> hintr : v{{ hintrVersions.hintr }} </a>
+    <a class="dropdown-item" href="#"> rrq : v{{ hintrVersions.rrq }} </a>
+    <a class="dropdown-item" href="#"
+      >traduire : v{{ hintrVersions.traduire }}</a
```
hillalex, line 7: something a bit weird with the formatting here?

src/app/static/src/app/store/hintrVersion/actions.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    async getHintrVersion(context) {
+        const { commit } = context;
+        const response = await api<HintrVersionMutation, HintrVersionMutation>(context)
+            .ignoreSuccess()
+            .ignoreErrors()
+            .get(`/meta/hintr/version`);
+
+        if (response) {
+            commit({ type: HintrVersionMutation.HintrVersionFetched, payload: response.data })
```
hillalex, line 24: So in this case, because we're just committing the response data, should use:
```
await api<HintrVersionMutation, HintrVersionMutation>(context)
            .withSuccess(HintrVersionMutation.HintrVersionFetched)
            .ignoreErrors()
            .get(`/meta/hintr/version`);
```

And the response data will automatically be committed to the mutation as required.

src/app/static/src/tests/localStorageManager.test.ts
```diff
@@ -39,10 +41,11 @@ describe("LocalStorageManager", () => {
             modelOptions: mockModelOptionsState(),
             modelOutput: mockModelOutputState(),
             stepper: mockStepperState(),
-            metadata: mockMetadataState(),
+            metadata: mockMetadataState({plottingMetadataError: mockError("metadataError")}),
```
hillalex, line 44: why has this been added here?

src/app/static/src/app/store/hintrVersion/hintrVersion.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const initialHintrVersionState = (): HintrVersionState => {
+    return {
+        hintrVersion: { hintr: "unknown", naomi: "unknown", rrq: "unknown", traduire: "unknown" }
+    }
+};
+
+export const hintrGetters = {
+    complete: (state: HintrVersionState) => {
```
hillalex, line 19: I don't see this being used anywhere?

hillalex (changes requested):
Widget looks great - nice one :)

I wouldn't have created a whole new Vuex module for this, but just added the hintr version details to the root state (or possibly to the metadata module.) But this certainly works fine and means you've got to grips with the module system, so I'm not sure it's worth re-writing. What does @EmmaLRussell think?

There is one type of test missing, which is a front-end to back-end integration test. See `metadata.itest.ts` for an example of how to add that. At the moment such a test will fail because the back-end ticket isn't merged - do you want to review the comments made on that PR about style and merge it in to get it working properly? See: https://github.com/mrc-ide/hint/pull/362

src/app/static/src/app/components/header/HintrVersionMenu.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <drop-down :text="`v${hintrVersions.naomi}`" :right="true" style="flex: none">
+    <a class="dropdown-item" href="#"> naomi: v{{ hintrVersions.naomi }} </a>
```
hillalex, line 3: These shouldn't actually be links! Maybe `<span class="dropdown-item">` with `style=cursor: default` to stop the cursor being styled like a link cursor.

hillalex (commented):
Great, thanks for this, but the integration test is failing. To run locally:
1. make sure the app itself is not running as the next script will bring it up inside a docker container
1. run `./src/app/static/scripts/run-dev-dependencies-for-integration-tests.sh`
1. run `npm run integration-test hintrVersion` to run just your new test suite

(these instructions are also documented in https://github.com/mrc-ide/hint/blob/master/src/app/static/README.md

EmmaLRussell:
Yeah, I think it's fine either way - as you say @hillalex it doesn't really need to be in its own module, but it doesn't hurt to keep the root module from expanding too much, as there's already quite a bit in there, in the actions/mutations.

> I wouldn't have created a whole new Vuex module for this, but just added the hintr version details to the root state (or possibly to the metadata module.) But this certainly works fine and means you've got to grips with the module system, so I'm not sure it's worth re-writing. What does @EmmaLRussell think?

EmmaLRussell (commented):
Widget looks good - as Alex says, will be less confusing to not style them as links. Currently seeing all versions as 'unknown', and "No mapping for GET /meta/hintr/version" error in the backend, is the endpoint not implemented yet?

README.md
```diff
@@ -10,6 +10,8 @@ Requirements:
 * npm
 * openjdk 8
 * coreutils (Mac users only)
+* realpath (Mac Users only)
```
hillalex, line 13: Thanks, could you change this to:

```coreutils or [realpath](https://github.com/harto/realpath-osx) (Mac users only)```

hillalex (changes requested):

hillalex (approved):

PR #364, opened 2020-10-21
mrc-1851 add Vue/Typescript linting 

hillalex:
This PR adds eslint to the build, and reformats all files in `src/app/static/src/app` to fix all linting errors. It also updates the .idea codeStyles file so that Vue files will be indented correctly when auto-formatting.

The set of rules is relatively minimal, so we might want to add more as we go. 

Only one notable change:

Using `{}` as a type is generally bad, even though it's actually fine in `Vue.extend<Data, {}, Computed, {}>`. But trying to tell eslint that it's ok in this one context is tricky (e.g. adding disable rule comments in every file, or just disabling the rule altogether which isn't recommended). The proposed solution is to use the `unknown` type rather than `{}`. So have changed that everywhere, and we'll have to just remember to use that approach going forward.

There are still a bunch of warnings (have set Travis build to ignore warnings) that we might want to fix at some point.

Have updated the README to explain what we're doing and how to run linting locally: https://github.com/mrc-ide/hint/pull/364/files#diff-e0ecdb9e46e4c91c3a778b8b6077636810993f454f435196b7cc9241a4216b9c


hillalex:
Obviously code coverage has not changed, despite what codecov is implying!!

EmmaLRussell (approved):

PR #366, opened 2020-10-26
Mrc 1842 Add front end for downloading summary report

carbowizzy:
PR for downloading summary report, I was unable to run the model for visualization of this branch. This branch is expected to show the download summary as seen in the below image.

image


src/app/static/src/tests/components/downloadResults/downloadResults.test.ts
```diff
@@ -35,10 +35,12 @@ describe("Download Results component", () => {
             "Télécharger les résultats grossiers du groupe d'âge", store);
 
         const links = wrapper.findAll("a");
-        expect(links.length).toBe(2);
+        expect(links.length).toBe(3);
         expectTranslated(links.at(0), "Export", "Exporter", store);
         expect(links.at(0).attributes().href).toEqual("/download/spectrum/testId");
         expectTranslated(links.at(1), "Download", "Télécharger", store);
         expect(links.at(1).attributes().href).toEqual("/download/coarse-output/testId");
+        expectTranslated(links.at(2), "Download", "Télécharger", store);
```
hillalex, line 43: Can we also test that the `h4` tag has the correct translation? In fact at the moment it is showing "downloadSummaryOutput" because of a mismatch between the key in the component ("downloadSummaryOutput") and the key in `locales` ("downloadSummaryReport")

hillalex (changes requested):
Nearly there, just need to fix the translation key. 

Sorry about the problem fitting the model, the problem is a bug in hintr - it brings up workers that have a timeout of 5 minutes after which they shut down! So if you have the docker containers running for more than 5 minutes before trying to run the model, you get no workers and no model run result. 

Rich is working on a fix for this today, but you can also get around it by bringing up dependencies right before testing.

hillalex (approved):

PR #367, opened 2020-10-26
mrc-1871 set loading status to true while loading current project

hillalex:
Project loading is broken because it relies on a page refresh, but on page refresh the `Stepper.vue` component is too eager to redirect the user back to `/projects` while the back-end is still loading the current project. Since the stepper respects the `loading` property, we just need to set this before and after loading the current project.


Nicolas-Dolan (approved):
Works well when tested locally.

PR #371, opened 2020-10-30
Disable Calibrate button while calibration in progress

EmmaLRussell:
Hide the form's button and provide a separate button in the component while is disabled while calibrating is true

### DIFF
--- a/src/app/static/src/app/components/modelCalibrate/ModelCalibrate.vue
+++ b/src/app/static/src/app/components/modelCalibrate/ModelCalibrate.vue
@@ -4,12 +4,19 @@
             <loading-spinner size="lg"></loading-spinner>
             <h2 id="loading-message" v-translate="'loadingOptions'"></h2>
         </div>
-        <dynamic-form v-if="!loading"
+        <dynamic-form ref="form"
+                      v-if="!loading"
                       v-model="calibrateOptions"
-                      :submit-text="submitText"
                       @submit="calibrate"
                       :required-text="requiredText"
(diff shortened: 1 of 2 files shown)


src/app/static/src/app/components/modelCalibrate/ModelCalibrate.vue
```diff
@@ -106,7 +114,10 @@
         methods: {
             update: mapMutationByName(namespace, ModelCalibrateMutation.Update),
             calibrate: mapActionByName(namespace,"calibrate"),
-            fetchOptions: mapActionByName(namespace, "fetchModelCalibrateOptions")
+            fetchOptions: mapActionByName(namespace, "fetchModelCalibrateOptions"),
+            submitForm() {
+                (this.$refs.form as any).submit()
```
EmmaLRussell, line 120: Yeah, I forgot about this separate button business when I blithely put this down as a one pointer! We do want submission to go via the form, because it reshapes DynamicFormMeta into DynamicFormOptions for us. This seems a bit ugly but can't think of a simple alternative.

Nicolas-Dolan (approved):
Calibrate button becomes an unresponsive grey button when calibrating as expected when tested on local.

src/app/static/src/app/components/modelCalibrate/ModelCalibrate.vue
hillalex: Yep, I think this is the way :+1:

hillalex (approved):

PR #372, opened 2020-10-30
mrc-1884 Pass format function to barchart component

EmmaLRussell:
I have published the changes from https://github.com/reside-ic/vue-charts/pull/6 to npm so that the formatFunction feature was available to HINT. You'll need to `npm install` to install this new version.

This branch uses the existing formatOutput method to tell the barchart how to format values of a given indicator. 

One thing that might need looking at in future at is the y-axis - it's currently showing percentage indicators to 2 decimal places, as we use the same indicator format for both axis ticks and value tooltips.  So you get a lot of ticks like 10.00% which seems a little verbose. But to fix that we would need separate formats for axes and values in the metadata.

### DIFF
--- a/src/app/static/package-lock.json
+++ b/src/app/static/package-lock.json
@@ -1863,9 +1863,9 @@
       }
     },
     "@reside-ic/vue-charts": {
-      "version": "0.0.4",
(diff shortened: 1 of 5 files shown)


Nicolas-Dolan (approved):
This works well on local. I see what you mean about the yaxis but it's definitely a big improvement to have percentages there instead of floats.

hillalex (approved):

PR #375, opened 2020-11-04
Mrc 1889 Add custom error pages for 404s and 500s

carbowizzy:
This PR is for a friendly 404 error page, 500 page was not included for technical reasons.


src/app/static/templates/404.ftl
```diff
@@ -0,0 +1,31 @@
...
+    <link rel="stylesheet" href="/public/css/style-33eddb4b7c3a4150e40d70284fba3cac.min.css">
+    <!-- endinject -->
+</head>
+<body>
+<div>
+<br>
+<br>
+<br>
+ <h1 class="text-center mt-3">404 Error Page </h1>
+ <p class="text-center mt-3">The requested page was not found, try again.</p>
```
EmmaLRussell, line 17: "Try again" seems a little brusque! Don't think we need it anyway. Would be nice to include a link to the homepage.

src/app/static/templates/404.ftl
```diff
@@ -0,0 +1,31 @@
...
+    <!-- inject:css -->
+    <link rel="stylesheet" href="/public/css/style-33eddb4b7c3a4150e40d70284fba3cac.min.css">
+    <!-- endinject -->
+</head>
+<body>
+<div>
+<br>
+<br>
+<br>
+ <h1 class="text-center mt-3">404 Error Page </h1>
```
EmmaLRussell, line 16: Maybe just 'Page not found'? 404 might not mean much to the user.

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt
```diff
@@ -3,41 +3,44 @@ package org.imperial.mrc.hint.exceptions
...
-                                         status: HttpStatus,
-                                         request: WebRequest): ResponseEntity<Any>
+    @ExceptionHandler(NoHandlerFoundException::class)
+    fun handleNoHandlerFoundException(): ModelAndView
     {
-        logger.error(e.message)
-        // this handles standard Spring MVC exceptions which do not contain
-        // sensitive info so we return the original error message here
-        return unexpectedError(status, request, e.message)
+        return viewErrorPage("404")
```
EmmaLRussell, line 39: This really should distinguish between api requests and web requests - we shouldn't really be getting any 404s over the api calls from the javascript, but if we do it's going confuse things for it to return an html page with a 200 rather than json with a 404. Should be able to tell if this is an api request from the request object. We haven't got a dedicated api route prefix in this application, but we should be including application/json in the Accept header of our api requests so you could check for that.

EmmaLRussell (commented):
Why is the 500 page not being included? Will it be in another ticket?

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt
```diff
@@ -111,12 +114,74 @@ class HintExceptionHandler(private val errorCodeGenerator: ErrorCodeGenerator,
...
+                .toString(Charsets.UTF_8)
+    }
+
+    fun handleOtherExceptions(error: Throwable?, request: WebRequest): ResponseEntity<Any>
+    {
+        var otherExceptions: ResponseEntity<Any>? = null
+
+        when (error)
+        {
+            is HttpRequestMethodNotSupportedException ->
```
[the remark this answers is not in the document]
carbowizzy: The requirement necessitate the need to return type of ModelAndView and ResponseEntity, which the old implementation doesn't support. The new implementation overrides a method with a return type of ResponseEntity that calls throwable classes automatically, so I had to implement each throwable to accept ModelAndView and ResponseEntity as return type.

hillalex:
> Why is the 500 page not being included? Will it be in another ticket?

We discussed 500 and realised that because it's a SPA, there's not really a way for a user to generate a 500 from a direct server request (as opposed to an axios call) so I don't think we need to worry about this case.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ExceptionHandlerTests.kt
```diff
@@ -36,11 +36,8 @@ class ExceptionHandlerTests : SecureIntegrationTests()
     fun `route not found errors are correctly formatted`()
     {
         val entity = testRestTemplate.getForEntity("/nonsense/route/", String::class.java)
-        Assertions.assertThat(entity.statusCode).isEqualTo(HttpStatus.NOT_FOUND)
-        JSONValidator().validateError(entity.body!!,
-                "OTHER_ERROR",
-                unexpectedErrorRegex,
-                "No handler found for GET /nonsense/route/")
+        Assertions.assertThat(entity.statusCode).isEqualTo(HttpStatus.OK)
```
hillalex, line 39: I think we still want the status code here to be a 404.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ExceptionHandlerTests.kt
```diff
@@ -56,12 +53,7 @@ class ExceptionHandlerTests : SecureIntegrationTests()
         val badPostEntity = HttpEntity(body, headers)
 
         val entity = testRestTemplate.postForEntity<String>("/disease/survey/", badPostEntity)
-
-        assertError(entity,
-                HttpStatus.BAD_REQUEST,
-                "OTHER_ERROR",
-                unexpectedErrorRegex,
-                "Required request part 'file' is not present")
+        Assertions.assertThat(entity.statusCode).isEqualTo(HttpStatus.BAD_REQUEST)
```
hillalex, line 56: We dont want this to change, the old test should still pass!

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ExceptionHandlerTests.kt
```diff
@@ -138,10 +133,10 @@ class ExceptionHandlerTests : SecureIntegrationTests()
     }
 
     @Test
-    fun `does not include original message from arbitrary exceptions`()
+    fun `does not include original message from handle exceptions`()
     {
         val sut = HintExceptionHandler(RandomErrorCodeGenerator(), ConfiguredAppProperties())
-        val result = sut.handleArbitraryException(PSQLException("some message", mock()), mock())
+        val result = sut.handleException(PSQLException("some message", mock()), mock())
```
hillalex, line 139: So this is also a behaviour change that I'm not sure whether we want - previously we were returning known SpringBoot error messages as "trace" objects which I think has been useful for debugging...

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt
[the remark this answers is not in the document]
hillalex: This class is no longer overriding the Spring built in ExceptionHandler. So This is just duplicating the handling of Spring exceptions.

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt
```diff
@@ -111,12 +114,74 @@ class HintExceptionHandler(private val errorCodeGenerator: ErrorCodeGenerator,
...
+
+    // resource bundles are encoded with ISO-8859-1
+    fun ResourceBundle.getUTF8String(key: String): String
+    {
+        return this.getString(key)
+                .toByteArray(Charsets.ISO_8859_1)
+                .toString(Charsets.UTF_8)
+    }
+
+    fun handleOtherExceptions(error: Throwable?, request: WebRequest): ResponseEntity<Any>
```
hillalex, line 130: Very nearly there, looking good! But I think we should pull out a method here `handleSpringException` which deals with all the specific errors below, and includes the original error message in the `unexpectedError`:

```
        // this handles standard Spring MVC exceptions which do not contain	
        // sensitive info so we return the original error message here	
        return unexpectedError(status, request, e.message)
```

whereas in the else clause, for all other exceptions, we don't include the original message.

EmmaLRussell (commented):
So it looks like these changes address returning a 404 response, and including original exception messages, but there are a couple of requested changes which you have marked as resolved but I can't see code changes for them: re-instating error logging,and making the page itself a bit friendler. Did you forget to push a commit?

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt
```diff
@@ -114,9 +115,19 @@ class HintExceptionHandler(private val errorCodeGenerator: ErrorCodeGenerator,
                 .toResponseEntity()
     }
 
-    private fun viewErrorPage(page: String): ModelAndView
+    private fun handleErrorPage(page: String, e: Exception, request: WebRequest): Any
```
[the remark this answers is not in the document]
carbowizzy: the implementation works fine on all browsers, is there a browser it doesn't work on ?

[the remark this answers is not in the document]
EmmaLRussell: Not that I know of, I just thought that in the call we'd agreed to leave this part out of the scope of this ticket as we might end up using a different approach. What browsers did you test, or did you use a cross browser tool?

EmmaLRussell (dismissed):
Thanks!

EmmaLRussell (commented):
Sorry, I jumped the gun there, as the build is failing - looks like a couple of linting issues.

EmmaLRussell (approved):

PR #376, opened 2020-11-04
mrc-1714 added French translations for ADR

Nicolas-Dolan:
Went through ADR components and used v-translate/i18 to add translation where needed. Also added missing translations to locales file. Currently translations are just google translations. I'll attach a spreadsheet of all new translations (ie, all translations excluding those sent to UNAIDS for checking previously) to the youtrack ticket.

I'm unsure why buildkite is failing but I suspect it has to do with teething problems with buildkite, not this branch.


src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -10,15 +10,16 @@
...
         <modal id="dataset" :open="open">
-            <h4 v-if="!loading">Browse ADR</h4>
-            <p v-if="loading">Importing files - this may take several minutes. Please do not close your browser.</p>
+            <h4 v-if="!loading" v-translate="'browseADR'"></h4>
+            <p v-if="loading" v-translate="'importingFiles'"></p>
             <div v-if="!loading">
                 <tree-select :multiple="false"
                              :searchable="true"
                              :options="datasetOptions"
+                             :placeholder="select"
```
Nicolas-Dolan, line 22: I tried using `v-translate:placeholder="'select'"` here but it didn't show up in the browser for some reason.

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -351,6 +355,7 @@ describe("select dataset", () => {
         expect(rendered.find(Modal).findAll("button").length).toBe(2);
         expect(rendered.findAll("p").length).toBe(0);
         expect(rendered.find("h4").text()).toBe("Browse ADR");
+        // expectTranslated(rendered.find("h4"), "Browse ADR", "Parcourir ADR", store);
```
Nicolas-Dolan, line 358: It doesn't seem to want to pick up the translations on these tests for some reason (I just get English where I should get French). Does anybody have any suggestions?

Nicolas-Dolan:
I've now added amends to some of the French translations based on Rich's comments.

src/app/static/src/app/components/adr/SelectDataset.vue
EmmaLRussell: You might have to include it as a computed property. See how the ModelCalibrate component provides translations to DynamicForm e,g, https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/components/modelCalibrate/ModelCalibrate.vue#L12

EmmaLRussell (commented):
I think the branch just needs merging with master to pick up the buildkite scripts

src/app/static/src/app/components/adr/SelectDataset.vue
Nicolas-Dolan: That's what I ended up doing with this. See the 'select' computed property,

src/app/static/src/tests/components/adr/selectDataset.test.ts
EmmaLRussell: Taking out `sync: false` from the mount seemed to fix it. Not sure why that was in there? French text needs to be updated to 'Checher ADR'

EmmaLRussell (changes requested):

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -75,7 +77,7 @@ describe("select dataset", () => {
     }
 
     const getStore = (baselineProps: Partial<BaselineState> = {}, rootProps: Partial<RootState> = {}) => {
-        return new Vuex.Store({
+            const store = new Vuex.Store({
```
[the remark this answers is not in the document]
Nicolas-Dolan: I'm not sure how else to apply registerTranslations without having a store variable to pass to it.

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -329,6 +347,8 @@ const fr: Partial<Translations> = {
     bar: "Barre",
     badUsernamePassword: "Le nom d'utilisateur ou le mot de passe est incorrect",
     browse: "Parcourir",
```
Nicolas-Dolan, line 349: Maybe this should be changed to Chercher for consistency? Rich suggested that Parcourir should be changed to Chercher: "I don’t think this is right. This is in the sense of “look in ADR” right? Chercher ADR might be better?"

src/app/static/src/tests/components/adr/selectDataset.test.ts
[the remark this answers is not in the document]
EmmaLRussell: I don't think you should need to do it because expectTranslated does it for you

src/app/static/src/app/store/translations/locales.ts
EmmaLRussell: So that's the text on the button? Yeah, should consistent here, but didn't that key already exist and had another use (or uses) elsewhere? If so, should check that Chercher makes sense for those. Browse and Search (chercher) do seem like slightly different concepts to me, so other uses of Parcourir might be valid? Ask Rich!

src/app/static/src/tests/components/adr/selectDataset.test.ts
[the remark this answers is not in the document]
Nicolas-Dolan: if I change it back the tests fail to render and say `Failed to resolve directive: translate`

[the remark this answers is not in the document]
EmmaLRussell: Oh fair enough, sorry

EmmaLRussell (commented):
@Nicolas-Dolan So I think the only outstanding comment still on this from me is resolving chercher/parcourir? Or are we just to wait and see what UNAIDS come up with?

src/app/static/src/tests/components/adr/selectDataset.test.ts
hillalex: I think `sync: false` was needed before the actions were wired up, to stop the component lifecycle running through and closing the modal before loading spinner state could be tested. If this test now passes without it, feel free to remove (and you can probably remove some or all of the `await Vue.nextTick()` in that case.)

Nicolas-Dolan:
> @Nicolas-Dolan So I think the only outstanding comment still on this from me is resolving chercher/parcourir? Or are we just to wait and see what UNAIDS come up with?

So I explained the context to Rich and he asked a native speaker, and both are fine apparently. After searching the codebase I can't find an instance where the 'browse' key is actually used, only 'browseADR'. So I think we should keep it as chercher for now and if UNAIDS have any objections they can tell us. I was thinking about deleting the 'browse' key from the locales file but I'm worried it is being somewhere and I just can't find it.

EmmaLRussell:
I can't find it being used anywhere either, I guess it's obsolete, should be safe to remove it. 

> After searching the codebase I can't find an instance where the 'browse' key is actually used, only 'browseADR'. So I think we should keep it as chercher for now and if UNAIDS have any objections they can tell us. I was thinking about deleting the 'browse' key from the locales file but I'm worried it is being somewhere and I just can't find it.

src/app/static/src/app/components/adr/SelectDataset.vue
hillalex: >I tried using v-translate:placeholder="'select'" here but it didn't show up in the browser for some reason.

For reference, that's because `placeholder` here is a custom prop being passed to the tree-select component, rather than the native HTML attribute of the same name.

hillalex (approved):

PR #377, opened 2020-11-05
mrc-1894 add build to buildkite

hillalex:
* adds buildkite pipeline
* removes unused Travis related scripts

Codecov is currently missing because we need to get codecov tokens into the buildkite agents. PR for that [here](https://github.com/reside-ic/reside-hyperv-scripts/pull/4) but having some problems deploying it, so I think in the interest of getting HINT builds running again we deploy this as is and add codecov once the agents are sorted.


README.md
```diff
@@ -36,6 +36,10 @@ If the database schema has changed, you can regenerate the database interface co
...
 
 ### Distribution
-A docker image containing the app is created by running `./scripts/build-app.sh`. This is run as part of 
-the Travis build. 
-Run `docker run -p 8080:8080 mrcide/hint:branch_name` to run a built image.
+A docker image containing the app is created as part of the BuildKite build. To create such an image locally,
+run `./buildkite/make-build-env.sh` followed by `./buildkite/build.sh`. A CLI image is also created as part of 
+the BuildKite build, using `./buildkite/build-cli.sh`.
+
+Run `docker run -p 8080:8080 --name hint mrcide/hint:branch_name` to run a built image. The app will not start until
```
hillalex, line 43: I noticed this README info was out of date and I think this is what Lekan referenced yesterday - you can't just run an app image in this way as it waits endlessly for config to be provided.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/HintApplicationTests.kt
```diff
@@ -37,8 +37,7 @@ class HintApplicationTests : SecureIntegrationTests()
     {
         val response = testRestTemplate.getForEntity<String>("/metrics")
         assertThat(response.statusCode).isEqualTo(HttpStatus.OK)
-        assertThat(response.body).isEqualTo("running 1\nbusy_workers 0\nidle_workers 2\n" +
-                "paused_workers 0\nexited_workers 0\nlost_workers 0\nlive_workers 2")
+        assertThat(response.body).contains("running 1")
```
hillalex, line 40: This was not always the case, sometimes the workers were busy when this endpoint was hit. The more specific message is tested in the unit tests, so here just check we're accessing the endpoint.

src/app/static/src/app/store/metadata/metadata.ts
```diff
@@ -31,7 +31,7 @@ export const metadataGetters = {
 
         const selectedDataType = rootState.surveyAndProgram.selectedDataType;
 
-        let metadataForType = null;
+        let metadataForType: Metadata | null = null;
```
hillalex, line 34: When I ran a fresh `npm install`, `ts` compilation threw up a couple of errors. I'm actually suprised it wasn't complaining about these before, but just fixed them here.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/clients/ADRClientTests.kt
```diff
@@ -15,7 +15,7 @@ class ADRClientTests
     fun `can parse successful response from ADR`()
     {
         val sut = ADRFuelClient(ConfiguredAppProperties(), "fakekey")
-        val response = sut.get("/organization_list_for_user")
+        val response = sut.get("organization_list_for_user")
```
hillalex, line 18: turns out the latest version of the ckan API is sensitive to extra slashes, unlike the previous version

Dockerfile
```diff
@@ -2,10 +2,10 @@ FROM openjdk:8u121
 
 RUN mkdir /static/public -p
 
-COPY ./src/app/static/public /static/public
+COPY ./app/static/public /static/public
```
EmmaLRussell, line 5: So this has changed because it's now happening within the build docker image?

EmmaLRussell (commented):
Looks good to me. 

Are you going to move on to sorting code coverage next? it's a shame to lose it even temporarily. The travis builds should just be flaky/should come back, right? I wonder if we should just hang on to the travis script for now, so we can see codecov if and when it returns, but otherwise just ignore travis build failures so long a the bk build is green.

hillalex:
> Looks good to me.
> 
> Are you going to move on to sorting code coverage next? it's a shame to lose it even temporarily. The travis builds should just be flaky/should come back, right? I wonder if we should just hang on to the travis script for now, so we can see codecov if and when it returns, but otherwise just ignore travis build failures so long a the bk build is green.

No idea if Travis will start up again, but yes could keep those. We'll have to rely on code review to look properly at tests (which we should be doing anyway!) but as soon as the flaky buildkite repo problem is resolved the agents will have the codecov tokens and then we can uncomment those lines.

Dockerfile
hillalex: Exactly, but note that it's still used from `scripts/build-app.sh` which itself is called by `src/app/static/scripts/run-dependencies-for-integration-tests.sh` and should still work.

EmmaLRussell (approved):

PR #378, opened 2020-11-05
mrc-1904 Use 'modified' fields when checking for ADR resource changes

EmmaLRussell:
Instead of using the deprecated `resource_id` field, check if either `last_modified` or `resource_modified` have changed.

Can be tested by changing the dev ADR Malawi dataset by eg deleting or adding a resource or changing a description - the 'Refresh' button should appear shortly after change was made (this now polls for changes every 10s, so shouldn't need to refresh page).


Nicolas-Dolan (approved):
Tested this locally, modifying the descriptions of various files at https://dev.adr.fjelltopp.org/inputs-unaids-estimates/malawi-inputs-unaids-estimates-2020 .  Refresh button appeared as expected with an info tooltip to say what files had been modified. Files refreshed successfully when clicking button. fjelltopp website did crash several times when I edited files though.

PR #380, opened 2020-11-05
mrc-1903 Switch on share projects

EmmaLRussell:
All functionality for share projects should now be working -should be able to create an additional user, and share a project with them so they can see it in their projects list


Nicolas-Dolan (approved):
The validation on the share project feature is a bit slow but works eventually (ie, I entered the email address of the user I wanted to share with and it took quite a while for the ok button to become usable - so long that the first time I tried, I thought it wasn't working). Otherwise the file/project sharing works as expected.

EmmaLRussell:
Interesting, I haven't noticed that being slow before, wonder what caused that. 

> The validation on the share project feature is a bit slow but works eventually (ie, I entered the email address of the user I wanted to share with and it took quite a while for the ok button to become usable - so long that the first time I tried, I thought it wasn't working). Otherwise the file/project sharing works as expected.

PR #382, opened 2020-11-10
mrc-1901 applying number formatting to legend of choropleth and bubble plot

Nicolas-Dolan:
Colour scale for choropleth is now formatted to correspond to correspond to the formatting of the choropleth tooltips and table. Note that percentages are still presented as fractions in the input to adjust the scale. Also indicators that have a scale of modifier in their metadata are not adjusted in the input as I couldn't see a way to have the user's inputted value translated to the scaled value that was actually shown. Instead I have put some text next to the input box in these instances to indicate how much the value is being scaled by.


src/app/static/src/app/components/plots/MapAdjustScale.vue
```diff
@@ -24,21 +24,23 @@ import {ColourScaleType} from "../../store/colourScales/colourScales";
                 <form novalidate>
                     <div class="row p-0 mb-2">
                         <label for="custom-min-input" class="col col-form-label col-2"><span v-translate="'min'"></span></label>
-                        <div class="col pt-1">
+                        <div class="col pt-1" :style="{paddingRight: 5 + 'px'}">
```
Nicolas-Dolan, line 27: Not ideal to use in line style. I'll try to figure out what bootstrap classes to use.

src/app/static/src/app/components/plots/MapAdjustScale.vue
```diff
@@ -24,21 +24,23 @@ import {ColourScaleType} from "../../store/colourScales/colourScales";
...
                             <input id="custom-min-input" type="number" :step="step"
                                    v-model.number="colourScaleToAdjust.customMin"
                                    :max="colourScaleToAdjust.customMax"
                                    @change="update" @keyup="update" :disabled="disableCustom">
                         </div>
+                        <p v-if="colourScaleToAdjust.customMin" class="col col-form-label" :style="{paddingLeft: 0}">{{ scaleText}}</p>
```
Nicolas-Dolan, line 33: I had a lot of issues trying to get the user's input to accurately reflect what was being shown when the values had been scaled. In the end I just add this info next to the input to show how much their entered value is being scaled by.

src/app/static/src/app/components/plots/MapLegend.vue
```diff
@@ -103,8 +105,22 @@
                             valAsProportion = 1 - valAsProportion;
                         }
 
-                        if (val >= 1000) {
-                            val = numeral(val).format("0a")
+                        val = formatOutput(val, format, scale, accuracy)
+
+                        if (typeof(val) === 'number') {
```
hillalex, line 110: Sorry I'm really confused here - why aren't we just using the result of `formatOutput`?

hillalex (commented):
This might be mostly or entirely a feature of the metadata needing to be changed, but there is some really weird behaviour going on with the scaling/rounding, see the following screenshots:

This is looking at proportion recently infected in the survey data:

Shouldn't the upper bound be 60 here?

Screenshot from 2020-11-11 14-20-07

This seems like a result of the metadata being wrong, i.e aggressive rounding making every value 0 (the tooltips also all show 0) - but seems pretty important to fix as this data is useless!

Screenshot from 2020-11-11 14-19-47

Nicolas-Dolan:
> This might be mostly or entirely a feature of the metadata needing to be changed, but there is some really weird behaviour going on with the scaling/rounding, see the following screenshots:
> 
> This is looking at proportion recently infected in the survey data:
> 
> Shouldn't the upper bound be 60 here?
> 
> Screenshot from 2020-11-11 14-20-07
> 
> This seems like a result of the metadata being wrong, i.e aggressive rounding making every value 0 (the tooltips also all show 0) - but seems pretty important to fix as this data is useless!
> 
> Screenshot from 2020-11-11 14-19-47

You're right, the metadata on some of the indicators is way off. I believe this instance is due to accuracy being set to too high a number when all the actual values are below the accuracy threshold so they just get rounded to 0. Also, one of the indicators that should be formatted as percentage  isn't. Jeff/Rob are aware of this and Jeff needs to go through each indicator and adjust but I don't think this has happened yet. I might make a list of suggested changes to help this along. I think the frontend is doing what it's supposed to though.

src/app/static/src/app/components/plots/MapLegend.vue
Nicolas-Dolan: I'm following what was there before. I believe format `0a` was being used to make the numbers a little smaller on the scale (eg, make 10000 into 10k). It wouldn't be appropriate to do that on the table or tooltips but for the scale it makes sense. I've set it up so that it does the scaling/formatting of the `formatOutput` util and then applies the `0a`/`0.0a` formatting on top of that for the very large numbers.

Nicolas-Dolan: Again, sometimes issues with the metadata such an accuracy that is too high will interfere here causing all the numbers to be rounded down to 0. Fixing the metadata should prevent this from happening.

Nicolas-Dolan: I think it might be a good idea not to pass an accuracy parameter into formatOutput here. We're already rounding because of format("0a") so no need to round twice. This would avoid the scale being affected by some of the weird numbers coming in from the metadata as well.

hillalex: Ok, got you.

I think that you could simplify a little here with:

```
if (typeof(val) === "string" && !val.includes('%')) {
     val = parseFloat(val)
}
if (typeof val == "number") {
    ... formatting logic
}

```

src/app/static/src/app/components/plots/MapAdjustScale.vue
```diff
@@ -78,12 +80,14 @@ import {ColourScaleType} from "../../store/colourScales/colourScales";
     interface Props {
         show: boolean,
         colourScale: ColourScaleSettings,
-        step: number
+        step: number,
+            metadata: any
```
hillalex, line 84: could you also do an auto-format on this file? spacing has gone a bit weird

hillalex:
> Jeff/Rob are aware of this and Jeff needs to go through each indicator and adjust but I don't think this has happened yet. I might make a list of suggested changes to help this along. I think the frontend is doing what it's supposed to though.

Ok great, glad they're aware of this issue. That sounds like it would be very helpful if you could do that!

Nicolas-Dolan:
> > Jeff/Rob are aware of this and Jeff needs to go through each indicator and adjust but I don't think this has happened yet. I might make a list of suggested changes to help this along. I think the frontend is doing what it's supposed to though.
> 
> Ok great, glad they're aware of this issue. That sounds like it would be very helpful if you could do that!

I've posted on the hint-dev channel

src/app/static/src/app/components/plots/MapAdjustScale.vue
Nicolas-Dolan: I'm struggling to configure my autoformatter so that it doesn't but every html attribute on a separate line so I cannot format without making big changes to the file.

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -276,8 +276,7 @@
             },
             options() {
                 const featureIndicators = this.featureIndicators;
-                const indicatorMetaData = findMetaData(this.indicators, this.selections.indicatorId);
-                console.log('indicatorMetaData', indicatorMetaData, 'colourIndicator', this.colorIndicator)
+                const { format, scale, accuracy } = this.colorIndicator!;
```
hillalex, line 279: really like how you've improved code readability with this type of construct :heart:

hillalex (commented):
Lovely tidy code! Just need to fix some file conflicts...

PR #385, opened 2020-11-11
mrc-1909 updates filters to include calendar_quarter filter

Nicolas-Dolan:
Updated to reflect changes to schema: https://github.com/mrc-ide/hintr/pull/208/files adding calendar_quarter (shown as Period) to ART page of sap. ~~Also removes filters from sap that are disabled and relabels calendar_quarter and year as Period.~~


src/app/static/src/tests/mocks.ts
```diff
@@ -309,6 +309,7 @@ export const mockModelResultResponse = (props: Partial<ModelResultResponse> = {}
             age_group: "1",
             calendar_quarter: "1",
             indicator_id: 1,
+            indicator: 'mock',
```
Nicolas-Dolan, line 312: Seemed to need this property although the value 'mock' I picked at random. Still works though.

src/app/static/src/app/components/plots/FilterSelect.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div>
+    <div  v-if="!disabled">
```
hillalex, line 2: was this requested by Jeff? I think the reason we had it the other way was so that filters didnt "jump" around when you toggle between datasets...

Nicolas-Dolan: Not directly. They wanted calendar quarter and year to both be Period, but they didn't want two Periods on the same page (even if one was disabled) so one needed to be hidden. It seemed to make sense to remove all disabled filters if I was doing that. If you think it would be better, I could think of a way to just hide the disabled calender quarter/year and leave all the other disabled filters on the page.

Nicolas-Dolan: @r-ash What do you think?

hillalex: > Not directly. They wanted calendar quarter and year to both be Period, but they didn't want two Periods on the same page (even if one was disabled) so one needed to be hidden. It seemed to make sense to remove all disabled filters if I was doing that. If you think it would be better, I could think of a way to just hide the disabled calender quarter/year and leave all the other disabled filters on the page.

Ok, I see how Period and Period is not a good UI. But why not label them Quarter/Year?

r-ash: I don't think we should merge them into a single period column if it causes problems for other parts of the UI. The filters jumping around doesn't sound good. I think 2 separate filters is fine and have mentioned it to Jeff that it might be what he gets.

Implementing 2 separate filters use the same drop down in a non-hacky way might need some metadata change in which case I think we should leave it for now and it can be something we put into the backlog.

EmmaLRussell (approved):
It does seem a more consistent UI to disable rather than hide.

PR #387, opened 2020-11-12
Mrc 1891 add icon to ADR retrieved files

carbowizzy:
This PR adds ADR tag indicating whether a file has come from the ADR. if it has, it will show red text "ADR" next to the file name.


src/app/static/src/app/components/files/ManageFile.vue
```diff
@@ -60,6 +61,7 @@
         valid: boolean,
         error: Error,
         existingFileName: string,
+        hasValidAdr: boolean,
```
hillalex, line 64: this prop name doesn't match the prop you're passing to it here - https://github.com/mrc-ide/hint/pull/387/files#diff-ee212c213967126a949944618cc04cbb58942c8165a8ed85b07790a4f4056c74R30 

as a result it will never be true

src/app/static/src/tests/components/baseline/baseline.test.ts
```diff
@@ -194,6 +194,34 @@ describe("Baseline upload component", () => {
         expectDeleteToDispatchAction(2, () => actions.deletePopulation, done);
     });
 
+    it("upload can set validAdr to true if data is from ADR", async () => {       
+        const store = createSut({country: "Malawi"});
+        const wrapper = shallowMount(Baseline, {store});
+
+        wrapper.findAll(ManageFile).at(0).setProps({hasValidAdr : true});
+        expect(wrapper.findAll(ManageFile).at(0).props().hasValidAdr).toBe(true);
```
hillalex, line 202: Rather than testing whether props are as you expect, it's always better to test that the UI element itself is rendered correctly - in this case, we should test whether the tag with "ADR" in it is rendered. In fact, the code has a bug in it and the expected element is not rendered, but you can't pick that up from this test.

src/app/static/src/app/components/files/ManageFile.vue
```diff
@@ -2,6 +2,7 @@
     <div class="form-group">
         <label class="font-weight-bold mb-0" v-translate="label"></label>
         <tick color="#e31837" v-if="valid" width="20px"></tick>
+        <a class="mb-0" v-if="hasValidAdr" href="#">ADR</a>
```
hillalex, line 5: This isn't actually a link, should just be a `span` tag (with `.color-red` for styling)

src/app/static/src/app/components/baseline/Baseline.vue
```diff
@@ -65,16 +68,19 @@
                 pjnz: state => ({
                     valid: !!state.country,
                     error: state.pjnzError,
+                    adrValid: state.pjnz?.fromADR,
                     existingFileName: state.pjnz && state.pjnz.filename
                 } as PartialFileUploadProps),
                 shape: state => ({
                     valid: state.shape != null,
                     error: state.shapeError,
+                    adrValid: state.shape?.fromADR,
```
hillalex, line 77: This prop name doesn't quite make sense - the flag is just showing whether files comes from ADR vs from the user's local computer, has nothing to do with validity, so can we change `adrValid` to `fromADR`?

hillalex (changes requested):

src/app/static/src/tests/components/baseline/baseline.test.ts
hillalex: We should also have similar tests for the `SurveyAndProgram.vue` component

src/app/static/src/app/components/files/ManageFile.vue
carbowizzy: I initially thought this would eventually be a link back to the ADR as seen in comment by Rich on the mockup. It is now changed to span.

hillalex (approved):
this now works nicely, thanks

PR #389, opened 2020-11-13
mrc-1914 - fix import button loading endlessly

carbowizzy:
This PR fixes error which was generated when a user clicks "Import" to import files from ADR when no file is selected, it hangs at the "Importing files - this may take several minutes screen.


src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -210,6 +210,7 @@
                     outOfDate: false} : null
             },
             async importDataset() {
+                if(this.newDatasetId){
```
hillalex, line 213: This needs auto-formatting/proper indenting! And a unit test. Can we also mark the button as disabled when `this.newDatasetId` is null, so the user can't even click it (and add a unit test for this too)?

hillalex (changes requested):

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -588,7 +588,7 @@ describe("select dataset", () => {
 
         await Vue.nextTick();
 
-        expect(rendered.findAll(LoadingSpinner).length).toBe(1);
+        expect(rendered.findAll(LoadingSpinner).length).toBe(0);
```
hillalex, line 591: why has this been changed? it changes the meaning of this test...

hillalex (changes requested):

src/app/static/src/app/components/adr/SelectDataset.vue
carbowizzy: I auto-format, I think the issue has to do with how IntelliJ and VSCode do formatting. I will explore IntelliJ for formatting.

src/.idea/codeStyles/Project.xml
```diff
@@ -19,18 +19,9 @@
       </option>
       <option name="LBRACE_ON_NEXT_LINE" value="true" />
     </JetCodeStyleSettings>
-    <VueCodeStyleSettings>
```
hillalex, line 22: This should not be removed - it contains a rules for indenting Vue files...

hillalex (commented):
The build is failing - I know it's easy to miss sometimes but as part of our PR protocol you should only request a review once you've confirmed that the build is passing!

hillalex (approved):

PR #390, opened 2020-11-16
mrc-1920 fix bug with multiple datasets

hillalex:
FIxes bug which causes the first dataset to always be selected.

### DIFF
--- a/src/app/static/src/app/components/adr/SelectDataset.vue
+++ b/src/app/static/src/app/components/adr/SelectDataset.vue
@@ -91,7 +91,7 @@
         currentLanguage: Language,
         select: string
     }
-    
+
 
     interface Data {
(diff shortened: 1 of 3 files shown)


src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -141,7 +141,7 @@
                 }))
             },
             newDataset() {
-                const fullMetaData = this.datasets.find(d => d.id = this.newDatasetId);
+                const fullMetaData = this.datasets.find(d => d.id == this.newDatasetId);
```
hillalex, line 144: this was the culprit :woman_facepalming:

EmmaLRussell (approved):

PR #391, opened 2020-11-16
mrc-1918 intermittent decryption error

hillalex:
I think I got to the bottom of the intermittent decryption errors - the cipher class is not thread safe and should be instantiated at point of use.

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt
@@ -39,7 +39,7 @@ class HintExceptionHandler(private val errorCodeGenerator: ErrorCodeGenerator,
     @ExceptionHandler(NoHandlerFoundException::class)
     fun handleNoHandlerFoundException(error: Exception, request: WebRequest): Any
     {
-        logger.error(error.message)
+        logger.error(error.message, error)
 
         val page = "404"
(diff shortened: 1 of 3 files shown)


src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt
```diff
@@ -39,7 +39,7 @@ class HintExceptionHandler(private val errorCodeGenerator: ErrorCodeGenerator,
     @ExceptionHandler(NoHandlerFoundException::class)
     fun handleNoHandlerFoundException(error: Exception, request: WebRequest): Any
     {
-        logger.error(error.message)
+        logger.error(error.message, error)
```
hillalex, line 42: using the second argument to pass the error object means the stack trace gets logged

src/app/src/main/kotlin/org/imperial/mrc/hint/security/tokens/KeyHelper.kt
```diff
@@ -12,7 +12,7 @@ object KeyHelper
 {
 
     private const val KEY_PATH = "/etc/hint/token_key"
-    private const val KEY_SIZE = 1024
+    private const val KEY_SIZE = 4096
```
hillalex, line 15: also just bumping this up - it's the key size we use in production

EmmaLRussell (approved):
Works for me!

PR #392, opened 2020-11-17
mrc-1911 Add labels for indicators being displayed on the choropleth/bubble plot legends

Nicolas-Dolan:
Add labels for indicators being displayed on the choropleth/bubble plot legends. Let me know if you think the labels should be styled differently.


EmmaLRussell (approved):
Looks good to me. I wondered about styling the labels to be a bit bolder but I also don't think they should be too shouty, so let's leave them as they are for now.

PR #393, opened 2020-11-17
mrc-1916 add loading spinner to model option validating

carbowizzy:
PR to add loading spinner to the 'Validate' button on the model options page which takes a bit of time now before the check mark appears.


hillalex (changes requested):
Approach looks good, just a little tweaking required.

hillalex (approved):

PR #396, opened 2020-11-17
Mrc-1928 Make load history more prominent in project history

carbowizzy:


EmmaLRussell (changes requested):
Load from name works well. Looks like the load icon for version has ended up in the rename column though. 
There's a failing front end test I think.

EmmaLRussell:
Not sure what you mean,  is there a fix for this in another branch you could merge in?
Still seeing the load icon in the wrong column for the versions (project is ok) - the load column is first already. 
Screenshot from 2020-11-18 15-32-29

> 
> Ohhh yes, it was part of the requirement for the load button to appear first, wasn't just included in the ticket.

EmmaLRussell (approved):

PR #398, opened 2020-11-18
mrc-1945 Add hint version to version menu

EmmaLRussell:
Arguably HintrVersionMenu is no longer a great name for this component, but it was showing more than hintr's version anyway! ('PackageVersionMenu'?)


hillalex (approved):
Can't think of a better name really, `appVersionMenu` or `packageVersionMenu` seem ok?

PR #399, opened 2020-11-19
mrc-1948 Versioning guidance

EmmaLRussell:
We're going to be updating HINT's version number on all functionality changes from now on, which should just mean remembering to update a couple of places as described here. (We'll also add a PR template to make it easier to remember.) Can you check if this makes sense and is clear?

One reason for requiring all relevant branches to provide a new version number is that we're going to add a buildkite step to apply git tags to master branch on merge for each new version, so we really only want that to happen once per version.


EmmaLRussell:
Ha ha, failed ADR integration tests - the system works!

Nicolas-Dolan (approved):
Sounds good to me.

hillalex (approved):
This sounds good to me.

PR #400, opened 2020-11-19
mrc-1567 Make "fit to filtered data" default colour scale

Nicolas-Dolan:
Make “Fit to filtered data” default colour scale or make “fit to entire dataset” the default value for min/max boxes

Does this mean the static defaults are kind of useless?


hillalex (approved):
Good q and... probably? I think that means removing them in hintr

PR #401, opened 2020-11-19
mrc-1931 Show remove file link when there is an error

EmmaLRussell:
I *think* this change is all we need to do to address the issue of not being able to remove files with errors - it allows the delete request to be sent to the backend when there was any error with the upload. 

I believe this works in all cases, both when file save fails (I had to temporarily sabotage the controller method to test this) and when the subsequent validate fails. In both cases the delete request triggered by the 'remove' click returns a 200 so there are no further errors, the existing error is cleared and subsequent step is enabled 

It does look a little bit odd because you have the 'remove' link as you do when you have a (successful) file name, but no file name (I've hidden the File label in this case, but the control still expands to add the remove link). So it does look like you're being invited to remove something that isn't there - we don't actually get the filename back from the server if there's an error. An alternative might be to use a 'clear error' link next to the error instead.

### DIFF
--- a/src/app/static/src/app/store/baseline/baseline.ts
+++ b/src/app/static/src/app/store/baseline/baseline.ts
@@ -9,6 +9,7 @@ import {localStorageManager} from "../../localStorageManager";
 export interface BaselineState extends ReadyState {
     selectedDataset: Dataset | null
     pjnzError: Error | null
+    pjnzErroredFile: string | null
     country: string
     iso3: string
     pjnz: PjnzResponse | null
@@ -16,8 +17,10 @@ export interface BaselineState extends ReadyState {
     regionFilters: NestedFilterOption[]
     flattenedRegionFilters: Dict<NestedFilterOption>
     shapeError: Error | null
+    shapeErroredFile: string | null
     population: PopulationResponse | null,
     populationError: Error | null,
+    populationErroredFile: string | null,
     validating: boolean,
     validatedConsistent: boolean,
     baselineError: Error | null
@@ -29,13 +32,16 @@ export const initialBaselineState = (): BaselineState => {
         country: "",
         iso3: "",
         pjnzError: null,
+        pjnzErroredFile: null,
         pjnz: null,
         shape: null,
         regionFilters: [],
         flattenedRegionFilters: {},
         shapeError: null,
+        shapeErroredFile: null,
         population: null,
         populationError: null,
+        populationErroredFile: null,
         ready: false,
         validating: false,
         validatedConsistent: false,
--- a/src/app/static/src/tests/integration/helpers.ts
+++ b/src/app/static/src/tests/integration/helpers.ts
@@ -5,5 +5,6 @@ export function getFormData(fileName: string): FormData {
     const file = fs.createReadStream(`../testdata/${fileName}`);
     const formData = new FormData();
     formData.append('file', file);
+    formData.get = (name: string) => { return name == 'file' ? File : null };
     return formData
 }
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.0.1
+
+* Fix bug to allow removal of files with errors
+
 # hint 1.0.0
 
 * ADR integration
--- a/src/app/static/src/app/components/baseline/Baseline.vue
+++ b/src/app/static/src/app/components/baseline/Baseline.vue
@@ -69,19 +69,19 @@
                     valid: !!state.country,
                     error: state.pjnzError,
                     fromADR: !!state.pjnz?.fromADR,
-                    existingFileName: state.pjnz && state.pjnz.filename
+                    existingFileName: (state.pjnz && state.pjnz.filename) || state.pjnzErroredFile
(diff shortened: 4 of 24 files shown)


Nicolas-Dolan (commented):
This might be a lot more work but would it be possible to make it so that the stepper can advance even if there are errors in the non-compulsory files and just not include those files in the model? You could have the usual message about the file error + an additional warning with non-compulsory files that this file will not be included in the model, please upload a new file to include, etc. 

I think having a remove link next to a non-existent file looks odd; maybe just an 'x' icon in the corner of the error message or a 'clear error' link as you said.

EmmaLRussell:
I think it's good to require the user to take some action to address the error before moving on, either clearing the error or uploading a valid file - otherwise they might click on without really noticing the error and then get a run that doesn't include all the files they expect. 

But I'll try giving the error its own equivalent of the 'remove' link, I agree that would make more sense. 

> This might be a lot more work but would it be possible to make it so that the stepper can advance even if there are errors in the non-compulsory files and just not include those files in the model? You could have the usual message about the file error + an additional warning with non-compulsory files that this file will not be included in the model, please upload a new file to include, etc. The additional warning could be a different colour maybe.
> 
> I think having a remove link next to a non-existent file looks odd; maybe just an 'x' icon in the corner of the error message or a 'clear error' link as you said. Could use an 'x' icon with a 'clear error' tooltip even.

hillalex:
> I think it's good to require the user to take some action to address the error before moving on, either clearing the error or uploading a valid file - otherwise they might click on without really noticing the error and then get a run that doesn't include all the files they expect.
> 
> But I'll try giving the error its own equivalent of the 'remove' link, I agree that would make more sense.
> 
> > This might be a lot more work but would it be possible to make it so that the stepper can advance even if there are errors in the non-compulsory files and just not include those files in the model? You could have the usual message about the file error + an additional warning with non-compulsory files that this file will not be included in the model, please upload a new file to include, etc. The additional warning could be a different colour maybe.
> > I think having a remove link next to a non-existent file looks odd; maybe just an 'x' icon in the corner of the error message or a 'clear error' link as you said. Could use an 'x' icon with a 'clear error' tooltip even.

I agree with Emma that an explicit action is preferred, but seems like we should if we can return the filename from the backend?

EmmaLRussell:
That would work too - it would mean working round the api class a bit, which I think doesn't give the response back to the caller if there was an error. I'll look into it. 
Sorry, somehow messed up replying to this comment earlier. 
>
> I agree with Emma that an explicit action is preferred, but seems like we should if we can return the filename from the backend?

EmmaLRussell:
OK, file name of errored file is now visible along wth the remove link. 

Rather than changing backend code to get the filename from the backend I've taken the approach of finding the file name from either the upload file form data or the import url provided to the action - then if there is an upload or import error, it sets an 'erroredFile' property in the state, which the component displays. 

It's a bit boring having to have all these new little properties, when we already have response.filename for all the file types - but trying to bodge the response types to include some sort of errored mode would be worse I think. 

This does mean that a file in error isn't flagged as being an ADR file or not to the user - not sure if that's an issue.

src/app/static/src/tests/integration/helpers.ts
```diff
@@ -5,5 +5,6 @@ export function getFormData(fileName: string): FormData {
     const file = fs.createReadStream(`../testdata/${fileName}`);
     const formData = new FormData();
     formData.append('file', file);
+    formData.get = (name: string) => { return name == 'file' ? File : null };
```
EmmaLRussell, line 8: Using FormData with jest seems to need a bit of a helping hand.

src/app/static/src/app/store/baseline/baseline.ts
```diff
@@ -9,15 +9,18 @@ import {localStorageManager} from "../../localStorageManager";
 export interface BaselineState extends ReadyState {
     selectedDataset: Dataset | null
     pjnzError: Error | null
+    pjnzErroredFile: string | null
```
hillalex, line 12: Perhaps you tried this and it threw up other problems but it there a need for extra properties on the state? Couldn't you just construct a fake `PjnzResponse` object with the filename and the `fromADR` flag and save if to the existing `pjnz` property with the existing mutation `PJNZUpdated`? And then that would just filter through to the component without needing to make any changes there?

hillalex (commented):
I like this approach as would be a real pain to change the way the API behaves. One thought though - could it be simplified? See inline comment

src/app/static/src/app/store/baseline/baseline.ts
EmmaLRussell: I did consider this, but I thought there were probably multiple places in the code where we are doing checks on `if (state.pjnz)` etc in order to check for successful responses, which would then fail and be a pain to be debug. I didn't actually try it, might be worth it to overcome the ADR issue though.

hillalex: No worries, lets leave this as is then as it works fine :+1:

hillalex (approved):

PR #404, opened 2020-11-19
Mrc 1929 Add no data message when filter selection results in no data

carbowizzy:


Nicolas-Dolan (changes requested):
I ran into a few bugs with this branch. In the household survey tab I'd get the 'No data to display...' message even when there were data, and eventually after some fiddling around with the Detail level the choropleth corrupted and the legends moved to the top left corner. See screenshot:
image

EmmaLRussell (changes requested):
This looks good, I like the styling. 

It would be nice to also hide the `map-legend` control if emptyFeature is true.

Thinking about the position of the message: while adding this message to the bottom left of the map is ok for the choropleth, if we do need to add it to the bubble plot too (which is possible - Jeff seemed to think that some combinations of filters have no values in output data), that won't work as the size legend is in the bottom left. So we should probably standardise on adding the no data message to the centre of the map. (Which should be possible: https://stackoverflow.com/questions/23762176/leaflet-custom-control-position-center)

Could you maybe just double check with Jeff that his should be added to bubble plot too? And if so, update the message position to be centred, and add another ticket to add this message to the bubble plot.

EmmaLRussell:
See discussion here: https://github.com/mrc-ide/hint/pull/407#pullrequestreview-536339694 
Looks like we are going to have to do the same thing for the bubble plot

carbowizzy:
> See discussion here: [#407 (review)](https://github.com/mrc-ide/hint/pull/407#pullrequestreview-536339694)
> Looks like we are going to have to do the same thing for the bubble plot

I know , working on that.

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -30,7 +30,9 @@
                     </button>
                 </div>
                 <div class="col-md-2 project-cell">
-                    {{ p.name }}
+                    <span
+                    v-on:click="loadVersion($event, p.id, v.id)">
```
EmmaLRussell, line 34: Sorry, I'm still not sure what's going on with this link - it's not actually in master. Did the branch with the load version link never get merged? Or did it get merged and something overrode it? Can you remember what ticket that was in?

src/app/static/src/scss/partials/maps.scss
```diff
@@ -7,6 +7,15 @@
   display: table-cell;
 }
 
+.emptyFeature-center {
```
EmmaLRussell, line 10: Sorry I'm going to be a complete pedant and ask you to change this to kebab-case like the other css classes (`empty-feature-center`)!

EmmaLRussell (changes requested):
Great, thank you. Just a tiny class renaming change, and still a bit confused about that mysterious link from, I assume, another branch.

I do still think it would be good to hide the legends when  no data, but I'll add that as a separate ticket.

EmmaLRussell (approved):
Great, thanks. Just some conflicts to fix then feel free to merge once the build passes.

PR #407, opened 2020-11-20
mrc-1952 User can choose to scale bubble size by filtered dataset or entire dataset

EmmaLRussell:
# Pull Request Template

## Description

Sorry, this is a big changeset, but a lot of the changes are just from renaming 'ColourScale..' types to just 'Scale..' as they are being re-used for size scale now as well.

I was able to completely re-use the logic for finding the colour range, and pulled it out into a method to be used both for finding colour range and size range.  I did put a couple of properties into the MapAdjustScale component to allow the static range options to be hidden for now, but the other tickets for enabling the two static ranges should just be a case of switching off these properties for the Bubble Plot and testing that the scaling works. 

The only other significant changes is adding bubble size scales to the plotting selections modules and percolating updates from MapAdjustScale. 

This PR is using the new template which I merged in order to try it out, but I think we should change its checklist, since at the moment the expectation is that the user won't check all the boxes to complete the PR (e.g it'll be one of major, minor or patch version update). However, github helpfully labels the PR as having completed 'x of y' items so this will look incomplete. So I think we should either put a note into the template to delete options which are not relevant, or combine options (e.g. 'The build is passing, or failing only because of ADR tests'). Could probably also lose the 'Pull Request Template' heading, doesn't seem necessary. What do you think?

## Type of changes

- [ ] Major
- [x] Minor
- [ ] Patches

## Checklist

- [x] I have incremented version number
- [ ] The build passed successfully
- [x] The build failed because of ADR tests

### DIFF
--- a/src/app/static/src/app/components/plots/MapAdjustScale.vue
+++ b/src/app/static/src/app/components/plots/MapAdjustScale.vue
@@ -81,7 +81,7 @@ import {ScaleType} from "../../store/colourScales/colourScales";
 
     interface Props {
         show: boolean,
-        colourScale: ScaleSettings,
+        scale: ScaleSettings,
         step: number,
         metadata: any,
         hideStaticDefault: boolean,
@@ -95,7 +95,7 @@ import {ScaleType} from "../../store/colourScales/colourScales";
     }
 
     interface Data {
-        colourScaleToAdjust: ScaleSettings
+        scaleToAdjust: ScaleSettings
     }
 
     interface Methods {
@@ -106,7 +106,7 @@ import {ScaleType} from "../../store/colourScales/colourScales";
         name: "MapAdjustScale",
         props: {
             show: Boolean,
-            colourScale: Object,
+            scale: Object,
             step: Number,
             metadata: Object,
             hideStaticDefault: Boolean,
@@ -114,15 +114,15 @@ import {ScaleType} from "../../store/colourScales/colourScales";
         },
         data(): any {
             return {
-                colourScaleToAdjust: {...this.colourScale},
+                colourScaleToAdjust: {...this.scale},
                 ColourScaleType: ScaleType
             };
         },
         computed: {
             invalidMsg() {
                 let result = null;
-                if (this.colourScaleToAdjust.type == ScaleType.Custom) {
-                    if (this.colourScaleToAdjust.customMin >= this.colourScaleToAdjust.customMax) {
+                if (this.scaleToAdjust.type == ScaleType.Custom) {
+                    if (this.scaleToAdjust.customMin >= this.sScaleToAdjust.customMax) {
                         result = i18next.t("maxMustBeGreaterThanMin");
                     }
                 }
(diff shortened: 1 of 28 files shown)


src/app/static/src/app/components/plots/MapAdjustScale.vue
```diff
@@ -101,48 +107,52 @@ import {ColourScaleType} from "../../store/colourScales/colourScales";
     export default Vue.extend<Data, Methods, Computed, Props>({
         name: "MapAdjustScale",
         props: {
+            name: String,
```
EmmaLRussell, line 110: I added this name property to allow the component to use it to construct a unique name for its radio button group -  without this, you could have both components open at once and acting as the same group (selecting in one component cleared selection in the other).

hillalex:
looks like the ADR tests failed because the app wasn't ready to accept connections - I guess try bumping up  the `sleep` here: https://github.com/mrc-ide/hint/blob/master/buildkite/test-adr-front-end-integration.sh#L16

hillalex:
Re the template - looks good! But yes I think lose the heading, combine options in final checklist like you suggest, and make "type of changes" into just a section without a checklist, e.g. just put

```
## Type of changes

Major - Minor - Patch 
```

and the author can delete as relevant?

hillalex (commented):
Code looks good and works as expected. One slight oddity - if the filtered dataset has no values in it, get something like this:

Screenshot from 2020-11-23 10-23-47

This kinda looks broken, and I'm not quite sure where the smallest bubble size (or smallest colour scale) comes from? Not sure what we should do in this case - show nothing? Show a "no data" message?

EmmaLRussell:
Ah, so you do get output filter combinations that yield no values! I wasn't sure. See conversation on https://github.com/mrc-ide/hint/pull/404  - I think if there are no values, as well as putting up a message, both the choropleth and bubble plot should hide their legends. 

Yeah, not sure why a value is shown there, I think in choropleth it shows a stacks of 0s. 

Can you remember what filter values you used to get no output values? Was that with the Malawi data?

> Code looks good and works as expected. One slight oddity - if the filtered dataset has no values in it, get something like this:
>

hillalex:
> I think if there are no values, as well as putting up a message, both the choropleth and bubble plot should hide their legends.

that makes sense to me

> Can you remember what filter values you used to get no output values? Was that with the Malawi data?

it was after importing the ADR test data set and replacing the broken programme file with the one in the local test data. I chose "Southern" as the region filter and then selected "Region" as the area detail level.

hillalex:
> it was after importing the ADR test data set and replacing the broken programme file with the one in the local test data. I chose "Southern" as the region filter and then selected "Region" as the area detail level.

which now I think about it seems like a bug! if you don't have a region filter you see data for each region - if you select a specific region with the filter you see no data

hillalex:
Ha, now sleep time looks ok but looks like a legit ADR failure. Happy for this to be merged anyway, seems like the no data question is somewhat independent.

EmmaLRussell:
OK, great, I'll merge this and make another ticket for the PR template changes.

> Ha, now sleep time looks ok but looks like a legit ADR failure. Happy for this to be merged anyway, seems like the no data question is somewhat independent.

PR #408, opened 2020-11-23
mrc-1981 Fix bubble plot for value ranges with one value only

EmmaLRussell:
## Description

As I suspected, the problem was that the bubble plot wasn't prepared for a value range with a single value, i.e. where min == max, and was getting itself into all sorts of divide-by-0 difficulty. 

In fixing this, I've introduced a slight inconsistency - it felt natural to make the only bubble size for the sole value the maximum bubble size. However I noticed that the colour we use for such a value is the colour representing the lower end of the colour scale. What do you think? Should it be consistified by taking min bubble size, or max colour value? Or is no-one going to care?

## Type of changes

- [ ] Major
- [ ] Minor
- [x] Patches

## Checklist

- [x] I have incremented version number
- [ ] The build passed successfully
- [x] The build failed because of ADR tests

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,6 @@
+# hint 1.1.1
+* Fixed bug where bubble plot failed to display bubble for single-value range.
+
 # hint 1.1.0
 
 * Bubble plot size can be scaled by filtered dataset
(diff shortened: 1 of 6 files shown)


hillalex (approved):
Agree max bubble size feels more natural here. I say leave as is and see if anyone complains.

PR #411, opened 2020-11-25
Mrc 1991 - Hide map legends when no data to display

carbowizzy:
## Description

This PR hides legend control pane in maps when selections have not data to display. This PR doesn't hide indicator pane as it is useful for filtering prevalence even when there are not data to display on a particular prevalence.

## Type of version change

Minor

## Checklist

### DIFF
--- a/src/app/static/src/app/components/plots/choropleth/Choropleth.vue
+++ b/src/app/static/src/app/components/plots/choropleth/Choropleth.vue
@@ -21,7 +21,7 @@
                              :level-labels="featureLevels"
                              @detail-changed="onDetailChange"
                              @indicator-changed="onIndicatorChange"></map-control>
-                <map-legend :metadata="colorIndicator"
+                <map-legend v-show="!emptyFeature" :metadata="colorIndicator"
                             :colour-scale="indicatorColourScale"
                             :colour-range="colourRange"
                             @update="updateColourScale"></map-legend>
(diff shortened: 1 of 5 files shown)


src/app/static/src/app/hintVersion.ts
```diff
@@ -1,2 +1,2 @@
 
-export const currentHintVersion = "1.1.3";
+export const currentHintVersion = "1.1.4";
```
EmmaLRussell, line 2: I think we should get into the way of incrementing minor version (i.e. the second number) when we put in new behaviour changes which aren't just bug fixes, as described in the readme. It's maybe debatable which this is, but I think it's less of a bug fix and more of an improvement, which would make this 1.2.0

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -21,7 +21,7 @@
                              :level-labels="featureLevels"
                              @detail-changed="onDetailChange"
                              @indicator-changed="onIndicatorChange"></map-control>
-                <map-legend :metadata="colorIndicator"
+                <map-legend v-show="!emptyFeature" :metadata="colorIndicator"
```
EmmaLRussell, line 24: I'd probably just do do this as a v-if, no reason to render the control at all here.

EmmaLRussell (changes requested):
Looks good, need to add it for the bubble plot too - both size legend and map legend (which is the colour legend).

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
carbowizzy: The v-if behaviour was erratic on the map, freezing indicators control, while v-show happens to work well.

EmmaLRussell (approved):

PR #412, opened 2020-11-26
mrc-1953 Add custom values to bubble scales, and do not exceed max/min bubble size

EmmaLRussell:
## Description
Show custom values for range, and fix bug in bubble plot which meant wrong range was being used to initialise size scale values! Added missing test for this. 

Also ensure we clip bubble size to min and max values. 

## Type of version change
_Delete as appropriate_

Minor

## Checklist

### DIFF
--- a/src/app/static/src/app/components/plots/bubble/BubblePlot.vue
+++ b/src/app/static/src/app/components/plots/bubble/BubblePlot.vue
@@ -339,7 +339,7 @@
                 if (current) {
                     return current
                 } else {
-                    const newScale = initialiseScaleFromMetadata(this.colorIndicator);
+                    const newScale = initialiseScaleFromMetadata(this.sizeIndicator);
                     this.updateSizeScale(newScale);
                     return newScale;
                 }
@@ -452,9 +452,9 @@
                 this.$emit("update-colour-scales", newColourScales);
             },
             updateSizeScale: function (scale: ScaleSettings) {
-                const newColourScales = {...this.sizeScales};
-                newColourScales[this.selections.sizeIndicatorId] = scale;
-                this.$emit("update-size-scales", newColourScales);
+                const newSizeScales = {...this.sizeScales};
+                newSizeScales[this.selections.sizeIndicatorId] = scale;
+                this.$emit("update-size-scales", newSizeScales);
             },
         },
         watch:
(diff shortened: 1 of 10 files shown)


src/app/static/src/app/components/plots/bubble/BubblePlot.vue
```diff
@@ -452,9 +452,9 @@
                 this.$emit("update-colour-scales", newColourScales);
             },
             updateSizeScale: function (scale: ScaleSettings) {
-                const newColourScales = {...this.sizeScales};
-                newColourScales[this.selections.sizeIndicatorId] = scale;
-                this.$emit("update-size-scales", newColourScales);
+                const newSizeScales = {...this.sizeScales};
+                newSizeScales[this.selections.sizeIndicatorId] = scale;
+                this.$emit("update-size-scales", newSizeScales);
```
carbowizzy, line 457: Looks good, render well together. 

I found one weird behaviour in the console though, I believe it has something to do with the scale size. When a user selects an option from size indicator drop down menu which has empty circle radius, the bubble returns NaN. I think we could have a conditional statement checking if selections have radius before trying to display or compute bubbles in min/max control legend.

carbowizzy (changes requested):

src/app/static/src/app/components/plots/bubble/BubblePlot.vue
EmmaLRussell: Ah, great spot. Hm, interesting, they should really all have a calculated radius for all values. Can you tell me how to recreate this with the test data, or maybe upload a screenshot so I can see all the settings?

carbowizzy: Screenshot 2020-11-26 at 18 24 06

carbowizzy: This bug isn't actually caused by your ticket. Its from master.

EmmaLRussell: Yeah, it's because when there's no data, the data range includes NaN and it attempts to draw SizeLegend bubbles with radius NaN... I've fixed it in this branch, though actually your branch to hide the legends when no data should make this go away too I think.

carbowizzy (approved):

PR #413, opened 2020-11-26
mrc-1992 Don't show ADR key to user after it has been entered

carbowizzy:
## Description

This PR removes a link that shows 'edit' after entering ADR key, users can edit or remote the key
Edit allows them to see the unencrypted key. The PR removes the edit action, which leaves users with the option to remove the key if they want to alter/add a new key.

## Type of version change

Minor

## Checklist

### DIFF
--- a/src/app/static/src/tests/components/adr/ADRKey.test.ts
+++ b/src/app/static/src/tests/components/adr/ADRKey.test.ts
@@ -48,12 +48,11 @@ describe("ADR Key", function () {
         expect(rendered.find("span").text()).toBe("none provided");
     });
 
-    it("shows edit/remove links if key exists", () => {
+    it("shows remove links if key exists", () => {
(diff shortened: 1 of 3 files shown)


src/app/static/src/tests/components/adr/ADRKey.test.ts
```diff
@@ -80,7 +79,7 @@ describe("ADR Key", function () {
     it("can edit key", async () => {
```
EmmaLRussell, line 79: I guess this test should be renamed to 'can add key if key exists' or similar

EmmaLRussell (changes requested):
Looks good to me, just need to update NEWS.md with a description of the version change, and I think this is also minor rather than a patch.

EmmaLRussell (approved):

PR #415, opened 2020-11-26
mrc-1994 Make 'Download results' step red when advance to that step

carbowizzy:
## Description

This PR mark download step as complete so that it doesn't appear that user still needs to do something

## Type of version change

Minor

## Checklist


Nicolas-Dolan (approved):
This works well. I had a think about edge cases when the logic might cause issues but none came to mind.

EmmaLRussell (approved):

PR #417, opened 2020-11-27
mrc-2005 Show spinner in Browse ADR while fetching datasets

EmmaLRussell:
## Description

Give user feedback that we're still fetching datasets, rather than just showing no options in dropdown

## Type of version change
_Delete as appropriate_

Minor

## Checklist


carbowizzy (commented):
Looks good, work well,

While loading the dropdown menu data, the dropdown doesn't appear. I think it would be nice to show disabled drop down menu while data is being loaded. As soon as the data is loaded then it should become enabled.

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -45,6 +45,10 @@
                     >
                     </label>
                 </tree-select>
+                <div v-if="fetchingDatasets" class="pt-2" id="fetching-datasets">
+                    <loading-spinner  size="xs"></loading-spinner>
```
Nicolas-Dolan, line 49: I hadn't realised adding a loading spinner would be this straightforward. My illusion that the speed the spinner spins gives any indication about what is going on under the hood has been shattered!

Nicolas-Dolan (approved):
This works well. I tried making it fail and it handles that well too.

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -45,12 +46,12 @@
                     >
                     </label>
                 </tree-select>
-                <div v-if="fetchingDatasets" class="pt-2" id="fetching-datasets">
-                    <loading-spinner  size="xs"></loading-spinner>
+                <div :class="fetchingDatasets ? 'visible' : 'invisible'" style="margin-top:15px;" id="fetching-datasets">
```
EmmaLRussell, line 49: Modifying the visible/invisible class here rather than using v-if so that the height of the dialog doesn't jump about when finish fetching datasets. Setting the margin style to overcome the loading spinner giving itself a negative top margin. 

The styling of the drop down when disabled is almost identical to when it's enabled - the drop icon is a marginally different shade of grey. So I'm not sure that really helps. But see if you think it's better!

carbowizzy: I think it's brilliant...

New users will be aware that they are waiting for drop down data.

carbowizzy (approved):

src/app/static/src/app/components/adr/SelectDataset.vue
carbowizzy: Resolve the conflict and feel free to merge

PR #418, opened 2020-11-27
mrc-2007 Set a separate (long) timeout for front end ADR integration tests

EmmaLRussell:
## Description

We needed to set a long timeout (up to 20s) for slow ADR dev server responses in integration tests. These don't apply to non-ADR integration tests, so separate out the config for ADR tests here, and reset the timeout for non-ADR tests to more reasonable 6s.

## Type of version change
_Delete as appropriate_

None

## Checklist

### DIFF
--- a/src/app/static/package.json
+++ b/src/app/static/package.json
@@ -9,7 +9,7 @@
     "build": "gulp copy-templates && gulp styles && webpack",
     "test": "jest -c jest.config.js",
     "integration-test": "jest --testPathIgnorePatterns adr -c jest.integration.config.js",
-    "adr-integration-test": "jest --testPathPattern adr -c jest.integration.config.js",
+    "adr-integration-test": "jest --testPathPattern adr -c jest.integration.adr.config.js",
     "integration-test-docker": "jest --testPathIgnorePatterns adr -c jest.integration.docker.config.js",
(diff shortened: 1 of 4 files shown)


carbowizzy (approved):
Looks good, well done.

Nicolas-Dolan (approved):
Looks good. This should help speed things up when ADR fails. I know it's not connected but the backend integration tests take longer than ever though.

src/app/static/package.json
```diff
@@ -9,9 +9,9 @@
     "build": "gulp copy-templates && gulp styles && webpack",
     "test": "jest -c jest.config.js",
     "integration-test": "jest --testPathIgnorePatterns adr -c jest.integration.config.js",
-    "adr-integration-test": "jest --testPathPattern adr -c jest.integration.config.js",
+    "adr-integration-test": "jest --testPathPattern adr -c jest.integration.adr.config.js",
```
hillalex, line 12: (could move the `--testPathPattern`/`--testPathIgnorePatterns` flags into the config files now that they've been separated)

EmmaLRussell: This makes sense, but because the configs build on each other, and because there's no equivalent of testPathPattern in config (so I think you'd need to use testRegex instead), you'd end up with the config overriding each other in a couple of ways, which I think would be more confusing that this is...

PR #419, opened 2020-11-27
mrc-1995 Swap button colours in Save version confirmation dialog

EmmaLRussell:
## Description

Make the Save button the more obvious selection, should be red while cancel button should be white

## Type of version change
_Delete as appropriate_

Patches

## Checklist


Nicolas-Dolan (approved):
Yep, that's definitely a red button. :)

carbowizzy (approved):

PR #420, opened 2020-11-27
mrc-1966 Auto advance to next step when run model completes

Nicolas-Dolan:
## Description

This change makes the model options page automatically progress to the calibrate page once it successfully completes fitting the model. However, there is still an issue with the dialogue box that saves changes not appearing when you attempt to go back and re-run the model or change the calibration settings. The initial reasoning behind this ticket was to solve the first issue as it was believed that the stepper simply needed to have progressed for the dialogue to become active (see: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1966), but this doesn't seem to be enough, so its absence seems to be a separate issue for another ticket.

There is also an issue regarding the model options that has been introduced at some stage:  https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2008

## Type of version change
Minor

## Checklist


src/app/static/src/app/components/Stepper.vue
```diff
@@ -146,6 +146,11 @@
             VersionStatus
         },
         watch: {
+            complete: function (){
+                if (this.activeStep === 4){
```
Nicolas-Dolan, line 150: Initially I had the if statement be slightly stricter for this, requiring calibrate options step to be enabled. 
`if (this.activeStep === 4 && this.isEnabled(5)){`
However, I was not able to get the tests for this to pass as the criteria for getting the calibrate to be enabled were very strict and I failed at mocking them up. Should everything else in the app be working correctly, this shouldn't be an issue as you would normally expect the calibrate step to be enabled once the model run completed. However, ideally the stricter criteria would be maintained so I'd welcome any feedback about the tests.

EmmaLRussell: I think the problem with isEnabled is that it requires all previous steps to be complete, but in the case of your test, the earlier steps like baseline are not complete, as defined in the stepper getter 'complete' (which farms out module level logic to getters in baseline etc). You could bodge up baseline state etc which does satisfy the getters that it's complete, or you could just mock the stepper getters to give back the complete values you want. This would mean breaking open createSut though - let's talk about this if not clear.

EmmaLRussell:
Yes. Looking at the logic for this (see `editsRequireConfirmation` in stepper getters) it seems that this is only triggered if there are later *complete* steps. Here the calibrate step is not complete until it's been executed.  We'd forgotten that..

It feels like the editsRequireConfirmation needs a rethink - it's too crude to check if later steps are complete, different subsequent steps may have different  ways of registering edits which necessitate save points. In fact, in the case of model run which prompted all this, it's not really about subsequent steps but about the current step and wanting to save its previous state. 

So maybe steps should have a confirmEdit state and any changes should check confirmEdit on this and subsequent steps..?

> However, there is still an issue with the dialogue box that saves changes not appearing when you attempt to go back and re-run the model or change the calibration settings. The initial reasoning behind this ticket was to solve the first issue as it was believed that the stepper simply needed to have progressed for the dialogue to become active (see: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-1966), but this doesn't seem to be enough, so its absence seems to be a separate issue for another ticket.

Nicolas-Dolan:
> So maybe steps should have a confirmEdit state and any changes should check confirmEdit on this and subsequent steps..?

Adding a confirmEdit state makes sense to me. In terms of the UI, are you proposing a confirm edit dialogue box once the model run has completed? Would clicking confirm advance to the calibrate step or is that no longer necessary?

EmmaLRussell:
We can keep this branch as it seems pretty sensible to advance the stepper when the run completes from UI perspective, but as you say it doesn't sort out the problem we were trying to solve with the save point, which I think we fix in another branch. 

You wouldn't see the confirm dialog on run complete, but instead if you go back to the run step and request another run. My suggestion is that we replace the current logic to check if the dialog is required whenever a change is made (which is: are there any subsequent complete steps?) with: does the current (changing) step or any subsequent step have changes such that editing this step should prompt confirmation? 

Not completely sure how to capture this, but we want it to be triggered by re-running model (current step) and by editing model options then going back and making a change (subsequent step) but not by just editing model options (current step). This might just be a case of deriving the confirmEdit state of the step after model run from whether we have a model run response (so we return to only checking subsequent steps).. 

I think this could improve several steps - for example, at the moment, I suspect the dialog isn't triggered by adding an ART file only in step 2, but is by adding a survey file (required for step 2 to be complete). 

> Adding a confirmEdit state makes sense to me. In terms of the UI, are you proposing a confirm edit dialogue box once the model run has completed? Would clicking confirm advance to the calibrate step or is that no longer necessary?

EmmaLRussell:
hasChanges is probably a more intuitive name than confirmEdit actually

> Adding a confirmEdit state makes sense to me. In terms of the UI, are you proposing a confirm edit dialogue box once the model run has completed? Would clicking confirm advance to the calibrate step or is that no longer necessary?

EmmaLRussell (commented):
I've reinstated isEnabled check and rigged up a test store to satisfy the completed getters so the test now passes. 

(It would be nice to do a maintenance ticket to make that createSut method less ugly, maybe by passing in a single parameter with all the populated partial test states as needed, rather than that huge long parameter list.)

If you're happy with that, could you just finish off by updating the version?

Nicolas-Dolan:
> I've reinstated isEnabled check and rigged up a test store to satisfy the completed getters so the test now passes.
> 
> (It would be nice to do a maintenance ticket to make that createSut method less ugly, maybe by passing in a single parameter with all the populated partial test states as needed, rather than that huge long parameter list.)
I can create a ticket for this if you haven't already?
> 
> If you're happy with that, could you just finish off by updating the version?

@EmmaLRussell The frontend integration tests on this are failing. Do you want me to take a look?

EmmaLRussell:
> I can create a ticket for this if you haven't already?

That would be great, thanks Nick :)

> @EmmaLRussell The frontend integration tests on this are failing. Do you want me to take a look?

Should just be the ADR ones? I think they were timing out earlier

EmmaLRussell:
Actually, I think if you merge with master, that might fix the ADR test errors, looks like it's still using the shorter timeout

EmmaLRussell (approved):

PR #421, opened 2020-11-30
Mrc 1961 stale login doesn't inform user - just shows errors

carbowizzy:
## Description

Display error message and auto redirect stale login to login page. I have commented some blocks of codes, they are to be removed but I wanted to be sure if they are referenced somewhere else although I didn't see them been used else where.

## Type of version change

Minor

## Checklist


src/app/static/src/app/apiService.ts
```diff
@@ -132,10 +137,12 @@ export class APIService<S extends string, E extends string> implements API<S, E>
                 data: {}
             }
         }
+         */
 
-        if (!isHINTResponse(failure)) {
+        if(!isHINTResponse(failure)){
             this._commitError(APIService.createError("apiCouldNotParseError"));
```
Nicolas-Dolan, line 143: Does your average user understand what this error means? Would plain English like: "You are not currently logged in", not be clearer

src/app/static/src/tests/apiService.test.ts
```diff
@@ -148,6 +148,24 @@ describe("ApiService", () => {
             detail: "Your session has expired. Please refresh the page and log in again. You can save your work before refreshing."
         });
     });
+    */
+
+    it("commits da default error message if an empty 401 response is received", async () => {
```
Nicolas-Dolan, line 153: Small typo - da

Nicolas-Dolan (commented):
I tested this by deleting the api cookie and it seems to work as intended, producing the could not parse api message and redirecting to the login screen. Just need to clean up the code by removing comments.

src/app/static/src/app/apiService.ts
EmmaLRussell: This is just a whitespace change, not related to this ticket.

src/app/static/src/app/apiService.ts
```diff
@@ -120,7 +120,12 @@ export class APIService<S extends string, E extends string> implements API<S, E>
             return
         }
 
-        let failure = e.response && e.response.data;
+        if (e.response && e.response.status == 401) {
+            window.location.assign("/login")
+        }
+
+        const failure = e.response && e.response.data;
+        /*
```
EmmaLRussell, line 128: You're right, you shouldn't need this block any more.

src/app/static/src/app/apiService.ts
```diff
@@ -120,7 +120,12 @@ export class APIService<S extends string, E extends string> implements API<S, E>
             return
         }
 
-        let failure = e.response && e.response.data;
+        if (e.response && e.response.status == 401) {
+            window.location.assign("/login")
```
EmmaLRussell, line 124: I think we also said that we wanted the login page to show a message to the user, something your 'Your session expired. Please login again to continue.'  I think you can do this by including the message you want to show in a query string parameter 'error' to the login endpoint. (As handled by the backend here: https://github.com/mrc-ide/hint/blob/master/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/LoginController.kt)
This is nice, because at this point, we still know what the user's preferred language is, so we can keep the message in a locale string..

EmmaLRussell (changes requested):

src/app/static/src/tests/apiService.test.ts
carbowizzy: Spot on, thanks.

src/app/static/src/app/apiService.ts
carbowizzy: removed

src/app/static/src/app/apiService.ts
carbowizzy: Implemented

EmmaLRussell (commented):
Looks great - nearly there!

EmmaLRussell (approved):

PR #422, opened 2020-12-01
update to vue dynamic form version

Nicolas-Dolan:
## Description

Updates to reside-ic/vue-dynamic-form": "^0.4.0",

## Type of version change
Minor

## Checklist

- [x ] I have incremented version number, or version needs no increment
- [ x] The build passed successfully, or failed because of ADR tests

### DIFF
--- a/src/app/static/package-lock.json
+++ b/src/app/static/package-lock.json
@@ -3738,9 +3738,9 @@
       "dev": true
     },
     "@reside-ic/vue-dynamic-form": {
-      "version": "0.0.5",
-      "resolved": "https://registry.npmjs.org/@reside-ic/vue-dynamic-form/-/vue-dynamic-form-0.0.5.tgz",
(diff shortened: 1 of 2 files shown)


EmmaLRussell (changes requested):
Not quite sure why, but only some fields are getting the red 'required' label - see screenshot. The text-danger class is not being applied to the black labels. Any ideas?
Screenshot from 2020-12-01 14-32-50

Nicolas-Dolan:
> Not quite sure why, but only some fields are getting the red 'required' label - see screenshot. The text-danger class is not being applied to the black labels. Any ideas?

@EmmaLRussell It looks like it's doing what it's supposed to. We only wanted empty required fields to turn red, right?

EmmaLRussell:
Ha! Sorry, ignore me, I completely forgot that. 

> 
> @EmmaLRussell It looks like it's doing what it's supposed to. We only wanted empty required fields to turn red, right?

EmmaLRussell (approved):

PR #423, opened 2020-12-01
mrc-1993 makes ADR import button red

Nicolas-Dolan:
## Description

makes ADR import button red as is a call to action

## Type of version change
Patches

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.6.1
+
+* Makes ADR import button red
+
 # hint 1.6.0
 
 * Show spinner when fetching ADR datasets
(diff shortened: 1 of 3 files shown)


EmmaLRussell (changes requested):
Style change looks great, just need to update the version - we're generally saying that any visible changes to the app at all should get a new version. Usually 'Minor' (what you've put in the template) means new functionality, and gets the second number updated. 'Patches' is usually bug fixes, and updates the third number. Maybe style fixes like this should be patch change?

Nicolas-Dolan:
> 'Patches' is usually bug fixes, and updates the third number. Maybe style fixes like this should be patch change?

I've put the style change as Patch

EmmaLRussell (changes requested):
Nearly there - need to remember to update it in hintVersion.ts too!

PR #424, opened 2020-12-01
Mrc 2012 Bug: 400 error from `/model/options` results in hanging spinner

carbowizzy:
## Description

This PR displays error message to users when encountered options model error:

To reproduce, load Uganda datasets from ADR
the datasets will pass validation in baseline but on getting to model options step, an error is encounter. The page displays hanging spinner.

This PR was implemented to display error message and code so that the technical team can troubleshoot with a known code as the default message doesn't include error code/key for this error. However, I have also taken into consideration, a case where the error message might include error code/key in future.

## Type of version change

Patches

## Checklist


src/app/static/src/app/components/modelOptions/ModelOptions.vue
```diff
@@ -12,6 +12,13 @@
                       :required-text="requiredText"
                       :select-text="selectText"></dynamic-form>
 
+        <div v-if="hasOptionsError" id="model-option">
+            <div v-if="!otherError" class="error-message" id="model-option-error">
+                <span v-translate="'loadOptionsError'"></span>
+                {{ optionsError.key }}
+            </div>
+            <error-alert v-else :error="optionsError"></error-alert>
```
EmmaLRussell, line 20: Keep this, but won't need `v-else`

src/app/static/src/app/components/modelOptions/ModelOptions.vue
```diff
@@ -12,6 +12,13 @@
                       :required-text="requiredText"
                       :select-text="selectText"></dynamic-form>
 
+        <div v-if="hasOptionsError" id="model-option">
+            <div v-if="!otherError" class="error-message" id="model-option-error">
```
EmmaLRussell, line 16: As per chat with @r-ash we can take out this markup dealing with special case to show key, and all associated code (e.g. otherError, locale strings), and just retain the `<error-alert>`, and trust hintr to pass us a sensible error detail.

EmmaLRussell (changes requested):
Good stuff, sorry you're going to have to remove some of it! :grimacing:

Nicolas-Dolan (commented):
I wasn't able to replicate the error in master or trigger this condition in this branch (perhaps the Uganda dataset has been fixed since). Because I don't completely understand what this is trying to achieve, my input on this code is limited, but its approach in producing a validate error for the modelOptions seems sensible.

EmmaLRussell (changes requested):
This is much cleaner, thanks. I think we should also hide the `<dynamic-form>` if hasOptionsError - I tried sabotaging the options endpoint to throw an error, and the error displays nicely, but we also get Validate button visible, because the form is displaying with empty options set. 
Screenshot from 2020-12-03 15-31-39

Also need to remember to update NEWS.md please

EmmaLRussell (approved):

PR #428, opened 2020-12-04
mrc-2014 Trigger confirm save dialog if subsequent steps have any changes

Nicolas-Dolan:
## Description

The current way that the confirm save dialog is triggered is if there are any complete future steps, but this means that incomplete steps will get their changes wiped silently.

We should replace this with a check if any subsequent steps have any change.

We currently have 'complete' getters for each step. We should add 'hasChanges' getters as well, and use these in the stepper getter editsRequireConfirmation instead of 'complete'.

hasChanges should be true for each step as follows:

- Baseline - NA as never called, but conceptually if any files are set
- Survey and Program - if any files are set
- Model Run Options - if the Update mutation has been triggered, indicating options have changed (set a flag)
- Model Run - we actually want changes here (re-run) to trigger the dialog if there is a modelRunResult, but since we want to check future steps only, roll that into..
- Calibrate - if there is a model run result (actual calibration settings don't need to be checked)
- Review output - NA, always false
- Download results - NA, always false

We should also check the running the model invokes editsRequireConfirmation - it may not do currently.

**Note:** there is still a small bug in the modelOptions page where a change to multi-options fields will not trigger the confirm save dialog. This will need to be addressed in a separate ticket targeting the vue dynamic form.

Any recommendations for integration tests to add would be appreciated, if needed.

I've left most of the code regarding the complete status as is for now, but there may be a number of places where it has become redundant.

## Type of version change
 Patches 

## Checklist


src/app/static/src/app/components/ResetConfirmation.vue
```diff
@@ -87,6 +89,13 @@
                     this.newVersion();
                 }
             },
+            changesToShow: function (){
```
Nicolas-Dolan, line 92: The purpose of this function is to remove the calibrate model step from the list of subsequent steps with changes when edits require confirmation. When the user attempts to rerun the model, it will show the model run step as a step with changes that would be lost.

src/app/static/src/app/store/modelOptions/modelOptions.ts
```diff
@@ -33,6 +33,9 @@ export const initialModelOptionsState = (): ModelOptionsState => {
 export const modelOptionsGetters = {
     complete: (state: ModelOptionsState) => {
         return state.valid
+    },
+    hasChanges: (state: ModelOptionsState) => {
+        return JSON.stringify(state.optionsFormMeta) !== JSON.stringify(initialModelOptionsState().optionsFormMeta)
```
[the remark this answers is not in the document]
EmmaLRussell: Confirmed that this is a bug - if you go forward to options then back to step 2 without changing any options, and delete a file, the dialog is triggered.

EmmaLRussell (changes requested):
It's not exactly a 'fatal flaw', but I think there's two issues with the approach:

1. Current implementation will not list genuine changes to Calibrate step (I had forgotten about the step listing in the dialog, sorry).
2. The ugliness about 'later step' actually including 'this step' for modelRun has currently been forced out to the edit dialog component, and that feels like the wrong place for it. See suggestion about acknowledging that modelRun is a special case in the editsRequireConfirmation method instead. 

Other than that, works really nicely! Except I think there is a bug with the model options' hasChanges (see below). 

We can probably do without laterCompleteSteps now?

Happy to chat about this if you'd like, and would understand if you can't face making all these changes in this branch. I think this definitely an improvement on what's there, and actually the calibrate step issue is moot until Calibration is restored...

carbowizzy (changes requested):
Works well, just minor changes and formatting.

1. Changes to same file causes dialog box to open, it would be nice for re-uploading of the same file to behave as though no changes.

2. There should be a dialog box informing users that going back to change the calibration will discard map filtered selection.

3. The message on dialog box, these steps will automatically be saved in a version as “{version number}”. You will be able to reload this version from the Projects page. Should include the version number. I think users need to know the version they are saving.

4. if calibration is running and a user re-runs model, steppers' subsequent behaviour doesn't get simulated. Ideally calibration should be stopped and set to incomplete before refitting a model.

5. Dialog box should appear when users emit an event, for example, there are too many dialog boxes when editing model options. Each field triggers a dialog box. Perhaps on change event will be ideal instead of on click event on the text field.

Nicolas-Dolan:
@EmmaLRussell I've implemented all of your requested changes but I have two questions about the stepper. 1) Should run model fit cause all subsequent steps (namely calibrate, review output) to be reset / marked as incomplete? 2) Similarly, should edits to the calibrate model cause subsequent steps to be reset / incomplete? Currently, there is no reset confirmation in the calibrate model. I've left these as they are for now and have removed warnings that the calibrate model and review outputs will have their changes wiped if the fit model is run.

@carbowizzy I've implemented your second point but the others I think are features that can go into separate tickets. I wasn't able to replicate the bug you mentioned on point 5 (I wasn't able to trigger multiple dialog boxes at once). There is a bug in modelOptions where interacting with the tree select options does not trigger the reset confirmation; after doing some research, I think this needs to be addressed in the vue dynamic form.

src/app/static/src/app/store/modelOutput/mutations.ts
```diff
@@ -3,12 +3,34 @@ import {ModelOutputState} from "./modelOutput";
...
+        const {selectedTab, selections} = state
+        // As the barchart does not update on render, any updates from the bar tab are true updates
+        if (selectedTab === 'bar'){
+            state.changes = true;
+        } else {
+            // As the bubble and map update on render, this stores their initial updates in state
+            // and then references new updates against those initially logged inorder to detect
+            // changes made by the user
+            const originKey = Object.keys(payload)[0]
+            const originTab = originKey + '_' + selectedTab
```
EmmaLRussell, line 26: If we keep this bit of code: Can you explain the significance of originKey please, and the need for the conditional on l.28? Not really sure what the need for it is. If you need to clone the object in some cases (?), would it be easier/safer just to do that in all cases, and cut down the complexity of this code? 

Isn't the payload the full dynamic form config here with hierarchical data? So unlikely to be a string or number? Should it be doing a deepClone? (I think we have a util method for that somewhere.)

EmmaLRussell (commented):
As per discussion with Rob, probably yes to wiping and marking incomplete Calibration and Review Output step changes on re-run model fit - but don't need to include in warning dialog, or keep track of hasChanges for these steps - but check with Jeff.

Other than that, code looks good in general, not completely sure what's going on in the Update mutation.

EmmaLRussell (commented):
See suggested changes in  https://github.com/mrc-ide/hint/pull/444/

EmmaLRussell (approved):

PR #429, opened 2020-12-04
mrc-2026 Bug: Cancel button on share project dialog is broken

carbowizzy:
## Description

Bug: Cancel button on share project dialog is broken

To reproduce do the following: 
Open share dialog
Click cancel
Open share again and box for entering email has disappeared

## Type of version change

Patches 

## Checklist


EmmaLRussell (changes requested):
There still seems to be a bit of an issue here - might be a slightly different bug. If you fill in an invalid address (or maybe a valid one too, haven't tested that), and hit Cancel, without pressing Enter, the dialog remains open, although it adds an extra input. You need to hit Cancel a second time to actually close the dialog.

carbowizzy:
> There still seems to be a bit of an issue here - might be a slightly different bug. If you fill in an invalid address (or maybe a valid one too, haven't tested that), and hit Cancel, without pressing Enter, the dialog remains open, although it adds an extra input. You need to hit Cancel a second time to actually close the dialog.

Fixed this issue now.

EmmaLRussell (approved):
Nice :+1:

PR #431, opened 2020-12-07
Avoid opening links with js

EmmaLRussell:
## Description

Our previous workaround for links in dropdown menu items not working was to open the links using javascript. However, I found that this caused the default Firefox settings to regard this behaviour as a pop-up and not to open it by default, warning the user instead. (Chrome didn't mind though!)

This didn't seem like an ideal user experience, so I dug into the cause of the links not working and found that it was because we've implemented DropDown so as to close (hide) the component on blur. The upshot of this seemed to be that the blur event triggered by the link click hid the component containing the link, and actually stopped the click from causing the link to be followed. 

I've put in a workaround in the form of an optional 'delay' property on DropDown. If this is true, the close behaviour is delayed by 1 second, which seems to be long enough for the link to be followed. I tried just making this asynchronous, with no particular delay, but the delay did seem to be required. It's not terribly elegant, but since the link opens a new tab, the user is not aware of the delay to closing the drop down. 

## Type of version change
_Delete as appropriate_

Patches

## Checklist


carbowizzy (approved):
Looks greats !

PR #432, opened 2020-12-07
Mrc 2033 Bug: hint cannot recover model run when hintr is torn down

carbowizzy:
## Description

## Type of version change
_Delete as appropriate_

Patches

## Checklist


src/app/static/src/app/store/modelRun/mutations.ts
```diff
@@ -51,8 +51,7 @@ export const mutations: MutationTree<ModelRunState> = {
     },
 
     [ModelRunMutation.RunStatusError](state: ModelRunState, action: PayloadWithType<Error>) {
-        state.errors.push(action.payload);
-        if (state.errors.length >= maxPollErrors) {
+        if (state.pollingCounter >= maxPollErrors) {
```
EmmaLRussell, line 54: I think we also need to reset pollingCounter as well don't we, for next time we want to run. Should do that in stopPolling, so it happens for all cases.

src/app/static/src/app/store/modelRun/mutations.ts
```diff
@@ -61,6 +60,7 @@ export const mutations: MutationTree<ModelRunState> = {
                 detail: null
             });
         }
+        state.pollingCounter += 1
```
EmmaLRussell, line 63: Should be an `else` here.

src/app/static/src/tests/modelRun/mutations.test.ts
```diff
@@ -98,12 +91,11 @@ describe("Model run mutations", () => {
...
             status: {done: false} as ModelStatusResponse,
-            modelRunId: "123"
+            modelRunId: "123",
+            pollingCounter: 9
         });
-
         mutations.RunStatusError(testState, {payload: "1"});
-
-        expect(testState.errors).toStrictEqual(["1"]);
+        expect(testState.errors).toStrictEqual([]);
```
EmmaLRussell, line 98: This test should be renamed - no longer adds errors if if max not exceeded.

EmmaLRussell (changes requested):

src/app/static/src/app/store/modelRun/mutations.ts
carbowizzy: There's no need for else if the counter is above.

src/app/static/src/app/store/modelRun/mutations.ts
```diff
@@ -51,6 +51,7 @@ export const mutations: MutationTree<ModelRunState> = {
     },
 
     [ModelRunMutation.RunStatusError](state: ModelRunState, action: PayloadWithType<Error>) {
+        state.pollingCounter += 1
```
EmmaLRussell, line 54: Of course, incrementing the poll here means you'll hit the limit one iteration earlier. I guess that's ok at this sort of scale.

src/app/static/src/tests/modelRun/mutations.test.ts
```diff
@@ -66,7 +66,7 @@ describe("Model run mutations", () => {
         expect(testState.errors).toStrictEqual(["Test Error"]);
     });
 
-    it("run status error adds error and stops polling if max errors exceeded", () => {
+    it("run status error and stops polling if maxPollErrors exceeded", () => {
```
EmmaLRussell, line 69: This one didn't need to be renamed as the error is added!

EmmaLRussell (approved):

PR #436, opened 2020-12-09
mrc-2021 make share project ok button work properly

carbowizzy:
## Description

This PR reviews how ok button interacts with email validation, maybe validation could be triggered on clicking OK? Main thing to fix is OK button should be enabled and it is confusing when OK button is disabled

The validation will lose its asynchronous behaviour if validation is implemented upon clicking ok button hence I adopted the approach of initialising blur event when mouse is out of the field, I notice this approach was initially adopted for on key enter as well. This way we can still maintain validation happening on each fields accordingly rather than validating upon clicking ok button which will show (n) number of messages at once.

## Type of version change

 Minor 

## Checklist

### DIFF
--- a/src/app/static/src/app/components/projects/ShareProject.vue
+++ b/src/app/static/src/app/components/projects/ShareProject.vue
@@ -1,9 +1,9 @@
 <template>
     <div>
-        <button class="btn btn-sm btn-red-icons" 
-        v-tooltip= "tooltipShare"
-        @click="shareProject">
-        <share-2-icon size="20"></share-2-icon>
+        <button class="btn btn-sm btn-red-icons"
+                v-tooltip="tooltipShare"
+                @click="shareProject">
+            <share-2-icon size="20"></share-2-icon>
         </button>
         <modal :open="open">
             <h4 v-translate="'shareProject'"></h4>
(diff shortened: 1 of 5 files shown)


src/app/static/src/app/components/projects/ShareProject.vue
```diff
@@ -17,6 +17,7 @@
                                class="form-control"
                                :class="{'is-invalid': email.valid === false}"
                                @blur="() => addEmail(email, index)"
+                               @mouseout="$event.target.blur()"
```
Nicolas-Dolan, line 20: This is a good solution. One potential downside is the user might get confused about why they are no longer selecting the input if the mouse moves out of it for whatever reason, which isn't typical behaviour when browsing the web. I still think this is the best solution proposed so far though.

Nicolas-Dolan (approved):
Looks good and works well. Something to perhaps bear in mind in a future ticket is that while I was testing it I noticed that the validation doesn't prevent you from sharing projects with yourself, resulting in two projects with the same name. It also doesn't stop you from putting in the same email more than once so you could send multiple copies to the same person by accident.

EmmaLRussell (changes requested):
This seems like a good approach, works nicely.

Just need to add a test for the component changes.

I'll add tickets for the two validation issues @Nicolas-Dolan mentions - shouldn't be able to share with yourself or with the same email twice.

src/app/static/src/app/components/projects/ShareProject.vue
carbowizzy: I will create a ticket to look into this. thanks...

EmmaLRussell (approved):
Looks good - I've fixed the version conflict. Can be merged when the build passes.

PR #437, opened 2020-12-14
mrc-2028 Add a map reset button

Nicolas-Dolan:
## Description

Adds a button to bubbleplot and choropleth that updates the bounds of the map, resetting the zoom and position of the view to default.

image
image

Any comments on style/layout would be appreciated. Alternatively, I thought the button could be an icon button directly below the zoom (+/-) buttons, with 'Reset view' as a tooltip.

## Type of version change
Minor

## Checklist


EmmaLRussell (commented):
This is great, works well. Yes, I agree, I think it would look better below or to the right of the zoom button, with an icon to keep it tidy (the refresh icon maybe?).

Nicolas-Dolan:
Updated button to look like this:
image

src/app/static/src/tests/components/plots/choropleth/choropleth.test.ts
```diff
@@ -388,8 +389,12 @@ describe("Choropleth component", () => {
             fitBounds: mockMapFitBounds
         };
 
-        const button = wrapper.find(LControl).find('button')
-        expectTranslated(button, 'Reset view', 'Réinitialiser la vue', store);
+        const button = wrapper.find(LControl).find('div').find('a')
+        expect(button.attributes("aria-label")).toBe("Reset view");
```
EmmaLRussell, line 393: You should be able to do this with `expectTranslated` - it optionally takes an attribute name param if you want to check an attribute rather than `text()`

EmmaLRussell (commented):
This looks much better, works well. One small issue is that, if you're scrolled down a little to get the whole map in view, clicking on the reset button punts you back up to the top of the page - it's because it follows the link to "#". So you could make a `resetView` method or similar which first calls `event.preventDefault()` then calls updateBounds. 

But actually, rather than having to add this in both the Choro and Bubble plot, maybe it would be cleaner to make a ResetMap component which emits a `resetMap` event in the button's click handler, having first called `preventDefault()`. Add this to both the Choro and Bubble, and set the event handler to `updateBounds` in each.

Nicolas-Dolan:
> This looks much better, works well. One small issue is that, if you're scrolled down a little to get the whole map in view, clicking on the reset button punts you back up to the top of the page - it's because it follows the link to "#". So you could make a `resetView` method or similar which first calls `event.preventDefault()` then calls updateBounds.

Ah, I hadn't noticed that as the whole map is always visible on my screen.

EmmaLRussell (approved):
Nice. I've just updated the version conflict and merged from master, will merge on build complete.

PR #441, opened 2020-12-17
mrc-2060 Regenerate front end types

EmmaLRussell:
## Description

Regen front end types for async calibration. The output barchart looks odd in this branch - I think this is probably because we're still using the old calibrate endpoint, and perhaps it has lost the plotting metadata from its result. Will check this is resolved in the next branch which will use the new endpoints. 

This should the console errors you were seeing in the last branch.  

## Type of version change
_Delete as appropriate_

None

## Checklist

### DIFF
--- a/src/app/static/src/app/components/plots/table/Table.vue
+++ b/src/app/static/src/app/components/plots/table/Table.vue
@@ -153,7 +153,8 @@
                 if (this.selections.detail === 0) {
                     return [areaArray[0]]
                 } else if (!this.selections.detail) {
-                    return this.selections.selectedFilterOptions.area.map(row => row.id)
+                    const selectedAreaOptions = this.selections.selectedFilterOptions.area || [];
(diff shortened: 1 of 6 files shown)


src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -153,7 +153,8 @@
                 if (this.selections.detail === 0) {
                     return [areaArray[0]]
                 } else if (!this.selections.detail) {
-                    return this.selections.selectedFilterOptions.area.map(row => row.id)
+                    const selectedAreaOptions = this.selections.selectedFilterOptions.area || [];
```
Nicolas-Dolan, line 156: Smart! I need to remember to use this syntax more often.

Nicolas-Dolan (approved):
Works great on local with no more errors in the console when I run calibrate.

PR #444, opened 2020-12-18
Use resetState to wipe downstream steps from modelRun

EmmaLRussell:
## Description

This really comprises only two small changes from your approach. 
1. Don't need `ResetCalibrate` mutation - `ResetOutputs` covers calibrate state too
2. Rather than a specific action to resetFromFit, make use of the existing `resetState` mechanism to designate clearing the model run result as an update which should wipe downstream state. 

The cause of your trouble with `validate` was my poor advice - I got confused between `validate` action, which is really about loading state correctly (hence the fact that it triggers load errors if something goes wrong) - and `resetState` which is triggered on every mutation and checks for 'update' mutations which should trigger downstream state wipes.  

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/app/components/Stepper.vue
+++ b/src/app/static/src/app/components/Stepper.vue
@@ -157,7 +157,7 @@
         },
         watch: {
             complete: function (){
-                if (this.activeStep === 4 && this.isEnabled(5)){
+                if (this.activeStep === 4 && this.isComplete(4) && this.isEnabled(5)){
                     this.next()
                 }
             },
--- a/src/app/static/src/app/components/Stepper.vue
+++ b/src/app/static/src/app/components/Stepper.vue
@@ -157,7 +157,7 @@
         },
         watch: {
             complete: function (){
-                if (this.activeStep === 4 && this.isComplete(4) && this.isEnabled(5)){
+                if (this.activeStep === 4 && this.isEnabled(5)){
                     this.next()
                 }
             },
--- a/src/app/static/src/app/root.ts
+++ b/src/app/static/src/app/root.ts
@@ -32,6 +32,7 @@ import {ADRSchemas} from "./types";
 import {modelCalibrate, initialModelCalibrateState, ModelCalibrateState} from "./store/modelCalibrate/modelCalibrate";
 import { initialHintrVersionState, hintrVersion, HintrVersionState } from "./store/hintrVersion/hintrVersion";
 import {currentHintVersion} from "./hintVersion";
+import {ModelRunMutation, ModelRunUpdates} from "./store/modelRun/mutations";
 
 export interface TranslatableState {
     language: Language
@@ -100,6 +101,10 @@ const resetState = (store: Store<RootState>) => {
             }
 
             if (type[0] == "modelOptions" && ModelOptionsUpdates.includes(type[1] as ModelOptionsMutation)) {
+                store.commit(RootMutation.ResetOutputs);
+            }
+
+            if (type[0] == "modelRun" && ModelRunUpdates.includes(type[1] as ModelRunMutation)) {
                 store.commit(RootMutation.ResetOutputs);
             }
         }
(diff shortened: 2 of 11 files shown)


src/app/static/src/app/components/Stepper.vue
```diff
@@ -157,7 +157,7 @@
         },
         watch: {
             complete: function (){
-                if (this.activeStep === 4 && this.isComplete(4) && this.isEnabled(5)){
```
Nicolas-Dolan, line 160: This change was to prevent a potential bug in the code which I hadn't considered. The function was being triggered whenever complete state changed in either direction from true to false. This meant that when step 4 was made incomplete (and step 5 was still enabled), the stepper would advance, which is not ideal. The above code made sure that the watch function only cared about the complete state going from false to true and not true to false.

src/app/static/src/app/root.ts
```diff
@@ -102,6 +103,10 @@ const resetState = (store: Store<RootState>) => {
             if (type[0] == "modelOptions" && ModelOptionsUpdates.includes(type[1] as ModelOptionsMutation)) {
                 store.commit(RootMutation.ResetOutputs);
             }
+
+            if (type[0] == "modelRun" && ModelRunUpdates.includes(type[1] as ModelRunMutation)) {
+                store.commit(RootMutation.ResetOutputs);
```
Nicolas-Dolan, line 108: This makes things more straightforward!

src/app/static/src/app/store/modelRun/mutations.ts
```diff
@@ -68,6 +73,10 @@ export const mutations: MutationTree<ModelRunState> = {
         stopPolling(state);
         Object.assign(state, initialModelRunState());
         state.ready = true;
+    },
+
+    [ModelRunMutation.ClearResult](state: ModelRunState) {
+        state.result = null;
```
Nicolas-Dolan, line 79: I'm surprised this is the only state that needs to be reset here.

Nicolas-Dolan (approved):
Thanks, this is a much more straight forward and less specific approach than I took. Works well locally too. There is still a bug in modelOptions where interacting with the tree select options does not trigger the reset confirmation. I think this needs to be addressed in the vue dynamic form in another ticket.

EmmaLRussell:
Do you want to make a ticket for that if there isn't one already?

> Thanks, this is a much more straight forward and less specific approach than I took. Works well locally too. There is still a bug in modelOptions where interacting with the tree select options does not trigger the reset confirmation. I think this needs to be addressed in the vue dynamic form in another ticket.

src/app/static/src/app/store/modelRun/mutations.ts
EmmaLRussell: Should maybe do the plotting selections too, is that what you mean? Possible that filter options selected for previous run won't be relevant any more. Not sure we've been resetting those anyway - will have a look.

src/app/static/src/app/components/Stepper.vue
EmmaLRussell: Ah ok, will reinstate

src/app/static/src/app/store/modelRun/mutations.ts
EmmaLRussell: I think we're good with plotting selections actually - there's some code to update for the barchart specifically, and I think the Choropleth and Bubble chart will work around selected filter options which are no longer valid.

Nicolas-Dolan: OK, thanks for checking. I wasn't sure if other state that got updated by the modelrun would need to be reset also or if that was occurring automatically downstream (if that makes sense).

EmmaLRussell: Yeah, you're right, the ResetOutputs mutation in root gets triggered by this mutation, which wipes everything upstream.

PR #446, opened 2020-12-22
mrc-2062 Show progress label while calibrating

EmmaLRussell:
## Description

Shows new progress information from calibrate status response. Status returns an array of strings or ProgressPhases. For simple first version, we just show first of these next to progress spinner - we may replace this with something more elaborate later. 

Not updating version, as already updated in mrc-2061

## Type of version change
_Delete as appropriate_
None

## Checklist


src/app/static/src/app/components/modelCalibrate/ModelCalibrate.vue
```diff
@@ -85,7 +87,15 @@
             ...mapStateProps<ModelCalibrateState, keyof Computed>(namespace, {
                 loading: s => s.fetching,
                 calibrating: s => s.calibrating,
-                error: s => s.error
+                error: s => s.error,
+                progressMessage: s => {
+                    if (s.status && s.status.progress && s.status.progress.length > 0) {
```
EmmaLRussell, line 92: `status.progress` is an array but in practice it currently only ever contains one item

src/app/static/src/tests/components/modelCalibrate/modelCalibrate.test.ts
```diff
@@ -122,6 +122,18 @@ describe("Model calibrate component", () => {
         expect(wrapper.find("button").classes()).not.toContain("btn-submit");
     });
 
+    it("renders string progress message", () => {
```
Nicolas-Dolan, line 125: Do you need to add a test for when it should be rendering 'Calibrating...'

Nicolas-Dolan (changes requested):
So the calibrating step works well and the messages show up nicely. I then tried to go back and rerun the model and then recalibrate and that's where I started having issues. The model reran fine but didn't reset the calibrate/review output steps, suggesting that this hasn't been merged with hint 1.10.1. When I merged in the latest version of master I got conflicts, notably in modelCalibrate actions (I noticed that mrc-2058 had already been merged with master so perhaps that has something to do with it). Also when I reran calibrate the second time it timed out and failed.

EmmaLRussell:
I was intending to sort out all the merging with latest master from the main feature branch, mrc-2058.  I don't think mrc-2058 has the latest master changes, and I'm not that surprised merging master directly into mrc-2062 causes problems. 

I'll try merging through from mrc-2058 to pick up latest changes. 

Odd about second calibrate failing, I haven't seen this behaviour, though possibly I haven't been re-fitting, just re-calibrating. Will check it out. 

> So the calibrating step works well and the messages show up nicely. I then tried to go back and rerun the model and then recalibrate and that's where I started having issues. The model reran fine but didn't reset the calibrate/review output steps, suggesting that this hasn't been merged with hint 1.10.1. When I merged in the latest version of master I got conflicts, notably in modelCalibrate actions (I noticed that mrc-2058 had already been merged with master so perhaps that has something to do with it). Also when I reran calibrate the second time it timed out and failed.

src/app/static/src/tests/components/modelCalibrate/modelCalibrate.test.ts
EmmaLRussell: THere's actually a test already, but I'll rename it make it clear it only applies when no progress information is available

EmmaLRussell:
For reasons which aren't clear to me, the merge from master to mrc-2058 was a disaster, and wiped out a bunch of branch changes (:shrug:). This should now be fixed and merged into this branch, so all expected master behaviour should now show up here. I wasn't able to reproduce the timeout on second calibrate that you saw, so see if you can still break it!

Nicolas-Dolan (approved):
This works well. I wasn't able to break it :)

PR #448, opened 2021-01-05
mrc-2074 More useful "No data" message on map

Nicolas-Dolan:
## Description

'No data' message on map to direct user what to do

## Type of version change
Patches 

## Checklist


EmmaLRussell (approved):
Looks good, just need to fix version conflict in NEWS.md and hintVersion.ts, then go ahead and merge.

PR #451, opened 2021-01-06
mrc-2045 Share project validation shouldn't allow user to share with self or the same email twice in one action

Nicolas-Dolan:
## Description

Also provides an alternate warning message when the user enters their own email into a share project input.

**Update:** now combined with updates for mrc-2046 (Share project dialog validation shouldn't allow share with same email twice in one action)

## Type of version change
Patches

## Checklist


src/app/static/src/app/components/projects/ShareProject.vue
```diff
@@ -109,33 +112,48 @@
...
-                        this.emailsToShareWith.push({
-                            value: "",
-                            valid: null
-                        });
-                    }
-                    this.userExists(e.value)
-                        .then((result: boolean) => {
-                            this.emailsToShareWith[index].valid = result;
+            addEmail() {
+                this.emailsToShareWith.map((email: EmailToShareWith, index: number) => {
```
Nicolas-Dolan, line 124: Because the validation of duplicate emails is dependent on changes to other emails, validation must be run over the entire list every time the blur event fires

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -12,6 +12,7 @@ export interface Translations {
     badUsernamePassword: string,
     bar: string,
     back: string,
+    blank: string,
```
Nicolas-Dolan, line 15: This might not be necessary, but v-translate was complaining when it received an empty string.

Nicolas-Dolan:
Update: now combined with updates for mrc-2046 (Share project dialog validation shouldn't allow share with same email twice in one action)

EmmaLRussell (commented):
Works really well - just got some suggestions to improve readability

src/app/static/src/app/components/projects/ShareProject.vue
EmmaLRussell: Huh, weirdly didn't include my review comment.. 
This works, but think readability could be improved. Also don't need to consider adding control in the map function, if reinstate method params. Something like this? 
```addEmail(e: EmailToShareWith, index: number) {
                if (e.value && index == this.emailsToShareWith.length - 1) {
                    this.emailsToShareWith.push({
                        value: "",
                        valid: null,
                        validationMessage: "blank"
                    });
                }
                this.emailsToShareWith.map(async (email: EmailToShareWith, index: number) => {
                    if (email.value) {
                        let invalidMsg = null;

                        if (email.value == currentUser) {
                            invalidMsg = "projectsNoSelfShare";
                        } else if (this.emailsToShareWith.filter(val => val.value === email.value).length > 1) {
                            invalidMsg = "duplicateEmails";
                        } else {
                            await this.userExists(email.value)
                                .then((result: boolean) => {
                                    if (!result) {
                                        invalidMsg = "emailNotRegistered"
                                    }}
                                );
                        }

                        this.emailsToShareWith[index].validationMessage = invalidMsg || "";
                        this.emailsToShareWith[index].valid = invalidMsg === null;

                    } else {
                        email.valid = null;
                    }
                });
                this.showValidationMessage = this.invalidEmails;  //Could this be a computed property?
            }
```

EmmaLRussell:
Trying this for third time: too long?
```addEmail(e: EmailToShareWith, index: number) {
                if (e.value && index == this.emailsToShareWith.length - 1) {
                    this.emailsToShareWith.push({
                        value: "",
                        valid: null,
                        validationMessage: "blank"
                    });
                }
                this.emailsToShareWith.map(async (email: EmailToShareWith, index: number) => {
                    if (email.value) {
                        let invalidMsg = null;

                        if (email.value == currentUser) {
                            invalidMsg = "projectsNoSelfShare";
                        } else if (this.emailsToShareWith.filter(val => val.value === email.value).length > 1) {
                            invalidMsg = "duplicateEmails";
                        } else {
                            await this.userExists(email.value)
                                .then((result: boolean) => {
                                    if (!result) {
                                        invalidMsg = "emailNotRegistered"
                                    }}
                                );
                        }

                        this.emailsToShareWith[index].validationMessage = invalidMsg || "";
                        this.emailsToShareWith[index].valid = invalidMsg === null;

                    } else {
                        email.valid = null;
                    }
                });
                this.showValidationMessage = this.invalidEmails;  //Could this be a computed property?
            }

**Comment by Nicolas-Dolan** [comment]
> Possible readability improvement? Logic quite repetitive and a bit confusing before. And don't need to consider adding control in map function if reinstate method params
> 
@EmmaLRussell I've not been able to get this code to work 100% correctly. I think the issue is that only putting `this.showValidationMessage` at the end means the order of operations can be messed up by the async part of the code, resulting in it temporarily failing while it waits for the result to return, so what should be a valid response returns as negative. This in turn upsets the v-translate method.

**Comment by EmmaLRussell** [comment]
@Nicolas-Dolan Could you try it with these changes? I did end up making `showValidationMessage` a computed property, but it needs to be subtly different from `invalidEmails` - see comment in code. I think the bug where showValidationMessage was hanging over from previous display of dialog was probably already in master.

You're right about the async method not playing nicely I think - although the internal logic should all be synchronous with itself, I think it maybe affects the way that other computed properties get updated, hence `invalidEmails` not giving value expected. I think same issue was affecting tests - putting in a `setTimeout` to replace `Vue.nextTick()` seemed to fix those.

File: src/app/static/src/tests/components/projects/shareProject.test.ts [hunk 3]
```diff
@@ -296,10 +371,10 @@ describe("ShareProject", () => {
         input.setValue("testing");
         input.trigger("blur");
 
-        await Vue.nextTick();
-
-        wrapper.find(Modal).find("button").trigger("click");
-        expect(cloneProject.mock.calls[0][1]).toEqual({projectId: 1, emails: ["testing"]});
+        setTimeout(() => {
```
Nicolas-Dolan, line 374: Interesting, I'm not familiar on the nuance of how `Vue.nextTick()` and `setTimeout()` differ.

src/app/static/src/app/components/projects/ShareProject.vue
```diff
@@ -170,11 +182,16 @@
                 return i18next.t('shareProjectInstructions', {project: this.project.name, lng: this.currentLanguage});
             },
             invalidEmails() {
+                //Invalid state until all emails evaluated as valid = true (may not have blurred yet)...
                 return this.emailsToShareWith.filter(e => e.value && !e.valid).length > 0
             },
+            showValidationMessage() {
+                //...however only show error message if any confirmed to be valid = false
+                return this.emailsToShareWith.filter(e => e.value && e.valid === false).length > 0
```
Nicolas-Dolan, line 190: So presumably this excludes valid values that are null.

Nicolas-Dolan (commented):
I couldn't find any of the previous bugs when I checked this out locally so should be fine to merge.

src/app/static/src/app/components/projects/ShareProject.vue
EmmaLRussell: Exactly - they're the ones that haven't been validated yet

PR #452, opened 2021-01-08
mrc-2105 Get model results from new calibrate endpoint on load

EmmaLRussell:
## Description

This fixes bug where refreshing page (and loading project) would not fetch model outputs. This is because the load action on page load was using the old modelRun/getResult action, which just provides the pre-calibrate 'result'. Now using the post-calibrate result action modelCalibrate/getResult. 

Also needed to update the action itself to check if status indicated any results available (since this will also be called for projects where calibrate has not yet run), and to set Ready on the modelRun state when complete. (We should probably change the logic to use Ready on modelCalibrate state here, but this will require further changes, and may be better done at same time as mrc-2064.

## Type of version change
_Delete as appropriate_

Patches

## Checklist

### DIFF
--- a/src/app/static/src/app/index.ts
+++ b/src/app/static/src/app/index.ts
@@ -30,7 +30,7 @@ export const app = new Vue({
     methods: {
         ...mapActions({loadBaseline: 'baseline/getBaselineData'}),
         ...mapActions({loadSurveyAndProgram: 'surveyAndProgram/getSurveyAndProgramData'}),
-        ...mapActions({loadModelRun: 'modelRun/getResult'}),
+        ...mapActions({loadModelRun: 'modelCalibrate/getResult'}),
         ...mapActions({getADRSchemas: 'getADRSchemas'}),
         ...mapActions({getCurrentProject: 'projects/getCurrentProject'}),
     },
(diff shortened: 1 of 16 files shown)


src/app/static/src/app/index.ts
```diff
@@ -30,7 +30,7 @@ export const app = new Vue({
     methods: {
         ...mapActions({loadBaseline: 'baseline/getBaselineData'}),
         ...mapActions({loadSurveyAndProgram: 'surveyAndProgram/getSurveyAndProgramData'}),
-        ...mapActions({loadModelRun: 'modelRun/getResult'}),
+        ...mapActions({loadModelRun: 'modelCalibrate/getResult'}),
```
Nicolas-Dolan, line 33: So I think you alluded to this in your comment, but what is supposed to happen when the user fits the model but does not calibrate and then refreshes? When I tested that scenario myself, I got a load error and the model fit was wiped.

Nicolas-Dolan (commented):
Does what it's supposed to do when I tested it but I have one question below. I think you have this in hand so otherwise ok to merge.

src/app/static/src/app/index.ts
EmmaLRussell: Oh good catch, wasn't expecting an error there..

EmmaLRussell: Yeah, the above doesn't work because it does need the modelRun result to be available for validate logic where it expects the modelRun step to be complete. So I've ended up making modelCalibrate state a readyState after all, with resultant boilerplate, and calling both these actions on page load..

Nicolas-Dolan (approved):
I wasn't able to find any bugs when I ran it locally and refreshed at various points.

PR #455, opened 2021-01-14
mrc-2111 Fix integration test failures on axios upgrade

Nicolas-Dolan:
## Description

## Type of version change
Patches

## Checklist


Nicolas-Dolan:
Mark has a partial fix https://github.com/mrc-ide/hint/compare/mwoodbri
Based on this https://github.com/axios/axios/issues/2654

mwoodbri:
The remaining test now passes, so this PR reinstates the behaviour we were relying on from Axios <0.19.1 while resolving the Dependabot alert. As mentioned above, it may be that the tests should pass using the new (xhr) adapter, but I think that's a separate discussion.

Nicolas-Dolan:
> The remaining test now passes, so this PR reinstates the behaviour we were relying on from Axios <0.19.1 while resolving the Dependabot alert. As mentioned above, it may be that the tests should pass using the new (xhr) adapter, but I think that's a separate discussion.

Thanks for your work here @mwoodbri. This seems to have solved the main issue of this ticket. I'll have it reviewed by a third party and then it can be merged.

EmmaLRussell (approved):

PR #457, opened 2021-01-15
mrc-2115 - Hide project history column(shared by) when no data

carbowizzy:
## Description

This PR hides shared by column header and rows if a project hasn't been shared by someone. 

## Type of version change

Minor 

## Checklist


Nicolas-Dolan (changes requested):
Thanks for your work here. I think perhaps the columns are getting too crowded and I don't think we can fit another containing an email address. There's also the issue of the version columns becoming misaligned (which I'm not sure if that predates this pr or not). Anyway perhaps another solution would be to have 'Shared by: xxxxx' appear under the project name if it has been shared. @EmmaLRussell what do you think? Sorry Lekan if this creates extra work for you.

image

EmmaLRussell:
Good idea @Nicolas-Dolan about putting 'shared by' underneath the project, I think that could work too, although might be fiddly to get the vertical spacing right to not look messy with a mix of shared/unshared projects. I don't think the extra column looks *too* bad, particularly since we're going to hide it for users who have no projects shared with them (probably most users other than admins), though I agree it is getting pretty busy. Happy with either approach, and to await user feedback, 

> 
> Thanks for your work here. I think perhaps the columns are getting too crowded and I don't think we can fit another containing an email address. There's also the issue of the version columns becoming misaligned (which I'm not sure if that predates this pr or not). Anyway perhaps another solution would be to have 'Shared by: xxxxx' appear under the project name if it has been shared. @EmmaLRussell what do you think? Sorry Lekan if this creates extra work for you.
> 
> image

EmmaLRussell (commented):
Actually, one advantage to putting the sharedBy underneath as @Nicolas-Dolan suggests is that currently the email address does (usually) overflow the container, which means it will be asymmetrically hidden in narrow windows. So maybe that is the best approach to go with if you can bear it. In which case, it would be probably be better to abandon this PR and make a new ticket to reimplement shared by display .

NEWS.md
```diff
@@ -1,6 +1,16 @@
+
 # hint 1.14.0
 * Hide project history column(shared by) when no data
```
EmmaLRussell, line 3: This comment should be updated to more clearly reflect the actual changes now being made.

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -393,8 +384,8 @@
                 "deleteVersion"
             ),
 
-            tooltipContent(tooltipValue: string) {
-                return i18next.t(tooltipValue, {
+            getTranslatedValue(key: string) {
```
EmmaLRussell, line 387: Nice rename :+1:

EmmaLRussell (commented):
Thanks for this - looks good to me, I think the small text is pretty neat. Could you just update the version comment, then go ahead and merge.

PR #459, opened 2021-01-18
Update translations with text from UNAIDS

r-ash:
## Description

Updates translations with text from UNAIDS translator - see task for file from them

## Type of version change
_Delete as appropriate_

Patches

## Checklist


EmmaLRussell (commented):
Actually, this still needs a bit of work as the longer strings are overflowing the colour scale panel

PR #460, opened 2021-01-20
Mrc 2124 validate Survey & Program file on change Baseline file

carbowizzy:
## Description

This PR replaces removal of surveyAndProgram data upon re-uploading of Baseline data. each time a user uploads baseline data instead of removing survey and programs data, validation is requested on each of the survey and program data. If success, the model gets reloaded and when failed a message is displayed below each file upload. 

Please note that when a user delete any baseline data then the survey and program data reset itself, while uploading and re-uploading validate itself.

## Type of version change

Minor

## Checklist


EmmaLRussell (approved):
This looks good - and not much code required!

However, I wasn't able to test the case where the new baseline file is not compatible with the SAP files (but is compatible with the other baselines), and hence the SAP files should be be removed (I think they should be removed and not just marked as invalid). Were you? I tried sabotaging the population file by deleting a lot of rows but that didn't seem to make any difference. If you have some example file which tests this case, could you attach it here, or ask Jeff for an example. 

Also, I think that the Confirm Edit dialog should not be triggered from baseline changes for Step 2 only, and should not be shown in the list of steps where there are other triggering steps (e.g. if model fit has run). This is because, in the valid case, the SAP files will be retained. I think this will need changing in the `changesToRelevantSteps` getter in stepper module.

EmmaLRussell (changes requested):
Sorry i meant that to be a request changes!

EmmaLRussell:
Hmm, thinking about this again, I don't think the behaviour should be as I described. If we're going to no longer warn about potentially wiping the data, we probably shouldn't remove the files silently if they fail validation, but should retain them with an error message indicating what the issue was. That makes more sense, and will be easier to implement. Apologies for the indecision.

carbowizzy:
> Hmm, thinking about this again, I don't think the behaviour should be as I described above. If we're going to no longer warn about potentially wiping the data, we probably shouldn't remove the files silently if they fail validation, but should retain them with an error message indicating what the issue was. That makes more sense, and will be easier to implement. Apologies for the indecision.

Step 2 is no longer being checked if it has changes. The confirmation modal will not trigger from baseline.

EmmaLRussell (commented):
This seems like the right behaviour to me. Regarding my previous question, I think I've found a way to test it manually - if you initially import all the Antarctica data from ADR then replace all the baseline files with the old Malawi files in the testdata folder, you get a set of baseline data which is consistent with itself but does not match the ADR SAP files, so these now show errors (correctly) when you advance to Step 2. 

One behaviour which I think isn't right - in this situation, I found that the Choropleth was visible, but showing not data. However, when you upload files manually, if no SAP file is valid, the choropleth is not displayed. Could we make this the behaviour when SAP files are made invalid by baseline changes too?

EmmaLRussell (commented):
Let's leave this til the next sprint, as I think this approach might have unwelcome side effects. 

I think I understand the need to reset selectedDataType on error in this case of revalidating - but I don't really like that happening in the mutation, as (I think) this would mean wiping a valid selectedDataType when you uploaded an invalid file (e.g. if you're sitting with the survey tab map happily showing then upload an invalid ART file, you wouldn't want your survey to disappear). 

Instead of setting selectedDataType to null in the error mutation, could you do it in validateSurveyAndProgramData, then set selectedDataType on successful upload? Ideally it would retain the previously selected data type if it was successfully validated, but would choose any other valid type if it wasn't (if any).

Sorry this is so fiddly!

Also - there's code changes here without tests, so remember to add some next sprint...

src/app/static/src/app/store/surveyAndProgram/actions.ts
```diff
@@ -188,23 +187,39 @@ export const actions: ActionTree<SurveyAndProgramState, RootState> & SurveyAndPr
...
             [
                 api<SurveyAndProgramMutation, SurveyAndProgramMutation>(context)
                     .withError(SurveyAndProgramMutation.SurveyError)
                     .withSuccess(SurveyAndProgramMutation.SurveyUpdated)
                     .freezeResponse()
-                    .get<SurveyResponse>("/disease/survey/"),
+                    .get<SurveyResponse>("/disease/survey/")
+                    .then((response) => {
+                    if (response) {
+                        commitSelectedDataTypeUpdated(commit, DataType.Survey);
```
EmmaLRussell, line 200: As per previous comment
>Ideally it would retain the previously selected data type if it was successfully validated, but would choose any other valid type if it wasn't (if any).

So I think rather than setting the value three times you might want to do something like push to an array of successful datatypes as you get  succesful responses. Then after all promises are complete, set selected data type to the previously selected type (if it's in the array) or to the first data type in the array (if there are any) if it's not

EmmaLRussell (changes requested):
Confirmed choropleth issue now fixed :+1: 

Still no tests for the new code?

src/app/static/src/app/store/surveyAndProgram/actions.ts
```diff
@@ -186,8 +187,10 @@ export const actions: ActionTree<SurveyAndProgramState, RootState> & SurveyAndPr
     },
 
     async validateSurveyAndProgramData(context) {
-        const {commit} = context;
-        commit({type: SurveyAndProgramMutation.SelectedDataTypeUpdated, payload: null});
+        const {commit, rootState} = context;
+        const selectedDatType: DataType[] = []
```
EmmaLRussell, line 191:
```suggestion
        const successfulDataTypes: DataType[] = []
```

src/app/static/src/tests/surveyAndProgram/actions.test.ts
```diff
@@ -293,6 +293,33 @@ describe("Survey and programme actions", () => {
 
     });
 
+    it("it validates, commits and marks state ready", async () => {
```
EmmaLRussell, line 296: This test looks great. Would be nice to have test of error responses coming back for some / all of the requests, and that the selected datatype is upated  as expected.

EmmaLRussell (changes requested):

EmmaLRussell (approved):
Great, thanks for this. Go ahead and update the version to fix the conflict and merge.

PR #461, opened 2021-01-21
add uncertainty ranges to choropleth tooltips

carbowizzy:
## Description

Screenshot 2021-01-22 at 10 00 41

## Type of version change

Minor

## Checklist


src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -293,11 +293,20 @@
...
 
+                        console.log(featureIndicators[area_id])
                         const values = featureIndicators[area_id];
                         const value = values && values!.value;
-
+                        const lower_value = values && values!.lower_value;
+                        const upper_value = values && values!.upper_value;
                         const stringVal = (value || value === 0) ? value.toString() : "";
-                        if (stringVal) {
+                        if (stringVal && lower_value > 0) {
```
EmmaLRussell, line 302: Couldn't a lower value in a range legitimately be 0? Would it be safer here to switch on whatever falsy value you actually get for values without ranges (the input data)?

src/app/static/src/app/components/plots/choropleth/utils.ts
```diff
@@ -12,10 +12,12 @@ export const getFeatureIndicator = function (data: any[],
 
     const result = {} as IndicatorValuesDict;
     iterateDataValues(data, [indicatorMeta], selectedAreaIds, filters, selectedFilterValues,
-        (areaId: string, indicatorMeta: ChoroplethIndicatorMetadata, value: number) => {
+        (areaId: string, indicatorMeta: ChoroplethIndicatorMetadata, value: number, values: any) => {
             result[areaId] = {
                 value: value,
-                color: getColor(value, indicatorMeta, colourRange)
+                color: getColor(value, indicatorMeta, colourRange),
+                lower_value: values['lower']? values['lower']: 0,
```
EmmaLRussell, line 19: Ah you're setting this here - maybe let this be null or undefined to distinguish from real lower 0 value? Could make lower_value and upper_value optional properties in the type and not set them at all if not required.

src/app/static/src/tests/components/plots/table/table.test.ts
```diff
@@ -125,7 +125,9 @@ describe('Table from testdata', () => {
         expect(wrapper.findAll('th').at(2).text()).toBe('Sex (Click to sort Ascending)');
         expect(wrapper.findAll('td').at(2).text()).toBe('Female');
         expect(wrapper.findAll('th').at(3).text()).toBe('HIV prevalence (Click to sort Ascending)');
-        expect(wrapper.findAll('td').at(3).text()).toBe('1.00%');
+        const div = wrapper.findAll('td').at(3).findAll("div")
```
EmmaLRussell, line 128: This was just a missing test? You didn't actually change the table behaviour.

EmmaLRussell (commented):
Great stuff!
Could you check with Jeff if he also wants to see ranges in the bubble plot? I assume he will, but don't think we have a separate ticket for that. Could you either add one, or just add to bubble plot in this branch?

src/app/static/src/tests/components/plots/table/table.test.ts
carbowizzy: test needed to be updated because I added more data to the testdata, which the table uses as well.

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
carbowizzy: Modified to accept 0 as a starting value

carbowizzy: I have also added the ranges to bubble map with help from Jeff

EmmaLRussell (commented):
Works great, thanks for adding that! :+1: 

Could you add a test for the actual tooltip generated for the bubble plot? And for both choro and bubble, would be good to test that it now works as expected with 0s.

I'm finding the test data with all 0s for upper and lower values a bit confusing. Might be nice to make this a bit more realistic?

src/app/static/src/app/components/plots/bubble/BubblePlot.vue
```diff
@@ -393,16 +393,18 @@
                 const sizeIndicatorName = this.indicatorNameLookup[sizeIndicator];
                 const { format, scale, accuracy } = this.colorIndicator!;
                 const { format: formatS, scale: scaleS, accuracy: accuracyS } = this.sizeIndicator!;
+                const stringUpper_value = (upper_value || upper_value === 0) ? upper_value.toString() : "";
```
EmmaLRussell, line 396: Bit confused about this - see my comment in utils method - don't think these values will ever be 0 based on that (though it should be as 0 may be valid!) Also, why is this not needed for lower values, only for upper?

EmmaLRussell (commented):
Let's leave this to the next HINT sprint. I'm a bit unclear about the new logic you've put in on this latest commit, could you explain why it needed to be added?

Don't think you've added this?:
> And for both choro and bubble, would be good to test that it now works as expected with 0s.

src/app/static/src/app/components/plots/bubble/utils.ts
```diff
@@ -38,14 +38,21 @@ export const getFeatureIndicators = function (data: any[],
             if (indicatorMeta.indicator == colorIndicator.indicator) {
                 result[areaId].value = value
                 result[areaId].color = getColor(value, indicatorMeta, colourRange)
-                result[areaId].lower_value = (values['lower'] || values['lower'] === 0) ? values['lower'] : ""
-                result[areaId].upper_value = values['upper']
+                if(values['lower'])
```
EmmaLRussell, line 41: This comment is still unaddressed, and it still seems me that 0s will remain undefined, which we don't want as 0s are valid vald values. Could you please explain your thinking here, and also add a test for tooltips where lower or upperbound is 0? 

Why not just pass these values through without the check, and then do the `value ||  value === 0` checks in the tooltip code?

Sorry if this has got really confusing, let's chat about it if not clear. 

>So this won't include 0s? lower_value or upper_value will remain undefined if data value is 0? I thought we did want to pass those through?

src/app/static/src/app/components/plots/bubble/BubblePlot.vue
```diff
@@ -393,17 +393,21 @@
                 const sizeIndicatorName = this.indicatorNameLookup[sizeIndicator];
                 const { format, scale, accuracy } = this.colorIndicator!;
                 const { format: formatS, scale: scaleS, accuracy: accuracyS } = this.sizeIndicator!;
-                const stringUpper_value = (upper_value || upper_value === 0) ? upper_value.toString() : "";
-                const stringSizeUpper = (sizeUpper || sizeUpper === 0) ? sizeUpper.toString() : "";
-                if(lower_value && sizeLower) {
+
+                const stringLower_value = lower_value ? lower_value.toString() : "";
```
EmmaLRussell, line 397: Again, this is going to cause problems for 0 values. I think the code as you had it before would be fine (if it was for both lower and upper values) if 0s were being returned from the utils methods - which they should be! (but aren't at the moment, I don't think)

src/app/static/src/tests/components/plots/bubble/utils.test.ts
```diff
@@ -220,4 +220,29 @@ describe("Bubble plot utils", () => {
         });
     });
 
+    it("it compute prevalence and plhiv uncertainty where ranges are zeros", () => {
```
EmmaLRussell, line 223: Great, this is now testing that the features come out of the indicators as expected. Could you also add a test that the tooltip is generated correctly?

EmmaLRussell (approved):
Great, thanks :+1:

PR #462, opened 2021-01-29
mrc-2159 Use model calibrate id for downloads, not initial model fit id

EmmaLRussell:
## Description

As reported by Jeff:
It appears that the download buttons on the last page, are not downloading the 'calibrated' version of results.

The outputs on the 'Results' page (calibrated results), do not match the results in the download package, which looks to be the 'uncalibrated' results.

The total PLHIV (all ages, Sep 2023) on the results page is 1,010,000:

If I go to following steps:

'Download results' ->
'Export' button ->
Unzip and open the file 'indicators.csv'
Filter that sheet to 'MWI', 'Y000_999', indicator = "plhiv", "CY2020Q3"

The total in the spreadsheet is 1,070,458.23. Those numbers should match**

This is because the download component was using the modelRunId (used to generate the initial fit) rather than the model calibrateId.
NB for testing - plhiv is being rounded to nearest 100

## Type of version change
_Delete as appropriate_
Patches

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,6 @@
+# hint 1.15.1
+* fix download of wrong results set
+
 # hint 1.15.0
 * add custom event to open confirmation modal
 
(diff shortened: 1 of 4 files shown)


Nicolas-Dolan (approved):
Using a different dataset to Jeff, but getting consistent results between the Review Output and the exported excel:
image
image

PR #464, opened 2021-02-03
mrc-2164 - Bug: Calibration error not returned to user

Nicolas-Dolan:
## Description

Steps:
1. Run a model
2. Pick some calibration options which will error (eg, set 'Adjust to Spectrum number aware of HIV status' stratification to '5-year age group')
3. Click calibrate, UI never reports error and spinner never disappears so this project is bugged out and now can't be used

More info [here](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2164)

I believe the issue was that the error state and error message relied on the modelCalibrate action `getResult` being called, which would never happened as it previously needed `status.success` to be true; a condition which wouldn't occur if there was error and therefore the logic would just get stuck. I've removed that condition from stopping it being called and this has prevented the bug and allowed expected behaviour (error message is displayed in UI and calibrate run is aborted). 

Perhaps there was valid reasoning for the prior conditions, in which case this bug fix may need to be revised. I was also wondering if any additional tests should be added, such as in integration test to mock this this error state scenario.

## Type of version change
Patches

## Checklist


EmmaLRussell (approved):
This seems like the right approach since it seems we do need to fetch the result in order to see the error (I had assumed error would be returned at the status stage). 
This issue: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2157 may need a rethink in light of this. (Confusing error was being displayed on reload after model run failure) The suggested fix (basically the opposite of this change) probably needs a rethink! Wil make a note in that ticket.

PR #465, opened 2021-02-11
mrc-2184 display project name as default when rename is triggered with existin…

carbowizzy:
…g project name

## Description

This PR shows project name as default value when a user clicks rename link and copy project.
Adding focus on input upon modal opening has it's own maintenance ticket here https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2219

## Type of version change
_Delete as appropriate_

Minor

## Checklist

### DIFF
--- a/src/app/static/src/tests/components/projects/projectHistory.test.ts
+++ b/src/app/static/src/tests/components/projects/projectHistory.test.ts
@@ -302,7 +302,7 @@ describe("Project history component", () => {
         }
     };
 
-    it("shows modal when rename project link is clicked kand removes it when cancel is clicked", async () => {
+    it("does show project name as default value when a user clicks rename link", async () => {
         if (switches.renameProject) {
             const wrapper = getWrapper(testProjects);
             const renameLink = wrapper.find("#p-1").findAll(".project-cell").at(5).find("button");
@@ -313,6 +313,29 @@ describe("Project history component", () => {
             const proj1 = modal.find("input")
             const projectName1 = proj1.element as HTMLInputElement
             expect(projectName1.value).toBe("proj1")
+        }
+    });
+
+    it("does show project name as default value when a user clicks copy project", async () => {
+        if (switches.promoteProject && !switches.renameProject) {
+            const wrapper = getWrapper();
+            const copyLink = wrapper.find("#p-1").findAll(".project-cell");
+
+
+            copyLink.at(6).find("button").trigger("click")
+            await Vue.nextTick();
+            const modal = wrapper.findAll(".modal").at(1);
+            const proj1 = modal.find("input")
+            const projectName1 = proj1.element as HTMLInputElement
+            expect(projectName1.value).toBe("proj1")
+
+
+            copyLink.at(5).find("button").trigger("click")
+            await Vue.nextTick();
+            const modalVersion = wrapper.findAll(".modal").at(1);
+            const projVersion = modalVersion.find("input")
+            const projectNameVersion = projVersion.element as HTMLInputElement
+            expect(projectNameVersion.value).toBe("proj1")
         }
     });
 
--- a/src/app/static/src/tests/components/projects/projectHistory.test.ts
+++ b/src/app/static/src/tests/components/projects/projectHistory.test.ts
@@ -464,7 +464,7 @@ describe("Project history component", () => {
     });
 
     it("can use carriage return to invokes promoteVersion action when confirm copy", async () => {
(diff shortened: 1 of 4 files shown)


EmmaLRussell (commented):
The pre-populate words great. :+1: 

In the ticket Jeff also requests that hitting Return should also OK from the dialog, which this doesn't do yet. If we do add that in, should check that the Share dialog has consistent behaviour too. 

It would probably be good to give focus to the input on dialog open too.

src/app/static/src/tests/components/projects/projectHistory.test.ts
```diff
@@ -463,6 +463,32 @@ describe("Project history component", () => {
...
+            const copyLink = wrapper.find("#v-s11").findAll(".version-cell").at(5).find("button");
+            copyLink.trigger("click");
+            await Vue.nextTick();
+
+            const modal = wrapper.findAll(".modal").at(1);
+            const input = modal.find("input");
+            const copyBtn = modal.find(".modal-footer").findAll("button").at(0);
+            input.setValue("newProject");
+            expect(copyBtn.attributes("disabled")).toBe(undefined);
+            await copyBtn.trigger("keyup.enter");
```
EmmaLRussell, line 478: Shouldn't this trigger be on the input?

src/app/static/src/tests/components/projects/projectHistory.test.ts
```diff
@@ -463,6 +463,32 @@ describe("Project history component", () => {
         }
     });
 
+    it("can use carriage return to invokes promoteVersion action when confirm copy", async () => {
+        if (switches.promoteProject && !switches.renameProject) {
```
EmmaLRussell, line 467: Not sure what `!switches.renameProject` is doing here? Those switches are really obsolete now so we shouldn't use them  in new tests

src/app/static/src/tests/components/projects/projectHistory.test.ts
```diff
@@ -534,6 +560,33 @@ describe("Project history component", () => {
         }
     });
 
+    it("can use carriage return to invoke renameProject action", async () => {
+        if (switches.renameProject) {
```
EmmaLRussell, line 564: Shouldn't need this

EmmaLRussell (commented):
Good stuff, couple of comments on the tests. 

As per my previous comments: 
- share dialog has different behaviour and deals with enter key, so that's fine. 
- setting focus on inputs on dialog open?

src/app/static/src/tests/components/projects/projectHistory.test.ts
carbowizzy: I am not totally sure what you meant here. Are you saying the enter key should not be tested ?

src/app/static/src/tests/components/projects/projectHistory.test.ts
carbowizzy: Me too, but without !switches.renameProject the test fails. This is the pattern found in other tests relating to this control.

src/app/static/src/tests/components/projects/projectHistory.test.ts
carbowizzy: Spot on, thanks...

src/app/static/src/tests/components/projects/projectHistory.test.ts
EmmaLRussell: No I'm saying just take out that conditional on the switches

src/app/static/src/tests/components/projects/projectHistory.test.ts
EmmaLRussell: The switches are completely obsolete as both promoteProject and renameProject are completely implemented and always included in all deployments of HINT. If this test only passes with that conditional, it's because the conditional is always false since both those switches are in fact true, so the code inside the braces is never run. So there must be some other issue with the code I'm afraid. This pattern wouldn't have been a valid use of the switches anyway - it would only have been useful to test a situation where behaviour was particularly expected when promoteProject was true but renameProject was false. 

We should remove all references to those switches. I'll add a ticket to do that.

EmmaLRussell (changes requested):
See remarks on the switches - they really aren't relevant any more and shouldn't be referenced in new tests.

EmmaLRussell (commented):
Great, that test's looking good now.

Still an outstanding comment on setting focus on inputs in both dialogs on open.

EmmaLRussell (approved):

PR #466, opened 2021-02-24
mrc-2166 Emails are case sensitive when sharing a project

Nicolas-Dolan:
## Description

Details about the bug and replicating it can be found [here](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2166).

The system was treating only the domain of the email as case insensitive when really it should all be case insensitive.

Note: this highlights another issue with how users are added to the system and log in. Currently, logging in email is still case sensitive. This will need to be address in another ticket ([here](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2210)).

## Type of version change
Patches

## Checklist

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt
@@ -74,10 +74,11 @@ class DbProfileServiceUserLogic(private val userRepository: UserRepository,
             throw UserException("invalidEmail")
         }
 
-        val caseInsensitiveDomainRegex = Regex("(?-i)${emailArray[0]}@(?i)${emailArray[1]}")
+        // val caseInsensitiveDomainRegex = Regex("(?-i)${emailArray[0]}@(?i)${emailArray[1]}")
+        val caseInsensitiveEmail = Regex("(?i)${email}")
 
         val username = userRepository.getAllUserNames()
-                .find { caseInsensitiveDomainRegex.matches(it) }
+                .find { caseInsensitiveEmail.matches(it) }
                 ?: return null
 
         return profileService.findById(username)
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt
@@ -79,6 +79,7 @@ class DbProfileServiceUserLogic(private val userRepository: UserRepository,
 
         val username = userRepository.getAllUserNames()
                 .find { caseInsensitiveEmail.matches(it) }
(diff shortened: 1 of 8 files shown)


EmmaLRussell (approved):
This works really nicely :+1: 

We deliberately were insensitive to case for domain only, as can be seen from the previous code, as technically it is possible for emails differing by case only to exist in domains, though this is rarely allowed in practice by hosts. Domains can safely be treated as case-insensitive.

However, since this has evidently caused problems as people are cavalier in practice with the cases they use, I think it's ok to modify to be case sensitive to avoid these issues but we should be aware that this is why it was implemented as it was, and there is a tiny chance it could bite us. 

Could you create the ticket for addressing the corresponding with login if it doesn't exist, and add a link here?

src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt
```diff
@@ -74,10 +74,10 @@ class DbProfileServiceUserLogic(private val userRepository: UserRepository,
             throw UserException("invalidEmail")
         }
 
-        val caseInsensitiveDomainRegex = Regex("(?-i)${emailArray[0]}@(?i)${emailArray[1]}")
+        val caseInsensitiveEmail = Regex("(?i)${email}")
```
carbowizzy, line 77: works well, but this regex doesn't check If this is indeed a valid email address format ? Is this being checked somewhere else?

Nicolas-Dolan:
> This works really nicely 
> 
> We deliberately were insensitive to case for domain only, as can be seen from the previous code, as technically it is possible for emails differing by case only to exist in domains, though this is rarely allowed in practice by hosts. Domains can safely be treated as case-insensitive.
> 
> However, since this has evidently caused problems as people are cavalier in practice with the cases they use, I think it's ok to modify to be case sensitive to avoid these issues but we should be aware that this is why it was implemented as it was, and there is a tiny chance it could bite us.
> 
I wasn't aware of this. I think this might be a case of we're dammed if we do, dammed if we don't. I think go case insensitive and monitor if this causes issues. We can review again if it does.

> Could you create the ticket for addressing the corresponding issue with login if it doesn't exist, and add a link here?
I created one. I also add a link in the description of this ticket: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2210

src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt
Nicolas-Dolan: The lines immediately above do some basic validation, throwing an error if there aren't two strings around an @, but validation isn't really the role of this function, which is just to check if an email exists in the database. I think there are also some cases where we wouldn't want validation here, such as with guest users. The validation is primarily done when a user is created, I believe.

PR #467, opened 2021-02-24
mrc-2207 Remove obsolete feature switches

EmmaLRussell:
## Description

Removes obsolete feature switches for rename, copy and delete project, and all references to them in code and test.

A lot of bother around dealing with these switches in the tests involved them making assumptions about which column index in projectHistory component at which to find elements, which changed depending on switch values. To make these tests more readable and less brittle, I have added meaningful classes to the elements in question and selected on those in the tests. 

I've included checking that these are at the expected indexes in the render test only, so we're still checking these, but only in one place. 

Not updating version as no behaviour changes.

## Type of version change
_Delete as appropriate_

None

## Checklist

### DIFF
--- a/src/app/static/src/app/components/projects/ProjectHistory.vue
+++ b/src/app/static/src/app/components/projects/ProjectHistory.vue
@@ -29,42 +29,41 @@
                         ></chevron-down-icon>
                     </button>
                 </div>
-                <div class="col-md-3 project-cell">
+                <div class="col-md-3 project-cell name-cell">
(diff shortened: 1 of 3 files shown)


mwoodbri (approved):
I haven't tested this manually (I'm not quite set up for that yet) but code looks good - we all benefit from spring cleaning like this!

carbowizzy (approved):
I was looking at wrong code. Looks good.

PR #469, opened 2021-02-26
mrc-2214 Move all ADR functionality in Vuex into its own module

EmmaLRussell:
## Description

ADR functionality was previously part of RootState, RootMutations and RootActions. Now that we're going to be adding more ADR functions, it seemed to time to create a dedicated ADR module.

This makes for a very long and tedious changeset, apologies. There's no new logic, code has just moved to new location, and references in its clients updated. 

This change has meant a possible confusion has crept into our 'ADR' integration tests which are run separately from the others as they frequently fail due to reliance on ADR's dev server to interact with datasets. However we have some 'adr' tests which do not require this reliance, eg. fetching and saving keys. These were previously in `root.itest.ts`. I have moved them into `adr.itest.ts` to reflect the new module name, and have renamed the old `adr.itest.ts` to `adr-dataset.itest.ts`. I've updated the npm task and renamed the step in the buildkite pipeline to clarify that the 'adr tests' are actually 'adr dataset tests'

I've also bumped up the test timeout limit. Again. 

Not updating version as no behaviour change.

## Type of version change
_Delete as appropriate_

None

## Checklist


src/app/static/jest.integration.adr.config.js
```diff
@@ -1,4 +1,4 @@
 var config = require('./jest.integration.config');
 
-config.testTimeout = 20000;
+config.testTimeout = 120000;
```
carbowizzy, line 3: I am not complete sure of this integration test behaviour but bumping from 2000 to 12000 seems like a big plus to me.

src/app/static/src/app/store/root/actions.ts
```diff
@@ -10,11 +10,6 @@ import qs from "qs";
```
carbowizzy, line 10: I think you forgot to remove unused imports

carbowizzy (changes requested):
Works great ! 

Good to go, just remove unused imports and maybe reduce the timeout duration.

src/app/static/jest.integration.adr.config.js
EmmaLRussell: It is, but that seems to be how long it takes :(
I'll try it on 90..

EmmaLRussell:
Sadly, the timeouts really do seem to need to be that huge to enable the front end ADR tests communicating with ADR's dev server to pass. I think it is important for this ticket, and for this epic in general, to get the ADR tests working most of the time, since that's what's being changed.

carbowizzy (approved):

PR #470, opened 2021-02-26
mrc-2156 create consistency on how enter button works

Nicolas-Dolan:
## Description

Aims to create consistency with how the enter button is used across the app; namely, when a dialog appears with an ok or confirm button that is highlighted in red, pressing enter should cause that button to be pressed. I have implemented this by auto-focusing the button, rather than using event listeners where possible. Buildkite frontend tests are currently failing.

Functionality has been added for:
- Login screen
- Creating projects
- Reset confirmation
- Share project (see comments on code for details)
- Delete project/version

Possible additional candidates:
- Validate
- Fit model
- Cancel fitting
- Calibrate model
- Continue

## Type of version change
Minor

## Checklist


src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -392,6 +392,23 @@
...
+        //             }
+        //             if (self.newProjectName && !self.disableCreate && self.versionToPromote){
+        //                 self.confirmPromotion(self.newProjectName)
+        //             }
+        //             if (self.projectToDelete || self.versionToDelete){
+        //                 self.confirmDelete()
+        //             }
+        //         }
+        //     });
+        // },
```
Nicolas-Dolan, line 411: These captured enter presses for the rename, copyproject/version, and delete project dialogs. I've commented them out as this functionality was being covered by a ticket by Lekan. After merging the latest master, I see enter presses are being registered for rename and copy, but not delete. If delete functionality hasn't been covered I could put a focus on the ok button, similar to how I've done in resetConfirmation, rather than use an event listener.

src/app/static/src/tests/components/resetConfirmation.test.ts
```diff
@@ -171,6 +171,38 @@ describe("Reset confirmation modal", () => {
...
+        rendered.setProps({
+            open: true,
+            continueEditing: mockContinueEdit,
+            cancelEditing: jest.fn()
+        })
+        setTimeout(() => {
+            const okBtn = rendered.findAll("button").at(0)
+            expect(okBtn.element).toBe(document.activeElement);
+            // okBtn.trigger("keyup.enter"); // ideally we'd trigger an enter press here but this doesn't do anything in vue tests; however,...
+            okBtn.trigger("click"); // in the browser, pressing enter on an active element seems to produce the same behaviour as clicking it
```
Nicolas-Dolan, line 196: ideally we'd trigger an enter press here but this doesn't do anything in vue tests; however, in the browser, pressing enter on an active element seems to produce the same behaviour as clicking it

src/app/static/src/tests/components/projects/shareProject.test.ts
```diff
@@ -628,4 +628,91 @@ describe("ShareProject", () => {
...
+            expect(input2.element).toBe(document.activeElement);
+            input2.trigger("keyup.enter");
+
+            setTimeout(() => {
+                const okBtn = modal.find("button")
+                expect(okBtn.element).toBe(document.activeElement);
+                expect(okBtn.attributes("disabled")).toBeUndefined();
+                expect(modal.find(".help-text").isVisible()).toBe(false);
+                // okBtn.trigger("keyup.enter");
+                okBtn.trigger("click");
```
Nicolas-Dolan, line 664: Ideally we'd trigger an enter press here but this doesn't do anything in vue tests; however, in the browser, pressing enter on an active element seems to produce the same behaviour as clicking it

src/app/static/src/tests/components/projects/shareProject.test.ts
```diff
@@ -628,4 +628,91 @@ describe("ShareProject", () => {
         expect(mockTooltip.mock.calls[0][1].value).toBe("Partager");
     });
 
+    it("if email entered is valid, cycles to empty input then ok button on enter presses", async (done) => {
+        const cloneProject = jest.fn();
+        const wrapper = mount(ShareProject, {
+            attachToDocument: true,
```
Nicolas-Dolan, line 634: This is deprecated but typescript doesn't recognise attachTo as a type on mount.

src/app/static/templates/login.ftl
```diff
@@ -19,6 +19,12 @@
                 form.classList.add('was-validated');
             }
         }
+
+        window.addEventListener('keyup', function(e) {
+            if (e.key === 'Enter') {
+                document.querySelector("#loginBtn").click();
+            }
+        });
```
Nicolas-Dolan, line 27: There doesn't seem to be any testing suite for this ftl page. If testing ftl is possible, would someone be able to point me to an example.

src/app/static/src/app/components/ResetConfirmation.vue
```diff
@@ -79,6 +80,13 @@
             })
         },
         methods: {
+            toggleFocus: function (element: HTMLElement, toggle: boolean) {
+                if (element){
+                    this.$nextTick(() => {
+                        toggle ? element.focus() : element.blur();
```
Nicolas-Dolan, line 86: The blur in this case is perhaps redundant as the element is removed from the DOM when it closes.

src/app/static/src/app/components/projects/Projects.vue
```diff
@@ -12,6 +12,7 @@
             </div>
             <div class="my-3 col-6 clearfix">
                 <input type="text" class="form-control"
+                       @keyup.enter="!disableCreate ? createProject(newProjectName) : ''"
```
Nicolas-Dolan, line 15: I could take this logic out of the html, but it would require me to add another method with the disableCreate condition

src/app/static/src/app/components/projects/ShareProject.vue
```diff
@@ -119,6 +122,13 @@
         methods: {
             cloneProject: mapActionByName("projects", "cloneProject"),
             userExists: mapActionByName("projects", "userExists"),
+            enterEmails(email: EmailToShareWith, index: number){
```
Nicolas-Dolan, line 125: Expected behaviour: user enters email in first input > validation > new input created and is focused automatically > if user does not enter anything into the empty input and presses enter the focus moves to the ok button (if not disabled, otherwise just blurs the input) > if user presses enter again the ok button is triggered.

src/app/static/src/tests/components/projects/shareProject.test.ts
```diff
@@ -175,7 +175,7 @@ describe("ShareProject", () => {
         input2.trigger("blur");
 
         setTimeout(() => {
-            expect(input.classes()).toContain("is-invalid");
+            // expect(input.classes()).toContain("is-invalid");
```
Nicolas-Dolan, line 178: I think the first input is still showing up as valid because of how the mount endpoints for the validation are set up.

src/app/static/src/app/components/projects/ShareProject.vue
Nicolas-Dolan: After reviewing the accessibility ticket, it occurs to me that the first input should be automatically focused when the dialog is opened.

src/app/static/src/app/components/ResetConfirmation.vue
mwoodbri: This isn't an expression, so using the ternary operator is a bit confusing - a plain if/else would be clearer.

src/app/static/src/app/components/projects/Projects.vue
mwoodbri: Making this a statement rather than expression would make more sense (I don't think we care about returning a value):
```suggestion
                       @keyup.enter="!disableCreate && createProject(newProjectName)"
```
More generally (and I'm not suggesting making this change here!) this should probably be an HTML form. If it was the we'd get automatic form submission when enter was pressed (which we could prevent by the usual means in the form was invalid) and it would also benefit accessibility (by being more compatible with screen readers).

src/app/static/src/tests/components/projects/shareProject.test.ts
mwoodbri: Are we aware of any potential solutions to this? Updating type definitions?

src/app/static/src/tests/components/projects/shareProject.test.ts
mwoodbri: Out of curiosity, is this issue specific to buttons, rather than text inputs?

src/app/static/src/tests/components/projects/shareProject.test.ts
mwoodbri: Best to add another comment in the code explaining why you're commenting this out and what needs to resolved before it can be re-instated. Or if it's no longer relevant then remove it altogether - otherwise we end up with "ghost" code for no obvious reason!

src/app/static/src/app/components/projects/ProjectHistory.vue
mwoodbri: I think this is https://github.com/mrc-ide/hint/pull/465. Given that it has been merged, can the question of outstanding functionality be addressed here?

src/app/static/src/app/components/projects/ShareProject.vue
mwoodbri: Arguably for accessibility purposes (and to benefit screen readers etc) we should be following the default behaviour of HTML forms i.e. pressing enter on any input submits the form, and tab cycles through inputs + buttons.

src/app/static/src/app/components/projects/ShareProject.vue
```diff
@@ -119,6 +122,13 @@
...
+                this.addEmail(email, index)
+                let self = this
```
mwoodbri, line 127: I think this is the cause of build failure - you can maybe just do this:
```
this.cycleInputs(email, index)
```
because the arrow function should bind `this` appropriately.

src/app/static/src/app/components/projects/ShareProject.vue
```diff
@@ -150,6 +160,20 @@
...
+                    const lastInput = this.$el.querySelectorAll("#shareInputs > div > input")[lastInputIndex]! as HTMLElement
+                    const currentInput = this.$el.querySelectorAll("#shareInputs > div > input")[index]! as HTMLElement
+                    const okBtn = this.$refs.okBtn as HTMLElement
+
+                if (index === lastInputIndex && !email.value){
+                    currentInput.blur()
+                    okBtn.focus()
+                } else {
+                    currentInput.blur()
+                    lastInput.focus()
```
mwoodbri, line 174: Should this not focus index + 1?

mwoodbri (changes requested):
More consistency always good!

A few questions/suggestions to be addressed.

src/app/static/src/app/components/ResetConfirmation.vue
```diff
@@ -79,6 +80,13 @@
             })
         },
         methods: {
+            toggleFocus: function (element: HTMLElement, toggle: boolean) {
+                if (element){
```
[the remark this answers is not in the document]
Nicolas-Dolan: Yes, when the dialog closes the element is removed from the DOM.

EmmaLRussell (commented):
Just to add to Mark's comments:
- I agree that it would be good to use forms where possible to get the default Enter key behaviour
- Ftl template tests are in `org.imperial.mrc.hint.unit.templates` - but it sounds like those changes might not have been needed?
- Sounds like we could do with bumping the version of vue/test-utils to use `attachTo`. Maybe in a separate ticket though.

src/app/static/src/tests/components/projects/shareProject.test.ts
Nicolas-Dolan: Yes, this is a button specific issue. Vue tests recognise keyup triggers on inputs.

src/app/static/src/app/components/projects/ShareProject.vue
Nicolas-Dolan: I think this is a special case because pressing enter submits the inputs for validation against the database and each other (ie, checks for duplicates, etc), and as such isn't as passive as simply tabbing. Jeff requested this functionality, but I agree tabbing to move through the inputs would work as well. We can present to Jeff as is and review in future if there is a better approach, taking into account accessibility concerns in a future ticket.

src/app/static/src/app/components/projects/ShareProject.vue
Nicolas-Dolan: I made the design choice to move to the last (empty) input rather than the next after validation as I figured this is where the user would want to make future input if any.

src/app/static/src/tests/components/projects/shareProject.test.ts
Nicolas-Dolan: I updated the type definitions to include attachTo and tried with it but it didn't work with it either. I think the version of Vue tests we have doesn't recognise attachTo.

src/app/static/templates/login.ftl
Nicolas-Dolan: It just captures enter presses outside of the inputs so that when you go to login and you haven't selected the inputs (perhaps because they have been auto filled by your browser), you can just press enter to click the login button.

src/app/static/src/tests/components/projects/shareProject.test.ts
mwoodbri: Ah - yes. It wasn't introduced until the [final 1.0.0 release](https://github.com/vuejs/vue-test-utils/releases/tag/v1.0.0). Suggest we update via another ticket - I just tried it and it causes lots of test failures.

src/app/static/templates/login.ftl
mwoodbri: Oh great - that makes sense. Thanks!

src/app/static/src/tests/components/projects/shareProject.test.ts
mwoodbri: Ok, thanks. Best to add a comment in the code explaining this, as it probably won't be obvious to the whoever might be editing these tests in the future.

Nicolas-Dolan:
> * I agree that it would be good to use forms where possible to get the default Enter key behaviour
@EmmaLRussell In practice it is proving difficult to add a form tag wrapper around the various components as many contain `v-slot` templates, which need to be kept at the root of the component.

EmmaLRussell:
> > * I agree that it would be good to use forms where possible to get the default Enter key behaviour
> 
> @EmmaLRussell In practice it is proving difficult to add a form tag wrapper around the various components as many contain `v-slot` templates, which need to be kept at the root of the component.

OK, if this is proving non-trivial I'd suggest parking it and moving on to an ADR ticket, as finishing off ADR integration should be prioritiy for this sprint.

PR #471, opened 2021-02-26
mrc-2213 Add output file types to schemas endpoint

EmmaLRussell:
## Description

Store and provide the ADR resource types for the output files which we'll need to upload them. 

Awaiting confirmation from ADR of resource type for output summary - will update resource file in a separate ticket if necessary.

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ADRTests.kt
```diff
@@ -75,11 +75,8 @@ class ADRTests : SecureIntegrationTests()
         assertSuccess(result, null)
         var data = ObjectMapper().readTree(result.body!!)["data"]
         assertThat(data["pjnz"].textValue()).isEqualTo("inputs-unaids-spectrum-file")
-
-        result = testRestTemplate.getForEntity<String>("/adr/schemas/")
-        assertSuccess(result, null)
-        data = ObjectMapper().readTree(result.body!!)["data"]
-        assertThat(data["pjnz"].textValue()).isEqualTo("inputs-unaids-spectrum-file")
+        assertThat(data["outputZip"].textValue()).isEqualTo("inputs-unaids-naomi-output-zip")
```
EmmaLRussell, line 78: Not sure why this test was the same thing twice - fetching the result and checking the pjnz, but didn't seem necessary!

mwoodbri:
@EmmaLRussell The diffs between this branch and either master or mrc-2214 are a bit too large to comfortably digest. But happy to take a look after the other PR is merged (you bring the branches back in sync).

mwoodbri (approved):
LGTM. These look like the types I need to specify when uploading artefacts to ADR, so I'll re-use them for mrc-2200. When I was looking into that I noticed that there is inconsistency in ADR wrt these types e.g. `naomi-output-zip` is used in some places, but I assume scoping to `inputs-unaids` is the right thing to be doing.

EmmaLRussell:
> These look like the types I need to specify when uploading artefacts to ADR, so I'll re-use them for mrc-2200

Cool - the actual strings might change, but if you use AppProperties, that should just need to change in one place.

PR #473, opened 2021-03-03
mrc-2215 Construct upload files

EmmaLRussell:
## Description

Adds an ADR action which fetches dataset details and constructs uploadFile metadata objects for the files which the user can choose to upload.  These will be used both by the front end UI when presenting the files (hence they contain displayName translation key) and by the upload action itself (they also contain resource type, resource id etc which will be required to do the upload to ADR) - the UI component will call this action with a list of keys from this dictionary which user has selected. 

I haven't added an integration test, as the fetch of dataset details is already tested, and we don't yet have a dev dataset with output resources we can test against. 

## Type of version change
_Delete as appropriate_

None (No behaviour changes)

## Checklist


src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -197,12 +193,12 @@
                     title: fullMetaData.title,
                     url: `${this.schemas.baseUrl}${fullMetaData.type}/${fullMetaData.name}`,
                     resources: {
-                        pjnz: this.findResource(fullMetaData, this.schemas.pjnz),
-                        shape: this.findResource(fullMetaData, this.schemas.shape),
-                        pop: this.findResource(fullMetaData, this.schemas.population),
-                        survey: this.findResource(fullMetaData, this.schemas.survey),
-                        program: this.findResource(fullMetaData, this.schemas.programme),
-                        anc: this.findResource(fullMetaData, this.schemas.anc)
+                        pjnz: findResource(fullMetaData, this.schemas.pjnz),
```
EmmaLRussell, line 196: I discovered that this component had a completely duplicated version of `findResource`, and it might as well use the util

src/app/static/src/tests/adr/actions.test.ts
```diff
@@ -1,11 +1,29 @@
-import {mockADRState, mockAxios, mockError, mockRootState, mockSuccess} from "../mocks";
+import {
+    mockADRState,
+    mockAxios,
+    mockBaselineState,
+    mockError, mockFailure,
```
Nicolas-Dolan, line 5: May as well put mockFailure on its own line for consistency.

Nicolas-Dolan (approved):
Looks good. Everything seems to be covered by a unit test.

mwoodbri (approved):
Looks good! The fields that will be used to invoke an upload are the following:
- resourceType
- resourceFilename
- resourceId
- datasetId <- is this stored somewhere?
- modelRunId

EmmaLRussell:
> 
>     * datasetId <- is this stored somewhere?
> 

Yes, that's set in baseline state

mwoodbri:
> > ```
> > * datasetId <- is this stored somewhere?
> > ```
> 
> Yes, that's set in baseline state

Ah [gottit](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/baseline/baseline.ts#L10) - thanks! I was looking in adr...

PR #474, opened 2021-03-03
mrc-2200 Push outputs to ADR: backend for upload output file

mwoodbri:
## Description

Provide endpoint for uploading model outputs to ADR

Easiest way to test is to open developer console in browser, run the model, then re-send a previous POST request but with this URL: `localhost:8080/adr/datasets/hint_test/resource/inputs-unaids-naomi-output-zip/<modelCalibrateId>?resourceFileName=resourceFileName.html` (replace `modelCalibrateId`)

You can then visit https://adr.unaids.org/dataset/hint_test and check that a new resource has appeared

Similarly, you can update an existing resource: `localhost:8080/adr/datasets/hint_test/resource/inputs-unaids-naomi-output-zip/<modelCalibrateId>?resourceFileName=output.zip&resourceId=<resourceId>` (replace `modelCalibrateId` and `resourceId`)

The integration tests don't clean up uploaded files, but they are purged nightly when the test system is cloned from prod, and there won't be any clashes in the meantime because each upload has a unique ID.

~I know there is a minor issue with coverage - working on that now but comments on remainder would be appreciated.~ "Fixed" by adding a non-ADR test for the relevant code, albeit a test for failure. It's tested indirectly in ADR tests, but any suggestions welcome. I haven't had luck with mocking fuel.

## Type of version change

Minor

## Checklist


src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -151,4 +158,48 @@ class ADRController(private val encryption: Encryption,
...
+        // 2. Return error if artefact can't be retrieved
+        if (!artefact.first.statusCode.is2xxSuccessful)
+        {
+            val baos = ByteArrayOutputStream()
+            artefact.first.body?.writeTo(baos)
+            return ErrorDetail(artefact.first.statusCode, baos.toString()).toResponseEntity()
+        }
+
+        // 3. Stream artefact to file
+        val file = File(Files.createTempDirectory("adr").toFile(), resourceFileName).apply { deleteOnExit() }
```
EmmaLRussell, line 189: This will delete the file at the end of the request but leave the dir behind? That's what I'm seeing locally anyway. Shouldn't cause a problem.

src/app/src/main/kotlin/org/imperial/mrc/hint/models/ErrorDetail.kt
```diff
@@ -15,8 +15,8 @@ class ErrorDetail(private val httpStatus: HttpStatus,
         const val defaultError = "OTHER_ERROR"
     }
 
-    fun toResponseEntity() = ResponseEntity
+    fun <T> toResponseEntity() = ResponseEntity
```
EmmaLRussell, line 18: Yeah, this is nicer! :)

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ADRControllerTests.kt
```diff
@@ -305,6 +304,103 @@ class ADRControllerTests : HintrControllerTests()
...
+            on { downloadSpectrum("model1") } doReturn ResponseEntity.ok().body(StreamingResponseBody { it.write("".toByteArray()) })
+        }
+        val mockClient: ADRClient = mock {
+            on { postFile(eq("resource_create"), eq(listOf("name" to "output1.zip", "description" to "Naomi model outputs", "hash" to "D41D8CD98F00B204E9800998ECF8427E", "package_id" to "dataset1")), any()) } doReturn ResponseEntity.ok().body("whatever")
+        }
+        val mockBuilder: ADRClientBuilder = mock {
+            on { build() } doReturn mockClient
+        }
+        val sut = ADRController(mock(), mock(), mockBuilder, mock(), mockProperties, mock(), mockAPIClient, mock(), mock())
+        val result = sut.pushFileToADR("dataset1", "adr-output-zip", "model1", "output1.zip", null)
```
EmmaLRussell, line 320: I guess it would be good to test that the file parameter to `ADRClient.postFile` looks as expected, since that's just matching on `any()`. That's always super-fiddly though. ArgumentCaptor can be helpful.

EmmaLRussell (commented):
Confirmed that this works to upload to the dev server for both new and existing resources :+1: 

>Fixed" by adding a non-ADR test for the relevant code, albeit a test for failure.

I think that's probably ok, along with the 'indirect' tests. I guess we could add successful ADRClientTests by giving them the test key and uploading to the test dataset - but then we know we'd be testing the same thing twice, and that both tests will be really slow!

codecov.yml
```diff
@@ -4,3 +4,5 @@ coverage:
   status:
     project: no
     patch: yes
+fixes:
+  - "::src/"
```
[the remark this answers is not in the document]
mwoodbri: If I pick a [previous PR](https://app.codecov.io/gh/mrc-ide/hint/compare/471) on codecov I can't actually see annotated files under the "Coverage Changes" or "Files" tabs. This is because we run coverage from inside the `src` folder, but codecov needs to know that that the path on GitHub does start with `src`.

But I think this needs more thought (and comparison with other projects) so I'll remove it here and create a maintenance ticket.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
mwoodbri: I've switched this to delete the folder, after the file has been uploaded

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ADRControllerTests.kt
mwoodbri: Good call - I've made a change which I assume is consistent with your thinking

mwoodbri:
There is a further optimisation possible here, given that we're storing the file's checksum - if the file is unchanged we could either make uploading a no-op here, and/or alert the user in the front-end when selecting files. I'll create a follow-up ticket as it's an enhancement.

codecov.yml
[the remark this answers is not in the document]
mwoodbri: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2234

EmmaLRussell (approved):

PR #476, opened 2021-03-04
mrc-2209 Check if user has permission to upload to ADR dataset

EmmaLRussell:
## Description

Back end and front end to check if the user has permission to upload to the ADR dataset. Permissions are at organization level, so we use the organization id associated with the dataset. 

Front end: Adds a new action, invoked when the Download component (final step) loads. This gets the orgs with `update_dataset` permission and sets `userCanUpload` in ADR state if the selected dataset's org is one of these. 

Back end: Adds a new `adr/orgs` endpoint which takes a query parm for the permission to check for, returning all the orgs for which the user has that permission. This could have been a more specific endpoint e.g. userCanUpload but most other endpoints follow this pattern of the backend just passing through ADR responses, which the front end interprets.

The branch adds organization id to the data retained in selectedDataset in the baseline state. However for backward compatibility, we allow for this not being available if we're loading a project saved before this feature was added. In this case we might be able to get the organization out of `datasets` in ADR state, but more likely we'll need to reload dataset details from the backend (these are not persisted when state is saved as they are bulky and go stale).

## Type of version change

Minor

## Checklist

### DIFF
--- a/src/app/static/src/app/utils.ts
+++ b/src/app/static/src/app/utils.ts
@@ -1,9 +1,9 @@
 import * as CryptoJS from 'crypto-js';
 import {ActionMethod, CustomVue, mapActions, mapGetters, mapMutations, mapState, MutationMethod} from "vuex";
-import {DatasetResource, Dict, Version} from "./types";
+import {ADRSchemas, DatasetResource, Dict, Version} from "./types";
 import {Error, FilterOption, NestedFilterOption, Response} from "./generated";
 import moment from 'moment';
-import {DynamicControlGroup, DynamicControlSection, DynamicFormMeta} from "@reside-ic/vue-dynamic-form";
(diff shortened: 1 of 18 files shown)


src/app/static/src/app/utils.ts
```diff
@@ -173,6 +173,26 @@ export const findResource = (datasetWithResources: any, resourceType: string): D
         outOfDate: false} : null
 };
 
+export const datasetFromMetadata = (id: string, datasets: any[], schemas: ADRSchemas) => {
```
EmmaLRussell, line 176: This moved from SelectDataset component in order to reuse when re-fetching dataset details for backward compatibility

mwoodbri (approved):
This is really nice. And a very timely lesson for me in Vue testing!

Ready to go IMO (bar a typo).

Nicolas-Dolan (commented):
This was useful for me to look at to get a sense of another piece of the adr epic. I think I just about understand what it's doing but I don't have much of value to add to its review.

PR #478, opened 2021-03-05
mrc-2210 Make login email case insensitive

mwoodbri:
## Description

Make username comparisons case-insensitive by downcasing both the database and provided value when looking for user record. This is achieved by overriding the `read()` method of pac4j's `DbProfileService` (which ultimately uses a parameterised statement to avoid SQL injection attacks).

Alternative is one-off downcasing of all values in database with co-ordinated changes to front- or back-end value to downcase input before passing to authentication layer, with a change to deploy tool to downcase all new usernames.

This is easy to test manually e.g. by logging in as Test.User@example.com

## Type of version change

Minor

## Checklist


EmmaLRussell (approved):
Looks good to me

Nicolas-Dolan (approved):
This works well. I tried creating a second login with uppercase letters and logging in with lowercase and it worked, but I noticed the user's email was still displayed in the banner with the upper casing. I think this is desired behaviour: maintaining the original casing in the database and displaying it when requested, but being blind to casing in the login.

One thing that occurs to me now though is what if a user creates a duplicate of an existing email just with altered casing. Will the system allow duplicate emails? Is it possible that they might get access to the existing account? Perhaps something for another ticket if a concern.

I think once the conflict is fixed this can be merged.

mwoodbri:
> One thing that occurs to me now though is what if a user creates a duplicate of an existing email just with altered casing. Will the system allow duplicate emails? Is it possible that they might get access to the existing account? Perhaps something for another ticket if a concern.

Good point: I don't think hint deploy checks this, so logged as mrc-2245

PR #479, opened 2021-03-08
mrc-2168 Handle ADR load dataset timeouts

EmmaLRussell:
## Description

This branch adds a message and button to the Select ADR Dataset dialog to enable the user to try again if there is an error (probably a timeout) when loading datasets.  As part of this work, `ADRError` in the module state is checked, so I also made sure this is reset before every request that sets it. 

The branch also implements the suggestion in the ticket to detect the odd HTML 504 Gateway Time-out responses, and log a simple message rather than a parsing error with the whole response. It also logs a simple message when an ADR request succeeds. 

Testing the timeout behaviour is tricky as timeouts are intermittent. It's easy enough to sabotage the front or back end request to force the front end behaviour. You can see from the the ticket the previously logged response string received for the timeouts, which the new code handles.  

Not implemented from ticket - logging elapsed time for ADR request. Have spun this out into another ticket to do in another sprint: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2238 

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
```diff
@@ -46,10 +47,9 @@ fun headersToMultiMap(headers: Headers): MultiValueMap<String, String>
 }
 
 @Suppress("UNCHECKED_CAST")
-fun Response.asResponseEntity(): ResponseEntity<String>
+fun Response.asResponseEntity(logger: Log = LogFactory.getLog(HintExceptionHandler::class.java)): ResponseEntity<String>
```
EmmaLRussell, line 50: Makes the logging more testable

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -50,6 +50,10 @@
                     <loading-spinner size="xs"></loading-spinner>
                     <span v-translate="'loadingDatasets'"></span>
                 </div>
+                <div v-if="adrError" id="fetch-error">
+                    <div v-translate="'errorFetchingDatasetsFromADR'"></div>
```
EmmaLRussell, line 54: This could have been our standard error view I guess, but perhaps better to have a generic error message as this is an expected intermittent error.

src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
```diff
@@ -75,25 +76,36 @@ fun Response.asResponseEntity(): ResponseEntity<String>
...
-        logger.error(e.message)
-        ErrorDetail(HttpStatus.INTERNAL_SERVER_ERROR, "Could not parse response.")
-                .toResponseEntity<String>()
+        if (this.body().asString(null).contains("504 Gateway Time-out"))
+        {
+            //Special case of ADR Gateway Timeouts returning HTML responses which cannot be parsed as JSON
+            val message = "ADR request timed out"
+            logger.error(message)
+            ErrorDetail(HttpStatus.GATEWAY_TIMEOUT, message)
+                    .toResponseEntity<String>()
```
mwoodbri, line 91: Fortunately we don't need `<String>` any more, but no need to change - it's consistent with the existing code.

mwoodbri (approved):
Looks great. Some comments, but more relevant to the future so can certainly be merged as-is.

Tested by just replacing [ADRController.kt#L92](https://github.com/mrc-ide/hint/blob/1794c34c8561d543dcf661731221ff6666154d6e/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt#L92) with:
```kotlin
val response = adr.get("http://status.savanttools.com/?code=504%20Gateway%20Timeout")
```

Hopefully we won't have to apply the same UI to other places where we access ADR...

EmmaLRussell:
Thanks @mwoodbri - all your comments are really good points, though as you say somewhat outside the scope of this ticket. I'll make a maintenance ticket to address them separately. Good idea to test through savanttools! :)

> Looks great. Some comments, but more relevant to the future so can certainly be merged as-is.
> 
> Tested by just replacing [ADRController.kt#L92](https://github.com/mrc-ide/hint/blob/1794c34c8561d543dcf661731221ff6666154d6e/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt#L92) with:
> 
> ```kotlin
> val response = adr.get("http://status.savanttools.com/?code=504%20Gateway%20Timeout")
> ```
> 
> Hopefully we won't have to apply the same UI to other places where we access ADR...

EmmaLRussell:
Maintenance ticket to address issues raised by review: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2241

PR #481, opened 2021-03-24
mrc-2216 Push outputs to ADR: invokes uploads

Nicolas-Dolan:
## Description

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/app/store/adr/actions.ts
+++ b/src/app/static/src/app/store/adr/actions.ts
@@ -5,8 +5,13 @@ import qs from "qs";
 import {ADRState} from "./adr";
 import {ADRMutation} from "./mutations";
 import {constructUploadFile, datasetFromMetadata, findResource} from "../../utils";
-import {Organization} from "../../types";
+import {Organization, Dict, UploadFile} from "../../types";
 import {BaselineMutation} from "../baseline/mutations";
+
+export interface uploadFilesPayload {
+    // filesToBeUploaded: Dict<UploadFile>[]
+    filesToBeUploaded: UploadFile[]
+}
 
 export interface ADRActions {
     fetchKey: (store: ActionContext<ADRState, RootState>) => void;
@@ -140,6 +145,26 @@ export const actions: ActionTree<ADRState, RootState> & ADRActions = {
                     }
                 });
         }
+    },
+
+    async uploadFilestoADR(context, uploadFilesPayload) {
+        const {state, rootState, commit} = context;
+        const selectedDatasetId = rootState.baseline.selectedDataset?.id;
+        const {filesToBeUploaded} = uploadFilesPayload;
+        const modelCalibrateId = rootState.modelCalibrate.calibrateId;
+        let abort = false;
+        let i;
+        for (i = 0; i < filesToBeUploaded.length; i++) {
+            if (!abort){
+                const { resourceType, resourceFilename, resourceId } = filesToBeUploaded[i]
+                await api<ADRMutation, ADRMutation>(context)
+                    .withError(ADRMutation.SetADRError, abort = true)
+                    .ignoreSuccess()
+                    .postAndReturn(`/adr/datasets/${selectedDatasetId}/resource/${resourceType}/${modelCalibrateId}`, 
+                    qs.stringify({resourceFileName: resourceFilename, resourceId}));
+            }
+        }
+        
     }
 };
 
--- a/src/app/static/src/app/components/downloadResults/UploadModal.vue
+++ b/src/app/static/src/app/components/downloadResults/UploadModal.vue
@@ -37,6 +37,7 @@
                 <button
                     type="button"
                     class="btn btn-red"
+                    @click="confirmUpload"
                     v-translate="'ok'"></button>
                 <button
                     type="button"
@@ -53,10 +54,13 @@
     import Modal from "../Modal.vue";
     import {Dict, UploadFile} from "../../types";
     import {BaselineState} from "../../store/baseline/baseline";
-    import {formatDateTime, mapStateProp, mapStateProps} from "../../utils";
+    import {formatDateTime, mapActionByName, mapStateProp, mapStateProps} from "../../utils";
     import {ADRState} from "../../store/adr/adr";
+    import {uploadFilesPayload} from "../../store/adr/actions";
 
     interface Methods {
+        uploadFilestoADRAction: (uploadFilesPayload: uploadFilesPayload) => void;
+        confirmUpload: () => void;
         handleCancel: () => void
         lastModified: (date: string) => string | null
         setDefaultCheckedItems: () => void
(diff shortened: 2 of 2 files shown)


src/app/static/src/tests/integration/adr-dataset.itest.ts
```diff
@@ -204,4 +204,47 @@ describe("ADR dataset-related actions", () => {
...
+        } as UploadFilesPayload
+
+        await adrActions.uploadFilestoADR({commit, state: adr, rootState: root} as any,
+            uploadFilesPayload);
+
+        expect(commit.mock.calls[1][0]["type"]).toBe(ADRMutation.ADRUploadCompleted);
+        // expect(commit.mock.calls[1][0]["payload"]["filename"])
+        //     .toBe("anc.csv");
+    }, 7000);
+
```
Nicolas-Dolan, line 249: I'm not sure if I'm setting up this test correctly or what I should be receiving as a payload. Any help here would be appreciated.

src/app/static/src/app/store/adr/actions.ts
```diff
@@ -140,6 +145,36 @@ export const actions: ActionTree<ADRState, RootState> & ADRActions = {
...
+        for (i = 0; i < filesToBeUploaded.length; i++) {
+
+            if (!(adr.state as ADRState).abortUpload){ // aborts subsequent file uploads if a prior fails
+                const { resourceType, resourceFilename, resourceId } = filesToBeUploaded[i]
+                
+                if (i === filesToBeUploaded.length - 1){ // only a successful upload of the last file sets the completed state
+                    await api<ADRMutation, ADRMutation>(context)
+                        .withError(ADRMutation.SetADRUploadError)
+                        .withSuccess(ADRMutation.ADRUploadCompleted)
+                        .postAndReturn(`/adr/datasets/hint_test/resource/${resourceType}/${modelCalibrateId}`,
```
mwoodbri, line 166: This should be the dataset ID, not `hint_test` (likewise below)

mwoodbri (changes requested):
This worked well for me (albeit painfully slowly due to speed of ADR system!).

I had to change a couple of things in order to it to work as expected: I was using the dev ADR system with the generic test user and the Burundi dataset. I've added some corresponding comments to the code - I haven't reviewed all the tests because they may be affected by some of my notes.

src/app/static/src/app/store/adr/actions.ts
```diff
@@ -140,6 +145,36 @@ export const actions: ActionTree<ADRState, RootState> & ADRActions = {
                     }
                 });
         }
+    },
+
+    async uploadFilestoADR(context, uploadFilesPayload) {
```
mwoodbri, line 150: How does this `uploadFilesPayload` argument relate to `uploadFiles` in the store? Is it the same thing?

src/app/static/src/tests/integration/adr-dataset.itest.ts
EmmaLRussell: You're trying to upload a file for real here to a test dataset? You can either hardcode this I guess, or use the dataset information it's fetching at the top of the test to find the resource of an uploadable file. But you'll also need to have output data available on the server to upload (so the model run and calibrate will need to have completed). 

GIven that we're already doing upload integration tests from the backend, it mght be easier here to do a simple integration test that we're hitting the endpoint and getting an expected error.

EmmaLRussell (changes requested):

src/app/static/src/tests/integration/adr-dataset.itest.ts
```diff
@@ -204,4 +205,46 @@ describe("ADR dataset-related actions", () => {
...
+        ] as UploadFile[]
+
+        await adrActions.uploadFilestoADR({commit, state: adr, rootState: root} as any,
+            uploadFilesPayload);
+
+        expect(commit.mock.calls.length).toBe(2);
+        expect(commit.mock.calls[0][0]["type"]).toBe("ADRUploadStarted");
+        expect(commit.mock.calls[1][0]["type"]).toBe("SetADRUploadError");
+        expect(commit.mock.calls[1][0]["payload"]["error"]).toBe("OTHER_ERROR");
+    }, 7000);
```
Nicolas-Dolan, line 248: This always times out. I'm not sure if there is any way around this besides giving it a really long timer, which is not ideal as it will slow down these tests every time we test a build.

src/app/static/src/app/store/adr/actions.ts
Nicolas-Dolan: I wasn't sure you could use an array as a payload. I've now just simplified it to an array of UploadFile

mwoodbri (changes requested):
This works very nicely. A couple of really minor questions over the code, and tests could maybe be improved slightly -please see comments inline.

I noticed a quirk that is possibly out of scope for this ticket - if you upload twice from the "Save results" page then you get two files in ADR, whereas if you repeat from scratch you only get one - because in the latter case the `resourceId` is set correctly. Is it possible to get it into the state somehow if the upload was successful? **Edit: see mrc-2274**

src/app/static/src/tests/integration/adr-dataset.itest.ts
Nicolas-Dolan: I'm not sure if it will succeed or not. Perhaps there is some way to guarantee a fail and get a response quickly from adr. I recall giving it an incorrect `resourceType` produced a quick error response, but I would imagine `"type1"` is an incorrect `resourceType`.

EmmaLRussell: I think this should always fail won't it? It's got nonsense test data that won't correspond to any real files. I'd expect it to fail immediately in the backend because the resourceType isn't recognised: https://github.com/mrc-ide/hint/blob/master/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt#L184
So it shouldn't ever get as far as ADR and shouldn't require a long timeout.

EmmaLRussell: Also, if you are just  going for the failure case, you don't need all that set-up with getting a real dataset, you can just give the adr state a dummy dataset too.

EmmaLRussell: Oh, in fact, that's probably why it's timing out, it should fly if you get rid of all that.

EmmaLRussell (commented):
This worked great for me to upload a zip file. Got a 'Could not parse error' when tried to upload summary report also (to Antarctica dataset), which was caused by a 403 response in html form not being digestible by the kotlin endpoint, and I guess the root cause of that was that the summary  report resource is not supported yet in the dataset. It would be nice to be able to show a better error in that case (definitely not one for this ticket though). 

A tiny other thing I just noticed (sorry!), was that while the button is disabled it goes pink, which isn't consistent with other buttons (Calibrate, Validate) which go grey.

mwoodbri:
> This worked great for me to upload a zip file. Got a 'Could not parse error' when tried to upload summary report also (to Antarctica dataset), which was caused by a 403 response in html form not being digestible by the kotlin endpoint, and I guess the root cause of that was that the summary report resource is not supported yet in the dataset. It would be nice to be able to show a better error in that case (definitely not one for this ticket though).

ADR test system still doesn't support upload of any HTML files - it's a bug that you should be able to replicate by uploading the same file manually e.g. to hint_test dataset - it should work on prod but not on test.

EmmaLRussell:
> > This worked great for me to upload a zip file. Got a 'Could not parse error' when tried to upload summary report also (to Antarctica dataset), which was caused by a 403 response in html form not being digestible by the kotlin endpoint, and I guess the root cause of that was that the summary report resource is not supported yet in the dataset. It would be nice to be able to show a better error in that case (definitely not one for this ticket though).
> 
> ADR test system still doesn't support upload of any HTML files - it's a bug that you should be able to replicate by uploading the same file manually e.g. to hint_test dataset - it should work on prod but not on test.

Oh of course, thanks, didn't realise or forgot that was how that issue manifested.

mwoodbri (approved):

EmmaLRussell (approved):

PR #484, opened 2021-03-26
mrc-2230 - Check and display user permissions on dataset when dataset is first selected

carbowizzy:
## Description

This PR displays some feedback about user permission when data is pulled from ADR -- e.g. when a user 'select' the data package in step 1, display whether or not they have write access to that data package and tooltip for further explanation.

## Type of version change
_Delete as appropriate_

## Checklist


mwoodbri:
@carbowizzy Just to clarify the intention of this code: I was expecting the same permission currently displayed on the final page (i.e. whether the user has permission to upload new files to the selected dataset) to now be displayed on the first page too, when the dataset is selected. But at first glance it looks like it's displaying something different i.e. my role within the first entry in the list of organisations I'm a member of. This is independent of dataset - so seems less specific than the permission displayed on the last page. Am I understanding correctly?

carbowizzy:
> @carbowizzy Just to clarify the intention of this code: I was expecting the same permission currently displayed on the final page (i.e. whether the user has permission to upload new files to the selected dataset) to now be displayed on the first page too, when the dataset is selected. But at first glance it looks like it's displaying something different i.e. my role within the first entry in the list of organisations I'm a member of. This is independent of dataset - so seems less specific than the permission displayed on the last page. Am I understanding correctly?

Yeah, I missed the id selection. Will update now.

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -23,6 +23,11 @@ export interface Translations {
     cancelEdit: string,
     cancelEditLoggedIn: string,
     cancelFitting: string,
+    capacityAccessLevel: string,
```
EmmaLRussell, line 26: Are these keys remnants of when you were displaying user role rather than dataset permission? "Capacity" doesn't seem that helpful as a term here. Would `adrAccessLevel`, `adrRead`, `adrReadWrite`, `adrReadTooltip` (or message) and `adrReadWrite` be closer to actual usage?

EmmaLRussell (changes requested):
Nice re-use of existing action!

It seems a bit odd that I'm being shown a permission level when I haven't selected a dataset yet...  

Screenshot from 2021-03-29 17-25-54

OK, I think is because you're invoking getUserCanUpload when the key is set, but it should be invoked whenever a dataset is selected, and its data fetched, as this varies per dataset. Also, need to avoid showing "Read" as permission level when no dataset is selected (when selectedDataset in baseline state is null - should hide the permission elements in this case).

Also, do you think the tooltip is a little too subtle? Could be message on page, something like "You do not currently have write permission in this dataset, and will be unable to upload files to ADR"

src/app/static/src/app/store/translations/locales.ts
mwoodbri: I agree with this - and I don't think we need the admin/member distinction at the dataset level. Would also suggest a common prefix for inline and tooltip text e.g. `adrRead` + `adrReadTooltip`.

Main thing is that we clearly communicate the most important point - that without write access you won't be able to upload your model outputs. This should probably be on the screen - with the tooltip providing some help on how to resolve the issue (e.g. contacting relevant people). We have plenty of space to do this.

src/app/static/src/app/components/adr/ADRIntegration.vue
```diff
@@ -1,7 +1,17 @@
...
     <div v-if="loggedIn" class="mb-5">
         <adr-key></adr-key>
-        <select-dataset v-if="key"></select-dataset>
+        <div v-if="key">
+            <select-dataset></select-dataset>
+            <div id="adr-capacity">
+            <span class="font-weight-bold align-self-stretch"
+                  v-translate="'capacityAccessLevel'">
+           </span>
+                <span v-tooltip="handleUploadPermission(hasUploadPermission, true)">
```
mwoodbri, line 10: I think we could do with some extra whitespace - it's looking a bit cramped:

Screenshot from 2021-03-29 18-08-36

Also: minor point but the formatting of this code makes it a bit hard to read

mwoodbri (changes requested):
Looks pretty good. I think the message could be clearer, and I agree with Emma that the message should only appear after you have selected a dataset - but this is hopefully easy to rectify.

src/app/static/src/app/store/translations/locales.ts
carbowizzy: I tried to put more text on the screen but, it seems more to many text.

EmmaLRussell: Sorry, not sure what you mean, do you mean it looks too busy? There's not much else on the baseline screen so I'd be surprised if we couldn't find a way to display that information in a way that doesn't look cluttered. Could use `small` text styling if appropriate.

src/app/static/src/app/components/adr/ADRIntegration.vue
```diff
@@ -1,7 +1,15 @@
 <template>
     <div v-if="loggedIn" class="mb-5">
         <adr-key></adr-key>
-        <select-dataset v-if="key"></select-dataset>
+        <div v-if="key">
+            <select-dataset></select-dataset>
+            <div class="pt-3" id="adr-capacity" v-if="selectedDataset">
+                <span class="font-weight-bold align-self-stretch" v-translate="'adrAccessLevel'"></span>
+                <span v-tooltip="handleUploadPermission(hasUploadPermission, true)">
+                    <a href="#">{{ handleUploadPermission(hasUploadPermission, false) }}</a>
```
EmmaLRussell, line 9: It's a bit confusing for this to be a link, could it just be a red span?

EmmaLRussell (approved):
Works great, thanks. Just one tiny comment - sorry didn't notice that before.

src/app/static/src/app/components/adr/ADRIntegration.vue
Nicolas-Dolan: I agree, I found myself clicking on it to see what would happen.

mwoodbri (changes requested):
I like @Nicolas-Dolan's suggested wording, and agree with @EmmaLRussell regarding tooltip link - if you can fix that then this is ready to merge.

NEWS.md
```diff
@@ -2,6 +2,8 @@
 
 * Check and display user permissions on dataset when dataset is first selected
 
+* Invoke ADR uploads with actions
```
EmmaLRussell, line 5: Looks like the merge has combined two features into one version here.

EmmaLRussell (approved):
Looks good, should sort out the version merge then ready to go

mwoodbri (approved):
:+1:

PR #485, opened 2021-03-30
Mrc-2235 Avoid re-uploading unchanged files to ADR

carbowizzy:
## Description

This PR adds feature to avoider-uploading. When a user uploads files to ADR, it asks ADR for the checksum and do a transparent no-op if it matches. If it does not match then it will re-upload output else it will do nothing. 

I have a test for my changes already. I am not sure why the coverage is failing as no indication displayed for possible untested lines.

## Type of version change
_Delete as appropriate_

## Checklist


src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -205,16 +207,42 @@ class ADRController(private val encryption: Encryption,
                 listOf("name" to resourceFileName, "description" to artefact.second, "hash" to file.md5sum(),
                         "resource_type" to resourceType)
         val adr = adrClientBuilder.build()
+        val newDatasetHash = commonParameters[2].second
```
mwoodbri, line 210: Accessing by index is fragile... would be better to do:
```kotlin
val filePart = Pair("upload", file)
val hash = file.md5um()
val commonParameters =
                listOf("name" to resourceFileName, "description" to artefact.second, "hash" to hash,
                        "resource_type" to resourceType)
val adr = adrClientBuilder.build()
return try
{
    ...
```

EmmaLRussell (commented):
Codecov is flagging up partial cover in uploadFileHasNoChanges so maybe it thinks the error response path is not unit tested: https://app.codecov.io/gh/mrc-ide/hint/compare/485/diff#D2L242
Although I'd expect that to flag the line with `if` on it.. 

And it looks like that is being tested in `pushes file to ADR fails if retrieval from ADR fails` (this seems like not the best name for this test, it's really testing that the push fails) - non success response on get should mean `uploadFileHasNoChanges` returns false, so the upload is attempted, and fails...
So, not sure, sorry!

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ADRControllerTests.kt
```diff
@@ -358,6 +360,54 @@ class ADRControllerTests : HintrControllerTests()
...
+        val hasOldHash = sut.uploadFileHasNoChanges("resource1", "D41D8CD98F00B204E9800998ECF8427E")
+        assertThat(hasOldHash).isTrue
+
+        val result = sut.pushFileToADR("dataset1", "adr-output-zip", "model1", "output1.zip", "resource1")
+        assertThat(result.statusCode).isEqualTo(HttpStatus.OK)
+        assertThat(objectMapper.readTree(result.body)["data"].textValue()).isEqualTo(null)
+    }
+
+    @Test
+    fun `update ADR if upload file has changes`()
```
mwoodbri, line 388: This seems to be testing the same thing as `pushes updated file to ADR`? If so then that test can be removed.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
EmmaLRussell, line 210: This is a little bit obfuscated, could you just set `val newDatasetHash = file.md5sum()` before constructing commonParameters, then use that value when making the list.

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ADRControllerTests.kt
```diff
@@ -390,6 +440,7 @@ class ADRControllerTests : HintrControllerTests()
             on { downloadSpectrum("model1") } doReturn ResponseEntity.ok().body(StreamingResponseBody { it.write("".toByteArray()) })
         }
         val mockClient: ADRClient = mock {
+            on { get(any()) } doReturn ResponseEntity.status(HttpStatus.BAD_GATEWAY).body("Bad Gateway")
```
mwoodbri, line 443: This line renders the next one irrelevant. Suggest reverting this test and splitting hash retrieval failure into a new one.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -205,16 +207,42 @@ class ADRController(private val encryption: Encryption,
...
         return try
         {
             when (resourceId)
             {
                 null -> adr.postFile("resource_create", commonParameters + listOf("package_id" to id), filePart)
-                else -> adr.postFile("resource_patch", commonParameters + listOf("id" to resourceId), filePart)
+                else ->
+                {
+                    val fileHasNoChanges = uploadFileHasNoChanges(resourceId, newDatasetHash)
```
mwoodbri, line 218: Not sure we need to declare a variable for this

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -205,16 +207,42 @@ class ADRController(private val encryption: Encryption,
...
+                    if (fileHasNoChanges)
+                    {
+                        SuccessResponse(null).asResponseEntity()
```
mwoodbri, line 221:
```suggestion
                        EmptySuccessResponse.asResponseEntity()
```
More generally I guess we need to be careful here: this response probably looks very different to that from `resource_patch`. But as long as the front-end isn't expecting access to resource `id` etc from response ([it doesn't currently](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/adr/actions.ts#L168)) then it's fine.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -205,16 +207,42 @@ class ADRController(private val encryption: Encryption,
...
         }
     }
+
+    fun uploadFileHasNoChanges(resourceId: String, newDatasetHash: String): Boolean
+    {
+        val adr = adrClientBuilder.build()
+        val response = adr.get("resource_show?id=${resourceId}")
+        if (response.statusCode.is2xxSuccessful)
+        {
+            val parser = JSONParser()
```
EmmaLRussell, line 241: We're using `objectMapper` elsewhere in this controller to read JSON, probably best to be consistent.

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ADRControllerTests.kt
```diff
@@ -358,6 +360,54 @@ class ADRControllerTests : HintrControllerTests()
...
+    @Test
+    fun `does not update ADR if uploaded file has no changes`()
+    {
+        val hash = mapOf("hash" to "D41D8CD98F00B204E9800998ECF8427E")
+        val data = mapOf("data" to hash)
+        val mockAPIClient: HintrAPIClient = mock {
+            on { downloadSpectrum("model1") } doReturn ResponseEntity.ok().body(StreamingResponseBody { it.write("".toByteArray()) })
+        }
+        val mockClient: ADRClient = mock {
+            on { postFile(eq("resource_patch"), eq(listOf("name" to "output1.zip", "description" to "Naomi model outputs", "hash" to "D41D8CD98F00B204E9800998ECF8427E", "resource_type" to "adr-output-zip", "id" to "resource1")), any()) } doReturn ResponseEntity.ok().body("whatever")
```
EmmaLRussell, line 372: Don't need to include this if the code won't get as far as the push.

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ADRControllerTests.kt
```diff
@@ -358,6 +360,54 @@ class ADRControllerTests : HintrControllerTests()
...
+        val mockBuilder: ADRClientBuilder = mock {
+            on { build() } doReturn mockClient
+        }
+        val sut = ADRController(mock(), mock(), mockBuilder, mock(), mockProperties, mock(), mockAPIClient, mock(), mock())
+        val hasOldHash = sut.uploadFileHasNoChanges("resource1", "D41D8CD98F00B204E9800998ECF8427E")
+        assertThat(hasOldHash).isTrue
+
+        val result = sut.pushFileToADR("dataset1", "adr-output-zip", "model1", "output1.zip", "resource1")
+        assertThat(result.statusCode).isEqualTo(HttpStatus.OK)
+        assertThat(objectMapper.readTree(result.body)["data"].textValue()).isEqualTo(null)
```
EmmaLRussell, line 384: Maybe also `verify` that `postFile` was not called on the client?

EmmaLRussell (changes requested):
This works nicely :+1:  - if I upload then go back and forwards a step to reload the frontend upload files (as mrc-2274 isn't merged in yet), there is no new upload activity visible on ADR. If I recalibrate with different parameters, changing the output results and re-upload the files appear updated as expected. 

Couple of small suggestions below.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
carbowizzy: ObjectMapper() for some reason was not return any data, so after too many hours on it I changed to JsonParser().

mwoodbri (changes requested):
This worked well for me. Some minor suggestions for code, but nothing that should take very long to resolve.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -205,16 +207,42 @@ class ADRController(private val encryption: Encryption,
...
+            val json: JSONObject = parser.parse(response.body!!) as JSONObject
+            val data: JSONObject = json["data"] as JSONObject
+            return data["hash"] == newDatasetHash
```
EmmaLRussell, line 244:
```suggestion
            val hash = objectMapper.readTree(response.body!!)["data"]["hash"].asText()
            return hash != newDatasetHashh
```

This worked for me

mwoodbri: Nice... should help with coverage too!

mwoodbri (changes requested):
Looks good - just an issue with test naming and would still be good to use `objectMapper`, as @EmmaLRussell pointed out.

Also one other point that might be worth considering to avoid inadvertently overwriting uploads.

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ADRControllerTests.kt
```diff
@@ -371,7 +417,7 @@ class ADRControllerTests : HintrControllerTests()
     }
 
     @Test
-    fun `pushes file to ADR fails if retrieval from hintr fails`()
+    fun `pushes file to ADR fails retrieval from ADR fails`()
```
mwoodbri, line 420: Please revert this change - the original name is correct because this test appears to relate to retrieval from hintr, not ADR

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -201,20 +204,50 @@ class ADRController(private val encryption: Encryption,
...
+                    }
+                }
             }
         }
         finally {
             tmpDir.deleteRecursively()
         }
     }
+
+    fun uploadFileHasChanges(resourceId: String, newDatasetHash: String): ResponseEntity<Boolean>
```
mwoodbri, line 236: Hi @carbowizzy. I think this an improvement but it has also gotten a bit difficult to follow: particularly constructing a response just to indicate success or failure to another method. I suggest something like this:
```kotlin
    fun uploadFileHasChanges(resourceId: String, newDatasetHash: String): Boolean
    {
        val adr = adrClientBuilder.build()
        val response = adr.get("resource_show?id=${resourceId}")
        if (response.statusCode.isError) {
            throw IOException("Unable to retrieve hash")
        }
        val parser = JSONParser()
        val json: JSONObject = parser.parse(response.body!!) as JSONObject
        val data: JSONObject = json["data"] as JSONObject
        return data["hash"] != newDatasetHash
    }
```

Then you can handle in the method above as:
```kotlin
        return try
        {
            when (resourceId)
            {
                null -> adr.postFile("resource_create", commonParameters + listOf("package_id" to id), filePart)
                else ->
                {
                    if (uploadFileHasChanges(resourceId, fileHash))
                    {
                        adr.postFile("resource_patch", commonParameters + listOf("id" to resourceId), filePart)
                    }
                    else
                    {
                        EmptySuccessResponse.asResponseEntity()
                    }
                }
            }
        }
        catch (e: IOException) {
            ErrorDetail(HttpStatus.INTERNAL_SERVER_ERROR, e.message!!).toResponseEntity()
        }
        finally
        {
            tmpDir.deleteRecursively()
        }
```

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ADRControllerTests.kt
```diff
@@ -343,21 +342,68 @@ class ADRControllerTests : HintrControllerTests()
...
+        }
+        val mockBuilder: ADRClientBuilder = mock {
+            on { build() } doReturn mockClient
+        }
+
+        val sut = ADRController(mock(), mock(), mockBuilder, mock(), mockProperties, mock(), mockAPIClient, mock(), mock())
+        val result = sut.uploadFileHasChanges("resource1", "test")
+        assertThat(result.statusCode).isEqualTo(HttpStatus.BAD_GATEWAY)
+        assertThat(result.body!!).isFalse
+    }
```
mwoodbri, line 381: To reflect changes suggested above:
```suggestion
    @Test
    fun `uploading returns error and does not update ADR if exception occurs while comparing hashes`()
    {
        val mockAPIClient: HintrAPIClient = mock {
            on { downloadSpectrum("model1") } doReturn ResponseEntity.ok().body(StreamingResponseBody { it.write("".toByteArray()) })
        }
        val mockClient: ADRClient = mock {
            on { get("resource_show?id=resource1") } doReturn ResponseEntity.status(HttpStatus.BAD_GATEWAY).body(null)
        }
        val mockBuilder: ADRClientBuilder = mock {
            on { build() } doReturn mockClient
        }

        val sut = ADRController(mock(), mock(), mockBuilder, mock(), mockProperties, mock(), mockAPIClient, mock(), mock())
        val result = sut.pushFileToADR("dataset1", "adr-output-zip", "model1", "output1.zip", "resource1")
        verify(mockClient, never()).postFile(any(), any(), any())
        assertThat(result.statusCode).isEqualTo(HttpStatus.INTERNAL_SERVER_ERROR)
        assertThat(objectMapper.readTree(result.body)["data"].textValue()).isEqualTo(null)
    }
```

mwoodbri (changes requested):
This isn't too far from being mergeable IMO, but it merits a bit of simplification, just to make it more maintainable in the future.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -201,20 +205,48 @@ class ADRController(private val encryption: Encryption,
...
             tmpDir.deleteRecursively()
         }
     }
+
+    fun uploadFileHasChanges(resourceId: String, newDatasetHash: String): Boolean
+    {
+        val adr = adrClientBuilder.build()
+        val response = adr.get("resource_show?id=${resourceId}")
+        if (response.statusCode.isError) {
+            throw IOException("Unable to retrieve hash")
```
EmmaLRussell, line 245: This is looking good now. If this error message is going to make it back to the user (which I think it will), would suggest possibly being a little more descriptive here, if only to add "from ADR" to give some idea of what's gone wrong. 

And I'd still vote for using objectMapper!

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -201,20 +205,47 @@ class ADRController(private val encryption: Encryption,
...
+                        EmptySuccessResponse.asResponseEntity()
+                    }
+                }
             }
         }
-        finally {
+        catch (e: Exception)
+        {
+            ErrorDetail(HttpStatus.INTERNAL_SERVER_ERROR, e.message!!).toResponseEntity()
+        } finally
```
mwoodbri, line 234:
```suggestion
        }
        finally
```

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -201,20 +205,47 @@ class ADRController(private val encryption: Encryption,
...
+                    }
+                    else
+                    {
+                        EmptySuccessResponse.asResponseEntity()
+                    }
+                }
             }
         }
-        finally {
+        catch (e: Exception)
```
mwoodbri, line 231:
```suggestion
        catch (e: IOException)
```

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -167,7 +171,7 @@ class ADRController(private val encryption: Encryption,
     }
 
     @PostMapping("/datasets/{id}/resource/{resourceType}/{modelCalibrateId}")
-    @Suppress("ReturnCount")
+    @Suppress("ReturnCount", "UnsafeCallOnNullableType", "TooGenericExceptionCaught")
```
mwoodbri, line 174:
```suggestion
    @Suppress("ReturnCount", "UnsafeCallOnNullableType")
```

mwoodbri (changes requested):
Works very nicely. I think we can be more specific about Exception type (my fault: sorry!). But will be ready to merge after that.

mwoodbri (approved):
Works very nicely in my testing - good job!

EmmaLRussell (approved):

PR #487, opened 2021-03-31
mrc-2279 Push input files to ADR if uploaded from local

EmmaLRussell:
## Description

Sorry for the large changeset, quite a lot of it is test data.. 

To test this functionality, you'll need to switch on the featureSwitch by adding qs parameter i.e.: http://localhost:8080/?adrPushInputs=1

You can test that pushing input changes works by downloading an input file from ADR - one of the csv files is easiest - and editing one of the values locally. Make a project in HINT and import from dataset as usual, then upload the edited input file. When you get to the last step in the wizard, as long as qs param has been set, you should see your edited input file in the list of files available to upload. Select that file and upload. You should then see in ADR that the resource timestamp has been updated and if you download and open the file you should see your value change. It takes a couple of minutes for ADR to revalidate the uploaded file. 

In the course of adding inputs I realised that we've been somewhat conflating resource name and resource filename. Name is displayed in ADR and is something like 'Spectrum File' or 'Survey Data'. It shouldn't really be identical with the filename, but we have been using filename as the name for the output files. So I've added name as a separate parameter to send to the upload file endpoint....

...and actually, for the input files we already know the filename without providing it in the parameters to the endpoint because we want to use the filename the user used to upload, and we already have that in the version file on the server. But I've included the code to pass it through anyway to be consistent, since we do need the param for the output files.

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ADRControllerTests.kt
```diff
@@ -328,32 +370,147 @@ class ADRControllerTests : HintrControllerTests()
...
+    {
+        val sut = ADRController(mock(), mock(), mock(), mock(), mockProperties, mock(), mock(), mock(), mock())
+        val result = sut.pushFileToADR("datasetId", "adr-pjnz", "calId",
+                "testResFilename.", "testResName", "testResId")
+        assertThat(result.statusCode).isEqualTo(HttpStatus.BAD_REQUEST)
+        assertThat(result.body!!).contains("File does not exist")
+    }
+
+    @Test
+    fun `throws exception on make input file to push to ADR if not input resource type`()
```
EmmaLRussell, line 485: Adding this to satisfy code coverage by covering the default line of the `when` mapping string resource type to file type in the private `makeInputFileToPushToADR` method (which is there to break up the long `pushFileToADR` method to keep linting happy!). Logically, it shouldn't ever (!) hit that line, given the way the `pushFileToADR` distinguishes input from output files, hence using reflection here to call the private method separately. 

There may be a more elegant way to deal with all this, and I think changes in Lekan's metadata branch is also going to result in some restructuring of the method so should possibly await the outcome of that before settling on this approach.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -172,38 +174,48 @@ class ADRController(private val encryption: Encryption,
...
-            artefact.first.body!!.writeTo(fis)
+        if (errorDetail != null)
+        {
+           return errorDetail.toResponseEntity()
         }
 
-        // 4. Checksum file and upload with metadata to ADR
-        val filePart = Pair("upload", file)
+        // 3. Checksum file and upload with metadata to ADR
+        @Suppress("UnsafeCallOnNullableType")
```
carbowizzy, line 209: I think this @suppress("UnsafeCallOnNullableType") can be avoided.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -213,8 +225,64 @@ class ADRController(private val encryption: Encryption,
...
+        finally
+        {
             tmpDir.deleteRecursively()
         }
     }
+
+    @Suppress("ReturnCount")
+    private fun makeOutputFileToPushToADR(artefact: Pair<ResponseEntity<StreamingResponseBody>, String>,
+                                          tmpDir: File,
+                                          resourceFileName: String): Pair<File?, ErrorDetail?>
```
carbowizzy, line 237: This method returns multiple values. I wonder if this could just be an Object instead of having to use Pair with null value.

src/app/static/src/app/store/adr/actions.ts
```diff
@@ -122,20 +131,56 @@ export const actions: ActionTree<ADRState, RootState> & ADRActions = {
...
                                 metadata,
                                 1,
                                 schemas.outputSummary,
                                 `${project.name}_naomi_summary.html`,
+                                `${project.name} Naomi Summary`,
                                 "uploadFileOutputSummary")
-                        };
+                        } as Dict<UploadFile>;
+
+                        if (switches.adrPushInputs) {
```
carbowizzy, line 151: How would this feature switch be used in production ?

carbowizzy (commented):
Looks great, works perfect just some comments about certain logic.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
EmmaLRussell: How would you avoid it? Need to do the unsafe call, either here or when we call md5sum - we know from the logic of the called method that exactly one of errorDetail and file will be null, but the compiler doesn't pick that up.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
EmmaLRussell: Could make a new type just for the return type of these methods, with file and errorDetail props, which would shut up the linting error, but feels like a bit of heavy overhead itself. Don't really mind either way, they're both kind of icky.

src/app/static/src/app/store/adr/actions.ts
EmmaLRussell: When we want to switch the feature on, we just need to switch the default value from false to true. Periodically we remove old switches when features are bedded in - as here it typically just means removing a containing conditional and removing the property from the switches object

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -213,8 +225,64 @@ class ADRController(private val encryption: Encryption,
...
+        if (resourceId == null)
+        {
+            return Pair(null,
+                    ErrorDetail(HttpStatus.BAD_REQUEST, "resourceId must be provided for input resourceType"))
+        }
+
+        // Map adr schema to version file type
+        val fileType = when (resourceType)
+        {
+            appProperties.adrPJNZSchema -> FileType.PJNZ
```
mwoodbri, line 269: Slightly surprising that we haven't needed this mapping from schema to filetype elsewhere. If there's a problem here it's that we're using (untyped) properties values when they should really be an enum. But even if that's worth addressing then it's another ticket.

mwoodbri (changes requested):
This works really well. And well spotted on filename vs name!

I have a couple of concerns, mainly about ADRController, simply because I had to read it three or four times in order to follow the logic. I don't think it was ever perfect, but maybe now is the time to ensure that it's a solid foundation. However, as you say, this will also depend on changes to metadata branch - where code is already simplified by dropping the `Pair`s of response and description.

EmmaLRussell:
Thanks @carbowizzy and @mwoodbri - yeah, it's a real plate of spaghetti that controller method, especially carved up into separate functions without due care and attention. I'll wait til Lekan's branch is merged in and then try to rationalise it a bit with a firmer base..

EmmaLRussell:
NB Note changes in https://github.com/mrc-ide/hint/pull/492 and their relevance for this branch - once mrc-2322 is merged in, this should use name for the matching rather than filename, since filename for an input resource can vary on new upload, but name should not.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
EmmaLRussell: I'm going to make this mapping into a single method called at the start of the post method, and use the FileType enum subsequently. I think we've avoided the mapping previous because most requests up til now are for single files of a known type.

EmmaLRussell:
> NB Note changes in #492 and their relevance for this branch - once [mrc-2322](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2322) is merged in, this should use name for the matching rather than filename, since filename for an input resource can vary on new upload, but name should not.

I've modified `findResource` in the front end to match on `name` rather than filename, however this is possibly a little misleading: either name or filename is fine for matching output files; name is not actually provided to the utils method for input files, since we haven't extracted the name from the dataset. So what it's doing for input files is defaulting to match on resourceType only - which is fine for input files as we shouldn't have more than one file per input type in a dataset.

mwoodbri (changes requested):
I retested this and it still works very nicely. Well, apart from some unrelated issues with upload code, which has exposed a blind-spot in my understanding of scoped actions/mutations - see mrc-2403.

I've added some inline comments but nothing major.

Only broader observation is about the actual requirements. Seems slightly odd that if I want to experiment/fiddle with different input files then after upload they become the default for everyone else (in contrast to outputs, which are scoped per-project) - especially as we don't get the option to select the version of input files to use when importing. Maybe we should ultimately be able to select input files by project... but for now I assume this is all by design.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -175,43 +177,133 @@ class ADRController(private val encryption: Encryption,
                       @PathVariable modelCalibrateId: String,
                       @RequestParam resourceFileName: String,
                       @RequestParam resourceId: String?,
-                      @RequestParam description: String) : ResponseEntity<String>
+                      @RequestParam resourceName: String,
+                      @RequestParam description: String?) : ResponseEntity<String>
     {
+        return when (resourceType)
```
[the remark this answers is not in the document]
mwoodbri: I think so - whenever I see a list of string comparisons I assume that some point a random string will be passed in and/or we'll forget to include a string that should be on the list. But it's another ticket at best.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -175,43 +177,133 @@ class ADRController(private val encryption: Encryption,
...
+                        description)
+            appProperties.adrPJNZSchema,
+            appProperties.adrShapeSchema,
+            appProperties.adrPopSchema,
+            appProperties.adrSurveySchema,
+            appProperties.adrARTSchema,
+            appProperties.adrANCSchema ->
+                pushInputFileToADR(id, resourceType, resourceId, resourceName)
+            else -> return ErrorDetail(HttpStatus.BAD_REQUEST, "Invalid resourceType").toResponseEntity()
+
```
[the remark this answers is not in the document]
EmmaLRussell: I'm surprised detekt didn't moan about that one!

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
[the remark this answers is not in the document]
EmmaLRussell: Ticket for this: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2422

src/app/static/src/app/store/adrUpload/actions.ts
```diff
@@ -30,21 +32,54 @@ export const actions: ActionTree<ADRUploadState, RootState> & ADRUploadActions =
...
+                        const uploadFiles: Dict<UploadFile> =  {
+                            outputZip: constructUploadFileWithResourceName(
                                 metadata,
                                 0,
                                 schemas.outputZip,
                                 `${project.name}_naomi_outputs.zip`,
-                                "uploadFileOutputZip"),
-                            outputSummary: constructUploadFile(
+                                "uploadFileOutputZip",
+                                `${project.name} Naomi Outputs`),
```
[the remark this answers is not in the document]
EmmaLRussell: We should revisit the resource names when when we do https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2421 - should be able to avoid renaming the resources when we update them.

src/app/static/src/app/store/adrUpload/actions.ts
```diff
@@ -30,21 +32,54 @@ export const actions: ActionTree<ADRUploadState, RootState> & ADRUploadActions =
...
+                                `${project.name} Naomi Summary`)
                         };
 
+                        if (switches.adrPushInputs) {
+                            const addLocalInputFileToUploads = (
+                                key: string,
+                                schema: string,
+                                response: ValidateInputResponse,
+                                displayName: string) => {
+                                if (!response.fromADR) {
```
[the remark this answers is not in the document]
EmmaLRussell: This is where it decides whether or not to include each input file based on whether it's come straight from ADR or has been uploaded (edited, we assume) from the user's local machine. So files fromADR aren't make available for user to upload to ADR. This is method for all each file type from line 73 onwards.

src/app/static/src/app/utils.ts
```diff
@@ -198,8 +200,28 @@ export const datasetFromMetadata = (id: string, datasets: any[], schemas: ADRSch
...
+        return getUploadFileFromResource(resource, resourceName, index, resourceType, resourceFilename, displayName);
+    } else {
+        return null;
+    }
+};
+
+export const constructUploadFileWithResourceName = (datasetWithResources: any, index: number, resourceType: string,
+                             resourceFilename: string, displayName: string, resourceName: string): UploadFile => {
+    const resource = findResource(datasetWithResources, resourceType, resourceName);
+    return getUploadFileFromResource(resource, resourceName, index, resourceType, resourceFilename, displayName);
```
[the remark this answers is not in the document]
EmmaLRussell: Yes.

mwoodbri (approved):
Thumbs up from me!

mwoodbri (approved):

carbowizzy (changes requested):
Works nicely. However, I believe there should be a header separating input from output files as the input files just appear at the bottom of the modal. A regular user will not be able to differentiate output from input files and may be confused as to which files to check when uploading to ADR.

EmmaLRussell:
I added a bit of code to the component to separate input and output files, and to show separate headers for both only if there are any input files to upload. Tests now passing and input files seem to be uploading ok to dev.

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -61,7 +66,10 @@
 
     interface Computed {
         dataset: string
-        uploadFiles: Dict<UploadFile>
+        uploadFiles: Dict<UploadFile>,
+        uploadFileSections: Array<Dict<UploadFile>>,
+        outputUploadFiles: Dict<UploadFile>,
+        inputUploadFiles: Dict<UploadFile>
```
mwoodbri, line 72: Are these hangers-on?

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -7,22 +7,27 @@
...
-                        <input class="form-check-input"
-                               type="checkbox"
-                               :value="key"
-                               v-model="uploadFilesToAdr"
-                               :id="`id-${index}`">
+                <div v-for="(uploadFileSection, sectionIndex) in uploadFileSections" :key="sectionIndex">
+                    <h5 v-if="Object.keys(uploadFileSections[1]).length > 0"
+                        v-translate="sectionIndex === 0 ? 'outputFiles' : 'inputFiles'"
+                        class="mt-3"></h5>
+                    <div id="output-file-id" class="mt-3" v-for="(uploadFile, key, index) in uploadFileSection" :key="uploadFile.index">
```
mwoodbri, line 14: Looks like this will produce multiple HTML elements with the same `id`?

mwoodbri (changes requested):
This is very slick and definitely a worthy enhancement (kudos @carbowizzy too!).

Just spotted a couple of tiny issues.

src/app/static/src/app/components/downloadResults/UploadModal.vue
EmmaLRussell: Oops, yes, relics of an earlier attempt.

mwoodbri (approved):

carbowizzy (approved):

PR #488, opened 2021-04-08
mrc-2274 Rebuild ADR uploadFiles with new resourceIds after upload new files

Nicolas-Dolan:
## Description

## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -133,7 +136,9 @@
         },
         watch: {
             uploadFiles() {
-                this.setDefaultCheckedItems()
+                if (!this.defaultSet){
```
mwoodbri, line 139: If you really do want to select the zip and html files only when they are first added then it would be cleaner to do something like this, avoiding need for `defaultSet` variable:
```kotlin
uploadFiles(newValue, oldValue) {
    if (!oldValue) {
        this.setDefaultCheckedItems()
    }
}
```
Alternatively you could make setDefaultCheckedItems re-runnable... something like this:
```kotlin
setDefaultCheckedItems() {
    this.uploadFilesToAdr = ["outputZip", "outputSummary"].filter(key => this.uploadFiles.hasOwnProperty(key))
}
```

src/app/static/src/app/store/adr/actions.ts
```diff
@@ -144,32 +144,49 @@ export const actions: ActionTree<ADRState, RootState> & ADRActions = {
...
-        const {state, rootState, commit} = context;
+        const {state, rootState, commit, dispatch} = context;
         const selectedDatasetId = rootState.baseline.selectedDataset?.id;
         const modelCalibrateId = rootState.modelCalibrate.calibrateId;
-        commit({type: ADRMutation.ADRUploadStarted, payload: uploadFilesPayload.length});
 
-        for (let i = 0; i < uploadFilesPayload.length; i++) {
-            commit({type: ADRMutation.ADRUploadProgress, payload: i + 1});
-            const { resourceType, resourceFilename, resourceId } = uploadFilesPayload[i]
+        if (selectedDatasetId){
```
mwoodbri, line 151: Could you explain why this check is necessary? (Not saying it's wrong - just not obvious!)

src/app/static/src/app/store/adr/actions.ts
```diff
@@ -144,32 +144,49 @@ export const actions: ActionTree<ADRState, RootState> & ADRActions = {
...
+                .ignoreSuccess()
+                .get(`/adr/datasets/${selectedDatasetId}`)
+                .then((response) => {
+                    if (response) {
+                        datasets = [response.data];
+                    }
+                });
+    
+            const regenDataset = datasetFromMetadata(selectedDatasetId, datasets, state.schemas!);
+            commit(`baseline/${BaselineMutation.SetDataset}`, regenDataset, {root: true});
```
mwoodbri, line 188: Some obvious duplication with `getUserCanUpload()`... would it make sense to share any of this code? Could relate to next point too.

src/app/static/src/app/store/adr/actions.ts
```diff
@@ -144,32 +144,49 @@ export const actions: ActionTree<ADRState, RootState> & ADRActions = {
...
+            dispatch("getUploadFiles");
```
mwoodbri, line 189: I'm not completely of the right approach to this, but feels like the DownloadResults page should be the one dispatching this action, based on it watching changes to the baseline state. Otherwise we could end up with a tangle of actions calling other actions. Any thoughts on thiis @EmmaLRussell?

mwoodbri (changes requested):
This works nicely but I have a couple of concerns about implementation - please see comments below

src/app/static/src/app/store/adr/actions.ts
Nicolas-Dolan: Initially I had it wrapped around just the api request for the datasets, but then it occurred to me that the api request posting the upload files needed it too, so I felt the whole action should probably depend on it.

src/app/static/src/app/store/adr/actions.ts
EmmaLRussell: I think by and large I'd rather have actions dispatching actions which should always always follow on from them, rather than relying on components to have that knowledge - the exception would be a case where for some reason the context of a particular component's call would require the follow-on action, but in other context it would not be required. We already use this pattern elsewhere (e.g. in baseline actions related to uploads).

Completely agree that we should avoid it becoming a tangle though and limit length of action 'chains' where possible.

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# hint 1.29.1
```
EmmaLRussell, line 1: Re-used version number

src/app/static/src/app/store/adr/actions.ts
mwoodbri: Ok - makes sense, thanks. The thought arose because in reality many of these actions are specific to a particular component, so when the actions call each other it appears to be coupling the components. But the alternative would mean more action at a distance - which isn't necessarily easier to manage.

src/app/static/src/app/store/adr/actions.ts
EmmaLRussell: Yeah, feels like this could at least be an async method, called from here and getUserCanUpload, which takes a selectedDatasetId, regenerates the dataset and commits the mutation. getUserCanUpload should then be able to read the org Id it needs from the newly mutated state.

NEWS.md
EmmaLRussell: Oh sorry, it's not re-used is it, you've set patch version. OK, fair enough, I guess this really a bug fix.

src/app/static/src/app/store/adr/actions.ts
mwoodbri: I guess my question is can `selectedDatasetId` ever actually be unset at this point?

Nicolas-Dolan: Possibly not. I was partly influenced by the behaviour of the `getUserCanUpload` action, which makes this check, and partly just trying to make the intellisense happy

mwoodbri: Ok. `getUserCanUpload` is different because it's expected to be called when the dataset isn't set. This might all seem a bit irrelevant but I think it's better to remove the check here because it could actually mask errors in the future. You can do so and then just change the earlier line to:
```typescript
const selectedDatasetId = rootState.baseline.selectedDataset!.id;
```

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -112,12 +112,8 @@
...
             },
             setDefaultCheckedItems: function () {
-                const uploadFilesKeys = ["outputZip", "outputSummary"]
-                uploadFilesKeys.forEach(key => {
-                    if (this.uploadFiles.hasOwnProperty(key)) {
-                        this.uploadFilesToAdr.push(key)
-                    }
-                })
+                this.uploadFilesToAdr = ["outputZip", "outputSummary"]
+                    .filter(key => this.uploadFiles.hasOwnProperty(key))
```
mwoodbri, line 116: Looks like we're lacking a test that relies on this, but should be easy to remedy?

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -72,7 +72,7 @@
         uploadFiles: Dict<UploadFile>
     }
 
-    interface Data {
+    interface Data {
```
mwoodbri, line 75: I think we can do without changes that introduce extraneous whitespace

mwoodbri (changes requested):
This worked very well for me - the dialog box was updated instantaneously. And I know there's an issue with Buildkite but all tests passed locally

Would just like to see a test that the right checkboxes are ticked when `uploadFiles` changes - I'm not sure this exists at the moment? Codecov is complaining at least.

EmmaLRussell (commented):
Works for me, the shared method is definitely an improvement. Don't have anything to add to Mark's review.

src/app/static/src/app/components/downloadResults/UploadModal.vue
Nicolas-Dolan: There is a test called "checkboxes are set by default" but I don't know how to test a change in uploadFiles as it comes from the adr state.

Nicolas-Dolan: I believe I've found a solution now.

mwoodbri (approved):

PR #489, opened 2021-04-09
mrc-2300 Improve performance of ADR uploads

mwoodbri:
## Description

Update Fuel to version [2.2.x](https://github.com/kittinunf/fuel/releases/tag/2.2.1) to incorporate fix for kittinunf/fuel#667

Also remove redundant dependencies

To test: upload Burundi model output zip before and after applying this patch. On my machine it reduces from ~10 mins to ~1 min.

## Type of version change

Patch

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.30.0
+
+* Improve performance of ADR uploads
+
 # hint 1.29.0
 
 * Progress feedback on uploading files to ADR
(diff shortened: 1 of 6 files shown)


EmmaLRussell (approved):
Burundi was actually pretty quick for me even before (maybe because I'm working remotely so uploading with Imperial-sized wifi) - but this did still seem to make a noticeable improvement.

mwoodbri:
> Burundi was actually pretty quick for me even before (maybe because I'm working remotely so uploading with Imperial-sized wifi) - but this did still seem to make a noticeable improvement.

Thanks for checking - my connection is quite slow and very asymmetric, which might explain the big speed-up for me.

PR #492, opened 2021-04-27
mrc-2322 Prevent ADR upload of output files from overwriting those from other projects

mwoodbri:
## Description

Step to reproduce original bug:
1. Create a project `project1` in HINT and upload output zip to ADR. Check visible in ADR.
2. Create a project `project2` using same ADR dataset and upload output zip.

What you expect to see:
- Resources `project1_naomi_outputs.zip` and `project2_naomi_outputs.zip` in the relevant dataset i.e. it's possible to have one resource per type per project.

What you actually see:
- `project1_naomi_outputs.zip` is replaced by `project2_naomi_outputs.zip` because we populate `resourceId` by looking for the first resource of the relevant type, instead of also comparing by name. This means you can only have one resource per type, across all projects. This is revealed by the warning in the upload dialog that the file already exists, even if you're now in a different project.

This PR only populates `resourceId` if name also matches, meaning a new resource will be created where relevant.

Note that this fix will conflict slightly with mrc-2279, which populates ADR's `name` field with `resourceName` instead of `resourceFilename`. If this PR is merged first then that one can choose whether to match `resourceName` with ADR's `name` field or to match `resourceFilename` with the suffix of the ADR `url` field. Both would have the same effect.

## Type of version change

Minor

## Checklist


EmmaLRussell (approved):
This looks good to me, notwithstanding buildkite weirdness. Good spot on the relevance to mrc-2279. I think what we should do there is to match on name, because when we upload input files, we may change the file name (should be the same as the local file), but the input file name should be fixed. I'll make a note on the other PR.

Nicolas-Dolan (approved):
It works well with both files for both projects appearing in adr. I still got the premature upload complete message and file already exists warning when I was uploading the files for the second project though; I guess that will be covered by another ticket.

mwoodbri:
> It works well with both files for both projects appearing in adr. I still got the premature upload complete message and file already exists warning when I was uploading the files for the second project though; I guess that will be covered by another ticket.

Thanks Nick. Not sure what you mean by "premature upload complete message and file already exists warning". If this is a pre-existing bug then please open a ticket if one doesn't exist already.

Nicolas-Dolan:
> Thanks Nick. Not sure what you mean by "premature upload complete message and file already exists warning". If this is a pre-existing bug then please open a ticket if one doesn't exist already.

When I reached the upload page for the second project, it said upload complete even though I had not uploaded any files for that project yet. It was presumably still counting the state from the uploads of the first project.

mwoodbri:
> When I reached the upload page for the second project, it said upload complete even though I had not uploaded any files for that project yet. It was presumably still counting the state from the uploads of the first project.

Ah ok, that's mrc-2364.

PR #493, opened 2021-04-27
mrc-2301 Bug fix: Importing from adr after fitting model should prompt new version behaviour

Nicolas-Dolan:
## Description

Currently if you import from adr and then move through the later steps (selecting model options, fitting, etc) and then go back and edit the selected dataset and import from adr again, it will wipe the later steps without any warning. Expected behaviour would be for the save new version dialogue to appear when you attempt to import a new version and a new version to be created for these changes.

## Type of version change
Patches

## Checklist


src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -528,6 +544,37 @@ describe("select dataset", () => {
...
+            "La modification de l'ensemble de données sélectionné entraînera la suppression des modifications de toutes les étapes suivantes.",
+            store);
+        expectTranslated(rendered.findAll("p").at(1),
+            "These steps will automatically be saved in a version. You will be able to reload this version from the Projects page.",
+            "Ces étapes seront automatiquement sauvegardées dans une version. Vous pourrez recharger cette version depuis la page Projets.",
+            store);
+        const buttons = rendered.findAll("button");
+        expectTranslated(buttons.at(3), "Save version and keep editing",
+            "Sauvegarder la version et continuer à modifier", store);
+        expectTranslated(buttons.at(4), "Cancel editing", "Annuler l'édition", store);
```
Nicolas-Dolan, line 576: I probably don't need all of this

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -528,6 +544,37 @@ describe("select dataset", () => {
         expect(rendered.find("#loading-dataset").exists()).toBe(false);
         expect(rendered.find(Modal).props("open")).toBe(false);
     });
+    
+    it("renders reset confirmation dialog when changing selected dataset", async () => {
```
Nicolas-Dolan, line 548: I wanted to add a test to check that the modal was rendered once the user clicked "Save version and keep editing" in the confirmation dialog but this required me to set up so much state for the reset confirmation component that it essentially became an integration test. I don't see any integration tests for reset confirmation or selected dataset, unless I'm missing something. Let me know if you think if I should add something along these lines in.

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -23,7 +23,7 @@
         <button
             class="btn btn-red"
             :class="selectedDataset && 'ml-2'"
-            @click="toggleModal"
+            @click="selectedDataset? confirmEditing($event) : toggleModal()"
```
mwoodbri, line 26: Curious about the inconsistency here... why do we pass `$event` to `confirmEditing()` but not `toggleModal()`? Why does one have `preventDefault()` but not the other?

More generally, this feels slightly convoluted. Would be keen to hear others' thoughts, but maybe `toggleModal()` should be replaced by `openModal()` and `closeModal()`, where the former includes a check for whether a dataset is already selected?

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -309,6 +320,17 @@
             toggleModal() {
                 this.open = !this.open;
             },
+            confirmEditing(e: Event) {
+                e.preventDefault();
+                this.showConfirmation = true;
+            },
+            continueEditing() {
+                this.showConfirmation = false;
+                this.open = true;
```
mwoodbri, line 329: Looks like this path isn't covered by tests

mwoodbri (commented):
This works well. The approach is maybe a bit over-zealous: if I select a dataset and then immediately change my mind and select another then I end up with a warning and a new project version being created. But best to be conservative I guess - I'm not sure what how we could classify changes to projects as trivial vs meaningful.

I have added an inline comment regarding implementation. Would be good to at least fix the inconsistency there (unless it's warranted), but my other point is less clear-cut.

src/app/static/src/tests/components/adr/selectDataset.test.ts
EmmaLRussell: I think it's fine for testing this text to go into the render confirmation dialog tests (and you have added that test), and for this test just to check that the dialog component is rendered with the corrrect props, e.g. that `open` is set to `true` when expected

EmmaLRussell (changes requested):
This works really nicely. :+1:  I think it's ok to be trigger happy about warning on dataset change. 

However I don't think we want a completely different message displayed for this change, and should use the existing behaviour of the dialog instead. There's an argument that your way is more descriptive to the user, but we could say the same thing about having special messages for all the changes which prompt this dialog. I think we want to keep the dialog as generic a we can.

Nicolas-Dolan:
> This works well. The approach is maybe a bit over-zealous: if I select a dataset and then immediately change my mind and select another then I end up with a warning and a new project version being created. But best to be conservative I guess - I'm not sure what how we could classify changes to projects as trivial vs meaningful.

@mwoodbri  This was an oversight on my part that I've fixed by adding editsRequireConfirmation clause to the openModal method.
```
openModal(){
                if (this.selectedDataset && this.editsRequireConfirmation){
                    this.showConfirmation = true;
                } else {
                    this.open = true;
                }
            },
```

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -183,6 +195,21 @@ describe("select dataset", () => {
...
+                errors: {
+                    namespaced: true,
+                    state: mockErrorsState(),
+                },
+                projects: {
+                    namespaced: true,
+                    state: mockProjectsState({currentProject: {id: 1, name: "v1", versions: []}, currentVersion}),
+                    actions: {
+                        newVersion: jest.fn()
+                    }
```
mwoodbri, line 211: I think you could just import the real projects actions and use them here

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -528,6 +555,76 @@ describe("select dataset", () => {
...
+    it("renders reset confirmation dialog when changing selected dataset and then modal if click save", async () => {
+        let store = getStore({
+            selectedDataset: fakeDataset
+        },
+            {datasets: [{...fakeRawDatasets[0], ...fakeRawDatasets[1], resources: [shape]}]}, true
+        )
+        const rendered = mount(SelectDataset, {
+            store, stubs: ["tree-select"]
+        });
+        rendered.find("button").trigger("click");
```
mwoodbri, line 568:
```suggestion
        await rendered.find("button").trigger("click");
```

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -528,6 +555,76 @@ describe("select dataset", () => {
...
+
+        await Vue.nextTick();
```
mwoodbri, line 570: Please see whether `await`ing above enables you to remove this `Vue.nextTick()` - I didn't need any when I was playing with this.

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -528,6 +555,76 @@ describe("select dataset", () => {
...
+
+        expect(rendered.find(ResetConfirmation).exists()).toBe(true);
```
mwoodbri, line 572: This is redundant, given the following line

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -528,6 +555,76 @@ describe("select dataset", () => {
...
+        expect(rendered.find(ResetConfirmation).props("open")).toBe(true);
+        expect(rendered.find(Modal).props("open")).toBe(false);
+        const buttons = rendered.findAll("button");
+        expectTranslated(buttons.at(3), "Save version and keep editing",
+            "Sauvegarder la version et continuer à modifier", store);
+        buttons.at(3).trigger("click");
```
mwoodbri, line 578: If this line duplicates the next one then it should be removed

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -528,6 +555,76 @@ describe("select dataset", () => {
...
+        await buttons.at(3).trigger("click");
+        store.state.projects.currentVersion = {id: "id1"} as any;
+        await Vue.nextTick();
```
mwoodbri, line 581: Again, not sure this is really necessary?

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -528,6 +555,76 @@ describe("select dataset", () => {
...
+            selectedDataset: fakeDataset
+        },
+            {datasets: [{...fakeRawDatasets[0], ...fakeRawDatasets[1], resources: [shape]}]}
+        )
+        const rendered = mount(SelectDataset, {
+            store, stubs: ["tree-select"]
+        });
+        rendered.find("button").trigger("click");
+
+        await Vue.nextTick();
```
mwoodbri, line 597:
```suggestion
        await rendered.find("button").trigger("click");
```

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -528,6 +555,76 @@ describe("select dataset", () => {
...
+            selectedDataset: fakeDataset
+        },
+            {datasets: [{...fakeRawDatasets[0], ...fakeRawDatasets[1], resources: [shape]}]}, true
+        )
+        const rendered = mount(SelectDataset, {
+            store, stubs: ["tree-select"]
+        });
+        rendered.find("button").trigger("click");
+
+        await Vue.nextTick();
```
mwoodbri, line 614:
```suggestion
        await rendered.find("button").trigger("click");
```

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -528,6 +555,76 @@ describe("select dataset", () => {
...
+
+        expect(rendered.find(ResetConfirmation).exists()).toBe(true);
+        expect(rendered.find(ResetConfirmation).props("open")).toBe(true);
+        expect(rendered.find(Modal).props("open")).toBe(false);
+        const buttons = rendered.findAll("button");
+        expectTranslated(buttons.at(4), "Cancel editing",
+            "Annuler l'édition", store);
+        buttons.at(3).trigger("click");
```
mwoodbri, line 622: Why are we clicking this button?

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -528,6 +555,76 @@ describe("select dataset", () => {
...
+        await buttons.at(4).trigger("click");
+        await Vue.nextTick();
```
mwoodbri, line 624: Can this be removed?

mwoodbri (changes requested):
This looks good to me, bar some superfluous(?) button clicking in the tests.

I'm not too bothered about the `Vue.nextTick()`s... except if they're making up for missing `await`s when triggering button clicks. In that case they should be removed if possible, because they're just visual noise.

src/app/static/src/tests/components/adr/selectDataset.test.ts
Nicolas-Dolan: Not quite sure what you are referring to here

src/app/static/src/tests/components/adr/selectDataset.test.ts
Nicolas-Dolan: I've now removed

src/app/static/src/tests/components/adr/selectDataset.test.ts
mwoodbri: Lines 573 and 574 are identical... is that intentional?

mwoodbri (changes requested):

mwoodbri (approved):

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -70,18 +70,22 @@
...
                 ></button>
             </template>
         </modal>
+        <reset-confirmation v-if="showConfirmation"
+                            :continue-editing="continueEditing"
+                            :cancel-editing="cancelEditing"
+                            :open="showConfirmation"></reset-confirmation>
     </div>
 </template>
 <script lang="ts">
```
carbowizzy, line 83: Remove unused import

carbowizzy (changes requested):
Looks good to me and works nicely. However, the dialog is prompted prematurely. This dialog should appear when user triggers the `import` button on the "select ADR dataset" modal. It has to be this way because a user can decide to cancel the modal without importing anything whereas another version of the project would have been created.

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -114,6 +120,8 @@
         markResourcesUpdated: () => void;
         startPolling: () => void;
         stopPolling: () => void;
+        continueEditing: () => void
+        cancelEditing: () => void;
```
mwoodbri, line 124: Inconsistent semicolons

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -306,8 +318,22 @@
 
                 this.startPolling();
             },
-            toggleModal() {
-                this.open = !this.open;
+            openModal(){
+                if (this.selectedDataset && this.editsRequireConfirmation){
```
EmmaLRussell, line 322: I'm not convinced you need this check on selectedDataset - is editsRequireConfirmation not sufficient?

EmmaLRussell (commented):
I'm somewhat agnostic about the behaviour on this one since we're already inconsistent when other values are changed. If you select a new file, it waits until you select and OK it before showing the dialog. If you change a model option it warns as soon as you click the input. 

I agree that it would be preferable to wait until the user has selected and clicked the button before warning but I don't think it's a showstopper, particularly since it does wait until after the import to wipe the values in subsequent steps. I'd probably change it unless that's going to be horrendously difficult for some reason.

Nicolas-Dolan:
Behaviour of ticket has now been amended so that reset confirmation dialogue triggers when the user clicks the import button on a new selected dataset. Clicking 'Save version and keep editing' in the dialogue will trigger the new version behaviour and then the import new dataset behaviour.

Note: this ticket is a little noisy due to my auto formatter. The changes generally make sense but perhaps some of the changes that spread out code over multiple lines could be considered unnecessary. Let me know if you have any objections.

Nicolas-Dolan:
> > Note: this ticket is a little noisy due to my auto formatter. The changes generally make sense but perhaps some of the changes that spread out code over multiple lines could be considered unnecessary. Let me know if you have any objections.
> 
> Tidying up the formatting of the lines you're editing (and possibly the immediate vicinity) is generally a good thing - making the code much easier to review if it introduces consistency. But in this case I'm really struggling to find the lines I should be reviewing i.e. the ones that have seen more than superficial changes. If you're not using IDEA already then you should consider it, or at least ensure that in future PRs you only allow your editor to reformat lines that you've deliberately changed. We just need to make the review process is as efficient as possible for everyone!

I know what you mean. I'll investigate whether I can configure my IDE to only format lines I've worked on.

src/app/static/src/app/components/ResetConfirmation.vue
```diff
@@ -6,7 +6,7 @@
 
             <p v-translate="'discardWarning'"></p>
             <ul>
-            <li v-for="step in changesToRelevantSteps" :key="step.number">
```
mwoodbri, line 9: Please remove this file from this PR - the change doesn't seem relevant to the ticket in question (see [PR guidelines](https://mrc-ide.myjetbrains.com/youtrack/articles/RESIDE-A-13/Pull-Request-guidelines))

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -309,21 +384,38 @@
             toggleModal() {
                 this.open = !this.open;
             },
+            confirmImport() {
+                if (this.newDatasetId && this.editsRequireConfirmation) {
```
mwoodbri, line 388: Is it possible to click "Import" and trigger this method if `newDataId` isn't set? (I think @EmmaLRussell may already have raised this)

mwoodbri (changes requested):
Behaviour is much improved - good stuff. Aside from the earlier comment on unnecessary reformatting there looks to be an unaddressed comment from previous review and a superfluous file in the PR. These could probably have been picked up by a final check through before re-requesting review.

src/app/static/src/app/components/adr/SelectDataset.vue
EmmaLRussell: It shouldn't be - though I've just noticed we have a similar clause in importDataset so I guess it's consistent with existing code. We probably shouldn't have a policy of not trusting our own logic!

EmmaLRussell (commented):
I have to agree with Mark - this is really hard to review. 'Hunt the relevant code change'! Could you please see if you can get your auto-formatter to stop auto-formatting!

The behaviour looks good - I think it would be even better if cancelling from the confirmation dialog also closed the import dialog - I think we can assume we've scared the user off from editing at that point, and that they'd expect clicking 'Cancel editing' to close both dialogs. Would this just be a case of calling `toggleModal()` from `cancelEditing`? 

As I said, I think we can lose that newDatasetId check here, and probably also in importDataset - user shouldn't be able to invoke these if no dataset is selected.

src/app/static/src/app/components/adr/SelectDataset.vue
Nicolas-Dolan: I think you are referring to a slightly different situation previously involving `selectedDataset`. Technically, the user shouldn't be able to click the button if `newDatasetId` is not set as the button will be disable; however, I noticed that the `importDataset` method does check for `newDatasetId` so I was just following the existing pattern.

Nicolas-Dolan:
> As I said, I think we can lose that newDatasetId check here, and probably also in importDataset - user shouldn't be able to invoke these if no dataset is selected.

So apparently the check for `newDatasetId` was a type guard to appease ts. To remove it, I've had to add `as string` to a parameter. It's a pick your poison scenario.

mwoodbri:
> > As I said, I think we can lose that newDatasetId check here, and probably also in importDataset - user shouldn't be able to invoke these if no dataset is selected.
> 
> So apparently the check for `newDatasetId` was a type guard to appease ts. To remove it, I've had to add `as string` to a parameter. It's a pick your poison scenario.

Ah! I think `!` would do the trick too, and might be clearer in it's intent i.e. asserting non-null rather than casting something to a string.

mwoodbri (approved):

EmmaLRussell (approved):

carbowizzy:
> Looks good to me and works nicely. However, the dialog is prompted prematurely. This dialog should appear when a user triggers the `import` button on the "select ADR dataset" modal. It has to be this way because a user can decide to cancel the modal without importing anything whereas another version of the project would have been created, so your dialog confirmation should appear when an import button is triggered.

This issue still persists, am I missing something ? This behaviour hasn't changed. When I clicked on edit selected dataset save version dialog appears immediately. A user should be able to see dropdown list of datasets and click on import button before displaying save version dialog.

carbowizzy (changes requested):
Lots of improvement , I think you have not attended to my earlier comment.

mwoodbri:
> > Looks good to me and works nicely. However, the dialog is prompted prematurely. This dialog should appear when a user triggers the `import` button on the "select ADR dataset" modal. It has to be this way because a user can decide to cancel the modal without importing anything whereas another version of the project would have been created, so your dialog confirmation should appear when an import button is triggered.
> 
> This issue still persists, am I missing something ? This behaviour hasn't changed. When I clicked on edit selected dataset save version dialog appears immediately. A user should be able to see dropdown list of datasets and click on import button before displaying save version dialog.

@carbowizzy I only see the "Save version" dialog when I click Import, and not before. Could you ensure your branch is up-to-date before re-checking, and post a detailed list of steps to reproduce if the problem persists?

carbowizzy:
> > > Looks good to me and works nicely. However, the dialog is prompted prematurely. This dialog should appear when a user triggers the `import` button on the "select ADR dataset" modal. It has to be this way because a user can decide to cancel the modal without importing anything whereas another version of the project would have been created, so your dialog confirmation should appear when an import button is triggered.
> > 
> > 
> > This issue still persists, am I missing something ? This behaviour hasn't changed. When I clicked on edit selected dataset save version dialog appears immediately. A user should be able to see dropdown list of datasets and click on import button before displaying save version dialog.
> 
> @carbowizzy I only see the "Save version" dialog when I click Import, and not before. Could you ensure your branch is up-to-date before re-checking, and post a detailed list of steps to reproduce if the problem persists?

Brilliant, looks like I didn't pull updates.

carbowizzy (approved):

PR #498, opened 2021-05-11
mrc-2403 Correctly scope ADR mutations/actions in upload logic

mwoodbri:
## Description

* Fix error handling and dataset reload functionality that was broken by splitting ADR state.
* Set `ADRUploadError` rather than `ADRError`, as the former is scoped to (and displayed on) the upload page
* Easiest way to test is to observe (lack of) browser console messages when landing on upload page (when `ADRUploadError` is cleared via mutation), and after uploading (when `adr/getAndSetDatasets` is invoked).
* This could have been spotted via tests if it was checked that `getAndSetDatasets` in particular was actually executed (rather than the action being invoked on the mock). Or if we had an in-browser end-to-end test that checked for JS errors. But any practical suggestions very welcome here - I'm short on ideas.
* Decrease in coverage looks like a false positive related to removal of code

## Type of version change

Patches

## Checklist

### DIFF
--- a/src/app/static/src/app/store/adrUpload/actions.ts
+++ b/src/app/static/src/app/store/adrUpload/actions.ts
@@ -6,7 +6,6 @@ import {ADRUploadState} from "./adrUpload";
 import {ADRUploadMutation} from "./mutations";
 import {constructUploadFile} from "../../utils";
 import {Dict, UploadFile} from "../../types";
-import {ADRMutation} from "../adr/mutations";
 
 export interface ADRUploadActions {
     getUploadFiles: (store: ActionContext<ADRUploadState, RootState>) => void;
@@ -20,9 +19,9 @@ export const actions: ActionTree<ADRUploadState, RootState> & ADRUploadActions =
         const project = rootState.projects.currentProject;
 
         if (selectedDataset && project) {
-            context.commit({type: ADRMutation.SetADRError, payload: null});
+            commit({type: ADRUploadMutation.SetADRUploadError, payload: null});
             await api(context)
-                .withError(ADRMutation.SetADRError)
+                .withError(ADRUploadMutation.SetADRUploadError)
                 .ignoreSuccess()
                 .get(`/adr/datasets/${selectedDataset.id}`)
                 .then((response) => {
@@ -91,7 +90,7 @@ export const actions: ActionTree<ADRUploadState, RootState> & ADRUploadActions =
                 break
             }
         }
-        await dispatch("adr/getAndSetDatasets", selectedDatasetId);
-        dispatch("getUploadFiles");
+        await dispatch("adr/getAndSetDatasets", selectedDatasetId, {root: true});
+        await dispatch("getUploadFiles");
     }
 };
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.35.1
+
+* Correctly scope ADR mutations/actions in upload logic
+
 # hint 1.35.0
 
 * Roll-back making login email case insensitive (mrc-2210)
--- a/src/app/static/src/app/hintVersion.ts
+++ b/src/app/static/src/app/hintVersion.ts
@@ -1,1 +1,1 @@
-export const currentHintVersion = "1.35.0";
+export const currentHintVersion = "1.35.1";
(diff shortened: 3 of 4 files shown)


EmmaLRussell:
>     * This could have been spotted via tests if it was checked that `getAndSetDatasets` in particular was actually executed (rather than the action being invoked on the mock). Or if we had an in-browser end-to-end test that checked for JS errors. But any practical suggestions very welcome here - I'm short on ideas.
> 

Yeah, it's a good point - the connection between the actions and mutations is a blind spot in our tests, and even an integration test doesn't help here as that just tests that the actions can talk to the backend ok - if the tests and code are under the same misapprehension about mutation names they'll spuriously pass. 

As you say we could have a full end to end browser test to pick up these kinds of integration failures. `getAndSetDatasets` is kind of an edge case too, since that only really needs to be invoked to guard against changes to the datasets on ADR (and, I think, also in case the project has been reloaded since the baseline dataset was fetched, in order to repopulate it). We could certainly make a browser test for the latter scenario, though it's a little involved. 

Another option might be to explicitly have an integration test of actions, and mutations similar to what we've ended up doing for some components where we actually create a real store with real actions and mutations and hook that up to the component. Something like what we have for the stepper component: https://github.com/mrc-ide/hint/blob/master/src/app/static/src/tests/components/stepper.test.ts#L65
So you could give the action the real store when you make the test call rather than some mocked `{commit}` and then check if the state has been updated as expected by the expected mutation, or that mockAxios has been invoked by the child action you expected to be called.

src/app/static/src/app/store/adrUpload/actions.ts
```diff
@@ -91,7 +90,7 @@ export const actions: ActionTree<ADRUploadState, RootState> & ADRUploadActions =
                 break
             }
         }
-        await dispatch("adr/getAndSetDatasets", selectedDatasetId);
-        dispatch("getUploadFiles");
+        await dispatch("adr/getAndSetDatasets", selectedDatasetId, {root: true});
```
EmmaLRussell, line 93: Could add a test for the new `{root: true}` parameter.

src/app/static/src/app/store/adrUpload/actions.ts
```diff
@@ -91,7 +90,7 @@ export const actions: ActionTree<ADRUploadState, RootState> & ADRUploadActions =
...
+        await dispatch("getUploadFiles");
```
EmmaLRussell, line 94: Doesn't really matter, but why add this `await`?

mwoodbri:
> Another option might be to explicitly have an integration test of actions, and mutations similar to what we've ended up doing for some components where we actually create a real store with real actions and mutations and hook that up to the component. Something like what we have for the stepper component: https://github.com/mrc-ide/hint/blob/master/src/app/static/src/tests/components/stepper.test.ts#L65

Many thanks - this is really helpful advice. I might have a go at this during next sprint...

src/app/static/src/app/store/adrUpload/actions.ts
mwoodbri: Simply to silence IDE warnings. But even if it's fire-and-forget in this case (is there an alternative idiom?) I guess it's a good habit for cases when it does matter...

src/app/static/src/app/store/adrUpload/actions.ts
mwoodbri: Done

mwoodbri:
> Another option might be to explicitly have an integration test of actions, and mutations similar to what we've ended up doing for some components where we actually create a real store with real actions and mutations and hook that up to the component. Something like what we have for the stepper component: https://github.com/mrc-ide/hint/blob/master/src/app/static/src/tests/components/stepper.test.ts#L65

This is done in de3937d. Rather than checking that the store is updated and/or that relevant requests are made I'm just checking that the relevant actions are invoked. I got the former working but it effectively duplicated tests for the relevant actions and would ultimately be quite fragile as it was presuming details of their implementations.

I've done this a new test as it's quite different to existing one in that it's emulating a full store - and hopefully a good reference on how to do this.

Nicolas-Dolan (approved):
Works well. Silly question, but do only commits get logged to the console, not dispatches?

EmmaLRussell (approved):
Looks like a good approach to testing these chained actions. Ready to merge once the version conflicts are fixed.

PR #499, opened 2021-05-12
Run tests against sandbox ADR

r-ash:
This is only for running tests and shouldn't be merged


r-ash:
Changes now on staging ADR so this no longer needed

PR #500, opened 2021-05-20
mrc-2424 Remove usage of old calibrate endpoing

EmmaLRussell:
## Description

The old `/calibrate` endpoint has been removed from hintr and, while we weren't using it in the app any more, we did still have an method in the api client, and an old endpoint and controller method which used it, so the tests for these are now failing. Remove all references to these. Version not updated as no app behaviour changes. 

## Type of version change
_Delete as appropriate_

None

## Checklist

### DIFF
--- a/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ModelRunTests.kt
+++ b/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ModelRunTests.kt
@@ -73,7 +73,7 @@ class ModelRunTests : SecureIntegrationTests()
     fun `can submit calibrate`()
     {
         val entity = getModelRunEntity()
-        val responseEntity = testRestTemplate.postForEntity<String>("/model/calibrate/1234", entity)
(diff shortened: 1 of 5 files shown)


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ModelRunTests.kt
```diff
@@ -73,7 +73,7 @@ class ModelRunTests : SecureIntegrationTests()
     fun `can submit calibrate`()
     {
         val entity = getModelRunEntity()
-        val responseEntity = testRestTemplate.postForEntity<String>("/model/calibrate/1234", entity)
+        val responseEntity = testRestTemplate.postForEntity<String>("/model/calibrate/submit/1234", entity)
```
EmmaLRussell, line 76: Turns out this test wasn't actually testing submit calibrate, but it is now!

mwoodbri (approved):

PR #502, opened 2021-05-26
Mrc-2383 - Add notes text area to confirmation dialog when editing

carbowizzy:
When any changes discovered in the app, a dialog appears prompting the user to save as a new version and keep editing or cancel editing, under this dialog I have added a note text area where user can add/edit project notes.

## Type of version change
_Delete as appropriate_

Minor 

## Checklist


src/app/static/src/tests/components/resetConfirmation.test.ts
```diff
@@ -150,6 +150,41 @@ describe("Reset confirmation modal", () => {
...
+            propsData: {
+                continueEditing: mockContinueEdit,
+                cancelEditing: jest.fn()
+            },
+            store: createStore(mockNewVersion, {currentUser: 'test.user@example.com'})
+        });
+
+        await rendered.find("#projectNoteControl").setValue("new version note")
+        const textarea = rendered.find("#projectNoteControl").element as HTMLTextAreaElement;
+        expect(textarea.value).toBe("new version note")
```
mwoodbri, line 185: It's not obvious what this test is for - directly setting the value of the textarea and then checking the value doesn't seem to add anything over the "can render translated version note label" test.

mwoodbri (changes requested):
Worked well in my testing but a few questions + suggestions for you

src/app/static/src/app/store/projects/projects.ts
```diff
@@ -15,7 +15,8 @@ export interface ProjectsState {
     versionUploadPending: boolean,
     versionTime: Date | null,
     cloneProjectError: Error | null,
-    cloningProject: boolean
+    cloningProject: boolean,
+    updateNote: string | null
```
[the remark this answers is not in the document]
carbowizzy: Redundant now removed

src/app/static/src/app/components/ResetConfirmation.vue
```diff
@@ -59,32 +66,45 @@
 
     interface Data {
         waitingForVersion: boolean
+        versionNote: string
```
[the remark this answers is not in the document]
carbowizzy: renamed and more intuitive

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -389,6 +390,7 @@ const en: Translations = {
     projects: "Projects",
     projectsHeaderCreate: "Create a new project",
     projectsHeaderReturn: "return to current project",
+    projectNote: "Note (Your reason for saving as a new version)",
```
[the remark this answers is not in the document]
carbowizzy: Jeff agreed to have the revised text.

src/app/static/src/tests/components/resetConfirmation.test.ts
```diff
@@ -150,6 +150,33 @@ describe("Reset confirmation modal", () => {
...
+        const mockContinueEdit = jest.fn();
+        const mockNewVersion = jest.fn();
+        const rendered = mount(ResetConfirmation, {
+            propsData: {
+                continueEditing: mockContinueEdit,
+                cancelEditing: jest.fn()
+            },
+            computed: {
+                projectVersionNote() {
+                    return "textarea value"
```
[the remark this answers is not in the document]
EmmaLRussell: Agreed - this should just be a case of adding currentVersion to the mockProjectState used when creating the store.

EmmaLRussell (changes requested):
Works great. :+1: 

Could do with a test in `resetConfirmation.test.ts`that the note is passed to the `newVersion` action.

Nicolas-Dolan (commented):
This worked well and I couldn't find any bugs. I have a minor note about duplicating ids.

src/app/static/src/app/components/ResetConfirmation.vue
```diff
@@ -48,7 +53,9 @@
     interface Computed {
         changesToRelevantSteps: StepDescription[],
         currentVersionId: string | null,
-        errorsCount: number
+        errorsCount: number,
+        handleVersionNote: string,
+        currentVersionNote: string | null
```
mwoodbri, line 58: In what circumstance can `currentVersionNote` be null? The computed value either returns the current value if there is one, or the empty string.

src/app/static/src/app/components/ResetConfirmation.vue
```diff
@@ -59,32 +66,48 @@
...
+                return state.currentVersion?.note || "";
+            }),
             ...mapGetters(["isGuest"]),
             errorsCount: mapStateProp<ErrorsState, number>("errors", state => {
                 return state.errors ? state.errors.length : 0;
-            })
+            }),
+            handleVersionNote: {
+                get() {
+                    return this.currentVersionNote!
```
mwoodbri, line 94: Is `!` necessary? See comment above.

Produces a compiler warning.

src/app/static/src/app/components/ResetConfirmation.vue
```diff
@@ -59,32 +66,48 @@
...
             handleConfirm: function () {
                 if (this.isGuest) {
                     this.continueEditing();
                 } else {
                     this.waitingForVersion = true;
-                    this.newVersion();
+                    if(this.currentVersionNote && !this.versionNote) {
+                        this.versionNote = this.currentVersionNote
+                    }
+                    this.newVersion(this.versionNote)
```
mwoodbri, line 110: Let's be consistent with semicolons

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -372,6 +373,7 @@ const en: Translations = {
     noMapData: "No data are available for the selected combination. Please review the combination of filter values selected.",
     noneProvided: "none provided",
     noRecords: 'There are no records matching your request',
+    noteHeader: "Notes: (your reason for saving as a new version)",
```
mwoodbri, line 376: This would look more natural, but I'm not overly concerned:
```suggestion
    noteHeader: "Notes (your reason for saving as a new version):",
```

mwoodbri (changes requested):
Thanks for this. Some comments inline and a couple of bugs that will need fixes, tests etc:
* If you create a new version with a comment, and then create another version then the textarea is populated with the previous comment - all good. But if I delete the comment then the old comment still gets saved against the new version.
* If you add a comment such as `!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~` then things seem to go wrong - when you create another version then the dialog is only populated with `!"`. It wouldn't be unexpected for people to use random punctuation in notes, so we'll have to do something about this. Note sure if the note isn't being saved faithfully, or whether it's just rendered incorrectly.

src/app/static/src/app/store/translations/locales.ts
carbowizzy: The format I used was suggested by Jeff

src/app/static/src/app/components/ResetConfirmation.vue
carbowizzy: This is TypeScript, semicolons are  highly unnecessary .

src/app/static/src/app/store/translations/locales.ts
mwoodbri: D'oh, sorry - I should have checked mockups!

src/app/static/src/app/components/ResetConfirmation.vue
carbowizzy: > Thanks for this. Some comments inline and a couple of bugs that will need fixes, tests etc:
> 
> * If you create a new version with a comment, and then create another version then the textarea is populated with the previous comment - all good. But if I delete the comment then the old comment still gets saved against the new version.
> * If you add a comment such as `` !"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~ `` then things seem to go wrong - when you create another version then the dialog is only populated with `!"`. It wouldn't be unexpected for people to use random punctuation in notes, so we'll have to do something about this. Note sure if the note isn't being saved faithfully, or whether it's just rendered incorrectly.

I will investigate this regex like issue.

src/app/static/src/app/components/ResetConfirmation.vue
EmmaLRussell: Unnecessary, but we use them as standard, and we should really be consistent one way or the other.

mwoodbri (changes requested):
This is looking practically merge-ready. `ResetConfirmation` is much easier to understand now :+1: 

Just one issue I can see: if I type a note such as "changed scope + level" then the "'+" doesn't get saved. We need to ensure that it does, or if that is absolutely impossible then tell the user to amend their note.

mwoodbri (approved):

EmmaLRussell (approved):

PR #505, opened 2021-05-28
Mrc-2384 Add note text area to "copy to" dialog in project history

carbowizzy:
This PR adds note field to to copy modal in project history and update promote endpoints accordingly. When a user creates a project from an existing project version, notes text can be added to the project indicating why it was copied. The notes text will be attached to the newly copied project version.

## Type of version change
_Delete as appropriate_

 Minor 

## Checklist


src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ProjectsController.kt
```diff
@@ -80,9 +80,9 @@ class ProjectsController(private val session: Session,
         @RequestParam("name") name: String,
         @RequestParam("note") note: String?): ResponseEntity<String>
     {
-        val newProjectId = projectRepository.saveNewProject(userId(), name, note = note)
+        val newProjectId = projectRepository.saveNewProject(userId(), name)
```
mwoodbri, line 83: So, just to confirm: when creating a new project from an existing project version then the note attached to the version should be transferred to the first version of the new project. And the new project itself won't have its note populated, either from the original project or version?

carbowizzy: That's right. The project itself won't have its note populated but the version.

mwoodbri (commented):
This worked for me and code looks pretty sound: just a couple of comments/clarifications

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -273,7 +280,10 @@
             currentLanguage: mapStateProp<RootState, Language>(
                 null,
                 (state: RootState) => state.language
-            )
+            ),
+            projects: mapStateProp<ProjectsState, Project[] | null>(namespace, state => {
```
EmmaLRussell, line 284: I don't really understand why this is necessary - the ProjectsMixin was already providing projects.

EmmaLRussell (changes requested):
Works for me! :+1:

I know the integration test is testing the whole process, but could you also add db tests (that the VersionRepository saves the promoted version note, and that the ProjectRepository returns the version notes in getProjects.

EmmaLRussell (approved):

mwoodbri (approved):
Good stuff!

PR #506, opened 2021-06-02
mrc-2404 De-dupe HINT users with same case-insensitive email addresses

mwoodbri:
## Description

This is not intended to be merged, it's just a convenient way to structure the required one-off migration (though ideas about how to preserve this logic beyond retaining this branch welcome - maybe we could actually have a subproject for this purpose?)

Basic algorithm:
* If user has more than one account then assign account with most projects as primary, and merge secondary accounts into it
* If none of the user's accounts have projects then pick the one that is alphabetically last i.e. prefer lower case

Notes:
* There are three tables that have FK constraint on `USERS.USER_ID` (also see schema diagram on mrc-2404):
  * `PROJECT` - `USER_ID` updated, `SHARED_BY` not updated as no deleted accounts have previously shared projects
  * `USER_SESSION` - deprecated but updated to satisfy constraint
  * `ADR_KEY` - not updated but this is ok as it demonstrates that no deleted accounts have ADR keys configured
 * A manual check reveals that the account with most projects is always consistent with the account with the most recently updated project
 * We are not changing the username of any primary accounts, in particular we are not down-casing them. This feels like a less risky approach, especially given that it is now impossible to create new accounts with the same email address, and that we have already made sharing case-insensitive, and that we have a patch (previously reverted, but now safe to apply) to make login case insensitive. It also means that we don't need to worry about the opaque `USERS.SERIALIZEDPROFILE` column that is owned by pac4j.

To run against clone of production DB:
```
$ docker exec -i hint_db psql -U postgres hint <hint.pg_dump
$ src/gradlew -p src migrate:run
```

For output of this see ticket.

## Type of version change

None

## Checklist


r-ash (approved):
Thanks, this looks great

EmmaLRussell (approved):
Algorithm seems sensible to me, code looks good. We should be sure to take a backup before we start changing live db. 

Managed to test it locally by clobbering userCLI to let me add a user with existing username except for case, and migrate worked as expected, with primary user getting project belonging to secondary user. 

As you say, it would be good to preserve this migration somewhere other than just in this branch - don't have any very bright ideas really. We could merge it with the addition of scary warning messages and cancel option to unwary user, and in that case we'd probably want to rename it to something more specific to this ticket - 'migrate' is probably too easy to confuse with the main database schema migrations - and probably put it into a folder for 'one-offs' (there will probably be others).  Could be a subproject though that might be inviting pain unnecessarily..

Nicolas-Dolan (approved):
Looks good. In addition to creating a backup of the live database it might also be worth timing the database changes to happen at a low-use time, such as the middle of the night, as I'm not sure how it would affect someone who's currently in a session. Most users are in Africa, so have similar time zones to us.

PR #507, opened 2021-06-02
mrc-2223 add enter key event to create project input field 

carbowizzy:
Adds enter event to create project field. A user can create a new project when he/she press enter button.

## Type of version change
_Delete as appropriate_

Minor

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,6 @@
+# hint 1.37.0
+
+* User can create a new project when enter key event is pressed
 
 # hint 1.36.0
 
--- a/src/app/static/src/app/components/projects/Projects.vue
+++ b/src/app/static/src/app/components/projects/Projects.vue
@@ -13,6 +13,7 @@
             <div class="my-3 col-6 clearfix">
                 <input type="text" class="form-control"
                        v-translate:placeholder="'projectName'"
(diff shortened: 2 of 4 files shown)


mwoodbri (commented):
This all looks good and works well.

But not sure how it relates to mrc-2223? Doesn't appear to fix any accessibility issues described by the report attached to ticket.

EmmaLRussell:
> This all looks good and works well.
> 
> But not sure how it relates to [mrc-2223](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2223)? Doesn't appear to fix any accessibility issues described by the report attached to ticket.



> This all looks good and works well.
> 
> But not sure how it relates to [mrc-2223](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2223)? Doesn't appear to fix any accessibility issues described by the report attached to ticket.

Just what I was about to say! Is this branch actually for a different ticket?

Nicolas-Dolan (approved):
Works well. Just need to update the versions.

mwoodbri:
Another observation: there's a related ticket mrc-2156, which seems to suggest that consistent behaviour would be for enter button to select submit button, rather than to immediately submit the form. We need to decide which behaviour is preferred.

PR #509, opened 2021-06-03
mrc-2386 - Add note icon(s) and dialogs to project history for project/version

carbowizzy:
This PR allows users to edit saved notes or add new notes. the icon can be accessed from the project history page. When a user clicks on the icon next the project name, a dialog appears with notes textarea. This can either display an existing note or a blank textarea , where a new text can be added. 

## Type of version change

Minor

## Checklist


carbowizzy:
Review of this branch is dependent on Mrc-2384

EmmaLRussell:
> Review of this branch is dependent on [Mrc-2384](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2384)

It's probably best not to request reviews of branches until they're actually ready to review i.e. any parent branches merged to master then merged back into the given branch.

EmmaLRussell (changes requested):
I haven't looked over all the code for this, but I think there are some bugs in the behaviour which need to be looked at:

The tooltip over the notes icon in the project row says 'Add or edit version notes' - should be 'project notes'.

The dialog for adding both project notes and version notes have title 'Project notes for version {version id}' and text 'Add or edit version notes for project {project name}'. So it seems that we're getting the version notes dialog for project notes. From the mockups would expect title 'Project notes' and text 'This will add notes to {project name}' and make no mention of a particular version. 

However, I think the text in the two dialogs should probably be more consistent anyway ie. should both start with 'Add or edit..' or with 'This will add notes to' (probably the former). 

When I update the project note, this note appears on the latest version as well, and vice versa. I expected project and version notes to be independent. 

I think the dialog showing the project note is really just showing/editing the latest version note, because when I update an older version to be the latest, I then see its note appearing in the project note dialog rather than the previous latest version. 

When a version note is edited, the project row that was open collapses - I'd expect the project history table expanded row to be unaffected by setting the note. 

Could the notes icons be right aligned so that they all line up, as in the mockup? This will need to allow for the possibility of long project names which may wrap.

carbowizzy:
> showing/editing

Thanks for looking at this PR. I think you may have been misled by the mockups, as you know mockups are not always the exact implementation some times. During the implementation, I realised having project note icon on the current project may not be possible as current project can be any version due to it being promoted. Nevertheless I tried to display the project note  for current project using the same icon for project version but to no avail. Displaying project not will require a new icon and represented somewhere else. However, I will look at the formatting and collapsible.

carbowizzy:
> I haven't looked over all the code for this, but I think there are some bugs in the behaviour which need to be looked at:
> 
> The tooltip over the notes icon in the project row says 'Add or edit version notes' - should be 'project notes'.
> 
> The dialog for adding both project notes and version notes have title 'Project notes for version {version id}' and text 'Add or edit version notes for project {project name}'. So it seems that we're getting the version notes dialog for project notes. From the mockups would expect title 'Project notes' and text 'This will add notes to {project name}' and make no mention of a particular version.
> 
> However, I think the text in the two dialogs should probably be more consistent anyway ie. should both start with 'Add or edit..' or with 'This will add notes to' (probably the former).
> 
> When I update the project note, this note appears on the latest version as well, and vice versa. I expected project and version notes to be independent.
> 
> I think the dialog showing the project note is really just showing/editing the latest version note, because when I update an older version to be the latest, I then see its note appearing in the project note dialog rather than the previous latest version.
> 
> When a version note is edited, the project row that was open collapses - I'd expect the project history table expanded row to be unaffected by setting the note.
> 
> Could the notes icons be right aligned so that they all line up, as in the mockup? This will need to allow for the possibility of long project names which may wrap.

Thanks for looking at this PR. I think you may have been misled by the mockups, as you know mockups are not always the exact implementation some times. During the implementation, I realised having project note icon on the current project may not be possible as current project can be any version due to it being promoted. Nevertheless I tried to display the project note  for current project using the same icon for project version but to no avail. Displaying project not will require a new icon and represented somewhere else. However, I will look at the formatting and collapsible.

EmmaLRussell:
> 
> Thanks for looking at this PR. I think you may have been misled by the mockups, as you know mockups are not always the exact implementation some times. During the implementation, I realised having project note icon on the current project may not be possible as current project can be any version due to it being promoted. Nevertheless I tried to display the project note for current project using the same icon for project version but to no avail. Displaying project not will require a new icon and represented somewhere else. However, I will look at the formatting and collapsible.

Sorry, I'm not sure what you're saying here. Is it not a bug that changing the project note actually changes a version note? Project has its own note field in the database doesn't it, in the `project` table? I would expect that to be what gets changed when you edit project note, not the note for any version (which is in the `project_version` table)  - is that not the case

carbowizzy:
> > Thanks for looking at this PR. I think you may have been misled by the mockups, as you know mockups are not always the exact implementation some times. During the implementation, I realised having project note icon on the current project may not be possible as current project can be any version due to it being promoted. Nevertheless I tried to display the project note for current project using the same icon for project version but to no avail. Displaying project not will require a new icon and represented somewhere else. However, I will look at the formatting and collapsible.
> 
> Sorry, I'm not sure what you're saying here. Is it not a bug that changing the project note actually changes a version note? Project has its own note field in the database doesn't it, in the `project` table? I would expect that to be what gets changed when you edit project note, not the note for any version (which is in the `project_version` table) - is that not the case?

What you are referring to is a project version and not project note. It was intended to be project version note, and no it was not a bug. We can chat about this , I will give you a call.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ProjectsController.kt
```diff
@@ -107,7 +107,7 @@ class ProjectsController(private val session: Session,
             @RequestParam("note") note: String): ResponseEntity<String>
     {
         versionRepository.updateVersionNote(versionId, projectId, userId(), note)
-        return EmptySuccessResponse.asResponseEntity()
+        return SuccessResponse(projectRepository.getProjects(userId())).asResponseEntity()
```
EmmaLRussell, line 110: Same comment as for updateProjectNote

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ProjectsController.kt
```diff
@@ -96,7 +96,7 @@ class ProjectsController(private val session: Session,
             @RequestParam("note") note: String): ResponseEntity<String>
     {
         projectRepository.updateProjectNote(projectId, userId(), note)
-        return EmptySuccessResponse.asResponseEntity()
+        return SuccessResponse(projectRepository.getProjects(userId())).asResponseEntity()
```
EmmaLRussell, line 99: What's the need for this change? Shouldn't need to get all projects having just updated a note field. 

Is it because you want to refresh all projects having updated the note? Our standard pattern for that is to refetch from the projects endpoint when the initial request succeeds, as [here](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/projects/actions.ts#L190) and [here](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/projects/actions.ts#L206) etc. It does mean making a second request, but it means each endpoint has one job to do rather than sharing functionality. 

This is the sort of change it would be great to explain your reasoning for in the PR.

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -182,14 +198,40 @@
                 </button>
             </template>
         </modal>
+
+        <modal :open="versionNoteToEdit || projectNoteToEdit">
+            <h4 v-html="editVersionNoteHeader" id="editVersionNoteHeader"></h4>
```
EmmaLRussell, line 203: As mentioned above, I guess we need an `editProjectNoteHeader` too

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -289,6 +369,26 @@
...
                 this.loadAction({projectId, versionId});
             },
+            handleEditVersionNote(projectId: number, versionId: string, versionNumber: number) {
+                this.projects.filter(project => {
+                    if (project.id === projectId) {
+                        this.displayProjectName = project.name
+                        this.editedNote = project.versions
+                            .find(version => version.versionNumber === versionNumber)!.note || ""
+                    }
+                })
```
EmmaLRussell, line 379:
```suggestion
               const project = this.projects.find(p => p.id === projectId)!;
               this.displayProjectName = project.name;
               this.editedNote = project.versions
                            .find(version => version.versionNumber === versionNumber)!.note || ""
   
```
You could make this a little terser by using find to get the project (rather than filter) as well as to get the version. Something like this?

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -289,6 +369,26 @@
...
+                this.selectedVersionNumber = `v${versionNumber}`;
+            },
+            handleEditProjectNote(projectId: number) {
+                this.projects.find(project => {
+                    if (project.id === projectId) {
+                        this.displayProjectName = project.name
+                        this.editedNote = project.note || ""
+                    }
+                })
+                this.projectNoteToEdit = projectId
```
EmmaLRussell, line 390:
```suggestion
                const project = this.projects.find(p => p.id === project.id);
                this.displayProjectName = project.name;
                this.editedNote = project.note || "";
                this.projectNoteToEdit = projectId;
```

A little more readable?

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -313,6 +413,38 @@
...
+                    this.versionNoteToEdit = null;
+                    this.editedNote = "";
+                }
+
+                if (this.projectNoteToEdit) {
+                    this.projectNoteToEdit = null
+                    this.editedNote = "";
+                }
+            },
+            async confirmNoteEditing() {
```
EmmaLRussell, line 427: Don't think this  needs to be async?

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -313,6 +413,38 @@
...
+                    const versionPayload: versionPayload = {
+                        version: this.versionNoteToEdit!,
+                        note: this.editedNote
+                    };
+                    this.updateVersionNoteAction(versionPayload);
+                    this.versionNoteToEdit = null;
+                    this.editedNote = "";
+                }
+
+                if (this.projectNoteToEdit) {
```
EmmaLRussell, line 438: Could make this an `else if` just to emphasise to the reader that it ought to be one or the other.

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -355,7 +487,7 @@
                 }
             },
             async confirmPromotion(name) {
-                if (this.versionToPromote) {
+                if (this.versionToPromote && this.newProjectName) {
```
EmmaLRussell, line 490: What's this? An opportunistic bug fix, not related to this ticket? Should there be an associated test?

Again, would be good to just do a quick review of your code before submitting for PR and mentioning out of the ordinary bits like like to circumvent daft questions like this one! ;)

src/app/static/src/app/store/projects/actions.ts
```diff
@@ -117,6 +120,16 @@ export const actions: ActionTree<ProjectsState, RootState> & ProjectsActions = {
             }
         }
     },
+    
+    async updateVersionNote(context, versionPayload: versionPayload) {
+        const {projectId, versionId} = versionPayload.version
+        const note = versionPayload.note
+
+        await api<ProjectsMutations, ProjectsMutations>(context)
+            .withSuccess(ProjectsMutations.SetPreviousProjects)
```
EmmaLRussell, line 129: OK I see that you are committing this response - see above remark on using existing pattern to fetch projects in a separate request

src/app/static/src/app/store/projects/actions.ts
```diff
@@ -130,6 +143,15 @@ export const actions: ActionTree<ProjectsState, RootState> & ProjectsActions = {
             .postAndReturn(`project/${projectId}/version/?parent=${versionId}&note=${note}`)
     },
 
+    async updateProjectNote(context, payload) {
+        const projectId = payload.projectId
+        const note = payload.note
+        await api<ProjectsMutations, ProjectsMutations>(context)
+            .withSuccess(ProjectsMutations.SetPreviousProjects)
```
EmmaLRussell, line 150: As above

src/app/static/src/tests/components/projects/projectHistory.test.ts
```diff
@@ -463,14 +472,162 @@ describe("Project history component", () => {
...
+
+        const noteText = textarea.element as HTMLTextAreaElement
+        expect(noteText.value).toBe("new notes")
+        expect(wrapper.vm.$data.editedNote).toBe("new notes")
+
+        const okBtn = modal.find(".modal-footer").findAll("button").at(0);
+        await okBtn.trigger("click");
+
+        expect(mockUpdateVersion.mock.calls.length).toBe(1);
+        expect(mockUpdateVersion.mock.calls[0][1]).toStrictEqual(
```
EmmaLRussell, line 501: Maybe check that mockUpdateProject *hasn't* been called? And the opposite for the below test of updating project note. 
Should maybe also check here that the dialog is hidden, as you do for the cancel tests.

EmmaLRussell (changes requested):
This works really well now, thanks for making those changes. 

The dialog title for project notes is 'Project notes for version v1' - I think it's misleading to mention the version at all now that you've separated out project notes from version notes. The project name gets mentioned in the subtitle, so could we just change the title to "Project notes"?

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ProjectsController.kt
carbowizzy: This changes updates the list of project without having to despatch a getProject. While the getProject will retrieve all projects but it causes a refresh of project history page, hence closes the collapsible. I do not think we have any efficiency issue and I didn't set loading flag for setting previous project because I do not want any wait before rendering. I also didn't return a project because the list uses date meta to rank the displayed projects. This works nicely, what's the issue ?

src/app/static/src/app/components/projects/ProjectHistory.vue
carbowizzy: I thought this was obvious, when a confirm promotion button is triggered without any input entry for project name, the endpoint action gets called. A bug that is now fixed.

mwoodbri (changes requested):
This is working really well, and besides the points @EmmaLRussell has made the code looks good to me too. If you can address those issues then I'll happily take a final look.

src/app/static/src/app/components/projects/ProjectHistory.vue
EmmaLRussell: > I thought this was obvious. When a confirm promotion button is triggered without any input entry for project name, the endpoint action gets called. A bug that is now fixed.

Obvious to you, perhaps, @carbowizzy since you've been working on this piece of code, but not necessarily to a reviewer who is having to do a context-switch from whatever they've had their head in. It's probably a good rule of thumb to always comment on any change in a PR which is not directly related to the ticket, no matter how trivial it appears to you.

src/app/src/test/kotlin/org/imperial/mrc/hint/database/VersionRepositoryTests.kt
```diff
@@ -278,7 +278,6 @@ class VersionRepositoryTests
         assertThat(newVersion.note).isEqualTo(null)
     }
 
-
```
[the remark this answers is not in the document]
carbowizzy: Which change ?

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ProjectsController.kt
EmmaLRussell: > This works nicely, what's the issue ?

As I said, my issue is that it differs from the other endpoints which make changes to the projects so it breaks our general pattern where a second request is made to retrieve projects, and as @mwoodbri says, it also overloads the endpoint to do more than it should need to. 

You make a good point about the subsequent refresh of the projects page collapsing the open project but perhaps there's another way to approach that e.g. adding a parameter to `getProjects` to avoid doing `SetLoading` (it's that which causes the project on the page to collapse?)

src/app/src/test/kotlin/org/imperial/mrc/hint/database/VersionRepositoryTests.kt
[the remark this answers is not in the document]
mwoodbri: A blank line was seemingly removed from this file? Hence it has been pulled into this PR. I know it is a really small issue but it's something that can usually be spotted by having a final look through the PR on GitHub itself before you assign it to reviewers. This is a good habit in case there are other similar but more significant issues (i.e. files committed that shouldn't be).

src/app/static/src/app/components/projects/ProjectHistory.vue
carbowizzy: Sure, no validation was triggered when invoking carriage return on the input of promoteVersion modal and even rename modal has no validation. I will probably look into fixing more on maintenance day. Couldn't get a test to fail or pass for this behaviour, this is where selenium will come handy.

src/app/static/src/app/store/projects/actions.ts
```diff
@@ -85,9 +88,9 @@ export const actions: ActionTree<ProjectsState, RootState> & ProjectsActions = {
             .postAndReturn<string>("/project/", qs.stringify({name}));
     },
 
-    async getProjects(context) {
+    async getProjects(context, isLoading = true) {
         const {commit} = context;
-        commit({type: ProjectsMutations.SetLoading, payload: true});
+        commit({type: ProjectsMutations.SetLoading, payload: isLoading});
```
EmmaLRussell, line 93: This is fine, but I really meant to just not do the commit at all if the isLoading parameter is false - this achieves the desired behaviour, but is doing an unnecessary commit. Sorry, should have clarified that.

EmmaLRussell (commented):
This is looking really good now, I think the only one of my comments not yet addressed is to add a test for the `newProject` check you added. 

Also a tiny comment on the `getProjects` parameter, see below

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -143,11 +159,11 @@
             <input type="text"
                    class="form-control"
                    v-translate:placeholder="'projectName'"
-                   @keyup.enter="confirmPromotion(newProjectName)"
+                   @keyup.enter.prevent="confirmPromotion(newProjectName)"
```
mwoodbri, line 162: Could you clarify this - is it a change in behaviour, and if so why? Sorry - not obvious to me! Also differs from pattern used in L185.

src/app/static/src/app/components/projects/ProjectHistory.vue
mwoodbri: Not clear what the plan is here - are you saying this is a big enough piece of work to merit a new ticket? If so then please create it, clearly explaining which modals need to be updated and why, and add a link here. Just to avoid it slipping through the cracks.

mwoodbri (commented):
Code looks ready to go, just a couple of minor queries to help my understanding.

EmmaLRussell (approved):

src/app/static/src/app/components/projects/ProjectHistory.vue
mwoodbri: @carbowizzy Not clear of the outcome here: was expecting to see either a test here, or a via a new ticket linked from here

carbowizzy: Test was added on L658 of projectHistory.

mwoodbri: Aha thanks very much. I didn't spot the relevance first time round! Looks good.

mwoodbri (approved):
Looks like a NEWS conflict but otherwise ready to go!

PR #511, opened 2021-06-07
Mrc-2391 - Modify Rename dialog modal and add notes text area

carbowizzy:
This PR modifies an existing dialog, adding text-area to the modal. This text-area accepts pre-populated note, which can be edited or a new text can be added if rename does not have an existing note. 

This implementation can be found on project history page, at rename column. When a user clicks on rename icon, a dialog appears with rename field and the newly added text-area. 

## Type of version change
Minor 

## Checklist


carbowizzy:
dependent on mrc-2386

EmmaLRussell:
> dependent on [mrc-2386](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2386)

OK, will delay reviewing until mrc-2386 (which is itself dependent on mrc-2384 is merged) so we can see a clean diff.

mwoodbri:
@carbowizzy I think the prereqs for this have been merged so if you bring up to date with master and then request a review I'll happily take a look

src/app/static/src/app/store/projects/actions.ts
```diff
@@ -227,12 +227,12 @@ export const actions: ActionTree<ProjectsState, RootState> & ProjectsActions = {
...
     async renameProject(context, projectPayload: projectPayload) {
         const {state, dispatch} = context;
-        const {projectId, name} = projectPayload
+        const {projectId, name, note} = projectPayload
 
         await api<ProjectsMutations, ErrorsMutation>(context)
             .ignoreSuccess()
             .withError(`errors/${ErrorsMutation.ErrorAdded}` as ErrorsMutation, true)
-            .postAndReturn(`/project/${projectId}/rename`, qs.stringify({name}))
+            .postAndReturn(`/project/${projectId}/rename`, qs.stringify({name: name, note: note}))
```
mwoodbri, line 235: I know this is tested indirectly but you should also add a test to [actions.test.ts](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/tests/projects/actions.test.ts#L646) to ensure that a mock POST is made, given a payload containing a note.

mwoodbri (changes requested):
This all looks good to me - and works nicely. Would just benefit from one extra test.

src/app/static/src/tests/projects/actions.test.ts
```diff
@@ -651,19 +651,20 @@ describe("Projects actions", () => {
             currentVersion: mockProject.versions[0]
         });
 
-        const stateUrl = "/project/1/rename";
-        mockAxios.onPost(stateUrl, "name=renamedProject")
+        const stateUrl = "/project/1/rename?name=renamedProject&note=test%20notes";
```
mwoodbri, line 654: This line and the rest of the test would benefit from some tidying:
1. If you `await` on the action then you don't need the `setTimeout()`
2. `stateUrl` works but isn't correct - the `name` and `note` are actually sent as a form-encoded POST body, not via the URL
3. You should check that nothing is committed. This is important because the action calls `getCurrentProject` and `getProjects` regardless of whether any errors occur. Perhaps this should be resolved in the action itself, but it should definitely be checked here.

Feel to pick and choose from this version:
```typescript
    it("renameProject changes a project's name", async () => {
        const commit = jest.fn();
        const dispatch = jest.fn();
        const state = mockProjectsState({
            currentProject: mockProject,
            currentVersion: mockProject.versions[0]
        });

        mockAxios.onPost("/project/1/rename")
            .reply(200, mockSuccess("OK"));

        const projectPayload = {
            projectId: 1,
            name: "renamedProject",
            note: "test notes"
        }
        await actions.renameProject({commit, state, rootState, dispatch} as any, projectPayload);

        const posted = mockAxios.history.post[0].data;
        expect(posted).toEqual("name=renamedProject&note=test%20notes");
        expect(commit.mock.calls.length).toBe(0);
        expect(dispatch.mock.calls[0][0]).toBe("getCurrentProject");
        expect(dispatch.mock.calls[1][0]).toBe("getProjects");
        expect(dispatch.mock.calls.length).toBe(2);
    });
```

mwoodbri (changes requested):
Minor fixes/improvements to test, but otherwise all looks good

EmmaLRussell (commented):
All works well, don't have much to add. See the actions test above your one, `update version notes commits Project Error on failure` for example of posting form-encoded body. 

We have tended to use `setTimeout`s for these tests, but awaiting is probably more readable!

mwoodbri (approved):
A big :+1: from me

EmmaLRussell (approved):

PR #514, opened 2021-06-29
mrc-2480 Use async endpoints for Just-In-time generation of download files from HINT front end

carbowizzy:
## Description

This PR changes  current implementation, where downloadable output files are generated during calibrate stage. This method  of generating output files introduce some latency issues which the new implementation aspires to resolve.  Please note that ADR Upload behaviour has not been modified in this ticket and if there is a need for modification then this ticket mrc-2523 https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2523 will address it as ADR scope has changed rapidly since this ticket.

###### Backend
- I have removed /download/spectrum/, /download/coarse-output and /download/summary endpoints in favour of async download generate endpoints /download/submit/<type>/<id> where type can be summary, coarse-output or spectrum. Please see ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2480 for detailed spec.
- Remove uploadMetadata from calibrate response in front end - a new endpoint  /meta/adr/<id> has now been added to retrieve metadata for uploadToADR.
- Updated upload to ADR controller to reflect new changes

###### Frontend
- Removed redundant endpoints and updated frontend code to use the new endpoints for downloads. 
- Created new store for downloadResults - actions/mutations
- Updated upload to ADR action to reflect new changes

 
TODO 👍 
- [ ] hintr_version: mrc-2407 branch, this needs to be reset to master and inform @r-ash 

## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/app/components/downloadResults/DownloadResults.vue
```diff
@@ -2,28 +2,58 @@
...
+                    <error-alert id="spectrum-error" v-if="spectrum.error" :error="spectrum.error"></error-alert>
+                </div>
+                <div id="coarse-output-download">
+                    <h4 class="mt-4" v-translate="'downloadCoarseOutput'"></h4>
+                    <button class="btn btn-red btn-lg my-3"
+                            @click="downloadCoarseOutput">
+                        <span v-translate="'download'"></span>
+                        <download-icon size="20" class="icon ml-2" style="margin-top: -4px;"></download-icon>
+                    </button>
+                    <div v-if="!uploadModalOpen">
```
mwoodbri, line 27: To avoid the complexities of co-ordinating download and upload UI it might be better just to disable the button that opens the modal if a download is currently in progress?

mwoodbri (changes requested):
This was a monumental task... and not one where the correct approach was obvious - clearly there were options for how responsibility should be divided between front- and backend.

Having said that, it's looking very promising. Downloading outputs mostly worked well for me - including when downloading more than one simultaneously.

Some issues:
- Clicking "Upload to ADR" before clicking any of the Export/Download buttons causes an error
- Clicking Download (e.g. summary report) displays the relevant spinner... but also the ADR upload spinner on the other side of the page
- The spinner only seems to spin the first time I click one of the download buttons
- I get an error on screen (see below) and in the console ("Error: Request aborted") when the "Save file" pop-up appears
2021-07-27 12-11-27 HINT error

Some other observations:
- Downloading a file for a second time doesn't seem to be any quicker than the first time. Is this to be expected? Regardless, I'm not sure of the value in displaying the "Download complete" message at any time - what does it actually signify?
- Obviously there's a fair amount of repetition in various places due to the three output types having very similar code. However, I don't think much of this can be reduced will keeping type safety (or a much more structural change)
- I think this feature sorely needs some documentation - even a sequence diagram would help to explain the interaction between UI, store, backend and ADR itself. Otherwise maintaining this part of the app is going to be a challenge.
- Minor point, but with hindsight it might have it easier to implement (and review/test) this PR if the ADR and non-ADR parts were done separately. As things stand they've got quite coupled.

I have made other various comments inline, but mostly echo points made here.

src/app/static/src/app/store/metadata/mutations.ts
```diff
@@ -1,23 +1,38 @@
...
+    },
+
+    [MetadataMutations.AdrUploadMetadataFetched](state: MetadataState, action: PayloadWithType<AdrMetadataResponse>) {
+        const adrMetadataTypeExists = state.adrUploadMetadata
+            .findIndex(adrMeta => adrMeta.type === action.payload.type)
+
+        adrMetadataTypeExists > -1 ?
+            state.adrUploadMetadata[adrMetadataTypeExists].description = action.payload.description
+            : state.adrUploadMetadata.push({type: action.payload.type, description: action.payload.description})
+
```
EmmaLRussell, line 35: Would this be a bit less of a faff if `adrUploadMetadata` was a dictionary keyed on type rather than an array?

src/app/static/src/app/types.ts
```diff
@@ -153,4 +153,23 @@ export interface UploadFile {
...
+export interface DownloadResultsDependency {
+    downloadId: string
+    downloading: boolean
+    statusPollId: number
+    status: DownloadStatusResponse
+    complete: boolean
+    error: Error | null
+}
+
+export interface PoolingStarted {
```
EmmaLRussell, line 167: Polling not pooling?

carbowizzy (commented):
> This was a monumental task... and not one where the correct approach was obvious - clearly there were options for how responsibility should be divided between front- and backend.
> 
> Having said that, it's looking very promising 👍
> 
> Downloading outputs mostly worked well for me - including when downloading more than one simultaneously.
> 
> Some issues:
> 
> * Clicking "Upload to ADR" before clicking any of the Export/Download buttons causes an error
> * Clicking Download (e.g. summary report) displays the relevant spinner... but also the ADR upload spinner on the other side of the page
> * The spinner only seems to spin the first time I click one of the download buttons
> * I get an error on screen (see below) and in the console ("Error: Request aborted") when the "Save file" pop-up appears
>   2021-07-27 12-11-27 HINT error
> 
> Some other observations:
> 
> * Downloading a file for a second time doesn't seem to be any quicker than the first time. Is this to be expected? Regardless, I'm not sure of the value in displaying the "Download complete" message at any time - what does it actually signify?
> * Obviously there's a fair amount of repetition in various places due to the three output types having very similar code. However, I don't think much of this can be reduced whilst keeping type safety (or maybe making a much more structural change)
> * I think this feature sorely needs some documentation - even a sequence diagram would help to explain the interaction between UI, store, backend and ADR itself. Otherwise maintaining this part of the app is going to be a challenge.
> * Minor point, but with hindsight it might have it easier to implement (and review/test) this PR if the ADR and non-ADR parts were done separately. As things stand they've got quite coupled.
> 
> I have made other various comments inline, but mostly echo points made here.

I have tried to recreate the bugs that is causing the error message but I could not reproduce. Did you clear your docker images before testing as this relies on a different hintr build mrc-240 ?

EmmaLRussell (changes requested):
It took me a while to work out that all the logic for ensuring downloads were available for ADR was implemented in the front end rather than the back end where I expected it to be! But once i found it, it looks like all the key logic is there and correct, great stuff :+1:. This should probably have been multiple tickets really! 

I think it is ok for the front end to be managing this process, as we already have the front end managing polling for status and other long-running processes, so it fits with existing models. 

In general, I think I'd slightly prefer it if the actions/mutations etc for summary, spectrum and coarse outputs were generic and took a string parameter, and if the state was, say, a dictionary rather than individually named props. It would be easier that way to incorporate any new output types that come along, and would avoid duplicated code. But the current implementation is clear, so I wouldn't insist on refactoring it.

I think we also need to make sure that the new downloadResults state is cleared when the project is re-calibrated or the state is otherwise invalidated, and a new version generated - otherwise, I think the state will have out of date download ids which it will think are current. This will mean doing some work in `resetState` in `root.ts` I think - need to include the download state in `ResetOutputs` and also add a case for resetting on recalibrate. 

Also, should ensure state is saved in `localStorageManager` though the download ids here are probably more useful than more ephemeral state like 'downloading'  which could be omitted from saved state. 

When I upload to ADR without first downloading, that did work for me, and I saw the below view, which I think is as intended. However, I think it would be better if the 'Download complete' messages were not shown if the user has not actually downloaded to their local machine - it's going to be very confusing to a user who thinks they've just uploaded to ADR and hasn't in fact downloaded anything. 
Screenshot from 2021-07-27 15-50-39

carbowizzy:
> I think we also need to make sure that the new downloadResults state is cleared when the project is re-calibrated or the state is otherwise invalidated, and a new version generated - otherwise, I think the state will have out of date download ids which it will think are current. This will mean doing some work in `resetState` in `root.ts` I think - need to include the download state in `ResetOutputs` and also add a case for resetting on recalibrate.

I have not persisted downloadResults state in localStorage because interactions on this page is limited, download action would not be triggered every time and when ever it is triggered it fetches the data from backend afresh. This way we don't need to clear downloadResult state when model is re-calibrated as it uses the most unto date calibrateId retrieved from root state.

EmmaLRussell:
> > I think we also need to make sure that the new downloadResults state is cleared when the project is re-calibrated or the state is otherwise invalidated, and a new version generated - otherwise, I think the state will have out of date download ids which it will think are current. This will mean doing some work in `resetState` in `root.ts` I think - need to include the download state in `ResetOutputs` and also add a case for resetting on recalibrate.
> 
> I have not persisted downloadResults state in localStorage because interactions on this page is limited, download action would not be triggered every time and when ever it is triggered it fetches the data from backend afresh. This way we don't need to clear downloadResult state when model is re-calibrated as it uses the most unto date calibrateId retrieved from root state.

Fair enough on not persisting downloadResults state, but i think there is still an issue in not clearing the state when recalibration occurs. I tested uploading to ADR, and saw the download files being generated as expected, then went back to calibrate and recalibrated, then forward to Save results again. The state has not been reset, so still see the 'Download complete' messages, and when I upload to ADR again, the download files are not regenerated, but are immediately uploaded with the old IDs, hence will be picking up the old calibration results rather than the new ones. 

(NB This was with the version of the code I reviewed not your latest changes, you may have changed the download message behaviour since then)

src/app/static/src/app/store/downloadResults/mutations.ts
```diff
@@ -0,0 +1,134 @@
...
+            case DOWNLOAD_TYPE.SUMMARY: {
+                state.summary.statusPollId = action.payload.pollId
+                break
+            }
+        }
+    }
+};
+
+const stopPollingSpectrum = (state: DownloadResultsState) => {
+    clearInterval(state.spectrum.statusPollId);
```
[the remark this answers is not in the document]
mwoodbri: Agree with this. Mutations should really just be modifying the application state.

In fact, I'm not even sure that the action should be doing this (rather than the component) or whether it makes sense to store these IDs in the store at all, but that's more a matter of opinion!

src/app/static/src/app/store/adrUpload/actions.ts
```diff
@@ -91,9 +91,10 @@ export const actions: ActionTree<ADRUploadState, RootState> & ADRUploadActions =
 
     async uploadFilesToADR(context, uploadFilesPayload) {
         const {rootState, commit, dispatch} = context;
-        const uploadMetadata = rootState.modelRun.result?.uploadMetadata
         const selectedDatasetId = rootState.baseline.selectedDataset!.id;
-        const modelCalibrateId = rootState.modelCalibrate.calibrateId;
+        let downloadId;
+
+        const uploadMetadata = rootState.metadata.adrUploadMetadata
```
mwoodbri, line 97: Is there any chance of a race condition here? Should this function be invoking and waiting for `getUploadMetadata` rather than `DownloadResults` doing it pre-emptively? The latter seems to be mixing download and upload concerns.

mwoodbri (changes requested):
This works very nicely now - not sure of cause of previous issues. The problem with spinner only spinning the first time it's used seems to be limited to Firefox, so safe to ignore.

There are a couple of comments from my previous review remaining, and a couple of new ones, but nothing major.

One cosmetic thing - but only if it's easy to fix: the "Downloading" text is not centre-aligned like the "Uploading" text - and is therefore very close to the header below it:
Screenshot from 2021-07-29 14-24-18
Screenshot from 2021-07-29 12-54-16

src/app/static/src/app/root.ts
```diff
@@ -111,6 +112,10 @@ const resetState = (store: Store<RootState>): void => {
             if (type[0] == "modelRun" && ModelRunUpdates.includes(type[1] as ModelRunMutation)) {
                 store.commit(RootMutation.ResetOutputs);
             }
+
+            if (type[0] == "modelCalibrate" && ModelCalibrateUpdates.includes(type[1] as ModelCalibrateMutation)) {
```
EmmaLRussell, line 116: I think this needs this a test?

src/app/static/src/app/store/downloadResults/mutations.ts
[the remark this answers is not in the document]
EmmaLRussell: We're not very consistent unfortunately - we already have precedent for managing polling in both components and actions e.g. polling for run and calibrate both happen in actions

EmmaLRussell (commented):
Download state is resetting nicely on recalibrate now, thanks. A couple of comments still unresolved from last review.

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -239,15 +245,23 @@
...
           this.setDefaultCheckedItems();
         },
-        summary() {
-          if (this.downloadIsReady()) {
-            this.sendUploadFilesToADR();
-          }
+        summary: {
+          handler(summary) {
+            if (this.open && this.downloadIsReady()) {
+              this.getUploadMetadata(summary.downloadId)
```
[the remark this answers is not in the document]
carbowizzy: Following @mwoodbri's review on separation of concern. Upload metadata will now be awaited before sending files to ADR, which solves the race issue.

carbowizzy:
- stopPolling has been moved downloadResults component. 
- removed complete polling mutation
- removed storing of status
- add complete to state

src/app/static/src/app/store/downloadResults/actions.ts
```diff
@@ -0,0 +1,103 @@
...
+            else if (downloadType === DOWNLOAD_TYPE.SUMMARY) {
+                getSummaryDownloadStatus(context)
+            }
+        }, 2000);
+
+        commit({type: "PollingStatusStarted", payload: {pollId: id, downloadType: downloadType}});
+    },
+};
+
+export const getSummaryDownloadStatus = async function (context: ActionContext<DownloadResultsState, RootState>) {
```
mwoodbri, line 78: Would be good to add a return type to this function. And same for the other two.

src/app/static/src/app/store/downloadResults/mutations.ts
```diff
@@ -0,0 +1,92 @@
...
+    SpectrumDownloadStatusUpdated = "SpectrumDownloadStatusUpdated",
+    SpectrumError = "SpectrumError",
+    CoarseOutputDownloadStarted = "CoarseOutputDownloadStarted",
+    CoarseOutputDownloadStatusUpdated = "CoarseOutputDownloadStatusUpdated",
+    CoarseOutputError = "CoarseOutputError",
+    SummaryDownloadStarted = "SummaryDownloadStarted",
+    SummaryDownloadStatusUpdated = "SummaryDownloadStatusUpdated",
+    SummaryError = "SummaryError",
+    PollingStatusStarted = "PollingStatusStarted",
+    StopPolling = "StopPolling"
```
mwoodbri, line 17: This appears redundant?

src/app/static/src/app/root.ts
```diff
@@ -107,6 +115,10 @@ const resetState = (store: Store<RootState>): void => {
...
+                store.commit(RootMutation.ResetDownload);
```
EmmaLRussell, line 120: This line still seems to be untested: your should be able add a test [here](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/tests/components/app.test.ts)

src/app/static/src/app/components/downloadResults/DownloadResults.vue
```diff
@@ -126,35 +162,89 @@
...
+            },
+            handleDownloadResult(downloadResults) {
+              if(!this.uploadModalOpen) {
+                if (downloadResults.complete) {
+                  window.location.assign(this.downloadUrl(downloadResults.downloadId));
+                  this.getUploadMetadata(downloadResults.downloadId);
+                  this.stopPolling(downloadResults.statusPollId)
+                }
+                if (downloadResults.error && downloadResults.statusPollId > -1) {
+                  this.stopPolling(downloadResults.statusPollId);
```
EmmaLRussell, line 206: This line is flagged as untested.

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -100,10 +126,58 @@
...
+
+                if (summary && spectrum) {
+                    return !!this.summary.complete && !!this.spectrum.complete;
+                } else {
+                    if (summary && !spectrum) {
+                        return !!this.summary.complete;
+                    } else {
+                        return !!this.spectrum.complete;
+                    }
+                }
```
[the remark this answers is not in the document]
EmmaLRussell: Could you please respond to this suggestion, either by accepting it or saying why you disagree?

mwoodbri (changes requested):
Looking good. A couple of minor comments on the code provided inline, with one left over from last time. One UI issue that seems to hit me about 50% of the time - when the download completes i.e. the file download box appears it seems to interrupt a previous Axios (polling?) request, which gets displayed as an error. Need to work around this somehow - either by avoiding these errors, or by ignoring them (i.e. distinguishing these from real ones):

Screenshot from 2021-08-17 09-33-42

EmmaLRussell (changes requested):
When I tried to upload both output files, without having previously downloaded, the upload seemed to get stuck on the second file - and polling continued despite the status responses indicting that download generation was complete (see screenshot). 
Screenshot from 2021-08-17 09-48-47

EmmaLRussell:
> When I tried to upload both output files, without having previously downloaded, the upload seemed to get stuck on the second file - and polling continued despite the status responses indicting that download generation for both files was complete (see screenshot).

This console error may be related:
Screenshot from 2021-08-17 09-55-36

src/app/static/src/app/store/downloadResults/actions.ts
carbowizzy: I am not sure why we need a return type right now, what are you thinking ?

mwoodbri: Was thinking that with TypeScript it's good for functions to have return types. But admittedly not something we do everywhere. And just realised that in this case we don't care about the result, only the mutations.

carbowizzy:
> Looking good. A couple of minor comments on the code provided inline, with one left over from last time. One UI issue that seems to hit me about 50% of the time - when the download completes i.e. the file download box appears it seems to interrupt a previous Axios (polling?) request, which gets displayed as an error. Need to work around this somehow - either by avoiding these errors, or by ignoring them (i.e. distinguishing these from real ones):
> 
> Screenshot from 2021-08-17 09-33-42

I believe your dependencies were not fully loaded hence the API response error below the page. This might have caused the dialog to show as I have just tried it on Firefox and could not reproduce.

carbowizzy:
> When I tried to upload both output files, without having previously downloaded, the upload seemed to get stuck on the second file - and polling continued despite the status responses indicting that download generation for both files was complete (see screenshot).
> Screenshot from 2021-08-17 09-48-47

Spot on, I replaced ?description with !description thinking that should ordinarily work, resolved now.

src/app/static/src/app/root.ts
carbowizzy: Actually this was tested in roots/mutations but I have added another test to app.tests

EmmaLRussell:
>Could you please respond to this suggestion, either by accepting it or saying why you disagree?
>This is an important logic that checks which resources are selected for upload to ADR, if both are checked return complete state for both
else
if summary is checked and not spectrum return complete state for summary.
else return complete state for spectrum.

I think you probably meant to reply to this rather than editing! Yes, I think I understand the logic, and I'm suggesting [this change](https://github.com/mrc-ide/hint/pull/514/files#r677447054) to provide a terser version of it.

mwoodbri (changes requested):
Epic.

No recurrence of errors I saw last time. But there is an issue with polling: I think that as a result of `setInterval()` being in the action and the `clearInterval()` being in `DownloadResults.vue`, `UploadModal.vue` never stops polling. Would be great if you could fix this.

One comment inline.

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -100,10 +130,58 @@
...
+                /**
+                 * This logic checks and returns complete status (true/false)
+                 * of each selected resource being prepared for Upload to ADR
+                 */
+
+                if (summary && spectrum) {
+                    return !!this.summary.complete && !!this.spectrum.complete;
+                } else {
+                    return summary ? !!this.summary.complete : !!this.spectrum.complete;
+                }
```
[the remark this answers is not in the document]
carbowizzy: I will remove comments completely.

carbowizzy (commented):
> Epic.
> 
> No recurrence of errors I saw last time. But there is an issue with polling: I think that as a result of `setInterval()` being in the action and the `clearInterval()` being in `DownloadResults.vue`, `UploadModal.vue` never stops polling. Would be great if you could fix this.
> 
> One comment inline.
> 
> I'm hoping test failures are transient - something is definitely up with ADR at the moment (I can only see one dataset, and it's read-only).

I think you might have a problem with your environment, you will need to tear down docker and rerun to use hintr: mrc-2407... better still how do I reproduce this issue ?

mwoodbri:
> > I'm hoping test failures are transient - something is definitely up with ADR at the moment (I can only see one dataset, and it's read-only).
> 
> I think you might have a problem with your environment, you will need to tear down docker and rerun to use hintr: [mrc-2407](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2407)... better still how do I reproduce this issue ?

Sorry - should been clearer. I meant failure of ADR tests on Buildkite.

carbowizzy:
> > > I'm hoping test failures are transient - something is definitely up with ADR at the moment (I can only see one dataset, and it's read-only).
> > 
> > 
> > I think you might have a problem with your environment, you will need to tear down docker and rerun to use hintr: [mrc-2407](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2407)... better still how do I reproduce this issue ?
> 
> Sorry - should been clearer. I meant failure of ADR tests on Buildkite.

How do I reproduce issue with polling ? I tried different scenarios but didn't get any issue.

mwoodbri:
> > > > I'm hoping test failures are transient - something is definitely up with ADR at the moment (I can only see one dataset, and it's read-only).
> > > 
> > > 
> > > I think you might have a problem with your environment, you will need to tear down docker and rerun to use hintr: [mrc-2407](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2407)... better still how do I reproduce this issue ?
> > 
> > 
> > Sorry - should been clearer. I meant failure of ADR tests on Buildkite.
> 
> How do I reproduce issue with polling ? I tried different scenarios but didn't get any issue.

ADR upload without download. Polling continues after upload is complete.

src/app/static/src/app/components/downloadResults/UploadModal.vue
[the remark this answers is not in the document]
EmmaLRussell: If I have got it a bit wrong, would it still be possible to rework your code to be more in this style, as both Mark and I feel that it makes the logic more visible than multiple `if` statements. 
Which case are you thinking of that this would not work for? Is it the case where neither summary nor spectrum are checked ? That should return false because downloadIsReady should be false if neither are ticked? Could add a clause to check at least one of the two is true?:
`return (summary || spectrum) && (!summary || !!this.summary.complete) && (!spectrum || !!this.spectrum.complete);`

src/app/static/src/app/root.ts
EmmaLRussell: This is a plugin though, rather than a mutation, so it wouldn't be covered by the mutation test - codecov didn't seem to think so anyway! But fixed now by your new test :+1:

EmmaLRussell (commented):
Code coverage issues now sorted, thanks :+1: 

I am seeing same behaviour as Mark is - polling for download status never stops even though download/upload successfully completes.

carbowizzy:
> > > > > I'm hoping test failures are transient - something is definitely up with ADR at the moment (I can only see one dataset, and it's read-only).
> > > > 
> > > > 
> > > > I think you might have a problem with your environment, you will need to tear down docker and rerun to use hintr: [mrc-2407](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2407)... better still how do I reproduce this issue ?
> > > 
> > > 
> > > Sorry - should been clearer. I meant failure of ADR tests on Buildkite.
> > 
> > 
> > How do I reproduce issue with polling ? I tried different scenarios but didn't get any issue.
> 
> ADR upload without download. Polling continues after upload is complete.

Absolutely correct, I forgot to clear timer in uploadModal

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -193,11 +194,21 @@
...
+                if (this.open) {
+                    if (this.downloadIsReady()) {
+                        await this.getUploadMetadata(downloadResults.downloadId)
+                        this.stopPolling(downloadResults.statusPollId)
+                        this.sendUploadFilesToADR();
+                    }
+
+                    if (downloadResults.error && downloadResults.statusPollId > -1) {
+                        this.stopPolling(downloadResults.statusPollId)
+                    }
```
mwoodbri: One question on the conditionals here: if there a case where we get to this block and we _don't_ want to stop polling?

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
```diff
@@ -344,6 +379,32 @@ describe(`uploadModal `, () => {
...
+        const store = createStore();
+        const wrapper = mount(UploadModal, {store})
+        jest.useFakeTimers()
+
+        await wrapper.setProps({open: true})
+        wrapper.vm.$store.state.downloadResults = completeDownloadResults
+        await Vue.nextTick()
+
+        expect(clearInterval).toHaveBeenCalledTimes(2)
+        expect(wrapper.emitted().close.length).toBe(2)
```
mwoodbri, line 392: Just so I understand: why is `close` emitted twice here?

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
```diff
@@ -75,6 +80,37 @@ describe(`uploadModal `, () => {
...
+            downloading: false,
+            complete: false,
+            error: null,
+            downloadId: null
+        },
+        spectrum: {
+            downloading: false,
+            complete: true,
+            error: null,
+            downloadId: null
```
mwoodbri, line 94: To clarify the logic:
- `downloading`, `complete` and `error` are mutually exclusive?
- If `complete` is true then `downloadId` should always be non-`null`? (contrary to the example here - which I know is only for testing purposes, but is confusing me)

mwoodbri (changes requested):
This is working nicely now - with one issue. Minor, but would be good to fix:

If I click Download for one of the resources, I can then open the modal while the download is underway. At this point I think it would make sense to disable the Ok button in the modal. Currently I can click it but nothing apparently happens... until the download completes

Such a change might also solve a related issue that I saw once in the testing: I clicked Download, then opened the modal to upload, and when uploading was complete I got two files downloaded (zip and report) even though I only requested one.

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -100,10 +131,54 @@
...
+            },
+            downloadIsReady() {
+                const {summary, spectrum} = this.findSelectedUploadFiles();
+
+                if (summary && spectrum) {
+                    return !!this.summary.complete && !!this.spectrum.complete;
+                } else {
+                    return summary ? !!this.summary.complete : !!this.spectrum.complete;
+                }
+            },
```
EmmaLRussell, line 172: I think you might not have seen my last comment as github can sometimes hide replies to old conversations, so staring a fresh one:
>Unfortunately this suggestion is not going to work in this case but I have refactored the code to be clearer.

If I have got it a bit wrong, would it still be possible to rework your code to be more in this style, as both Mark and I seem to feel that it makes the logic more visible than multi-line with if statements, even after refactoring.
Which case are you thinking of that this suggestion would not work for? Is it the case where neither summary nor spectrum are checked ? That should return false because downloadIsReady should be false if neither are ticked? Could add a clause to check at least one of the two is true?:
`return (summary || spectrum) && (!summary || !!this.summary.complete) && (!spectrum || !!this.spectrum.complete);`

src/app/static/src/app/components/downloadResults/UploadModal.vue
carbowizzy: Yes, downloadIsReady is always false even when one out of two download resources is complete, which means we will never get to stop polling the completed resource.

src/app/static/src/app/components/downloadResults/UploadModal.vue
carbowizzy: I don't see how this code is clearer than the above.

mwoodbri: Obviously a degree of subjectivity here, but trying to explain why I find Emma's suggestion preferable:
- It's shorter. Not always the best metric, but in this case it's proportional to how long it took me to understand!
- It's symmetric i.e. it doesn't (slightly misleadingly) treat summary and spectrum differently
- It feels like it has less "branches", again meaning it's easier to understand the intention of the code

More generally, not sure why we need the `!!`... isn't `complete` a boolean anyway?

EmmaLRussell: Yes, to me it feels that expressing the logic as a boolean structure rather than branching ifs more transparent as well as terser - it's the structure that my brain tries to construct when looking at the branching ifs anyway to understand the meaning of what's being expressed. But agreed, it is quite subjective.

EmmaLRussell: > More generally, not sure why we need the !!... isn't complete a boolean anyway?

It is, so no we shouldn't need that!

carbowizzy: > > More generally, not sure why we need the !!... isn't complete a boolean anyway?
> 
> It is, so no we shouldn't need that!

!! is needed because I am getting them through Partial<>

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
```diff
@@ -18,10 +18,14 @@ import {DownloadResultsState} from "../../../app/store/downloadResults/downloadR
 
 describe(`uploadModal `, () => {
 
+    beforeEach(() => {
+        jest.useFakeTimers();
```
EmmaLRussell, line 22: This wasn't strictly to fix the tests, but I was getting what seemed like a weird race condition and I didn't think it would hurt to user fake timers in all tests to avoid any knock on effects real timers might have..

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
```diff
@@ -298,6 +287,14 @@ describe(`uploadModal `, () => {
...
         expect(modal.classes()).toContain("show");
 
+        //Refresh adrUpload upload files to trigger watch, which populates uploadFilesToADR
+        store.state.adrUpload.uploadFiles= {...fakeMetadata};
+
+        //Click ok to trigger population of uploadFilesPayload
+        const okBtn = modal.find("button.btn-red");
+        expect(okBtn.element)
+        await okBtn.trigger("click");
+
```
EmmaLRussell, line 297: This was the main reason why downloadIsReady wasn't ever returning true - it relies on `uploadFilesPayload` to be set, which in turn derived from `uploadFilesToADR`. This code triggers these to be populated. I'm not sure how it was working before with the previous code!

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
```diff
@@ -378,12 +375,23 @@ describe(`uploadModal `, () => {
...
+        //Refresh adrUpload upload files to trigger watch, which populates uploadFilesToADR
+        store.state.adrUpload.uploadFiles= {outputZip: fakeMetadata.outputZip};
+
+        //Click ok to trigger population of uploadFilesPayload
+        const okBtn = wrapper.find(".modal button.btn-red");
+        expect(okBtn.element)
+        await okBtn.trigger("click");
+
+        wrapper.vm.$store.state.downloadResults.spectrum.complete = true;
```
EmmaLRussell, line 394: Just setting the complete property in the download results rather than replacing the whole thing. This is to prevent `mockUploadFilesToADR` being invoked twice because of both `spectrum` and `summary` changing..

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
```diff
@@ -378,12 +375,23 @@ describe(`uploadModal `, () => {
     });
 
     it("can clear timer when download action is complete", async () => {
-        const store = createStore();
+        const store = createStore(fakeMetadata, {
+            summary: {complete: false, downloading: false } as any,
+            spectrum: {complete: false, downloading: false} as any
+        });
```
EmmaLRussell, line 381: Using local metadata here, as we will be editing to set complete in spectrum to true later in the test. Failing to do this was what caused the errors I had thought were down to a race condition!

mwoodbri (commented):
This worked very well for me. I was able to engineer a situation where it said it was uploading 4(!) files to ADR - but this was only as a result of cancelling the original upload, it still succeeded successfully, and I think this an unusual scenario so I don't think it's a significant issue.

The tests are looking very healthy now too.

Beyond that I just have two questions inline - I am not suggesting a problem, I just want to make sure I understand how this is all intended to work.

mwoodbri (changes requested):
Spotted an issue while testing: I engineered the backend to send an error when a download request is first submitted and noticed that the error field was set on all three resources, even though I only requested the summary download:
Screenshot from 2021-08-24 11-09-29

Possible cause in comments below, though if it is indeed the cause then I'm surprised there haven't been broader issues. Any fix will need a test.

carbowizzy (commented):
> Spotted an issue while testing: I engineered the backend to send an error when a download request is first submitted and noticed that the error field was set on all three resources, even though I only requested the summary download:
> Screenshot from 2021-08-24 11-09-29
> 
> Possible cause in comments below, though if it is indeed the cause then I'm surprised there haven't been broader issues. Any fix will need a test.

Thanks for spotting that, broader tests would have picked it up..

EmmaLRussell:
> This worked very well for me. I was able to engineer a situation where it said it was uploading 4(!) files to ADR - but this was only as a result of cancelling the original upload, it still succeeded successfully, and I think this an unusual scenario so I don't think it's a significant issue.

Should we/have we made a bug ticket for this, or has it been fixed?

mwoodbri:
> > This worked very well for me. I was able to engineer a situation where it said it was uploading 4(!) files to ADR - but this was only as a result of cancelling the original upload, it still succeeded successfully, and I think this an unusual scenario so I don't think it's a significant issue.
> 
> Should we/have we made a bug ticket for this, or has it been fixed? I agree it doesn't need fixing for this ticket, but should probably record it as an issue

I am re-testing this branch from scratch at the moment. Will create a ticket if I can reproduce this time around.

EmmaLRussell (approved):
I'd be happy to merge this now. 

There's still some slightly odd behaviour around failures, but I think that can be rolled into another ticket. I sabotaged the summary endpoint, then chose to upload both files. The summary error appeared on the main page, while the dialog was still showing. The 'Preparing' message disappeared but subsequently I could not choose to only upload the zip file (which should work) - dialog stayed open but nothing happened.  If I stayed on the page and opened the upload dialog again, the same thing happened when I chose to upload zip file only. If I then navigated to the previous step, then forward again, uploading zip file only did work. This is another rather obscure case, so if you agree, I will open a separate bug ticket.

carbowizzy:
> step

Thanks, I think we can get this merged and have hot fixes for it.

src/app/static/src/app/store/downloadResults/downloadResults.ts
```diff
@@ -0,0 +1,43 @@
...
+    return {
+        spectrum: {...initialDownloadResults},
+        coarseOutput: {...initialDownloadResults},
+        summary: {...initialDownloadResults}
+    }
+}
+
+export const downloadResults: Module<DownloadResultsState, RootState> = {
+    namespaced: true,
+    state: {...initialDownloadResultsState()},
```
mwoodbri, line 39: You could argue that `initialDownloadResults` should be a function like `initialDownloadResultsState` - to similarly help avoid state inadvertently being re-used as it was previously. That would remove need for various instances of spread operator. But that's never a bad idea as a defensive tactic - no need to change.

mwoodbri (approved):
All the failures I have previously seen are now fixed or logged.

I likewise have some misgivings about the modal - next time around we should consider simply making download and upload mutually exclusive, rather then changing behaviour based on whether it is open. But I think you do have to try quite hard to break it now, and even then the actual behaviour appears safe (i.e. non-destructive!). However, we should keep the relevant tickets near the front of the queue - in combination with any user feedback we get following release.

On that note: the only other thought at this stage is that retrieving outputs e.g. summary can _seem_ to take a lot longer than it used to - if you're not taking into account that the calibration step is accordingly faster. But we'll see what reception it gets!

PR #515, opened 2021-06-29
mrc-2499 Add @reside-ic/vue-charts to vue-shims.d.ts

mwoodbri:
## Description

Help IDEA find the type definitions for vue-charts. Making this change results in some new type errors when building via CLI, which this PR also fixes.

## Type of version change

None

## Checklist

### DIFF
--- a/src/app/static/src/app/components/baseline/Baseline.vue
+++ b/src/app/static/src/app/components/baseline/Baseline.vue
@@ -64,31 +64,31 @@
         name: "Baseline",
         computed: {
             ...mapState<BaselineState>(namespace, {
-                country: state => state.country,
-                pjnz: state => ({
+                country: (state: BaselineState) => state.country,
+                pjnz: (state: BaselineState) => ({
                     valid: !!state.country,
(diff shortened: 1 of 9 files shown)


EmmaLRussell (approved):
I don't really understand why adding the vue-shims change necessitated the change to mapStateProp, but that seems like a good thing to do anyway!

mwoodbri:
> I don't really understand why adding the vue-shims change necessitated the change to mapStateProp

Likewise... I dislike making apparently arbitrary changes, but the type checker wasn't happy and does seem like a logical change.

mwoodbri:
> > I don't really understand why adding the vue-shims change necessitated the change to mapStateProp
> 
> Likewise... I dislike making apparently arbitrary changes, but the type checker wasn't happy and does seem like a logical change.

Correction: at least some of these changes were due to me inadvertently type checking using a slightly newer version of TypeScript (3.9.10), which is stricter. But the changes in this PR can therefore be considered future-proofing ;)

Nicolas-Dolan (approved):

PR #516, opened 2021-06-29
mrc-2459 Reinstate login case insensitivity

Nicolas-Dolan:
## Description

Reinstates login case insensitivity originally added by mrc-2210.

Can be tested manually by logging in as Test.User@example.com

Note: the change is predicated on there being are no duplicates in the database.

## Type of version change
Minor 

## Checklist


src/app/src/main/kotlin/org/imperial/mrc/hint/security/HintDbProfileService.kt
```diff
@@ -25,4 +25,11 @@ class HintDbProfileService(dataSource: DataSource, passwordEncoder: PasswordEnco
             throw BadCredentialsException("Username and password must be provided")
         }
     }
+
+    override fun read(names: List<String?>?, key: String, value: String?): List<Map<String?, Any?>?>?
+    {
+        val attributesList = buildAttributesList(names)
+        val query = "select $attributesList from $usersTable where lower($key) = lower(:$key)"
+        return query(query, key, value)
+    }
```
mwoodbri, line 34: Note that this overrides pac4j's default implementation, which is (in Java):
```kotlin
    @Override
    protected List<Map<String, Object>> read(final List<String> names, final String key, final String value) {
        final String attributesList = buildAttributesList(names);

        final String query = "select " + attributesList + " from " + usersTable + " where " + key + " = :" + key;
        return query(query, key, value);
    }
```

mwoodbri (commented):
No problem with the code but I think this should really be a minor release because it goes beyond bug-fixing and results in a change in user-facing behaviour. Also a couple of new commits to fix.

EmmaLRussell (approved):
Works for me!

mwoodbri (approved):

PR #517, opened 2021-06-30
mrc-2485_tests Add Portuguese translation tests and missing translations

EmmaLRussell:
## Description

This branch builds on mrc-2485 (https://github.com/mrc-ide/hint/pull/512) (this PR's target branch) which provides all Portuguese translations by:
- adding all tests for Portuguese text (now included in the helper method `expectTranslated`)
- adding a small number of missing translations (commented inline)

I've checked over the app in Portuguese for any odd text wrapping etc, and it all looks OK to me, except there is a fairly ugly wrap in the header for the file upload for Household Survey (the ADR 🗸 gets wrapped onto a new line). I think trying to fix this e.g. by reducing the width of the map, or floating the ADR marker right, wouldn't necessarily be any better, so I'd propose leaving this unless you can think of a good solution. The 'Select new file' text is also partially hidden - but we already live with this for the French text!

## Type of version change
_Delete as appropriate_

None

## Checklist

### DIFF
--- a/src/app/static/src/app/components/modelOptions/ModelOptions.vue
+++ b/src/app/static/src/app/components/modelOptions/ModelOptions.vue
@@ -6,7 +6,7 @@
         </div>
         <dynamic-form v-if="!loading && !hasOptionsError"
                       v-model="modelOptions"
-                      submit-text="Validate"
+                      :submit-text="validateText"
                       @confirm="confirmEditing"
                       @submit="validate"
                       :required-text="requiredText"
@@ -63,6 +63,7 @@
         currentLanguage: Language
         selectText: string
         requiredText: string
+        validateText: string
     }
 
     interface Data {
@@ -95,6 +96,9 @@
             },
             requiredText() {
                 return i18next.t("required", {lng: this.currentLanguage})
+            },
+            validateText() {
+                return i18next.t("validate", {lng: this.currentLanguage})
             },
             editsRequireConfirmation: mapGetterByName("stepper", "editsRequireConfirmation"),
             modelOptions: {
--- a/src/app/static/src/app/components/projects/ProjectHistory.vue
+++ b/src/app/static/src/app/components/projects/ProjectHistory.vue
@@ -4,7 +4,7 @@
         <div id="headers" class="row font-weight-bold pt-2">
             <div class="col-md-1 header-cell"></div>
             <div class="col-md-3 header-cell" v-translate="'projectName'"></div>
-            <div class="col-md-1 header-cell">Versions</div>
+            <div class="col-md-1 header-cell" v-translate="'versions'"></div>
             <div class="col-md-2 header-cell" v-translate="'lastUpdated'"></div>
             <div class="col-md-1 header-cell" v-translate="'load'"></div>
             <div class="col-md-1 header-cell" v-translate="'renameProjectHistoryHeader'"></div>
--- a/src/app/static/src/app/components/projects/ProjectHistory.vue
+++ b/src/app/static/src/app/components/projects/ProjectHistory.vue
@@ -377,9 +377,10 @@
             createProject: mapActionByName(namespace, "createProject"),
             getProjects: mapActionByName(namespace, "getProjects"),
             versionCountLabel(project: Project) {
+                const lng = this.currentLanguage;
                 return project.versions.length == 1
-                    ? "1 version"
-                    : `${project.versions.length} versions`;
+                    ? `1 ${i18next.t("versionCountLabelSingle", {lng})}`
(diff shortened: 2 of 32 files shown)


src/app/static/src/app/components/modelOptions/ModelOptions.vue
```diff
@@ -6,7 +6,7 @@
         </div>
         <dynamic-form v-if="!loading && !hasOptionsError"
                       v-model="modelOptions"
-                      submit-text="Validate"
+                      :submit-text="validateText"
```
EmmaLRussell, line 9: Translate the 'Validate' button in the model options - apparently we weren't doing this for French yet.

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -4,7 +4,7 @@
         <div id="headers" class="row font-weight-bold pt-2">
             <div class="col-md-1 header-cell"></div>
             <div class="col-md-3 header-cell" v-translate="'projectName'"></div>
-            <div class="col-md-1 header-cell">Versions</div>
+            <div class="col-md-1 header-cell" v-translate="'versions'"></div>
```
EmmaLRussell, line 7: We didn't need to translate this for French, since version(s) is the same in French as in English...

src/app/static/src/app/components/projects/ProjectHistory.vue
```diff
@@ -377,9 +377,10 @@
             createProject: mapActionByName(namespace, "createProject"),
             getProjects: mapActionByName(namespace, "getProjects"),
             versionCountLabel(project: Project) {
+                const lng = this.currentLanguage;
                 return project.versions.length == 1
-                    ? "1 version"
-                    : `${project.versions.length} versions`;
+                    ? `1 ${i18next.t("versionCountLabelSingle", {lng})}`
+                    : `${project.versions.length} ${i18next.t("versionCountLabelPlural", {lng})}`;
```
EmmaLRussell, line 383: Again, translating '1 version'/'2 versions' wasn't necessary in French, but is in Portuguese. Also, 'versions' in Portuguese isn't just the world for 'version' with an s on the end, so we do need separate translations for singular and plural..

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -891,7 +900,7 @@ const pt: Partial<Translations> = {
     projectsHeaderCreate: "Criar um novo projeto",
     projectsHeaderReturn: "regressar ao projeto atual",
     projectsNoSelfShare: "Os projetos não podem ser partilhados com a conta do próprio utilizador",
-    promoteVersionHeader: "A copiar versão {{ versão }} para um novo projeto",
+    promoteVersionHeader: "A copiar versão {{ version }} para um novo projeto",
```
EmmaLRussell, line 903: The {{}} here is for interpolation so shouldn't be translated

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -911,7 +920,7 @@ const pt: Partial<Translations> = {
     saveVersionConfirm: "Guardar versão e continuar a editar",
     select: "Selecionar...",
     selectADR: "Selecionar conjunto de dados do ADR",
-    selectedDataset: "Conjunto de dados selecionado,",
+    selectedDataset: "Conjunto de dados selecionado:",
```
EmmaLRussell, line 923: I think this comma was probably a typo?

mwoodbri (approved):
Bom trabalho!

I think we might need to reconsider our approach if we add many more languages...

I was confused by the main forms being in English, but that's the same for French.

Are we eventually expecting a copy of the [troubleshooting guide](https://mrc-ide.github.io/naomi-troubleshooting/index-en.html) in Portuguese?

EmmaLRussell:
> Bom trabalho!
> 
:smile: 

> I think we might need to reconsider our approach if we add many more languages...
> 
Agreed!

> I was confused by the main forms being in English, but that's the same for French.
> 
Actually, @r-ash was right, if you switch to French before loading the options page, it does show French in the form. In Portuguese, it is passing the Accept-Language header as 'pt', so Rob should that be getting picked up by hintr to return the Portuguese options?

> Are we eventually expecting a copy of the [troubleshooting guide](https://mrc-ide.github.io/naomi-troubleshooting/index-en.html) in Portuguese?

Yes I believe that's going to be created after we're a bit more certain of which features are going to be available at the next workshop.

src/app/static/src/app/store/translations/locales.ts
r-ash: oops!

r-ash (approved):
Thanks for fixing all this

PR #520, opened 2021-07-20
mrc-2519 Import releases from ADR: fetch and display release data from ADR

Nicolas-Dolan:
## Description

I decided to merge the UI branch into this ticket. Allows user to select a release for a dataset as per page 3 of [mockup](https://docs.google.com/presentation/d/1l_cxM0K-B0lv_70P0mvSXohj3NHa5jQn54yz_bcHdFg/edit#slide=id.ge3db4969ea_0_8). You can add/remove additional releases  for each dataset [here](https://dev.adr.fjelltopp.org/organization/naomi-development-team).

Any guidance on whether I need additional integration tests or a selenium test would be appreciated. Note: the front-end ADR integration test will sometimes fail as it seems to be affected by people manually adding releases to datasets.

## Type of version change
Minor

## Checklist


mwoodbri (changes requested):
This looks very good. Needs an integration test though.

Doing it in two branches is fine with a couple of caveats:
- We should ensure that `master` is always deployable. This means that it should be possible to cleanly merge this branch, without any side effects. In particular:
  - No debug logging
  - Either no UI changes, or non-functional UI hidden behind a feature flag. I think you're ok here because you're just adding an empty `div` which doesn't affect layout.
- "Announcing" the feature in `NEWS.md` is a potential source of confusion because this only includes some of the foundational work. This is where having one branch per ticket keeps things simple, but here you can just make it clearer that this is really just adding the endpoint rather than the feature itself.
- You'll have to ensure that when this branch is merged then you manually re-assign/update the ticket, as the YouTrack hook will close it automatically.

Overall you might have been better keeping this branch to just backend changes.

I'm confused about your testing guidance - I get a console log message whether I use Burundi or not. It doesn't tell me more than the mutation logging (and less than I can get by looking at the network inspector).

src/app/static/src/app/components/adr/SelectRelease.vue
```diff
@@ -0,0 +1,19 @@
...
+    import { mapActionByName } from "../../utils";
+
+    export default Vue.extend({
+        props: ["selectedDataset"],
+        methods: {
+            getReleases: mapActionByName("adr", "getReleases"),
+        },
+        watch: {
+            selectedDataset(id) {
+                this.getReleases(id);
```
mwoodbri, line 15: I think there's a bug here - the perils of incomplete UI having unintended consequences!

If I click the cross in the datasets dropdown then I now get an ADR error displayed in the UI, because `id` is `undefined` here

Nicolas-Dolan: I've put a check in the watcher that id is truthy

EmmaLRussell:
>     * "Announcing" the feature in `NEWS.md` is a potential source of confusion because this only includes some of the foundational work. This is where having one branch per ticket keeps things simple, but here you can just make it clearer that this is really just adding the endpoint rather than the feature itself.
> 

I think it would be fine to not update the version at all, since there are no application behaviour change from this branch,

mwoodbri:
> > "Announcing" the feature in `NEWS.md` is a potential source of confusion because this only includes some of the foundational work. This is where having one branch per ticket keeps things simple, but here you can just make it clearer that this is really just adding the endpoint rather than the feature itself.
> 
> I think it would be fine to not update the version at all, since there are no application behaviour change from this branch,

Maybe. In the general case if a release was made and something broke (i.e. the application behaviour inadvertently changed, as could have happened here) then the NEWS file would be my first port of call to narrow down the cause. But git obviously provides a more reliable changelog!

src/app/static/src/tests/components/adr/selectRelease.test.ts
```diff
@@ -0,0 +1,16 @@
...
+import Vuex from "vuex";
+
+describe("select release", () => {
+    
+    const getStore = () => {
+        const store = new Vuex.Store({});
+        return store;
+    }
+
+    it("renders select release", () => {
```
EmmaLRussell, line 12: Should also include a test that the action is invoked when `selectedDataset` changes (if truthy)

EmmaLRussell (changes requested):
Code looks good to me. Just a missing component test I think.

I would also have been inclined to limit this branch to just adding the endpoint, and testing it through integration testing, rather than including calls to the endpoint from the front end which may fail and affect app behaviour.

src/app/static/src/app/components/adr/SelectRelease.vue
```diff
@@ -0,0 +1,167 @@
+<template>
+    <div id="selectRelease" v-if="newDatasetId && releases.length">
+        <div class="pt-1">
```
mwoodbri, line 3: On my screen this is quite cramped 
```suggestion
        <div class="pt-2">
```

mwoodbri (changes requested):
This works well, so my comments just relate to some implementation details - none of which should require major changes

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -408,6 +420,10 @@
                     window.clearInterval(this.pollingId);
                 }
             },
+        updateDatasetVersion(releaseId, enableImport){
```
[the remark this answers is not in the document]
Nicolas-Dolan: For `selectDataset` to be able deduce for itself, it would need some knowledge of whether `useRelease` or `useData` is selected in the UI. If it only received a `null` or `undefined` `datasetVersionId` from `selectRelease`, would it be able to distinguish on its own whether that was because no release had been selected or because `useData` was selected and the releases tree select was disabled?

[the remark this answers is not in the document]
mwoodbri: I see now, sorry. So `valid` or something similar might be better than `enableImport`. But maybe the child component should instead emit two events: `valid` events (to indicate wther the form is valid), and `releaseId` values (only emitted when the form is valid). It doesn't feel quite right that this component is overwriting its `newDatasetVersionId` version even when the child form isn't in a valid state.

src/app/static/src/app/components/adr/SelectRelease.vue
```diff
@@ -0,0 +1,167 @@
...
+            <input
+                type="radio"
+                id="useData"
+                value="useData"
```
EmmaLRussell, line 7: Would 'useLatest' be a little more meaningful than 'useData' - both choices use data...

src/app/static/src/app/components/adr/SelectRelease.vue
```diff
@@ -0,0 +1,167 @@
...
+                newReleaseId: null,
+                choiceADR: "useData",
+            };
+        },
+        computed: {
+            releases: mapStateProp<ADRState, any[]>(
+                namespace,
+                (state: ADRState) => state.releases
+            ),
+            selectedDatasetOrRelease() {
```
EmmaLRussell, line 114: This computed value is just used to determine if the selections are in a valid state isn't it? I don't really like bringing in `newDatasetId` into it - it seems a but irrelevant if it's just going to be used to generate a boolean. Would this be a little more explicable as something like 
```
valid() {
  return (this.choiceADR === "useData) || this.newReleaseId;
}
```
That could fit in nicely with Marks' suggestion of adding a `valid` event

EmmaLRussell (changes requested):
Looks good, I like the tooltips and the release notes in the drop down, and the validation logic works nicely. 

This isn't yet merged with the changes added in https://github.com/mrc-ide/hint/pull/521? Might be good to merge that in, and check that all works end to end i.e that the correct release is requested from the backend when selected in the dialog.

Also - would it be worth pre-selecting the selected release when you re-open the dialog after importing? Seems a bit odd to pre-select the dataset but not the selected release, if any.

I also wonder if we should show the imported release name  in the area where we currently display the "Selected dataset" (this isn't something that went in the mockups but would perhaps make sense - maybe one for another ticket?)

Nicolas-Dolan:
> Also - would it be worth pre-selecting the selected release when you re-open the dialog after importing? Seems a bit odd to pre-select the dataset but not the selected release, if any.
> 
> I also wonder if we should show the imported release name in the area where we currently display the "Selected dataset" (this isn't something that went in the mockups but would perhaps make sense - maybe one for another ticket?)

I've added these two features to a s[eparate ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2543)

src/app/static/src/app/components/adr/SelectRelease.vue
```diff
@@ -0,0 +1,167 @@
...
+            </span>
+            <br />
+        </div>
+
+        <label
+            for="releaseSelector"
+            class="font-weight-bold"
+            v-translate="'releases'"
+        ></label>
+        <tree-select
```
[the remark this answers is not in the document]
Nicolas-Dolan: I prefer for it to be disabled for the time being as there is already a lot of UI popping in and out as it is. This also follows the mock ups.

mwoodbri (changes requested):
A problem encountered when testing manually: starting from scratch, if I select Antarctica 1.0 and then switch to Burundi an error is shown in the console and the "Selected dataset" remains Antarctica. Could you take a look? Maybe that dataset ID is being updated but version ID isn't.

Nicolas-Dolan:
> A problem encountered when testing manually: starting from scratch, if I select Antarctica 1.0 and then switch to Burundi an error is shown in the console and the "Selected dataset" remains Antarctica. Could you take a look? Maybe that dataset ID is being updated but version ID isn't.

@mwoodbri is that after you have clicked import on the Antarctica release?

mwoodbri:
> > A problem encountered when testing manually: starting from scratch, if I select Antarctica 1.0 and then switch to Burundi an error is shown in the console and the "Selected dataset" remains Antarctica. Could you take a look? Maybe that dataset ID is being updated but version ID isn't.
> 
> @mwoodbri is that after you have clicked import on the Antarctica release?

Yes

src/app/static/src/app/store/adr/actions.ts
```diff
@@ -59,6 +60,13 @@ export const actions: ActionTree<ADRState, RootState> & ADRActions = {
             });
     },
 
+    async getReleases(context, selectedDatasetId) {
+        await api<ADRMutation, ADRMutation>(context)
+            .withError(ADRMutation.SetADRError)
```
mwoodbri, line 65: Need to be careful here: ADR errors are only displayed in the modal... which is hidden as soon as you click Import. So they are effectively invisible. Could either write to baseline errors or attempt to keep dialog open until dataset details are retrieved (latter sounds more complicated).

EmmaLRussell: ..or show ADR errors in baseline component, though that show more than we want if it includes all possible ADR errors

EmmaLRussell (commented):
Thanks, that resolves my comments from last time. Can confirm that release is now fetched as expected. 

I also get the error Mark describes when loading Burundi after Antarctica v1.0. It seems to be because it's trying to fetch Burundi with the release id previously selected for Antarctica, so I guess that's not getting cleared.

Nicolas-Dolan:
With the adr frontend integration tests on bk, I seem to be getting different releases returned every time. I tried to narrow down the number of options it would return by only getting releases for the Antarctica dataset but even that seems variable.

mwoodbri (changes requested):
Looking very good now. Three suggestions below.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/adr/ADRTests.kt
```diff
@@ -67,6 +67,24 @@ class ADRTests : SecureIntegrationTests()
...
+
+        val id = if (isAuthorized == IsAuthorized.TRUE)
+        {
+            val data = ObjectMapper().readTree(datasets.body!!)["data"]
+            data.first()["id"].textValue()
+        }
+        else "fake-id"
+
+        val result = testRestTemplate.getForEntity<String>("/adr/datasets/$id/releases")
+        assertSecureWithSuccess(isAuthorized, result, null)
```
mwoodbri, line 85: You could additionally check the response is actually an array using a similar pattern to the test above i.e:
```kotlin
        if (isAuthorized == IsAuthorized.TRUE)
        {
            val data = ObjectMapper().readTree(result.body!!)["data"]
            assertThat(data.isArray).isTrue
        }
```

src/app/static/src/tests/components/adr/selectRelease.test.ts
```diff
@@ -0,0 +1,178 @@
...
+        expect(select.props("multiple")).toBe(false);
+        expect(select.props("searchable")).toBe(true);
+
+        const expectedOptions = [
+            {
+                id: "releaseId",
+                label: "releaseName",
+                customLabel: `releaseName
+                <div class="text-muted small" style="margin-top:-5px; line-height: 0.8rem">
+                    releaseNotes<br/>
```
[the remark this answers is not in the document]
Nicolas-Dolan: @mwoodbri would you prefer it to be like the above with just an empty line if there are no notes, or like the below? 
image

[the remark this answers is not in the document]
mwoodbri: I think @EmmaLRussell did the mockups so would be good to get her opinion... though not sure notes were included at all

I'm not sure it matters - your screenshot looks good. Bear in the mind that if we're not adding any notes to releases when we create them then they're likely to mostly be empty, so we'll end up with a lot of whitespace if we leave an empty line for each one.

mwoodbri (approved):
LGTM!

src/app/static/src/tests/components/adr/selectRelease.test.ts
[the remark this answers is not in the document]
EmmaLRussell: Notes weren't in the mockups, I didn't think to include them. I think I'd slightly prefer having notes and no-notes releases have uniform heights. Probably worth checking it clips nicely if notes are long.

EmmaLRussell (approved):
Long notes don't clip, but they wrap ok. Good to go I think. :+1: 
Screenshot from 2021-08-02 09-32-12

mwoodbri:
> Long notes don't clip, but they wrap ok. Good to go I think. +1

Agreed - if we find notes on the scale of War and Peace then we can reconsider. But good to make full note accessible without having to visit ADR website, and as this ends up with variable heights anyway then it vindicates current approach to styling.

PR #525, opened 2021-07-30
mrc-2544 Combine input choropleth tabs 

EmmaLRussell:
## Description
As per the [mockups](https://imperiallondon.sharepoint.com/:p:/r/sites/HIVInferenceGroup-WP/_layouts/15/guestaccess.aspx?e=IxIMI9&share=EauUdjWV7W1EiN2ErHz4R-wBD2mN6QUh6d0b4iJBPeXOWQ), this branch combines the existing 'Household survey'/'ART'/'ANC testing' tabs in step 2 into a single 'Map' tab, and allow selected datasource to be chosen in a 'Data source' drop-down instead.

It also moves the file and filter selections under the tabs - the file controls will move as part of mrc-2518

This branch does not really enhance the UI of itself, but is required preparatory work for adding the Input Time Series aka ShinyRob charts to a second tab in this step. 

To be consistent with the appearance of the Filters drop-downs used elsewhere, this uses the same Treeselect component

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
```diff
@@ -168,16 +183,33 @@
...
+                (state: RootState) => state.language
+            ),
             filterTableIndicators() {
                 return this.sapIndicatorsMetadata.filter((val: ChoroplethIndicatorMetadata) => val.indicator === this.plottingSelections.indicatorId)
+            },
+            dataSourceOptions() {
+                const options = [];
+                const lang = {lng: this.currentLanguage};
+                if (this.survey.available) {
+                    options.push({id: "2", label: i18next.t("survey", lang)});
```
mwoodbri, line 197: I guess you could pass in the translation key rather than the translation itself, and then set up the `tree-select` `option-label` to translate it, but I see any advantage besides reducing a tiny bit of repetition.

mwoodbri (approved):
Cool stuff - works very well and code is clear

Nicolas-Dolan (approved):
Works nicely. Perhaps this brings us a step closer to having all the filters in the one place :)

src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
EmmaLRussell: Good idea

EmmaLRussell: Actually, I am just going to leave that is - worked ok to put the translate into the option and value slots of the treeselect, but that ended up being just as verbose as this, it's very nasty to unit test, and I've ended up with race condition failures I can't face chasing down right now!

mwoodbri: Sorry - one of those ideas that's only good in theory!

PR #527, opened 2021-08-16
mrc-2523 Upload to core output resources

Nicolas-Dolan:
## Description
Uploads to core set of resources (namely, the output summary and zip) rather than creating a new resource per project. To test, select a dataset, proceed to save results step, and click the upload button. It should successfully upload and the Naomi Output ZIP and Naomi Results and Summary Report resources should have updated on the dev adr website. If using the tester adr access key, Burundi should have write permissions and can be checked [here](https://dev.adr.fjelltopp.org/inputs-unaids-estimates/burundi-inputs-unaids-estimates-2021).

## Type of version change
Patches

## Checklist


mwoodbri (commented):
Looks good.

One question about deploying: in practice does it depend on the user being able to create a release at the same time, or can it be released prior to mrc-2522?

EmmaLRussell (changes requested):
I think is almost there, but not quite. `constructUploadFileWithResourceName` will look for a resource with the same name, unlike `constructUploadFile` which only matches on type. For the new output file behaviour, we do want to match on type (it doesn't matter if the core output resources in the dataset have different names from the ones we were expecting) - as we do for input files. But, unlike for the input files, we do still want to be able to create the core resource if it doesn't exist yet, so we do need to include the name when constructing the output file object. 

Since `constructUploadFileWithResourceName` is only used for the output files, I think the easiest fix for this is to change that method so that it will not match by resource name when finding the resource. I think we should also use the found resource's name when constructing the file, so that we don't try to rename an existing resource (though I'm not actually sure what the ADR behaviour would be in that situation). Comments below just to explain changes in code: 
```
export const constructUploadFileWithResourceName = (datasetWithResources: any, index: number, resourceType: string,
 resourceFilename: string, displayName: string, resourceName: string): UploadFile => {
 const resource = findResource(datasetWithResources, resourceType); //<- remove resourceName param
 const name = resource?.name || resourceName; //<- use name from found resource, if any
 return getUploadFileFromResource(resource, name, index, resourceType, resourceFilename, displayName);

};
```

EmmaLRussell:
> Looks good.
> 
> One question about deploying: in practice does it depend on the user being able to create a release at the same time, or can it be released prior to [mrc-2522](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2522)?

That's a good point. I think ideally all the upload to release changes would go at the same time, so we could use a feature switch. But if that's going to be fiddly here, I probably wouldn't insist on it. It won't be the end of the world for us to upload to core resource without releases for a few weeks. Users can still get to previous versions via the Activity Stream. But we should definitely get on to the remaining releases work asap in that case.

mwoodbri:
> For the new output file behaviour, we do want to match on type only

This completely passed me by - I suppose this is to handle the case where a summary or zip file had previously been uploaded manually, using a name different to the one we're using here?

EmmaLRussell:
> > For the new output file behaviour, we do want to match on type only
> 
> This completely passed me by - I suppose this is to handle the case where a summary or zip file had previously been uploaded manually, using a name different to the one we're using here?

Yes, or it could also, I think, be the case where the core resource has been created in the dataset as an empty resource (no upload yet), on dataset creation.

mwoodbri:
> Yes, or it could also, I think, be the case where the core resource has been created in the dataset as an empty resource (no upload yet), on dataset creation.

Interesting - I guess the normal state of affairs is that ADR presents the "Missing resources" list based on the resources being part of the `inputs-unaids-estimates` schema but not yet uploaded? Regardless, clearly ADR has default names for these resources (e.g. "Naomi Output ZIP") and ideally we would just inherit those names rather than providing our own. But surely a non-issue in practice.

Nicolas-Dolan:
I've added the requested updates. Something to note is that by not forcing a naming convention on the core resources, we are potentially making it harder for the user on the adr website to determine which resource actually is the core resource. This is compounded by the fact that if multiple sets of resources already exist, the util method will arbitrarily pick one to be the core resource from that point onwards, keeping whatever name that resource had.

EmmaLRussell:
> I've added the requested updates. Something to note is that by not forcing a naming convention on the core resources, we are potentially making it harder for the user on the adr website to determine which resource actually is the core resource. This is compounded by the fact that if multiple sets of a particular resource type already exist, the util method will arbitrarily pick one to be the core resource from that point onwards, keeping whatever name that resource had.

I think that's OK, though, since the whole point is that there should only be one resource per type. If they have multiple resources of the same type in the dataset, all bets are off, as they would be for input resources. We've been proliferating output resources since we introduced ADR uploads, but this functionality hasn't yet been used at any of the user workshops.

src/app/static/src/tests/utils.test.ts
```diff
@@ -190,14 +190,6 @@ describe("utils", () => {
...
-                    resource_type: "test-type",
-                    id: "456",
-                    name: "A different resource",
-                    last_modified: "2021-03-01",
-                    metadata_modified: "2021-02-28",
-                    url: "http://test2"
-                },
                 {
                     resource_type: "test-type",
                     id: "123",
```
Nicolas-Dolan, line 195: I should probably have passed a different resource name argument to this test, just to show it wasn't being overwritten in this case.

EmmaLRussell:
> > Looks good.
> > One question about deploying: in practice does it depend on the user being able to create a release at the same time, or can it be released prior to [mrc-2522](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2522)?
> 
> That's a good point. I think ideally all the upload to release changes would go at the same time, so we could use a feature switch. But if that's going to be fiddly here, I probably wouldn't insist on it. It won't be the end of the world for us to upload to core resource without releases for a few weeks. Users can still get to previous versions via the Activity Stream. But we should definitely get on to the remaining releases work asap in that case.

Maybe we should merge from here intp an 'epic' branch rather than directly into master?

src/app/static/src/tests/utils.test.ts
EmmaLRussell: Yes, that would be worth testing

EmmaLRussell (approved):
Looks good to me - I think you're right about modifying that test to check resource name isn't overwritten when it exists. And should consider merging this into an intermediate epic branch rather than directly into master.

Nicolas-Dolan:
> And should consider merging this into an intermediate epic branch rather than directly into master.
Ok I guess I do that by setting up an epic PR and somehow have this pointed at that instead of master?

mwoodbri:
> > And should consider merging this into an intermediate epic branch rather than directly into master.
> 
> Ok I guess I do that by setting up an epic PR and somehow have this pointed at that instead of master?

1. Create new epic issue in Buildkite - probably by splitting mrc-2421. Add relevant tickets as subtasks of this epic i.e. mrc-2523, mrc-2522
2. Create a new branch from `master`, named after epic issue
3. Target this PR at that branch, rather than `master`
4. Later, when other PR(s) have also been merged into epic branch, create a PR from epic branch to `master`

mwoodbri (approved):

PR #528, opened 2021-08-18
Mrc 2564 Move survey and program (SAP) UI component to baseline step

carbowizzy:
This PR moves input upload to baseline, including corresponding actions and tests.

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
```diff
@@ -85,9 +53,9 @@
 
 <script lang="ts">
     import Vue from "vue";
-    import Treeselect from '@riophae/vue-treeselect';
+    import TreeSelect from '@riophae/vue-treeselect';
```
mwoodbri, line 56: I was quite surprised this worked - but turns out the package does both a default and named export. We seem to use both `Treeselect` and `TreeSelect` at the moment - I have no preference.

src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
```diff
@@ -156,26 +122,14 @@
                     return surveyAndProgram.selectedDataType != null;
                 },
                 anc: ({surveyAndProgram}: {surveyAndProgram: SurveyAndProgramState}) => ({
-                    valid: !!surveyAndProgram.anc,
-                    fromADR: !!surveyAndProgram.anc?.fromADR,
-                    error: surveyAndProgram.ancError,
-                    existingFileName: (surveyAndProgram.anc && surveyAndProgram.anc.filename)|| surveyAndProgram.ancErroredFile,
                     available: !surveyAndProgram.ancError && surveyAndProgram.anc
-                } as PartialFileUploadProps),
+                } as any),
```
mwoodbri, line 126: Not sure you even need the `as any` here - given there's no need to explicitly broaden the type

mwoodbri (commented):
Obviously this is a transitional step so not the easiest thing to review, but it seems to roughly have a net zero effect on line count!

Do you have any further guidance on manual testing i.e. whether it makes sense at all and if so whether the app should still be fully functioning? Superficially it is but I guess some of the validation won't be, looking at the next ticket to be implemented?

src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
carbowizzy: You are right, any removed.

src/app/static/src/tests/components/surveyAndProgram/fileUploads.ts
```diff
@@ -110,25 +124,25 @@ export function testUploadComponent(name: string, position: number) {
 
     it(`${name} upload is valid if data is present`, () => {
         const store = createSut(successState);
-        const wrapper = shallowMount(SurveyAndProgram, {store, localVue});
+        const wrapper = shallowMount(Baseline, {store, localVue});
```
EmmaLRussell, line 127: These tests should really move into the `components/baseline` folder to test functionality which is now happening from the `Baseline` component

src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
EmmaLRussell: Me neither really - `TreeSelect` is a little easier to read

EmmaLRussell (commented):
LGTM, behaves how I would expect it to in this intermediate stage.

src/app/static/src/tests/components/surveyAndProgram/fileUploads.ts
carbowizzy: done

src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
```diff
@@ -9,47 +9,15 @@
         </div>
```
Nicolas-Dolan, line 9: Silly question, but is this component eventually going to be discontinued/replaced or will it still serve some function?

carbowizzy: I will serve as Review inputs where the charts will be displayed.

EmmaLRussell (approved):
Thanks!

Nicolas-Dolan (approved):

mwoodbri (approved):

PR #529, opened 2021-08-18
Mrc 2566 Remove validation, reformat display and rename stepper to Review Inputs

carbowizzy:
This PR majorly renamed steppers name and no need to remove validation as they all work together seamlessly. 

## Type of version change
_Delete as appropriate_

Minor

## Checklist


EmmaLRussell (changes requested):
Unfortunately, this causes display issues when loading pre-existing projects:
Screenshot from 2021-08-19 17-56-29

Simplest solution is to retain the old translation keys, even though they're now misleading, and just upload the text. Alternative would be to convert the keys on project load. Not actually sure where the issue is, probably stepper state? Another alternative would be to avoid dependency on loaded step text keys if possible.

I can't see any removal of validation or reformatting of display in this changeset, but I don't think either are needed? Validation is dealt with in https://github.com/mrc-ide/hint/pull/530 ?

carbowizzy:
> old translation keys

I know where the issue comes from, steppers are stored in browser storage so it retained the previous translation key. The questions is, when does browser storage gets cleared ? 

I am sure we shouldn't have old translate key restated.

EmmaLRussell:
> > old translation keys
> 
> I know where the issue comes from, steppers are stored in browser storage so it retained the previous translation key. The questions is, when does browser storage gets cleared ?
> 
> I am sure we shouldn't have old translation key restated.

Yeah, it's the textKeys in the stepper state as stored. It's not just an issue with local storage though, this will also apply to projects loaded from the backend, or saved to file. We could try modifying load so that it only ever uses the latest text keys, from initialState, rather than the steps in the saved state.

src/app/static/src/app/store/load/actions.ts
```diff
@@ -70,6 +71,7 @@ export const actions: ActionTree<LoadState, RootState> & LoadActions = {
             await (dispatch("projects/createProject", projectName, {root: true}));
             savedState.projects.currentProject = rootState.projects.currentProject;
             savedState.projects.currentVersion = rootState.projects.currentVersion;
+            savedState.stepper.steps = initialStepperState().steps.filter(step => step.number <= 2);
```
EmmaLRussell, line 74: This is close but not quite what's needed. We don't want to blitz all of the existing stepper state - this will remove all but the first two steps. We just to replace the text keys with those from `initialStepperState()`.
I think we can also move this outside that `if` clause - this should work when a guest is loading from a saved file, as well as when a logged in user loads from a project.  
UPDATE: Having tested this suggested change, it does work for loading a saved file but appears to have no effect when loading a previously saved project (we still see the untranslated new text keys) so there must be some difference in the load logic which I'm missing
```
const latestKeys = initialStepperState();
savedState.stepper.steps.forEach((step: StepDescription) => {
  step.textKey = latestKeys.steps.find(latestStep => latestStep.number == step.number)!.textKey;
});

if (!rootGetters.isGuest) {
    await (dispatch("projects/createProject", projectName, {root: true}));
    savedState.projects.currentProject = rootState.projects.currentProject;
    savedState.projects.currentVersion = rootState.projects.currentVersion;
}
```

EmmaLRussell: Furhter update: OK, I think I figured it out, it's because load from project goes through a different code path, and doesn't hit load/setFiles action. So we can take the change from there, and put it into a slightly later action which is shared by both methods. If I add the line `
savedState.stepper.steps = initialStepperState().steps; //ensure latest step translation keys
`
at the start of load/getFilesAndLoad, then the load with new step names works for me, for both loading files as guest and loading projects.

PR #530, opened 2021-08-18
Mrc 2565 Add validation to survey and program (SAP) input on Baseline step

carbowizzy:
Add validation to survey and program (SAP) input on Baseline step

## Type of version change
_Delete as appropriate_

Minor 

## Checklist

### DIFF
--- a/src/app/static/src/app/store/stepper/getters.ts
+++ b/src/app/static/src/app/store/stepper/getters.ts
@@ -18,7 +18,7 @@ export const getters: StepperGetters & GetterTree<StepperState, RootState> = {
     },
     complete: (state: StepperState, getters: any, rootState: RootState, rootGetters: any) => {
         return {
-            1: rootGetters['baseline/complete'] && rootGetters['metadata/complete'],
+            1: rootGetters['baseline/complete'] && rootGetters['metadata/complete'] && rootGetters['surveyAndProgram/complete'],
             2: rootGetters['surveyAndProgram/complete'],
             3: rootGetters['modelOptions/complete'],
             4: rootGetters['modelRun/complete'],
--- a/src/app/static/src/app/store/stepper/getters.ts
+++ b/src/app/static/src/app/store/stepper/getters.ts
@@ -19,7 +19,7 @@ export const getters: StepperGetters & GetterTree<StepperState, RootState> = {
     complete: (state: StepperState, getters: any, rootState: RootState, rootGetters: any) => {
         return {
             1: rootGetters['baseline/complete'] && rootGetters['metadata/complete'] && rootGetters['surveyAndProgram/complete'],
-            2: rootGetters['surveyAndProgram/complete'],
+            2: rootGetters['baseline/complete'] && rootGetters['metadata/complete'] && rootGetters['surveyAndProgram/complete'],
             3: rootGetters['modelOptions/complete'],
             4: rootGetters['modelRun/complete'],
             5: rootState.modelCalibrate.complete,
--- a/src/app/static/src/tests/components/stepper.test.ts
+++ b/src/app/static/src/tests/components/stepper.test.ts
@@ -164,7 +164,7 @@ describe("Stepper component", () => {
         validatedConsistent: true
     };
 
-    const completeSurveyAndProgramState = {
+    const completedSurveyAndProgramState = {
         survey: mockSurveyResponse(),
         programError: null,
         ancError: null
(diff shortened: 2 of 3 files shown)


carbowizzy:
I have also left the RootState  reset for Survey and Program as it because the states still work independently of each other and can be used together seamlessly.

mwoodbri (changes requested):
I'm lacking some guidance on how to test this, but it works how I'd assume - when all files have been uploaded  in step 1 you can go to either step 2 or 3.

A couple of minor comments inline.

Slight concern that when I first go to the app in my browser it's complaining about not be able to load my data - which implies some inconsistency in format, despite us aiming for backwards compatibility. But there might be a simple explanation for that?

Finally, ticket says:
> Modify root state, change “resetState” behaviour for Baseline and SAP to account for new validation behaviour

But I can't see anything related in this PR? So either the ticket or the code will need to be updated.

src/app/static/src/app/store/stepper/getters.ts
```diff
@@ -18,8 +18,8 @@ export const getters: StepperGetters & GetterTree<StepperState, RootState> = {
     },
     complete: (state: StepperState, getters: any, rootState: RootState, rootGetters: any) => {
         return {
-            1: rootGetters['baseline/complete'] && rootGetters['metadata/complete'],
-            2: rootGetters['surveyAndProgram/complete'],
+            1: rootGetters['baseline/complete'] && rootGetters['metadata/complete'] && rootGetters['surveyAndProgram/complete'],
+            2: rootGetters['baseline/complete'] && rootGetters['metadata/complete'] && rootGetters['surveyAndProgram/complete'],
```
EmmaLRussell, line 22: Seems a bit odd to include baseline and metadata here, but I guess this is going to get removed in the next ticket anyway.

src/app/static/src/tests/components/stepper.test.ts
```diff
@@ -279,20 +295,20 @@ describe("Stepper component", () => {
...
         const wrapper = createReadySut(completedBaselineState,
-            {},
+            completeSurveyAndProgramState,
             {plottingMetadata: "TEST DATA" as any});
         const steps = wrapper.findAll(Step);
         expect(steps.at(0).props().enabled).toBe(true);
         expect(steps.at(1).props().enabled).toBe(true);
         expect(steps.at(0).props().complete).toBe(true);
-        expect([2, 3, 4, 5].filter(i => steps.at(i).props().enabled).length).toBe(0);
+        expect([2, 3, 4, 5].filter(i => steps.at(i).props().enabled).length).toBe(1);
```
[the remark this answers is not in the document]
EmmaLRussell:
```suggestion
        expect(steps.at(2).props().enabled).toBe(true);
        expect([3, 4, 5].filter(i => steps.at(i).props().enabled).length).toBe(0);
```

src/app/static/src/tests/components/stepper.test.ts
```diff
@@ -279,20 +295,20 @@ describe("Stepper component", () => {
         expect([1, 2, 3, 4, 5].filter(i => steps.at(i).props().enabled).length).toBe(0);
     });
 
-    it("upload surveys step is enabled when baseline step is complete", () => {
+    it("Review inputs step is enabled when baseline step is complete", () => {
```
EmmaLRussell, line 298: Should really have two tests to test both that Review inputs is *not* enabled when we have only completedBaselineState, and also that it is enabled  and complete (and subsequent step is enabled) when we have both completedBaselineState and completeSurveyAndProgram state. And rename each test accordingly.

EmmaLRussell (commented):
Works for me, as I would expect it. I wonder if we do need to revisit the validation rules - as  things stand, if you remove a baseline file, it silently clears any SAP files which are loaded. This makes some sense for the pjnz/geojson, which are fundamental to the project, but less so for the population file, and it does seem more confusing that this happens now that all files are on the same step together. It would be good to discuss this with the Naomi team to check what their preferred behaviour would be - that could go into another ticket or this one.

src/app/static/src/app/store/stepper/getters.ts
EmmaLRussell: Ignore this comment, this is right enough

mwoodbri:
> Works for me, as I would expect it. I wonder if we do need to revisit the validation rules - as things stand, if you remove a baseline file, it silently clears any SAP files which are loaded. This makes some sense for the pjnz/geojson, which are fundamental to the project, but less so for the population file, and it does seem more confusing that this happens now that all files are on the same step together. It would be good to discuss this with the Naomi team to check what their preferred behaviour would be - that could go into another ticket or this one.

I ignored this due to consistency with current behaviour, and additionally I guess that the files are never "lost" i.e. they can either be re-attached from your local machine or from ADR. But agree that at least a warning would be more user-friendly - if the science team consider it a priority.

mwoodbri (approved):
I haven't been able to reproduce the "unable to load data error" so hopefully just a problem at my end (though I see it's a concern for mrc-2566).

So looking good!

EmmaLRussell (approved):

PR #532, opened 2021-08-19
mrc-2522 Upload release to ADR: front end UI and logic

Nicolas-Dolan:
## Description

Creates UI so user can choose between uploading all files with a release or just uploading individual files. Integrates mrc-2523 and both PRs are pointed at epic branch mrc-2572 so that they can be merged together. To test, select a dataset, proceed to save results step, click the upload button, choose the type of upload in the radial, and click ok. It should successfully upload and appropriate resources should have updated on the dev adr website. If using the tester adr access key, Burundi should have write permissions and the resources can be checked [here](https://dev.adr.fjelltopp.org/inputs-unaids-estimates/burundi-inputs-unaids-estimates-2021) and the releases [here](https://dev.adr.fjelltopp.org/inputs-unaids-estimates/burundi-inputs-unaids-estimates-2021/releases). Each release requires a unique name (derived from the project name and version number) and new activity (ie, changes to the output since the last upload). Error messages should reflect these reasons for failure. Note, I've noticed a bug where going back to recalibrate does not trigger the new version behaviour. I need to investigate whether the bug is unique to this branch.

Mock ups are [here](https://docs.google.com/presentation/d/1l_cxM0K-B0lv_70P0mvSXohj3NHa5jQn54yz_bcHdFg/edit#slide=id.ge3db4969ea_0_53).

Note: I've added a second status message after the successful upload to indicate whether a new release was successfully created. This was to give context to the accompanying error message, should the release fail.
image

## Type of version change
Minor

## Checklist


src/app/static/src/app/store/adrUpload/adrUpload.ts
```diff
@@ -11,6 +11,8 @@ export interface ADRUploadState {
     uploadError: Error | null,
     uploading: boolean,
     uploadComplete: boolean,
+    releaseCreated: boolean,
+    releaseNotCreated: boolean,
```
Nicolas-Dolan, line 15: `releaseFailed` might be more appropriate. Let me know if you would like me to change.

src/app/static/src/app/components/downloadResults/DownloadResults.vue
```diff
@@ -42,6 +42,18 @@
...
+                <div id="releaseCreated" v-if="releaseCreated || releaseNotCreated" class="d-flex align-items-end">
+                    <div class="d-flex align-items-center height-40 mr-1">
+                        <span class="font-weight-bold" v-translate="releaseCreated ? 'releaseCreated' : 'releaseNotCreated'"></span>
+                    </div>
+                    <div class="d-flex align-items-center height-40">
+                        <tick color="#e31837" v-if="releaseCreated" width="20px"></tick>
+                        <svg class="tick" id="cross" v-if="releaseNotCreated" width="20px">
+                            <path d="M 10,10 l 90,90 M 100,10 l -90,90" stroke="#e31837" stroke-width="15"
+                                fill="none"></path>
+                        </svg>
```
Nicolas-Dolan, line 54: I could put this in its own component, like tick, if you think that would be valuable.

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -6,15 +6,45 @@
                 <div id="dataset-id" class="mt-4">
                     <span v-translate="'uploadFileDataset'"></span>
                     <span>{{ dataset }}</span></div>
-                <div id="instructions" class="mt-3" v-translate="'uploadFileInstruction'"></div>
```
Nicolas-Dolan, line 9: These instructions were not in the mock so have been deleted.

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -136,16 +188,32 @@
...
+            HelpCircleIcon,
         },
         watch: {
             uploadFiles() {
                 this.setDefaultCheckedItems()
+            },
+            choiceUpload(){
+                if (this.choiceUpload === 'createRelease'){
+                    this.setDefaultCheckedItems()
+                }
```
Nicolas-Dolan, line 211: This makes sure all files are selected when toggling back to createRelease. This means, if you had unselected any files in uploadFiles and then toggled back to createRelease, those selections will be lost.

src/app/static/src/app/store/adrUpload/actions.ts
```diff
@@ -130,5 +132,23 @@ export const actions: ActionTree<ADRUploadState, RootState> & ADRUploadActions =
             }
         }
         await dispatch("getUploadFiles");
+
+        const uploadComplete = rootState.adrUpload.uploadComplete;
+        if (createRelease && uploadComplete){
```
Nicolas-Dolan, line 137: I'm not 100% sure if it is necessary to check if uploadComplete before dispatching createRelease. I wanted to avoid a create release being sent to adr if the proceeding upload had failed. This means the upload error message won't get overwritten with a release error message.

src/app/static/src/scss/style.scss
```diff
@@ -61,8 +61,8 @@ a {
   margin-left: 5px;
   margin-bottom: 5px;
   margin-top: -5px;
-  height: 20px;
-  width: 20px;
+  height: 21px;
+  width: 21px;
```
Nicolas-Dolan, line 65: Made some small tweaks here so that the class could be used for cross as well as the tick. Doesn't seem to have any deleterious effects on the tick.

EmmaLRussell (changes requested):
This is looking good, and creating releases nicely. Haven't gone through all the code, but there's  a few issues with behaviour which will need to be addressed. 

The 'Upload complete' and 'Release created' messages are retained when recalibration happens. More seriously, we don't get a new version when we recalibrate, which means we get a 'Version names must be unique per dataset' error when we recalibrate and try to create another release. We need to address this, as this should be a supported use case. Sorry, that's my fault, I forgot that we didn't make a new version on recalibrate. I think we should discuss how we work around that. 

The upload dialog needs to show all input files as well as the output files, as shown in the mockups. These files should already be uploadable by switching off the `adrPushInputs` feature switch, which can now be removed. The implementation of this needs to be incorporated into the new dialog in the same way as the output files. We no longer need to check whether input files have been changed (so that logic can be removed). This is because the input files may have been modified from elsewhere on ADR since we pulled our versions of them, and we want everything in the release to be consistent. 

We can also remove the "This file already exists on ADR and will be overwritten" messages (and the logic which generates them) because, now that we only have core resources, the expectation is that all uploads with overwrite existing resources (except for the first upload of output files to a dataset, usually). 

Not sure about the message "Version already exists for this activity" - I don't think that will mean much to users since 'Version' isn't a publicly surfaced term in ADR (unlike in HINT), even though that message has come from ADR. How about something like "A release already exists on ADR for the latest files"? I guess that would mean checking for that ADR error message though. 

In this error case, arguably the "Upload is complete" message is misleading too, since we won't actually have uploaded anything new (as hashes are compared before doing an upload) but I think it's probably fine to keep it to emphasise that the files are safely on ADR, it's just the release creation that didn't work.

src/app/static/src/app/store/adrUpload/actions.ts
EmmaLRussell: That makes sense to me

src/app/static/src/app/store/adrUpload/adrUpload.ts
EmmaLRussell: Yes, I prefer `releaseFailed`

mwoodbri (changes requested):
I don't have anything to add right now - I think it's best to focus on addressing input files and release naming before dealing with less important suggestions. Please just re-request review when ready.

src/app/static/src/tests/adrUpload/actions.test.ts
```diff
@@ -230,48 +173,37 @@ describe("ADR upload actions", () => {
...
         expect(commit.mock.calls[1][0].payload).toStrictEqual(expectedUploadFiles);
     };
 
-    it("getUploadFiles includes non-ADR input files when adrPushInputs feature switch is on", async () => {
-        await testGetUploadFilesInputs(true);
-    });
-
-    it("getUploadFiles does not include non-ADR input files when adrPushInputs feature switch is off", async () => {
-        await testGetUploadFilesInputs(false);
+    it("getUploadFiles includes non-ADR input files", async () => {
```
EmmaLRussell, line 202: Misleading test name now - all input files should be included, not just non-ADR ones
```suggestion
    it("getUploadFiles includes input files", async () => {
```

src/app/static/src/tests/adrUpload/actions.test.ts
```diff
@@ -230,48 +173,37 @@ describe("ADR upload actions", () => {
...
+        await testGetUploadFilesInputs();
     });
 
-    it("getUploadFiles can include full set of non_ADR input files when adrPushInputs feature switch is on", async () => {
-        switches.adrPushInputs =  "true";
+    it("getUploadFiles can include full set of non_ADR input files", async () => {
```
EmmaLRussell, line 206: ..and here too. In fact, probably don't need both these tests, since all input files should always be included now.

src/app/static/src/tests/adrUpload/actions.test.ts
```diff
@@ -531,6 +463,66 @@ describe("ADR upload actions", () => {
         expect(mockAxios.history.post[1]["url"]).toBe("/adr/datasets/datasetId/resource/inputs-unaids-naomi-report/1");
     });
 
+    it("uploadFilesToADR dispatches createRelease on upload of final file if instructed to", async () => {
```
EmmaLRussell, line 466: Also need a test that createRelease is not dispatched if the upload failed?

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
```diff
@@ -163,10 +164,11 @@ describe(`uploadModal `, () => {
         return store;
     };
 
-    const getWrapper = () => {
+    const getWrapper = (directives = {}) => {
```
EmmaLRussell, line 167: I'd expect to see tests in this file that the uploadFilesToADR action is dispatched with expected createRelease flag values.

EmmaLRussell (changes requested):
This is looking really good now :+1:.

I'm not completely sure I like clearing the status in `destroyed` but not sure I've got a great argument for why. I think I'd prefer the clear to happen in `beforeMount` - it feels like it's more the responsibility of the newly created component to make sure it's starting with a clear state, rather than relying on what the previous instance did on its way out.

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -249,31 +293,47 @@
...
         components: {
             Modal,
+            HelpCircleIcon,
             DownloadProgress
         },
-      watch: {
-        uploadFiles() {
-          this.setDefaultCheckedItems();
+        watch: {
+            uploadFiles() {
```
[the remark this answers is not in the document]
Nicolas-Dolan: I tried putting `v-if="uploadModalOpen"` on the modal and then calling `this.setDefaultCheckedItems()` on mount and it seems to work properly when I manually tested it, but half of the unit tests will fail for strange reasons, such as not getting the correct translations. Do you have any idea why that might be?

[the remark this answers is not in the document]
EmmaLRussell: Which unit tests were failing, the ones for the UploadModal or for DownloadResults?

src/app/static/src/app/store/adrUpload/mutations.ts
```diff
@@ -34,6 +39,30 @@ export const mutations: MutationTree<ADRUploadState> = {
...
+        state.releaseFailed = false;
+        state.uploadComplete = false;
+    },
+
+    [ADRUploadMutation.ReleaseFailed](state: ADRUploadState, action: PayloadWithType<Error | null>) {
+        state.releaseFailed = true;
+        const alteredPayload = action.payload
+        switch(action.payload?.detail) {
+            case "Version already exists for this activity":
+                alteredPayload!.detail = "A release already exists on ADR for the latest files"
```
mwoodbri, line 57: Can we localise these strings?

mwoodbri (changes requested):
Looking good and working well. Some potential minor tidy-up suggestions inline.

src/app/static/src/app/components/Cross.vue
```diff
@@ -0,0 +1,12 @@
+<template>
+    <svg class="tick">
+        <path d="M 10,10 l 90,90 M 100,10 l -90,90" :stroke="color" stroke-width="15"
+              fill="none"></path>
+    </svg>
+</template>
+
+<script lang="ts">
+    export default {
+        props: ["color", "width"]
```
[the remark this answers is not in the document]
Nicolas-Dolan: There didn't seem to be. I was copying the pattern on tick. Could remove the prop from the tick component too (and all its instances) but that's outside the scope of this ticket.

src/app/static/src/app/components/downloadResults/UploadModal.vue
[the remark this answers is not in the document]
Nicolas-Dolan: I've got all tests passing now. They were originally set up in quite a hacky way because they couldn't rely on the checkboxes being set by the watch method so I think they make more sense now. The translations not coming through was very odd though. In one case, I needed to put an `await Vue.nextTick()` to get them to register on time for the test.

src/app/static/src/app/store/adrUpload/mutations.ts
```diff
@@ -34,6 +40,30 @@ export const mutations: MutationTree<ADRUploadState> = {
...
+        state.releaseFailed = true;
+        const alteredPayload = action.payload
+        switch(action.payload?.detail) {
+            case "Version already exists for this activity":
+                alteredPayload!.detail = translate("releaseExists")
+              break;
+            case "Version names must be unique per dataset":
+                alteredPayload!.detail = translate("releaseNameUnique")
+              break;
+        }
```
mwoodbri, line 63:
```suggestion
        switch (action.payload?.detail) {
            case "Version already exists for this activity":
                alteredPayload!.detail = i18next.t("releaseExists", this.currentLanguage)
                break;
            case "Version names must be unique per dataset":
                alteredPayload!.detail = i18next.t("releaseNameUnique", this.currentLanguage)
                break;
        }
```

src/app/static/src/tests/adrUpload/mutations.test.ts
```diff
@@ -56,4 +58,33 @@ describe("ADR mutations", () => {
...
+        const state = mockADRUploadState();
+        mutations[ADRUploadMutation.ReleaseCreated](state);
+        expect(state.releaseCreated).toBe(true);
+    });
+
+    it("can set release failed with error", () => {
+        const state = mockADRUploadState();
+        mutations[ADRUploadMutation.ReleaseFailed](state, {payload: mockError("Version already exists for this activity")});
+        expect(state.releaseFailed).toBe(true);
+        expect(state.uploadError!!.detail).toBe("A release already exists on ADR for the latest files");
```
Nicolas-Dolan, line 80: I'm not sure how I would test the different translations here.

mwoodbri: I'm not greatly concerned by this, as long as you can modify `downloadResults.test.ts` to check for the translated error detail message (at the moment it only checks for the summary?).

But you should really do a test here - at least for one other language. I'm a bit hazy on this but I think the easiest might be to simply do `i18next.changeLanguage("fr");` before making the mutation.

src/app/static/src/app/store/adrUpload/mutations.ts
```diff
@@ -34,6 +40,30 @@ export const mutations: MutationTree<ADRUploadState> = {
...
+        state.releaseFailed = false;
+        state.uploadComplete = false;
+    },
+
+    [ADRUploadMutation.ReleaseFailed](state: ADRUploadState, action: PayloadWithType<Error | null>) {
+        state.releaseFailed = true;
+        const alteredPayload = action.payload
+        switch (action.payload?.detail) {
+            case "Version already exists for this activity":
+                alteredPayload!.detail = i18next.t("releaseExists", this.currentLanguage)
```
mwoodbri, line 58: Sorry - I misled you here. The following should be sufficient:
```suggestion
                alteredPayload!.detail = i18next.t("releaseExists");
```
What I suggested only works by accident: `this.currentLanguage` is a copy-and-paste that isn't valid (it's not defined here) but works because it's therefore ignored by the `t()` function, which instead uses the currently set language - which is fine. Same for L61.

In fact, I'm not sure why we ever pass a second argument to this function, because we always call `i18next.changeLanguage()` when the language changes - which should be sufficient.

mwoodbri (changes requested):
Looking good. Just a couple of inline comments.

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -6,27 +6,52 @@
...
+                        type="radio"
+                        id="createRelease"
+                        value="createRelease"
+                        v-model="choiceUpload"
+                    />
+                    <label for="createRelease" v-translate="'createRelease'" class="pr-1"></label>
+                    <span class="icon-small" v-tooltip="translate('createReleaseTooltip')">
+                        <help-circle-icon></help-circle-icon>
+                    </span>
+                    <br />
```
mwoodbri, line 20: The wrapping here is a bit odd - especially in non-English language. Radio button, text and help icon all appearing on different lines?
Screenshot from 2021-09-17 09-37-44

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -179,12 +220,12 @@
...
-            lastModified: function (date: string) {
-                return formatDateTime(date)
+            translate(text) {
+                return i18next.t(text, { lng: this.currentLanguage });
             },
             setDefaultCheckedItems: function () {
-                this.uploadFilesToAdr = outputFileTypes
-                    .filter(key => this.uploadFiles.hasOwnProperty(key))
+                this.uploadFilesToAdr = [...outputFileTypes, ...inputFileTypes]
+                    .filter(key => this.uploadableFiles.hasOwnProperty(key))
```
mwoodbri, line 228: I guess time will tell whether selecting all files by default rather than none) is the most helpful behaviour - I don't know whether users would generally want to upload a minority or the majority of the files.

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -233,48 +274,67 @@
...
+                            sections[1][key] = this.uploadableFiles[key];
                         }
                         return sections;
                     }, [{} as any, {} as any]);
                 } else {
                     return [];
                 }
             },
+            uploadDisabled(){
+                return (!this.createRelease && this.uploadFilesToAdr.length < 1) || this.downloadingFiles;
```
mwoodbri, line 302: I wonder whether we should be disabling the upload button if you've selected the "create a release" option, but no files have changed since your last upload. Because in this case it will harmlessly but pointlessly go through the process of "uploading" all the files, but then, more problematically, fail to create a release. We know in advance that this is likely to fail, so allowing the user to do it is suboptimal. But then again I guess the dataset may have been changed by someone else in the meantime - so not quite so clear cut.

src/app/static/src/app/components/downloadResults/UploadModal.vue
Nicolas-Dolan: I guess if it becomes a hassle in future, we could add 'select all/deselect all buttons'.

src/app/static/src/app/components/downloadResults/UploadModal.vue
EmmaLRussell: I think in that case, where a release already exists for the latest point in the activity stream, we should at least make sure we show a comprehensible error message. I agree that the case where upload might succeed makes it not clear that we should disable upload - it's rather unlikely but could happen.

src/app/static/src/tests/components/downloadResults/downloadResults.test.ts
```diff
@@ -36,7 +36,7 @@ describe("Download Results component", () => {
         jest.useRealTimers()
     })
 
-    const createStore = (hasUploadPermission = true, getUserCanUpload = jest.fn(), uploading = false, uploadComplete = false, uploadError: any = null, downloadResults?: Partial<DownloadResultsState>, releaseCreated = false, releaseFailed = false) => {
+    const createStore = (hasUploadPermission = true, getUserCanUpload = jest.fn(), uploading = false, uploadComplete = false, uploadError: any = null, downloadResults?: Partial<DownloadResultsState>, releaseCreated = false, releaseFailed = false, ClearStatus = jest.fn()) => {
```
EmmaLRussell, line 39: Should camel-case `ClearStatus` for consistency here I think, even though it makes setting the value a little more verbose.

EmmaLRussell (commented):
Looking good. One comment from last time unaddressed I think:

> I'd expect to see tests in this file that the uploadFilesToADR action is dispatched with expected createRelease flag values.

Still can't see tests to this effect in `uploadModal.test.ts`?

src/app/static/src/app/components/downloadResults/UploadModal.vue
Nicolas-Dolan: The error message the user would get back in this scenario is quite helpful though. A disabled ok button with no other context might leave the user scratching their head and reporting it as a bug.

mwoodbri: Happy to leave as-is. Just wanted to raise as a potential optimisation - e.g. we could do `getReleases()` when the user starts an upload to potentially avoid pointlessly uploading all the files if nothing has changed since last release. But agree not for this ticket.

src/app/static/src/tests/adrUpload/mutations.test.ts
mwoodbri: This isn't really resolved - we don't have any checks that the translations for "Release names must be unique per dataset..." and "A release already exists..." are actually displayed. As stated above, you could do this this here (by changing the language before invoking the mutation) and/or by checking the rendered messages in `downloadResults.test.ts`.

src/app/static/src/app/components/downloadResults/UploadModal.vue
mwoodbri: Up to you, but I found that the radio buttons rendered better if I tweaked the HTML to be more consistent with the file list (sorry this is raw HTML rather than Vue, but should be easy to translate back):
```
<div class="mt-3 form-check">
  <input type="radio" id="uploadFiles" value="uploadFiles" class="form-check-input">
  <span class="form-check-label">
    <label for="uploadFiles" class="d-inline">Créer une nouvelle version ADR (télécharger tous les fichiers)</label>
    <span class="icon-small d-inline has-tooltip" data-original-title="null">
      <svg xmlns="http://www.w3.org/2000/svg" width="24px" height="24px" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="feather feather-help-circle"><circle cx="12" cy="12" r="10"></circle><path d="M9.09 9a3 3 0 0 1 5.83 1c0 2-3 3-3 3"></path><line x1="12" y1="17" x2="12" y2="17"></line></svg>
    </span>
  </span>
</div>
```
Otherwise the button isn't vertically aligned with first line of text, and the horizontal spacing isn't consistent - see below:
Screenshot from 2021-09-22 17-33-35

mwoodbri (changes requested):
This is looking very good. I think earlier point about testing translations is still partly unresolved. Aside from that just one really minor comment about formatting, that only applies to non-English labels.

EmmaLRussell:
> I'd expect to see tests in this file that the uploadFilesToADR action is dispatched with expected createRelease flag values.

Sorry, I realise this was badly worded now - I really meant checks for all parameter values for this action. I can see a test for the createRelease parameter being true, but not for an upload where createRelease is false, and also don't think there are any tests for correct uploadFiles values?

src/app/static/src/app/components/downloadResults/DownloadResults.vue
```diff
@@ -41,24 +41,33 @@
...
+                    </div>
+                    <div class="d-flex align-items-center height-40">
+                        <component :is="releaseCreated ? 'tick' : 'cross'" color="#e31837"></component>
                     </div>
                 </div>
                 <error-alert v-if="uploadError" :error="uploadError"></error-alert>
             </div>
         </div>
-        <upload-modal id="upload-modal" :open="uploadModalOpen" @close="uploadModalOpen = false"></upload-modal>
+        <upload-modal id="upload-modal" v-if="uploadModalOpen" :open="uploadModalOpen" @close="uploadModalOpen = false"></upload-modal>
```
mwoodbri, line 58: Interested to know why you've added the `v-if` here, in addition to `:open`. Would normally do the latter - having both looks a bit redundant?

Nicolas-Dolan: To avoid having to watch the uploadFile
image

EmmaLRussell: The v-if was to avoid the modal needing to include some tedious watch on values (the upload files themselves I think), which would always be set whenever it was opened, but not if the component was mounted closed immediately. So I guess `:open` could just be `"true"` if we're rendering based on `uploadModalOpen`

src/app/static/src/app/components/downloadResults/UploadModal.vue
Nicolas-Dolan: Personally, I don't see an issue with the current formatting. I think having the radial button vertically centred on the label looks fine and reducing the horizontal space between the two makes it look crowded. I applied the html structure you suggested above and this is what resulted, which I don't really prefer.
image

src/app/static/src/tests/adrUpload/mutations.test.ts
Nicolas-Dolan: I was able to check the French translations but when it came round to the Portuguese, using `i18next.changeLanguage("pt");` still returned English. You mentioned testing one language would be enough though.

mwoodbri: You need to register `pt` [here](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/tests/setup.ts#L23)

src/app/static/src/tests/adrUpload/mutations.test.ts
```diff
@@ -56,4 +59,43 @@ describe("ADR mutations", () => {
...
+        expect(state.uploadError!!.detail).toBe("Release names must be unique per dataset and a release with the same name already exists on ADR. Try renaming the project to generate a new release name.");
+
+        mutations[ADRUploadMutation.ReleaseFailed](state, {payload: mockError("other error")});
+        expect(state.releaseFailed).toBe(true);
+        expect(state.uploadError!!.detail).toBe("other error");
+    });
+
+    it("can set release failed with error in French", () => {
+        const state = mockADRUploadState();
+        i18next.changeLanguage("fr");
```
mwoodbri, line 94: This returns a promise, so really should have an `await`:
```suggestion
        await i18next.changeLanguage("fr");
```

src/app/static/src/tests/adrUpload/mutations.test.ts
```diff
@@ -56,4 +59,43 @@ describe("ADR mutations", () => {
...
+        mutations[ADRUploadMutation.ReleaseFailed](state, {payload: mockError("Version names must be unique per dataset")});
+        expect(state.releaseFailed).toBe(true);
+        expect(state.uploadError!!.detail).toBe("Release names must be unique per dataset and a release with the same name already exists on ADR. Try renaming the project to generate a new release name.");
+
+        mutations[ADRUploadMutation.ReleaseFailed](state, {payload: mockError("other error")});
+        expect(state.releaseFailed).toBe(true);
+        expect(state.uploadError!!.detail).toBe("other error");
+    });
+
+    it("can set release failed with error in French", () => {
```
mwoodbri, line 92:
```suggestion
    it("can set release failed with error in French", async () => {
```

mwoodbri (changes requested):
See inline for suggested fix for Portuguese test

mwoodbri (approved):

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
```diff
@@ -333,7 +331,46 @@ describe(`uploadModal `, () => {
...
+        await okBtn.trigger("click");
+
+        store.state.downloadResults = downloadResults
+        await Vue.nextTick()
+        expect(wrapper.vm.$data.choiceUpload).toBe("uploadFiles")
+        expect(mockUploadFilesToADR.mock.calls.length).toBe(2)
+        const num = mockUploadFilesToADR.mock.calls[0].length -2
+        expect(mockUploadFilesToADR.mock.calls[0][num]["uploadFiles"]).toStrictEqual([fakeMetadata["outputZip"],fakeMetadata["outputSummary"]])
+        expect(mockUploadFilesToADR.mock.calls[0][num]["createRelease"]).toBe(false)
+        expect(mockUploadFilesToADR.mock.calls[1][num]["uploadFiles"]).toStrictEqual([])
```
EmmaLRussell, line 372: I was hoping that this test would be checking that the uploadFiles parameter was set to include all the files the user had left selected. Shouldn't this array be populated? And if it isn't, shouldn't the OK button be disabled?

I'm also confused about why you're expecting multiple calls to `mockUploadFilesToADR' here. Shouldn't a single click of the OK button trigger the action once, after download has completed? (Is it possible that mocks reset between tests is not happening as expected?)

Also, it would be good to include a test that the expected subset of files were sent to the action when the user unselected some before clicking OK. 

I might be making incorrect assumptions here (it's Monday morning so my brain isn't really working yet) - if so, I'd appreciate a few comments in the code to explain what's going on!

EmmaLRussell (commented):
Works great manually. I'm a bit confused about the new test.

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
Nicolas-Dolan: The multiple calls to `mockUploadFilesToADR` actually predates this PR and exists in master. However, it does seem to be an artefact of testing as I see that only one call is made when I `console.log` manually. I'll look into it.

PR #538, opened 2021-08-26
mrc-2543 show selected release name and preselected in modal

Nicolas-Dolan:
## Description

If the user selects a release, the release name should be displayed in addition to the dataset name after selected dataset. 

image

If the user goes back to edit the selected dataset, the modal should have the appropriate combination of datasets, radial selection, releases selected. This should persist if the page is refreshed.

image

Note: Will need a release to be created to manually test. Easiest way to do this is go to the [activity stream](https://dev.adr.fjelltopp.org/inputs-unaids-estimates/activity/burundi-inputs-unaids-estimates-2021) and click add release.

## Type of version change
Minor

## Checklist


src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -6,7 +6,10 @@
                 v-translate="'selectedDataset'"
                 id="selectedDatasetSpan"
             ></span>
-            <a :href="selectedDataset.url" target="_blank">
+            <a v-if="releaseName" :href="selectedDataset.url + '/releases'" target="_blank">
+                {{ selectedDataset.title }} — {{ releaseName }}
```
Nicolas-Dolan, line 10: This won't show if the release has since been deleted from adr. The upside is it will update should the release get renamed.

I'm not sure what the desired behaviour is here. If we wanted to uncouple the user's local release from ADR, we could store the release name in Vuex. On the other hand, maybe it is important that the user's release is synchronised with ADR's.

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -6,7 +6,10 @@
                 v-translate="'selectedDataset'"
                 id="selectedDatasetSpan"
             ></span>
-            <a :href="selectedDataset.url" target="_blank">
+            <a v-if="releaseName" :href="selectedDataset.url + '/releases'" target="_blank">
```
Nicolas-Dolan, line 9: The url targets the releases page for the dataset rather than that specific release. I would need ADR to supply the release's specific url with the release if I were to target it directly as I cannot generate it from the information I have.

mwoodbri: URL is of the form https://dev.adr.fjelltopp.org/dataset/burundi-inputs-unaids-estimates-2021?activity_id=5c579cc1-3a14-4f89-b0a6-8813363940a0 and you should be able to get `activity_id` from the relevant `Release` object (you might just need to add that field to `Release` in `types.ts` in order to make it accessible).

src/app/static/src/app/components/adr/SelectRelease.vue
```diff
@@ -159,10 +173,14 @@
             valid() {
                 this.$emit("valid", this.valid);
             },
+            releases(){
+                this.preSelectRelease();
```
mwoodbri, line 177: You should be able to fix coverage by checking method is called if releases changes

mwoodbri (changes requested):
Some comments inline

May need a broader discussion over whether we think that pre-selecting release is actually valuable enough to the user to justify complexity involved in handling correctly.

EmmaLRussell (commented):
As discussed on Teams, this ticket needs to deal gracefully with selected release being deleted - should not show it in the import dialog if it is reopened, instead should revert to latest data. However, if cancel from dialog at that point, the previous release name should still be visible in the page.

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -232,6 +244,14 @@
                 namespace,
                 (state: ADRState) => state.adrError
             ),
+            releaseName(){
+                const selectedDataset = this.selectedDataset
+                if (selectedDataset?.release && this.releases.length){
+                    return this.releases.find(release => release.id === selectedDataset.release)?.name || null
+                } else {
+                    return null
+                }
```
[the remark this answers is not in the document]
Nicolas-Dolan: The above won't compile in ts because it can return undefined due to the `?.name` and `this.selectedDataset.release` isn't guarded against the object being null.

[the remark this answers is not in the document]
mwoodbri: Sorry - I should have checked before spouting off!

mwoodbri (changes requested):
Working great - clearly a lot of work has gone into this.

Just a couple of minor corrections/suggestions.

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -232,6 +244,16 @@
                 namespace,
                 (state: ADRState) => state.adrError
             ),
+            releaseName(){
+                return this.selectedRelease?.name || null;
+            },
+            releaseURL(){
+                if (this.selectedDataset?.url){
+                    return this.selectedRelease?.activity_id ? this.selectedDataset.url.replace('inputs-unaids-estimates', 'dataset') + '?activity_id=' + this.selectedRelease.activity_id : "";
```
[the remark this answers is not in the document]
EmmaLRussell: Agreed, and I think this would be more readable too

EmmaLRussell (commented):
Nothing much to add to Mark's comments - this works really nicely :+1:

src/app/static/src/app/components/adr/SelectDataset.vue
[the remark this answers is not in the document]
mwoodbri: As discussed, something like this:
```javascript
new URL(this.selectedDataset.url).origin + '/dataset?activity_id=' + this.selectedRelease.activity_id
```

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -232,6 +244,16 @@
                 namespace,
                 (state: ADRState) => state.adrError
             ),
+            releaseName(){
+                return this.selectedRelease?.name || null;
+            },
+            releaseURL(){
+                if (this.selectedDataset?.url){
+                    return this.selectedRelease?.activity_id ? new URL(this.selectedDataset.url).origin + '/dataset/' + this.selectedDataset.title.replace(/\s+/g,"-").toLowerCase() + '?activity_id=' + this.selectedRelease.activity_id : "";
```
mwoodbri, line 252:
```suggestion
                if (this.selectedDataset && this.selectedRelease) {
                    return new URL(this.selectedDataset.url).origin + '/dataset/' + this.selectedDataset.id + '?activity_id=' + this.selectedRelease.activity_id;
```

Nicolas-Dolan: You've put me in a bit of bind for writing code coverage here as there is no (even theoretical) situation where the url is rendered with empty strings and the else side of this statement is reached.

mwoodbri: Surely it can be removed then?

mwoodbri (changes requested):
Code is looking good... but I'm selecting release for Burundi but it's not being displayed on page as expected. Could you check?

Nicolas-Dolan:
> Code is looking good... but I'm selecting release for Burundi but it's not being displayed on page as expected. Could you check?

I've just checked manually and I can't find a problem. Could you give me steps to recreate?

mwoodbri:
> > Code is looking good... but I'm selecting release for Burundi but it's not being displayed on page as expected. Could you check?
> 
> I've just checked manually and I can't find a problem. Could you give me steps to recreate?

Sorry - took me 10 minutes to realise I was staring at mrc-2534 rather than mrc-2543 :facepalm:

mwoodbri (approved):

EmmaLRussell (approved):

EmmaLRussell:
Just need to fix the NEWS conflict

PR #542, opened 2021-09-07
Mrc-2575 Fix erroneous ADR upload file count

carbowizzy:
This PR fixes erroneous ADR upload counts. 

When an ok button gets triggered multiple times, it multiplies the actual selected files before uploading to ADR. This error could also be reproduced as seen in this ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2575.

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/app/components/downloadResults/UploadModal.vue
+++ b/src/app/static/src/app/components/downloadResults/UploadModal.vue
@@ -41,11 +41,12 @@
                     type="button"
                     class="btn btn-red"
                     :disabled="uploadFilesToAdr.length < 1"
-                    @click.prevent="confirmUpload"
+                    @click.prevent="handleConfirmUpload"
                     v-translate="'ok'"></button>
                 <button
                     type="button"
                     class="btn btn-white"
+                    :disabled="downloadingFiles"
                     @click.prevent="handleCancel"
                     v-translate="'cancel'"></button>
             </template>
@@ -86,6 +87,7 @@
         getUploadMetadata: (id: string) => Promise<void>
         handleDownloadResult: (downloadResults: DownloadResultsDependency) => void,
         stopPolling:(id: number) => void
+        handleConfirmUpload: () => void
     }
 
     interface Computed {
@@ -130,6 +132,11 @@
                 "adrUpload",
                 "uploadFilesToADR"
             ),
+            handleConfirmUpload() {
+                if (!this.downloadingFiles) {
+                    this.confirmUpload()
+                }
+            },
             confirmUpload() {
                 this.uploadFilesToAdr.forEach(value => this.uploadFilesPayload.push(this.uploadFiles[value]));
                 const readyForUpload = this.prepareFilesForUpload();
--- a/src/app/static/src/app/components/downloadResults/UploadModal.vue
+++ b/src/app/static/src/app/components/downloadResults/UploadModal.vue
@@ -41,7 +41,7 @@
                     type="button"
                     class="btn btn-red"
                     :disabled="uploadFilesToAdr.length < 1"
-                    @click.prevent="handleConfirmUpload"
+                    @click.prevent="confirmUpload"
                     v-translate="'ok'"></button>
                 <button
                     type="button"
@@ -87,7 +87,6 @@
         getUploadMetadata: (id: string) => Promise<void>
         handleDownloadResult: (downloadResults: DownloadResultsDependency) => void,
         stopPolling:(id: number) => void
-        handleConfirmUpload: () => void
     }
 
     interface Computed {
@@ -132,13 +131,8 @@
                 "adrUpload",
                 "uploadFilesToADR"
             ),
-            handleConfirmUpload() {
-                if (!this.downloadingFiles) {
-                    this.confirmUpload()
-                }
-            },
             confirmUpload() {
-                this.uploadFilesToAdr.forEach(value => this.uploadFilesPayload.push(this.uploadFiles[value]));
+                this.uploadFilesPayload = this.uploadFilesToAdr.map(value => this.uploadFiles[value]);
                 const readyForUpload = this.prepareFilesForUpload();
 
                 if (readyForUpload) {
--- a/src/app/static/src/app/components/downloadResults/UploadModal.vue
+++ b/src/app/static/src/app/components/downloadResults/UploadModal.vue
@@ -40,7 +40,7 @@
                 <button
                     type="button"
                     class="btn btn-red"
-                    :disabled="uploadFilesToAdr.length < 1"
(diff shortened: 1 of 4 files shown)


mwoodbri (changes requested):
Works well, but it would be helpful if the PR description contained a brief summary of the root cause of the file multiplication issue, to make it clearer whether this fix addresses that (rather than just the symptom!).

A couple of other comments inline.

EmmaLRussell (commented):
I think the underlying issue here is that files are being pushed to `uploadFilesPayload` in `confirmUpload` without the payload being cleared on Cancel. 

The cleanest change here would probably just be to regenerate that payload on every `confirmUpload` rather than policing the buttons. I did quickly try the below change, and it seemed to work. I don't think that would have any deleterious effects?

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -46,6 +46,7 @@
                 <button
                     type="button"
                     class="btn btn-white"
+                    :disabled="downloadingFiles"
```
mwoodbri, line 49: There isn't an explanation for this change, so asking here: is this cosmetic, is it required to solve the erroneous upload file count? Either way, if we're making this change then does it make sense to do the same for the "Ok" button? Otherwise I can keep clicking it during upload - not sure if this could cause an actual problem, but is inconsistent, and slightly confusing.

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
```diff
@@ -490,4 +490,32 @@ describe(`uploadModal `, () => {
         expect(labels.at(1).attributes("for")).toBe("id-0-1");
         expect(labels.at(2).attributes("for")).toBe("id-1-0");
     });
+
+    it(`does not send upload files to ADR if files are being prepared for upload`, async () => {
```
mwoodbri, line 494: This looks like it tests the changed button behaviour... but not the change made to address the underlying issue i.e. the fix to `confirmUpload`?

So doesn't appear to offer assurance that that issue is resolved.

mwoodbri (changes requested):
Thanks - looks like a more meaningful patch, but a couple of comments/questions inline to confirm.

I have also observed an issue while testing - if you try to upload summary and zip, and then try to download summary while upload is in progress, then upload fails.

src/app/static/src/app/components/downloadResults/UploadModal.vue
EmmaLRussell: Now that the change has gone in to recreate the payload on each confirmUpload, it should be safe to allow the user to Cancel at any time, shouldn't it? 

I agree that it would make more sense to disable the OK button while Download is happening, if only as further confirmation to user that something is happening.

carbowizzy: Cancelling does not really cancel the download or upload, it only closes the modal. If we truly want a user to cancel then we should have cancel button abort download / upload. 

The above is the reason why cancel button should also be disabled while downloading...

EmmaLRussell: I see, yes, it's just cancelling the modal, so misleading to still have it enabled once the upload has been (irrevocably) started.

mwoodbri (changes requested):
We seem to be missing a test for the change to `confirmUpload()`

If this change is unnecessary then it should be removed. If it is necessary (to prevent `uploadFilePayload` multiplying if `confirmUpload()` is called more than once) then that should be tested.

mwoodbri:
> > We seem to be missing a test for the change to `confirmUpload()`
> > If this change is unnecessary then it should be removed. If it is necessary (to prevent `uploadFilePayload` multiplying if `confirmUpload()` is called more than once) then that should be tested.
> 
> I am not quite sure what you are referring to exactly. I have removed handleConfirmUpload , which was the changes I made and left confirmUpload as it was. ok and cancel buttons are disabled on downloading and they have both been tested. There are no cases to be tested, please point me to your case.

To be clear: there is a code change in this PR that affects `confirmUpload()`. But it doesn't seem to be accompanied by either a change to an existing test, or the addition of a new test. So the code change is untested.

carbowizzy:
> > > We seem to be missing a test for the change to `confirmUpload()`
> > > If this change is unnecessary then it should be removed. If it is necessary (to prevent `uploadFilePayload` multiplying if `confirmUpload()` is called more than once) then that should be tested.
> > 
> > 
> > I am not quite sure what you are referring to exactly. I have removed handleConfirmUpload , which was the changes I made and left confirmUpload as it was. ok and cancel buttons are disabled on downloading and they have both been tested. There are no cases to be tested, please point me to your case.
> 
> To be clear: there is a code change in this PR that affects `confirmUpload()`. But it doesn't seem to be accompanied by either a change to an existing test, or the addition of a new test. So the code change is untested.

I have added a test for that case but I don't think it makes any difference as users will never be able to click cancel or ok button twice as they are both disabled when download is in progress.

mwoodbri:
> > > > We seem to be missing a test for the change to `confirmUpload()`
> > > > If this change is unnecessary then it should be removed. If it is necessary (to prevent `uploadFilePayload` multiplying if `confirmUpload()` is called more than once) then that should be tested.
> > > 
> > > 
> > > I am not quite sure what you are referring to exactly. I have removed handleConfirmUpload , which was the changes I made and left confirmUpload as it was. ok and cancel buttons are disabled on downloading and they have both been tested. There are no cases to be tested, please point me to your case.
> > 
> > 
> > To be clear: there is a code change in this PR that affects `confirmUpload()`. But it doesn't seem to be accompanied by either a change to an existing test, or the addition of a new test. So the code change is untested.
> 
> I have added a test for that case but I don't think it makes any difference as users will never be able to click cancel or ok button twice as they are both disabled when download is in progress.

Thank you! I think it's worth doing because a) it was the original root cause and b) if we ever change the button behaviour then we want to ensure that the underlying issue doesn't re-occur.

mwoodbri (approved):
LGTM - I have verified UI behaviour, and that new test fails with old code.

Have reported issue identified during testing as mrc-2605 as I think it's unrelated.

EmmaLRussell:
Err. OK, I guess it took me since yesterday to get around to looking at this, which is kind of tardy, but in general we shouldn't be merging branches unless we have approvals from all reviewers who have participated in the review. I've added a branch protection rule to HINT to match the one in OrderlyWeb requiring 2 reviewers before merging to master, but this *could* apply to 3 reviewers if 3 people have contributed. There's no particular hurry to merge these tickets right now. If you're impatient waiting for a participating reviewer to respond, feel free to nudge them, rather than merging without their review.

PR #543, opened 2021-09-10
mrc-1435 Keep login page as landing page in HINT

carbowizzy:
This PR redirects users to login page when using Naomi, if a user continues as guest, browser session remembers that the user wants to use app as guest and will not redirect to login page. 

## Type of version change
_Delete as appropriate_

## Checklist


mwoodbri (changes requested):
I haven't had much luck with this... when I click on "Continue as guest" I get redirected back to the login page with the following printed on console:
```
Uncaught SyntaxError: import declarations may only appear at top level of a module
```

Perhaps I'm doing it wrong?

Otherwise you may be able to workaround by declaring script as a module, but if you intend to interact with `localStorage` from FreeMarker page then easiest to do it directly rather than trying to import webpacked code.

carbowizzy:
> I haven't had much luck with this... when I click on "Continue as guest" I get redirected back to the login page with the following printed on console:
> 
> ```
> Uncaught SyntaxError: import declarations may only appear at top level of a module
> ```
> 
> Perhaps I'm doing it wrong?
> 
> Otherwise you may be able to workaround by declaring script as a module, but if you intend to interact with `localStorage` from FreeMarker page then easiest to do it directly rather than trying to import webpacked code.

Oh I was setting item directly and changed implementation to reusing the LocalStorageManager but forgot to test manually. You can review now.

EmmaLRussell (changes requested):
I'd imagined this being implemented from the backend, but this is a neat implementation based in the front end. 

I wonder if we should use sessionStorage rather than localStorage, which never expires. Requirements aren't clear if clicking 'Continue as guest' means user should never land on the login page from now until perpetuity, or if they should for still be directed there next session. 

Should add some Selenium tests to check integration.

carbowizzy:
> I'd imagined this being implemented from the backend, but this is a neat implementation based in the front end.
> 
> I wonder if we should use sessionStorage rather than localStorage, which never expires. Requirements aren't clear if clicking 'Continue as guest' means user should never land on the login page from now until perpetuity, or if they should for still be directed there next session.
> 
> Should add some Selenium tests to check integration.

That's a good point. Session storage or LocalStorage. 

I feel LocalStorage is more appropriate in this context because user can always come back to their project without the session expiring. Session storage may not be ideal but I am open to suggestions.

mwoodbri:
> I feel LocalStorage is more appropriate in this context because user can always come back to their project without the session expiring. Session storage may not be ideal but I am open to suggestions.

I'd simply ask whoever orginated this request (Rob/Jeff/Ian?). I'm not sure of the intention - if it was simply to avoid confusion the very first time a user opens the app and is confronted by lots of file inputs then localStorage makes more sense. But if it's also a means of encouraging people to create an account and/or login then sessionStorage would be more appropriate.

mwoodbri (commented):
This works nicely and I can't see an issue with it being implemented in the front-end.

Besides Selenium tests (which I do think would be of real value here) I have just one potential suggestion. When redirecting we're currently doing more work than we need to in terms of loading the app - the stepper component is partially initialised by the time we redirect, and the user sees a flash of this, including its loading animation.

The alternative is to use a [navigation guard](https://router.vuejs.org/guide/advanced/navigation-guards.html#navigation-guards) in the router itself. This could be done using a global, route or component guard. Example for route guard is adding a `beforeEnter` clause to `index.ts` instead of modifying `Stepper.vue`:
```typescript
    {
        path: "/",
        component: Stepper,
        beforeEnter: (to, from, next) => {
            if (store.state.currentUser === "guest" && !localStorageManager.getContinueAsGuest()) {
                window.location.assign("/login");
            } else {
                next();
            }
        }
    },
```
I think this is logically cleaner and it makes the redirect a bit faster because it short-cuts earlier, but it's not a huge difference, and might be harder to test.

carbowizzy:
> This works nicely and I can't see an issue with it being implemented in the front-end.
> 
> Besides Selenium tests (which I do think would be of real value here) I have just one potential suggestion. When redirecting we're currently doing more work than we need to in terms of loading the app - the stepper component is partially initialised by the time we redirect, and the user sees a flash of this, including its loading animation.
> 
> The alternative is to use a [navigation guard](https://router.vuejs.org/guide/advanced/navigation-guards.html#navigation-guards) in the router itself. This could be done using a global, route or component guard. Example for route guard is adding a `beforeEnter` clause to `index.ts` instead of modifying `Stepper.vue`:
> 
> ```ts
>     {
>         path: "/",
>         component: Stepper,
>         beforeEnter: (to, from, next) => {
>             if (store.state.currentUser === "guest" && !localStorageManager.getContinueAsGuest()) {
>                 window.location.assign("/login");
>             } else {
>                 next();
>             }
>         }
>     },
> ```
> 
> I think this is logically cleaner and it makes the redirect a bit faster because it short-cuts earlier, but it's not a huge difference, and might be harder to test.

This is definitely faster, thanks. Selenium will cover the test.

mwoodbri (changes requested):
Even without Selenium I think we can do better with respect to testing.

As outlined in previous link, the guard can be extracted to a standalone function e.g:
```typescript
export const beforeEnter = (to: Route, from: Route, next: NavigationGuardNext) => {
    if (store.state.currentUser === "guest" && !localStorageManager.getContinueAsGuest()) {
        window.location.assign("/login");
    } else {
        next();
    }
}

router.addRoutes([
    {
        path: "/",
        component: Stepper,
        beforeEnter
    },
    {
        path: "/accessibility",
        component: Accessibility
    },
    {
        path: "/projects",
        component: Projects
    },
]);
```
You can then test this function e.g:
```typescript
    it("doesn't redirect returning guest to login page", () => {
        const realLocation = window.location
        delete window.location;
        window.location = {...realLocation, assign: jest.fn()};

        const next = jest.fn();

        store.state.currentUser = "guest";
        Storage.prototype.getItem = jest.fn((key) => key === "asGuest" ? "continueAsGuest" : null);

        beforeEnter({} as Route, {} as Route, next);

        expect(window.location.assign).not.toHaveBeenCalled();
        expect(next).toBeCalled();

        window.location = realLocation
    });
```
You will need at least two other similar tests: for new guest (being redirected) and for non-guest (not being redirected).

Nicolas-Dolan (commented):
This looks good. Just needs tests for the different scenarios Mark mentioned.

src/app/static/src/tests/localStorageManager.test.ts
```diff
@@ -114,4 +114,12 @@ describe("LocalStorageManager", () => {
...
         expect(spy.mock.calls[0][1]).toBe(JSON.stringify(testState));
     });
+
+    it("can save and get continue as guest from session", () => {
+        sessionStorage.setItem("asGuest", "continueAsGuest")
+        let result = sessionStorage.getItem("asGuest")
+
+        expect(result).not.toBe(null);
+        expect(sessionStorage.getItem("asGuest")).toBe("continueAsGuest");
+    });
```
mwoodbri, line 124: I can't see much value in this test any more, and certainly not in this file - it's just testing basic browser functionality rather than anything specific to `localStorageManager`

src/app/static/templates/login.ftl
```diff
@@ -50,7 +54,7 @@
     </div>
     <div class="text-center mt-3">
         <div class="mb-3">OR</div>
-        <a class="btn btn-red" type="submit" href="/">Continue as guest</a>
+        <a class="btn btn-red" onclick="continueAsGuest()" type="submit" href="/">Continue as guest</a>
```
mwoodbri, line 57: One thing you could do is check that this attribute is present in [LoginTests.kt](https://github.com/mrc-ide/hint/blob/master/src/app/src/test/kotlin/org/imperial/mrc/hint/unit/templates/LoginTests.kt#L23)

mwoodbri (changes requested):
This works very nicely and seems to meet the requirements as expected.

Just a couple of comments about tests.

mwoodbri (approved):
LGTM!

EmmaLRussell (approved):
Great :+1: - nice bit of mocking!

Nicolas-Dolan (approved):

PR #546, opened 2021-09-16
mrc-2613 Remove JCenter from Gradle repository list

mwoodbri:
## Description

- Bump fuel to a version whose dependencies are available in Maven Central. Correspondingly parse hintr response as utf-8 - the default in fuel became ASCII. Ideally hintr would express this the content type it sends, but that's a more invasive change (actually a change to porcelain).
- Remove unnecessary reference to Maven Central
- Use JetBrains repo for old version of kotlinx-html, which is required for version of detekt we're using. Upgrading detekt would be a better option - but this would also require a Gradle update, which again is a bigger change for another ticket.

## Type of version change

None

## Checklist

### DIFF
--- a/src/build.gradle
+++ b/src/build.gradle
@@ -3,7 +3,9 @@ buildscript {
 
     repositories {
         mavenCentral()
-        jcenter()
+        maven { // required for detekt < v1.17.0
+            url "https://maven.pkg.jetbrains.space/public/p/kotlinx-html/maven"
+        }
     }
 
     dependencies {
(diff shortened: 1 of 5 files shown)


EmmaLRussell (approved):
:+1:

src/build.gradle
```diff
@@ -3,7 +3,9 @@ buildscript {
 
     repositories {
         mavenCentral()
-        jcenter()
+        maven { // required for detekt < v1.17.0
+            url "https://maven.pkg.jetbrains.space/public/p/kotlinx-html/maven"
```
carbowizzy, line 7: Doesn't this need a version number ?

mwoodbri: Good question. No, because it's a repository not a package.

And the package version is defined implicitly by dependency tree of detekt (v1.14).

carbowizzy (approved):

PR #547, opened 2021-09-16
Mrc 2584 - Remove genericChart dataset when corresponding input file changes

carbowizzy:
This PR removes genericChart dataset when corresponding input file changes. Code is watching for changes in programme and and data, particularly 'programme.valid' and 'anc.valid' 

Changes occur in all the child elements, so I decided to only watch 'valid' so as to listen to changes once at a time. Once changes detected, clearDataset() empties cached genericChart dataset state.

## Type of version change
_Delete as appropriate_

## Checklist


mwoodbri (changes requested):
Hi @carbowizzy - changes from mrc-1435 seem to have crept into this branch. Could you remove?

Also seems to be a random change to `Choropleth.vue`

EmmaLRussell (changes requested):
The logic for this looks fine, but I don't really think the Baseline component should be coordinating making these state changes - I think it would be better if it was done in the actions - no need for a watch then. I'd be looking at `uploadOrImportANC` and `uploadOrImportProgram` in surveyAndProgram actions. Could invoke the clear in the `then` clause after `commitSelectedDataTypeUpdated`. 

Also, you shouldn't need to clear all datasets, just the relevant one for the file which has changed.

carbowizzy (commented):
> Baseline



> The logic for this looks fine, but I don't really think the Baseline component should be coordinating making these state changes - I think it would be better if it was done in the actions - no need for a watch then. I'd be looking at `uploadOrImportANC` and `uploadOrImportProgram` in surveyAndProgram actions. Could invoke the clear in the `then` clause after `commitSelectedDataTypeUpdated`.
> 
> Also, you shouldn't need to clear all datasets, just the relevant one for the file which has changed.

Thanks Emma, I believe uploadOrImport... actions is another way to implement this logic. However, I think, ideally, we should be implementing it in the component rather than action. secondly, the action is getting cluttered and thus does not support separation of concern. Thirdly, the recommended approach does not account for removal of a dataset.

Also, the behaviour I got from clearing the single dataset does not meet the requirements. 
`<generic-chart v-if="genericChartMetadata"
                           chart-id="input-time-series"
                           chart-height="600px"
                           :metadata="genericChartMetadata"></generic-chart>`

genericChartMetadata does not trigger a reload if only the affected dataset is cleared. Since, I do not see a critical downside to clearing all dataset then I don't think it is necessary to alter the whole structure to make clearing only the re-uploaded/removed dataset tigger the required behaviour.

EmmaLRussell:
> Thanks Emma, I believe uploadOrImport... actions is another way to implement this logic. However, I think, ideally, we should be implementing it in the component rather than action. secondly, the action is getting cluttered and thus does not support separation of concern.

I think putting the logic in the component breaks separation of concerns more than putting it in the action does. Components are for user interaction, and accessing the store for whatever it needs in order to achieve that. I don't think components should be implementing logic related to how actions and mutations should be coordinated.

Logically, that upload action could be invoked from some other component if the interface changed - we would still want the clearing of the generic chart state to occur in that case.. so the concern does not really belong with the component. 

I agree that actions become more complicated/cluttered when they spawn other actions/mutations, but equally components become cluttered when they include logic which is really nothing to do with their user interface or their own store interaction. 

We've used the model of actions spawning whatever changes they imply elsewhere, and I'd prefer to stick with that where possible. I know we also have sequences of actions managed by other components eg for upload to ADR, but those are generally also involving user feedback within the sequence so make more sense for the components to handle than this case. 

Another approach would be to use a plugin and `store.subscribe`, like we do for `persistState` to spot when the mutation to the input files are occurring, and reset genericChart dataset state when they do. That would save separation of concerns and would avoid the action becoming more cluttered, but would be rather inconsistent with our usual approach to the consequences of single actions, and would be somewhat more opaque than putting that code in the action itself.

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -195,7 +195,7 @@
                         if (!this.initialised) {
                             return {max: 1, min: 0}
                         }
-                        return {max: this.colorIndicator.max, min: this.colorIndicator.min}
+                        return {max: this.colorIndicator?.max, min: this.colorIndicator?.min}
```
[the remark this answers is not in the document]
mwoodbri: > This seems irrelevant to this ticket but a quick fix for max error showing on initiation of Choropleth.

Not entirely comfortable with this - would be better in a new ticket, especially as it would likely benefit from a regression test and some guidance on manual testing. Especially as it can now return `{max: null, min: null}`, which looks a bit dubious...

mwoodbri:
> I think putting the logic in the component breaks separation of concerns more than putting it in the action does. Components are for user interaction, and accessing the store for whatever it needs in order to achieve that. I don't think components should be implementing logic related to how actions and mutations should be coordinated.

I concur

EmmaLRussell:
<generic-chart v-if="genericChartMetadata" chart-id="input-time-series" chart-height="600px" :metadata="genericChartMetadata"></generic-chart>

> genericChartMetadata does not trigger a reload if only the affected dataset is cleared. Since I do not see a critical downside to clearing all dataset then I don't think it is necessary to alter the whole structure to make clearing only the re-uploaded/removed dataset tigger the required behaviour.

Sorry, should have responded to this yesterday. It's not a critical downside as such, but it's reloading data that you don't need to, so it impacts performance. 

I don't really understand the relevance of reloading the generic chart component. Clearing the ANC or ART data will happen in step 1, where the generic chart component is not loaded  anyway, so it will always be reloaded after this change.

It should just be a case of passing a dataset id to the clear dataset mutation and selectively removing that shouldn't it?

EmmaLRussell (commented):
Thank you, that makes much more sense to me, and works well. I still think we don't need to clear all datasets when change a single file though!

Difficult to test for the delete file case, as we get an error - but that should be addressed by https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2588

mwoodbri (changes requested):
This looks much cleaner :+1: 

There seems to a failing test though, and I've added another comment about testing (see inline)

src/app/static/src/app/store/surveyAndProgram/actions.ts
```diff
@@ -151,6 +164,7 @@ export const actions: ActionTree<SurveyAndProgramState, RootState> & SurveyAndPr
             .delete("/disease/anc/")
             .then(() => {
                 commit({type: SurveyAndProgramMutation.ANCUpdated, payload: null});
+                commitClearGenericChartDataset(commit, DATASET_TYPE.ART)
```
mwoodbri, line 167: This doesn't look right... you'll need a test that the payloads are correct here. Please see my [previous comment](https://github.com/mrc-ide/hint/pull/547#discussion_r713972333).

mwoodbri (changes requested):
Still need a test for payload of new mutation...

src/app/static/src/app/store/genericChart/mutations.ts
```diff
@@ -21,8 +21,8 @@ export const mutations: MutationTree<GenericChartState> = {
     [GenericChartMutation.SetDataset](state: GenericChartState, action: PayloadWithType<SetDatasetPayload>) {
         state.datasets[action.payload.datasetId] = action.payload.dataset;
     },
-    [GenericChartMutation.ClearDataset](state: GenericChartState) {
-        state.datasets = {};
+    [GenericChartMutation.ClearDataset](state: GenericChartState, action: PayloadWithType<string>) {
+        state.datasets[action.payload] = {} as GenericChartDataset;
```
EmmaLRussell, line 25: It would be better to remove this from the dictionary entirely rather than replacing it with {}, which will fail tests for falsiness - this will restore it to the state it was in before the dataset was fetched. 
```suggestion
        delete state.datasets[action.payload];
```

EmmaLRussell (changes requested):
Still some work to do on the tests and data type bug, as described by Mark, and suggested tweak to the clear mutation.

carbowizzy:
> Still some work to do on the tests and data type bug, as described by Mark, and suggested tweak to the clear mutation.



> Still some work to do on the tests and data type bug, as described by Mark, and suggested tweak to the clear mutation.

The error in console is really not a bug as mrc-2588 is address the state when data is cleared. I have used a delete keyword.

EmmaLRussell:
> > Still some work to do on the tests and data type bug, as described by Mark, and suggested tweak to the clear mutation.
> 
> The error in console is really not a bug as [mrc-2588](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2588) is address the state when data is cleared. I have used a delete keyword.

The bug I was referring to is here: https://github.com/mrc-ide/hint/pull/547/files/dc9f7de49ac14d4b474914264664639bbce373e7#diff-8bc6152ff0e3d1b893445cbfdefc5fa8e610b8f933339d3e04ec24e11eeaed3aR167

As referred to by Mark's comment:
> This doesn't look right... you'll need a test that the payloads are correct here. Please see my previous comment. 

It's committing a clear to the ART dataset when delete ANC file

EmmaLRussell (approved):
LGTM :+1:

mwoodbri (approved):

PR #548, opened 2021-09-16
mrc-2583 Add tooltips for indicators in Time series chart

mwoodbri:
## Description

To test:
- Go to _Review inputs_ > _Time series_ chart
- Check that help icon/tooltip appears next to _Plot type_ header
- Observe also that entries in _Plot type_ dropdown are human-readable i.e. "ART count" rather than "art_count"
  - The hintr PR that this PR depends on (mrc-ide/hintr#295) provides not only descriptions (as displayed in the tooltip) but also labels, so although the filter component was previously capable of displaying these, this is the first time they will become visible

## Type of version change

Minor

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.54.0
+
+* Add tooltips for indicators in Time series chart
+
 # hint 1.53.0
 
 * Remove genericChart dataset when corresponding input file changes
--- a/src/app/static/src/app/components/plots/FilterSelect.vue
+++ b/src/app/static/src/app/components/plots/FilterSelect.vue
@@ -1,6 +1,9 @@
 <template>
     <div>
-        <label :class="'font-weight-bold' + (disabled ? ' disabled-label' : '')" v-translate="label"></label>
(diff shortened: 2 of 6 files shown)


EmmaLRussell (approved):
Nice. I was envisaging this being part of Generic Chart Configuration potentially, but this is much cleaner, keeps the descriptions with the options they refer to, and gives us the ability to include descriptions for any filter options. I suppose we might one day want descriptions on a whole filter, not just the options, but let's cross that use case when we come to it..

carbowizzy (approved):
LGTM

Nicolas-Dolan (commented):
Looks good. I noticed that the tooltips were not updating when I changed language but I suspect that is due to a wider issue with filters not being updated when the language changes and not something to worry about here.

mwoodbri:
> Looks good. I noticed that the tooltips were not updating when I changed language but I suspect that is due to a wider issue with filters not being updated when the language changes and not something to worry about here.

mrc-2620, unless I'm mistaken

PR #549, opened 2021-09-17
mrc-2617 Fix warnings emitted during tests

mwoodbri:
## Description

Two warnings were being emitted during run of full test suite, both due to missing actions:
- First was inconsequential, but now fixed (`selectDataset.test.ts`)
- Second appears slightly more serious - missing action seems to mask two test failures. The first of these was (I think) due to test being out of date in relation to merger of steps 1 and 2 of stepper. Second seemed to be flat-out wrong: there's no way to get to step 7 without mocking a lot more of the wizard state. Both now fixed.

## Type of version change

None

## Checklist


EmmaLRussell (approved):
LGTM

Nicolas-Dolan (approved):

mwoodbri:
@EmmaLRussell Since I opened this PR a new warning has appeared related to generic chart. I've mocked it in similar way to other actions within same test, but would be great if you just check it's a sensible approach.

EmmaLRussell (approved):
LGTM

PR #550, opened 2021-09-17
Mrc 2588 Show Time series tab and data source picker only when relevant

carbowizzy:
This PR hides Input time series tab if neither ART nor ANC files have been uploaded.  Does not show data source picker but displays chart component if only one of ART or ANC have been uploaded. The file uploaded is automatically selected.

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/app/components/genericChart/GenericChart.vue
+++ b/src/app/static/src/app/components/genericChart/GenericChart.vue
@@ -128,7 +128,7 @@
                 .reduce((running: Record<string, DataSourceSelections>, dataSource: DataSourceConfig) => ({
                     ...running,
                     [dataSource.id]: {
-                        datasetId: this.defaultDataSource.datasetId,
+                        datasetId: this.defaultDataSource.datasetId || dataSource.datasetId,
                         selectedFilterOptions: null
                     }
                 }), {});
--- a/src/app/static/src/app/components/genericChart/GenericChart.vue
+++ b/src/app/static/src/app/components/genericChart/GenericChart.vue
@@ -3,7 +3,7 @@
         <div class="row" v-if="chartData">
             <div class="col-3">
                 <div v-for="ds in chartConfigValues.dataSourceConfigValues" :key="ds.config.id">
-                    <data-source v-if="ds.editable && defaultDataSource.showDataPicker"
+                    <data-source v-if="ds.editable && availableDatasetIds.length > 1"
                                  :config="ds.config"
                                  :datasets="chartMetadata.datasets"
                                  :value="ds.selections.datasetId"
@@ -45,7 +45,7 @@
         Dict, DisplayFilter,
         GenericChartDataset,
         GenericChartMetadata,
-        GenericChartMetadataResponse, DefaultDataSource
+        GenericChartMetadataResponse
     } from "../../types";
     import DataSource from "./dataSelectors/DataSource.vue";
     import Filters from "../plots/Filters.vue";
@@ -86,7 +86,7 @@
         metadata: GenericChartMetadataResponse
         chartId: string
         chartHeight: string
-        defaultDataSource: DefaultDataSource
+        availableDatasetIds: string[]
     }
 
     interface Computed {
@@ -113,7 +113,7 @@
             metadata: Object,
             chartId: String,
             chartHeight: String,
-            defaultDataSource: Object
+            availableDatasetIds: Array
         },
         components: {
             DataSource,
@@ -128,7 +128,7 @@
                 .reduce((running: Record<string, DataSourceSelections>, dataSource: DataSourceConfig) => ({
                     ...running,
                     [dataSource.id]: {
-                        datasetId: this.defaultDataSource.datasetId || dataSource.datasetId,
+                        datasetId: this.availableDatasetIds.find(id => id === dataSource.datasetId) || this.availableDatasetIds[0],
                         selectedFilterOptions: null
                     }
                 }), {});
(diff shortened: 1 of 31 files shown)


src/app/static/src/app/components/genericChart/GenericChart.vue
```diff
@@ -57,6 +57,7 @@
     import {FilterOption} from "../../generated";
     import Plotly from "./Plotly.vue";
     import {filterData} from "./utils";
+    import {config} from "@vue/test-utils";
```
mwoodbri, line 60: Looks like a rogue import

src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
```diff
@@ -190,6 +191,15 @@
                     options.push({id: "0", label: i18next.t("ANC", lang)});
                 }
                 return options;
+            },
+            defaultDataSource() {
+                const selectDefaultDataset = !this.programme.available && this.anc.available ? "anc" : null
+                const showDataSourcePicker = !!this.anc.available && !!this.programme.available
```
mwoodbri, line 197:
```suggestion
                const showDataSourcePicker = this.anc.available && this.programme.available
```
These are sure to be booleans, so removing `!!` is safe and improves readability

src/app/static/src/app/types.ts
```diff
@@ -226,3 +226,8 @@ export interface GenericChartDataset {
         }
     }
 }
+
+export interface DefaultDataSource {
+    datasetId: string | null,
+    showDataPicker: boolean
+}
```
mwoodbri, line 233: Two observations on this:
- `DefaultDataSource` isn't a great name - this isn't a data source (or directly related, really)
- Not sure of value in bundling these two props into a new type - the two fields aren't really attributes of a common entity. Could they just be two individual props?

mwoodbri (changes requested):
A couple of observations inline. Plus one possibly-naive question: can't the dataset picker just hide itself if there is only one dataset available?

src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
carbowizzy: I know right, sometimes returns undefined if nothing is selected, I just want boolean.

carbowizzy:
> Works very nicely. A couple of observations inline. Plus one possibly-naive question: can't the dataset picker just hide itself if there is only one dataset available? That would avoid having to hard-code dataset IDs. If not then, as the ticket suggests, maybe this should be part of metadata.

Well, I contemplated adding it to the component metadata but I decided against it as the data needed to decide if we should be displaying source picker or not is user's uploaded data which means adding metadata at runtime to soon to be hintr generated chart metadata will be disruptive. I can rename and separate the props, I initially thought it would be nice to have few props so I put them in an object.

EmmaLRussell:
> > Works very nicely. A couple of observations inline. Plus one possibly-naive question: can't the dataset picker just hide itself if there is only one dataset available? That would avoid having to hard-code dataset IDs. If not then, as the ticket suggests, maybe this should be part of metadata.
> 
> Well, I contemplated adding it to the component metadata but I decided against it as the data needed to decide if we should be displaying source picker or not is user's uploaded data which means modifying metadata at runtime to soon to be hintr generated metadata may be disruptive. I can rename and separate the props, I initially thought it would be nice to have few props so I put them in an object. It would be nice to hear what @EmmaLRussell thinks.

I haven't looked through all the code yet, but in principle I agree with Mark that it's good if the components can infer as much as possible without it having to be configured. We definitely don't want to be changing metadata at runtime (if anything it would need to go into 'selections' or similar) - but if the DataSource component, or its GenericChart component parent, could spot that there is only one available dataset and hide the drop-down in that case, that seems like the cleanest approach. 

I suppose in general we could consider replacing the drop-down with a read-only label to make it clear which dataset is being displayed. However, I think it's going to be obvious to the user for this use case which dataset they're looking at, since all the plot types are called 'ART something' or 'ANC something'.

EmmaLRussell (commented):
This works fine, but I don't think the 'defaultDataSource' prop is really ideal. I think I see why you've done it - it's because the generic chart datasets are separate from the usual anc and art datasets, and the call to fetch those will fail if the standard anc/art have not been uploaded. 

Could we instead, pass 'availableDatasetIds' to the component as a prop, based on which of 'anc' and 'art' are available? This would be an array of dataset ids. if the configured datasetId for the data source is not in this array, it should use the first one that is. This is quite similar to the 'defaultDataSource' prop id, but I think it has the advantage of leaving more up to the `GenericChart` component in that it should be deciding whether or not to show the data source picker (decided by number of ids in `availableDatasetIds` - don't show if only one), and which dataset id to use as default if all are available (based on metadata config). If 'availableDatasetIds' provided is null, component should assume all will be available.

EmmaLRussell (commented):
This is looking really good now, thanks. I think a couple of tests are now a little confusingly named, but otherwise works great.

mwoodbri (commented):
Very nice. Just one potential improvement inline.

src/app/static/src/app/components/surveyAndProgram/SurveyAndProgram.vue
```diff
@@ -8,7 +8,7 @@
                        v-translate="'map'"
                        v-on:click="selectTab(0)"></a>
                 </li>
-                <li class="nav-item">
+                <li class="nav-item" v-if="anc.available || programme.available">
```
mwoodbri, line 11: Thought I'd added a comment here but can't see it now: was just going to suggest that you could do the following as a way to avoid doing the "anc or art" check in two places? But completely up to you.
```suggestion
                <li class="nav-item" v-if="availableDatasetIds.length">
```

EmmaLRussell (approved):

Nicolas-Dolan (approved):

mwoodbri (approved):

PR #551, opened 2021-09-20
mrc-2587 Switch Naomi to pull and push to ADR 2022 package

mwoodbri:
## Description

- Switch schema from `inputs-unaids-estimates` to `country-estimates-22`
  - UNAIDS have confirmed that this is ready to go live. Though at present there are no genuine [country datasets for 2022](https://adr.unaids.org/country-estimates-22) - meaning HINT end-users won't be able to import anything from ADR, including previously imported datasets.
- Switch tests from `antarctica-inputs-unaids-estimates-2021` to newly-created `antarctica-country-estimates-2022-1`
  - Instructions for creating this dataset have been added to [KB article](https://mrc-ide.myjetbrains.com/youtrack/articles/mrc-A-41/AIDS-Data-Repository-(ADR)#testing), which has also been updated to expand on the purpose each of our organisations within ADR
  - According to previous instructions from ADR team this new dataset is `private`. For it to be visible in testing (both manual and automated) I have changed the ADR `package_search` query to include private datasets. My suspicion is that private datasets are only used by us (as far as I can see they can only be created via the API), so this won't affect what users generally in the dataset list in HINT. But even if private datasets are used more widely, I think you would expect to see them in your list - so that you can import them.
  - We currently don't have any other datasets for new schema (e.g. Burundi, Uganda etc) but private copies of these can easily created by following the instructions, adding resources from 2022 datasets when available (or by using 2021 resources)

To-do:
- [x] **Get confirmation from UNAIDS before merging/deploying**

## Type of version change

Minor

## Checklist


src/app/static/src/tests/integration/adr-dataset.itest.ts
```diff
@@ -98,7 +98,7 @@ describe("ADR dataset-related actions", () => {
         expect(commit.mock.calls[2][0].type).toStrictEqual(ADRMutation.SetDatasets);
         const datasets = commit.mock.calls[2][0]["payload"];
         const dataset = datasets.find((dataset: any) => dataset.organization.name === "naomi-development-team");
-        expect(dataset).not.toBeNull()
+        expect(dataset).not.toBeUndefined();
```
mwoodbri, line 101: This assertion was previously incorrect - `find()` returned `undefined` if not found, not `null`

NEWS.md
```diff
@@ -1,10 +1,14 @@
-# hint 1.49.1
+# hint 1.51.0
```
EmmaLRussell, line 1: You've designated this a Major change in the PR - which I guess makes sense, as it will very much not be backwards compatible. So should this be `2.0.0`?

EmmaLRussell (commented):
Works for me! I'm seeing two 2022 Antarctica datasets, one created by Fjelltopp, the other by Naomi Dev Team. I take it that is as expected?

NEWS.md
EmmaLRussell: ..at leas not backwards compatible for people hoping to reimport older datasets

mwoodbri: My mistake. It should be minor. I'm don't think it's backwards incompatible (or that semantic versioning really makes sense for end-user software!).

mwoodbri:
> Works for me! I'm seeing two 2022 Antarctica datasets, one created by Fjelltopp, the other by Naomi Dev Team. I take it that is as expected?

Yes. I'm expecting the Fjelltopp one to either be made private or be deleted. I will ask, but expected behaviour given that it's currently a public 2022 dataset.

EmmaLRussell (approved):

NEWS.md
mwoodbri: > ..at leas not backwards compatible for people hoping to reimport older datasets

Quite. But that's implicit in this change being more broadly approved by the relevant parties. To be discussed on Friday.

carbowizzy (approved):
Only Antarctica dataset for 2022...

Nicolas-Dolan (approved):

PR #553, opened 2021-09-21
Mrc-2606 Re-fetch all metadata when language changes

carbowizzy:
This PR fixes wrong language for prevalence indicators on choropleth, bubble map, etc. PR adds a method to refetch plotting metadata upon switching languages.

Problem : 
* when a project is imported, it uses the current language to fetch plotting metadata and
* if a user changes the current language after importing a project, plotting metadata does not update prevalence indicator menu items language, leaving users with an incorrect language. 

Solution:
* If a user changes the current language, a request should be sent to fetch plotting metadata using the newly selected language. 

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/app/store/root/actions.ts
+++ b/src/app/static/src/app/store/root/actions.ts
@@ -51,6 +51,8 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
     },
 
     async changeLanguage(context, payload) {
+        const {dispatch} = context
         await changeLanguage<RootState>(context, payload)
+        await dispatch("metadata/getPlottingMetadata", payload)
     },
 };
--- a/src/app/static/src/app/store/root/actions.ts
+++ b/src/app/static/src/app/store/root/actions.ts
@@ -5,7 +5,6 @@ import {RootMutation} from "./mutations";
 import {LanguageActions} from "../language/language";
 import {changeLanguage} from "../language/actions";
 import i18next from "i18next";
-import {ISO3} from "../translations/locales";
 
 export interface RootActions extends LanguageActions<RootState>{
     validate: (store: ActionContext<RootState, RootState>) => void;
@@ -52,8 +51,8 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
     },
 
     async changeLanguage(context, payload) {
-        const {dispatch} = context
+        const {dispatch, rootState} = context
         await changeLanguage<RootState>(context, payload)
-        await dispatch("metadata/getPlottingMetadata", ISO3[payload])
+        await dispatch("metadata/getPlottingMetadata", rootState.baseline.iso3)
     }
 };
--- a/src/app/static/src/app/store/root/actions.ts
+++ b/src/app/static/src/app/store/root/actions.ts
@@ -53,6 +53,9 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
     async changeLanguage(context, payload) {
         const {dispatch, rootState} = context
         await changeLanguage<RootState>(context, payload)
-        await dispatch("metadata/getPlottingMetadata", rootState.baseline.iso3)
+
+        if (rootState.baseline?.iso3) {
+            await dispatch("metadata/getPlottingMetadata", rootState.baseline.iso3)
+        }
     }
 };
--- a/src/app/static/src/app/store/root/actions.ts
+++ b/src/app/static/src/app/store/root/actions.ts
@@ -5,6 +5,7 @@ import {RootMutation} from "./mutations";
 import {LanguageActions} from "../language/language";
 import {changeLanguage} from "../language/actions";
 import i18next from "i18next";
+import {ISO3} from "../translations/locales";
 
 export interface RootActions extends LanguageActions<RootState>{
     validate: (store: ActionContext<RootState, RootState>) => void;
@@ -53,6 +54,6 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
     async changeLanguage(context, payload) {
         const {dispatch} = context
         await changeLanguage<RootState>(context, payload)
-        await dispatch("metadata/getPlottingMetadata", payload)
-    },
+        await dispatch("metadata/getPlottingMetadata", ISO3[payload])
+    }
 };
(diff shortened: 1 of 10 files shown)


mwoodbri (changes requested):
Some guidance on manual testing would help, but two observations nevertheless:
- Switching language invokes request to `http://localhost:8080/meta/plotting/undefined`. Is that correct? It looks like the action is [invoked with a country code](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/baseline/actions.ts#L50) elsewhere.
- I don't think a UI component for switching language should have any knowledge of plotting. It would make rather more sense for the `getPlottingMetadata` action to be invoked from [this one](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/root/actions.ts#L53) (unless there's an even more appropriate place).

carbowizzy:
> Some guidance on manual testing would help, but two observations nevertheless:
> 
> * Switching language invokes request to `http://localhost:8080/meta/plotting/undefined`. Is that correct? It looks like the action is [invoked with a country code](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/baseline/actions.ts#L50) elsewhere.
> * I don't think a UI component for switching language should have any knowledge of plotting. It would make rather more sense for the `getPlottingMetadata` action to be invoked from [this one](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/root/actions.ts#L53) (unless there's an even more appropriate place).

Thanks... 

refetching metadata from root/action instead of language/action. LanguageMenu component is using root namespace so I believe it's the right place to be.

src/app/static/src/app/store/root/actions.ts
```diff
@@ -51,6 +51,8 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
     },
 
     async changeLanguage(context, payload) {
+        const {dispatch} = context
         await changeLanguage<RootState>(context, payload)
+        await dispatch("metadata/getPlottingMetadata", payload)
```
EmmaLRussell, line 56: I think this is where the problem with the metadata url is coming from - the payload for this action should be country code, not language. I'm surprised this actually works! Must be getting default metadata from hintr.

EmmaLRussell (changes requested):
This works nicely for updating input metadata, and output bubble plot, but for some reason, when I change from fr to en, I still see French values in output choropleth and barchart filters.
Screenshot from 2021-09-22 10-34-17

There is definitely something weird going on with the re-fetch of `meta/plotting` - I'm seeing the paths `meta/plotting/en` and `meta/plotting/fr` so it seems we're refetching in the url path instead of just in the header (would expect the country code to be final part of the url).

EmmaLRussell:
This is also going to be an issue for input time series chart - I've created a separate ticket for that: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2620

Nicolas-Dolan (commented):
I don't know if you were ready for review at this stage, but there still seem to be issues with the translations on the calibrate page and the filters on the review output.

carbowizzy:
> I don't know if you were ready for review at this stage, but there still seem to be issues with the translations on the calibrate page and the filters on the review output.

This ticket is only meant to refetch metadata for indicators. I think you should create a ticket for the calibrate page issue.

mwoodbri (changes requested):
Please see comment inline.

src/app/static/src/tests/root/actions.test.ts
```diff
@@ -182,13 +182,31 @@ describe("root actions", () => {
         expect(mockContext.dispatch).not.toHaveBeenCalled();
     });
 
-    it("changes language", async () => {
+    it("changes language and fetch plotting metadata", async () => {
```
mwoodbri, line 185: I'd be inclined to keep checks for `ChangeLanguage` and `getPlottingMetadata` mutations separate, but not overly concerned.

mwoodbri: Very subjective, but in my mind there are three things to verify:
- Language mutation is fired if language is changed
- Plotting metadata is re-fetched if language changed and data has been loaded
- Plotting metadata is not re-fetched if language changed and not data has been loaded
- 
But this is really subjective, and no problem with your approach if everyone else is happy!

mwoodbri (approved):
Works nicely

EmmaLRussell:
> This ticket is only meant to refetch metadata for indicators. I think you should create a ticket for the calibrate page issue.

Actually, the description for this ticket does specify that calibrate metadata is included: "This applies to input metadata, modelCalibrate metadata and (soon) input time series metadata." 'Metadata for indicators' does not just mean input indicators - indicators are used in output metadata too. 

I don't have a problem with implementing this in a separate branch/ticket (there is now a separate ticket for the input time series change, as linked to above), but that should be accounted for in this PR, with a link to the new ticket.

src/app/static/src/tests/root/actions.test.ts
EmmaLRussell: We haven't historically been very prescriptive about each test only testing one thing, and often test multiple aspects of a feature in a single test. I think single-check tests are usually cleaner and more readable, but can proliferate, and it's often easier to do multiple checks together particularly if test set up is verbose (though obviously that could usually be factored out into a common method). 

I think multiple-check tests are generally ok, if not ideal, as long as the test title makes clear all the things which are going to be tested.

carbowizzy:
> > This ticket is only meant to refetch metadata for indicators. I think you should create a ticket for the calibrate page issue.
> 
> Actually, the description for this ticket does specify that calibrate metadata is included: "This applies to input metadata, modelCalibrate metadata and (soon) input time series metadata." 'Metadata for indicators' does not just mean input indicators - indicators are used in output metadata too.
> 
> I don't have a problem with implementing this in a separate branch/ticket (there is now a separate ticket for the input time series change, as linked to above), but that should be accounted for in this PR, with a link to the new ticket.

Thank you Emma, I have now update this ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2622 to account for calibrate as well.

EmmaLRussell:
> Thank you Emma, I have now update this ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2622 to account for calibrate as well.

Thanks Lekan - in light of our chat with Rob, I have updated that ticket to make it clear that it applies to barchart indicators for both the main output plots, and the calibrate plot.

EmmaLRussell (approved):

PR #555, opened 2021-09-22
mrc-2600 Allow existing release name to be replaced with new release

Nicolas-Dolan:
## Description

## Type of version change
_Delete as appropriate_

## Checklist


src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -186,7 +186,26 @@ class ADRController(private val encryption: Encryption,
...
+            }
+        } else {
+            error = releasesResponse
+        }
+        if (error != null){
+            return error
+        } else {
+            return adr.post("/dataset_version_create", listOf("dataset_id" to id, "name" to name));
+        }
     }
```
Nicolas-Dolan, line 209: @EmmaLRussell is this the kind of error handling that is expected? For example, is it right that if we get an error trying to get the existing releases or deleting a release, should we arrest the attempt to create a release and return the offending error? Or might that be confusing for the user if they get a failed to delete release error message after trying to create a release? What do you think the appropriate behaviour should be?

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ADRControllerTests.kt
```diff
@@ -752,11 +752,78 @@ class ADRControllerTests : HintrControllerTests()
...
+                mock(),
+                mockBuilder,
+                objectMapper,
+                mockProperties,
+                mock(),
+                mock(),
+                mockSession,
+                mock())
+        val result = sut.createRelease("dataset-1", "release-1")
+        assertThat(result.body!!).isEqualTo("whatever")
```
mwoodbri, line 813: Both endpoints are returning `"whatever"` so looks to me as if this test isn't really checking that `/version_delete` is being invoked. You can either change one of responses to a different string, or check that the mock function is actually called.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -186,7 +186,29 @@ class ADRController(private val encryption: Encryption,
...
+            }
+        } else {
+            error = releasesResponse
+        }
+        // returns error if either get releases or delete release failed; otherwise, attempts to create a release
+        if (error != null){
+            return error
+        } else {
+            return adr.post("/dataset_version_create", listOf("dataset_id" to id, "name" to name));
+        }
```
mwoodbri, line 211: Finding the error handling a bit convoluted here... something like this is easier to read?
```suggestion
        // checks for existing releases on ADR with the same name as the release being created
        val releasesResponse = adr.get("/dataset_version_list?dataset_id=${id}")
        if (releasesResponse.statusCode != HttpStatus.OK) 
        {
            return releasesResponse
        }
        val releases = objectMapper.readTree(releasesResponse.body!!)["data"]
        val duplicateRelease = releases.find { it["name"]?.asText() == name }
        if (duplicateRelease != null)
        {
            val duplicateReleaseId = duplicateRelease["id"].asText()
            // if a release of the same name exists on ADR, request that it is deleted
            val deleteResponse = adr.post("/version_delete", listOf("version_id" to duplicateReleaseId))
            if (deleteResponse.statusCode != HttpStatus.OK)
            {
                return deleteResponse
            }
        }
        return adr.post("/dataset_version_create", listOf("dataset_id" to id, "name" to name))
```

mwoodbri (changes requested):
A couple of suggestions inline.

Any reason why we don't have an integration test for this? Should be able to test that creating two releases with the same name in Antarctica is successful. Probably including timestamp in version name to avoid any clashes.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/adr/ADRTests.kt
```diff
@@ -295,6 +295,38 @@ class ADRTests : SecureIntegrationTests()
...
+            val data = ObjectMapper().readTree(result.body!!)["data"]
+            assertThat(data["name"].textValue()).isEqualTo(releaseName)
+
+            val result2 = testRestTemplate.postForEntity<String>(
+                "/adr/datasets/hint_test/releases",
+                getPostEntityWithKey(mapOf("name" to listOf(releaseName)))
+            )
+
+            val data2 = ObjectMapper().readTree(result2.body!!)["data"]
+            assertThat(data2["name"].textValue()).isEqualTo(releaseName)
```
mwoodbri, line 322: I think worth adding one extra line here, just to check that a new release has actually been created. Otherwise you might just be getting the old one of the same name:
```suggestion
            assertThat(data2["name"].textValue()).isEqualTo(releaseName)
            assertThat(data2["id"].textValue()).isNotEqualTo(data["id"].textValue())
```

mwoodbri:
> Sorry, I'm struggling to test this. Should using the 2022 Antarctica dataset with the Tester user API key work? I keep getting this error, and no activity in the activity stream, whatever calibrate options I choose. I'm probably forgetting something vital though... 

This is an unfortunate feature of private datasets: they can only have one release. So if there is an existing release and its name doesn't match that of your new release (i.e. it isn't deleted by this new logic) then there will be an error. The way around this when using Antarctica is to delete the "1.0" release inherited from production ADR, create a new project in HINT, upload a release to ADR, start HINT from scratch, create a project of the same name and again upload a release. It should then replace the first one.

EmmaLRussell:
> > Sorry, I'm struggling to test this. Should using the 2022 Antarctica dataset with the Tester user API key work? I keep getting this error, and no activity in the activity stream, whatever calibrate options I choose. I'm probably forgetting something vital though...
> 
> This is an unfortunate feature of private datasets: they can only have one release. So if there is an existing release and its name doesn't match that of your new release (i.e. it isn't deleted by this new logic) then there will be an error. The way around this when using Antarctica (2022 version - this wasn't a problem when I was testing because we were still on 2021) is to delete the "1.0" release inherited from production ADR before testing. It may be necessary to re-instate Version 1.0 afterwards - I can't remember if any tests depend on it.

Thanks @mwoodbri I'd forgotten that 2022 Antarctica was private, but thought it might be something like that!

PR #556, opened 2021-09-23
mrc-2608 Persist current language in browser storage

carbowizzy:
This PR reuses VUE subscribe plugin in root  to listen for changes in changeLanguage mutation, if language is changed, selected language will be persisted in browser storage.

How to test:
* change app language and refresh page, the chosen language should remain as the app language and should not revert to English, which is the default language.

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/app/root.ts
+++ b/src/app/static/src/app/root.ts
@@ -41,6 +41,7 @@ import {
 } from "./store/downloadResults/downloadResults";
 import {ModelCalibrateMutation, ModelCalibrateUpdates} from "./store/modelCalibrate/mutations";
 import {GenericChartState, initialGenericChartState, genericChart} from "./store/genericChart/genericChart";
+import {LanguageMutation} from "./store/language/mutations";
 
 export interface TranslatableState {
     language: Language
@@ -76,6 +77,10 @@ const persistState = (store: Store<RootState>): void => {
     store.subscribe((mutation: MutationPayload, state: RootState) => {
         console.log(mutation.type);
         localStorageManager.saveState(state);
+
+        if (mutation.type === LanguageMutation.ChangeLanguage) {
+            localStorageManager.saveLanguage(state.language)
+        }
 
         const {dispatch} = store;
         const type = stripNamespace(mutation.type);
@@ -127,7 +132,7 @@ declare const currentUser: string;
 
 export const emptyState = (): RootState => {
     return {
-        language: Language.en,
+        language: localStorageManager.getLanguage(),
         version: currentHintVersion,
         hintrVersion: initialHintrVersionState(),
         adr: initialADRState(),
--- a/src/app/static/src/app/root.ts
+++ b/src/app/static/src/app/root.ts
@@ -41,7 +41,6 @@ import {
 } from "./store/downloadResults/downloadResults";
 import {ModelCalibrateMutation, ModelCalibrateUpdates} from "./store/modelCalibrate/mutations";
 import {GenericChartState, initialGenericChartState, genericChart} from "./store/genericChart/genericChart";
-import {LanguageMutation} from "./store/language/mutations";
 
 export interface TranslatableState {
     language: Language
--- a/src/app/static/src/app/root.ts
+++ b/src/app/static/src/app/root.ts
@@ -127,7 +127,7 @@ declare const currentUser: string;
 
 export const emptyState = (): RootState => {
     return {
-        language: localStorageManager.getState()?.language || Language.en,
+        language: Language.en,
         version: currentHintVersion,
         hintrVersion: initialHintrVersionState(),
         adr: initialADRState(),
@@ -149,9 +149,10 @@ export const emptyState = (): RootState => {
         downloadResults: initialDownloadResultsState()
     }
 };
+const existingState = localStorageManager.getState();
 
 export const storeOptions: StoreOptions<RootState> = {
-    state: emptyState(),
+    state: {...emptyState(), ...existingState && {language: existingState.language}},
     modules: {
         adr,
         genericChart,
--- a/src/app/static/src/app/root.ts
+++ b/src/app/static/src/app/root.ts
@@ -78,10 +78,6 @@ const persistState = (store: Store<RootState>): void => {
         console.log(mutation.type);
         localStorageManager.saveState(state);
 
-        if (mutation.type === LanguageMutation.ChangeLanguage) {
-            localStorageManager.saveLanguage(state.language)
-        }
-
         const {dispatch} = store;
         const type = stripNamespace(mutation.type);
         if (type[0] !== "projects" && type[0] !== "errors") {
@@ -132,7 +128,7 @@ declare const currentUser: string;
 
 export const emptyState = (): RootState => {
     return {
-        language: localStorageManager.getLanguage(),
+        language: localStorageManager.getState()?.language || Language.en,
         version: currentHintVersion,
         hintrVersion: initialHintrVersionState(),
         adr: initialADRState(),
--- a/src/app/static/src/tests/localStorageManager.test.ts
+++ b/src/app/static/src/tests/localStorageManager.test.ts
@@ -14,11 +14,10 @@ import {
     mockSurveyAndProgramState
 } from "./mocks";
 import {localStorageManager, serialiseState} from "../app/localStorageManager";
-import {RootState} from "../app/root";
+import {emptyState, RootState} from "../app/root";
 import {DataType} from "../app/store/surveyAndProgram/surveyAndProgram";
 import {currentHintVersion} from "../app/hintVersion";
 import {Language} from "../app/store/translations/locales";
-import {store} from "../app/main";
 
 declare const currentUser: string; // set in jest config, or on the index page when run for real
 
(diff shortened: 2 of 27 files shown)


mwoodbri (commented):
Because I'm not afraid of asking stupid questions, I'm going to ask one here: why can't we use the existing `rootState.language` field for this? Basically add it to [serialiseState()](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/localStorageManager.ts#L7), and then in `emptyState()` do `language: existingState?.language || Language.en`?

Otherwise we end up with two copies of `language` in `localStorage`...

carbowizzy:
> Because I'm not afraid of asking stupid questions, I'm going to ask one here: why can't we use the existing `rootState.language` field for this? Basically add it to [serialiseState()](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/localStorageManager.ts#L7), and then in `emptyState()` do `language: existingState?.language || Language.en`?
> 
> Otherwise we end up with two copies of `language` in `localStorage`...

This approach looks cleaner to me.

mwoodbri (changes requested):
An extra thing you'll need to do: ensure that `i18next.changeLanguage()` is called when loading language from local storage (and test for this) - at the moment it's only [invoked](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/language/actions.ts#L8) via mutations, which isn't the case here.

src/app/static/src/app/root.ts
```diff
@@ -127,7 +127,7 @@ declare const currentUser: string;
 
 export const emptyState = (): RootState => {
     return {
-        language: Language.en,
+        language: localStorageManager.getState()?.language || Language.en,
```
EmmaLRussell, line 130: Because `emptyState` is used in the tests for mocking, I think it would be nice to separate out the fetch from `localStorageManager` from `emptyState` itself, as we do for the modules. 

That would mean leaving `language: Language.en` as the default here, then probably pulling out existing state (again, see modules like `modelOptions.ts`), then on line 154 something like 
```
state: {...emptyState(), ...existingState && {language: existingState.language}}
```

carbowizzy: I have avoided making language a module, your suggested implementation means I have to make language a module. I believe this will change the existing language structure, which doesn't have language as module even though I do not understand why but I am wary. I can refactor the whole thing and make language work as module, please confirm there was no reason for leaving it as is.

carbowizzy:
> An extra thing you'll need to do: ensure that `i18next.changeLanguage()` is called when loading language from local storage (and test for this) - at the moment it's only [invoked](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/language/actions.ts#L8) via mutations, which isn't the case here.



> An extra thing you'll need to do: ensure that `i18next.changeLanguage()` is called when loading language from local storage (and test for this) - at the moment it's only [invoked](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/language/actions.ts#L8) via mutations, which isn't the case here.

I believe this is called when a user changes the language, if a user has not changed the language again why should `i18next.changeLanguage(lang)` be called ? Browser is loading persisted language and not making a change. Am I missing something ?

src/app/static/src/app/root.ts
EmmaLRussell: I don't believe it would need to be a module -the changes I am suggesting would be in `root.ts`, as described above.

EmmaLRussell: Just to be clear: 
```
export const emptyState = (): RootState => {
    return {
        language: Language.en
        ...etc....
    } 
};

const existingState = localStorageManager.getState();

  state: {...emptyState(), ...existingState && {language: existingState.language}},
   ...etc...
```

Note the difference in the last line with what the modules do - we want to set a single prop value directly on rootState if is existingState is present, not replace all props in a module.

carbowizzy: Oh in root.ts

mwoodbri:
> > An extra thing you'll need to do: ensure that `i18next.changeLanguage()` is called when loading language from local storage (and test for this) - at the moment it's only [invoked](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/language/actions.ts#L8) via mutations, which isn't the case here.
> 
> I believe this is called when a user changes the language, if a user has not changed the language again why should `i18next.changeLanguage(lang)` be called ? Browser is loading persisted language and not making a change. Am I missing something ?

There are a few cases in the codebase where we call `i18next.t()` with only one argument i.e. omitting the language. For example, [here](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/components/plots/MapAdjustScale.vue#L125). `i18next.changeLanguage()` is how we tell `i18next` what the current language is, hence it being invoked whenever we change language. If you don't call it when pulling the language from local storage then it will be inconsistent with `RootState.language` and that widget won't render errors properly.

More broadly we can discuss which approach we want to take to ensure consistency:
1. Always provide the language to `i18next.t()` and never call `i18next.changeLanguage()`
2. Always call `i18next.changeLanguage()` and never provide the language to `i18next.t()`

The advantage of second option is that we can still translate from contexts in which we don't have access to the store e.g. mutations in modules.

carbowizzy:
> `RootState.language` and that widget won't render errors properly

Fantastic, thanks for explaining that.

src/app/static/src/app/root.ts
```diff
@@ -149,9 +149,10 @@ export const emptyState = (): RootState => {
         downloadResults: initialDownloadResultsState()
     }
 };
+const existingState = localStorageManager.getState();
 
 export const storeOptions: StoreOptions<RootState> = {
-    state: emptyState(),
+    state: {...emptyState(), ...existingState && {language: existingState.language}},
```
mwoodbri, line 155: Do we have a test for this? Would be good to check that if language setting has been serialised to localStorage then it is reflected in initial state of the store.

I think this is a different test to the new one in `localStorageManager.test.ts` - it checks deserialisation but not that the the store is correctly populated.

mwoodbri (changes requested):
A couple of suggestions regarding tests.

One thing I'm slightly confused about is if you set language to French, create a project, logout and log back in, then the projects page is in English. But goes back to French when you reload the project.

However, if you're on the projects page, set the language to French, and reload the page then the language stays in French - as expected following this PR. Any ideas?

src/app/static/src/tests/localStorageManager.test.ts
```diff
@@ -118,4 +128,15 @@ describe("LocalStorageManager", () => {
         expect(spy.mock.calls[0][0]).toBe(`hintAppState_v${currentHintVersion}`);
         expect(spy.mock.calls[0][1]).toBe(JSON.stringify(testState));
     });
+
+    it("can get language from local storage", () => {
```
mwoodbri, line 132: The name of this test is a bit misleading - you're really testing both save and load. To test just loading you'd want to either mock `getItem`, or pre-populate localStorage itself.

src/app/static/src/app/root.ts
carbowizzy: https://github.com/mrc-ide/hint/blob/97faa5b7a0ddec3c306219529b535c0e27ff0ba0/src/app/static/src/tests/components/header/languageMenu.test.ts#L23

mwoodbri: If I revert your change to L154 then `npm test` still passes - including that one in `languageMenuTest.test.ts` - which implies that at least one test is missing

carbowizzy: I think this case can only be tested in browser i.e selenium.

carbowizzy:
> Seems to work very nicely.
> 
> One thing I'm slightly confused about is if you set language to French, create a project, logout and log back in, then the projects page is in English. But goes back to French when you reload the project.
> 
> However, if you're on the projects page, set the language to French, and reload the page then the language stays in French - as expected following this PR. Any ideas?
> 
> A couple of suggestions regarding tests inline.

I see what you mean about the project history language not being persisted. I have created a new ticket to handle this case here  https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2636

src/app/static/src/app/root.ts
mwoodbri: I assume you mean because `storeOptions` isn't a function (unlike `emptyState`), so that it's already initialised by the time the tests are run.

Would be good to get @EmmaLRussell's opinion on this. If it was a no-arg function then test would be as simple as:
```typescript
       jest.spyOn(localStorageManager, "getState").mockImplementation(() => mockRootState({language: Language.pt}));
       expect((storeOptions().state as RootState).language).toBe("pt");
```

EmmaLRussell: That's a good point, it could be. It also applies to all the other modules where this pattern is used - I don't think we have tests for any of those. We'd want to keep `existingState` as an object of course, initialised when the app loads.

EmmaLRussell (approved):
Nothing to add from me, if project history page bug is to be addressed in a separate ticket.

Nicolas-Dolan (changes requested):
Looks good. You just need to increment the version number.

mwoodbri (approved):

Nicolas-Dolan (approved):

PR #557, opened 2021-09-27
mrc-2622 Update all output plot metadata when language changes. 

EmmaLRussell:
## Description

When the user changes language, the filters and indicators for all output charts (map, barchart and bubble chart) should be updated for the new language. This also applies to the calibrate plot - however, we are not yet receiving fully updated metadata for this (awaiting full implementation in the back end). 

When language changes, we simultaneously refetch plotting metadata (which gets all metadata for the input plots, including indicators which are reused by the output map plots), and the calibrate results (which gets the remaining metadata for the output plots). 

Fetching the model calibrate results triggers a fetch of the calibrate plot, so that is automatically updated too. 

Because the fetch of result data takes a noticeable amount of time, a strange delay in updating some of the plot metadata was observed. To avoid this, I added an 'updatingLanguage' flag to the state, and show the 'Loading your data' spinner in the stepper if language is being updated, so that we don't see plot metadata being updated piecemeal. 

A result of this change is that, if you change language while on model options or model calibrate options steps, the options are automatically refetched in the new language, because the options components are unmounted and then re-mounted once language update completes, and mounting the component triggers a re-fetch of the options for the new language. This is really a happy accident rather than being planned. It might be confusing since the re-fetch of options occurs *after* the `updatingLanguage` stops being true. But it seems a little pointless to re-fetch the options twice. (Options are currently fetched when the component reloads in order to load any potential changes to options metadata.)

However, this caused a bug, in that user-selected option values were wiped when the options were fetched for the new language. This was because the control sections and groups in the new dynamic form metadata were being matched to the old metadata by their labels, which change when the language changes. I have refactored the form update code to instead always update the new form metadata by first creating a dictionary of form values from the old form metadata and matching on that. This dictionary uses the control names as keys, and can do so safely, as these are fixed between languages. The dictionary is similar to the `options` values in both modules - but not identical, since this is only updated when the form is validated, and also has its values transformed when emitted from the form, so we can't use this to match values to new metadata. 

There are two outstanding issues, for which I have created new tickets:

1. Currently, when options are re-fetched we decide whether re-validation of options is required based on whether the form metadata has changed at all (using `JSON.stringify`), suggesting we are dealing with a new version of the options in the back end. However, the form metadata has always changed by this measure when language changes, because of the updated labels. Model Options version is available from the backend, so we should use this as a more reliable guide to whether options need to be re-validated - if the new form metadata is a different version from the old metadata, then revalidation will be required. [Ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2633)

2. The output tables are showing column headers in lower case for FR and PT. This is because the table always treats filter labels as translation keys - this is correct for input metadata, but wrong for output metadata, which contains translated labels. We'll need to add a configuration flag to Table component to allow it to use translation keys or not. [Ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2635)

## Type of version change
_Delete as appropriate_
Minor

## Checklist


src/app/static/src/app/store/root/actions.ts
```diff
@@ -55,7 +55,13 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
         await changeLanguage<RootState>(context, payload)
 
         if (rootState.baseline?.iso3) {
-            await dispatch("metadata/getPlottingMetadata", rootState.baseline.iso3)
+            dispatch("metadata/getPlottingMetadata", rootState.baseline.iso3)
         }
+
+        if (rootState.modelCalibrate.status.done) {
+            dispatch("modelCalibrate/getResult")
```
EmmaLRussell, line 62: This actions re-fetches the calibrate plot as well as the calibrate model result which provides the metadata for the main result sets.

src/app/static/src/tests/modelOptions/mutations.test.ts
```diff
@@ -101,7 +101,7 @@ describe("Model run options mutations", () => {
                         {
                             label: "g1",
                             controls: [
-                                {...mockControl, value: 10},
+                                {...mockControl, name: "n1", value: 20},
```
EmmaLRussell, line 104: This test was sharing control names between groups - that was OK with old implementation where controls were checked by group, but now we rely on control names being unique in the form. This will be safe since the form transforms it to a dictionary anyway when it emits option values.

src/app/static/src/app/utils.ts
```diff
@@ -247,48 +252,27 @@ export const validateEmail = (test: string): boolean => {
...
-
-    const oldGroupLabels = oldSection.controlGroups.map(g => g.label);
-    newSection.controlGroups = newSection.controlGroups.map(g => {
-        const oldGroupIndex = oldGroupLabels.indexOf(g.label);
-        if (oldGroupIndex == -1) {
-            return g
-        } else {
-            return updateGroup(oldSection.controlGroups[oldGroupIndex], g)
-        }
+    const selectedOptions = {} as Dict<string | string[] | number | null | undefined>;
```
mwoodbri, line 255: This is now a thing of beauty... but slightly surprised we (still) don't have any tests for this? Unless I'm looking in the wrong place.

mwoodbri (commented):
Looking good.

Aside from a couple of nitpicks inline I'd just note that we have lingering English in Sex, Age and Detail dropdowns. But this is not new.

src/app/static/src/app/utils.ts
EmmaLRussell: It is (and was) tested via the modelOptions mutations tests - but you're right, it's worth adding some explicit tests for this method.

mwoodbri: Ah ok - sorry

carbowizzy (changes requested):
Code looks good to me, just one issue with behaviour. 

When browser is refreshed, the page returns to English but indicators and HIV prevalence on table column remain a different language. I know a ticket to persist language is on the way but you need to ensure it works with this behaviour. Also, calls are made to update language on every refresh, I would think we should persist and not re-fetch when language has not changed.

EmmaLRussell:
>When browser is refreshed, the page returns to English but indicators and HIV prevalence on table column remain a different language. I know a ticket to persist language is on the way but you need to ensure it works with this behaviour. Also, calls are made to update language on every refresh, I would think we should persist and not re-fetch when language has not changed.

Don't these two ticket work together though? Once language is persisted in saved state, both the language and metadata will be up to date for the last selected language when a refresh is performed. I don't really see the point of putting in code to fetch metadata on reload for language which is wrongly not persisted, when that's going to be fixed in mrc-2608, but I'm happy to delay merging this one until we have code available from that branch merged into master.

>Also, calls are made to update language on every refresh, I would think we should persist and not re-fetch when language has not changed.

Are you still talking about the way these two tickets interact? Because `changeLanguage` will now be called as part of the reload procedure from that ticket, but with these code changes that will trigger a re-fetch of data when that won't be necessary. That is a good point, I think we possibly need a way of distinguishing the two uses for changing language in that case. Again, that's something I could roll into this branch once the other is merged.

EmmaLRussell:
> we have lingering English in Sex, Age and Detail dropdowns. But this is not new.

Yeah, we have those even if in non-English from the start. I'll check with Rob about those.

EmmaLRussell:
> we have lingering English in Sex, Age and Detail dropdowns

Sex - we are actually doing hard-coded population of these for the input data. We should sort out this mess but for the moment I'll add the translations for these in this branch. For the output data, these come from hintr/naomi - I've mentioned it to Rob.

Age - I've flagged 'all ages' not being translated to Rob

Area detail - this text comes from the area hierarchy that the users upload, so we should be able to expect it to be in the language they expect.

EmmaLRussell:
> > Also, calls are made to update language on every refresh, I would think we should persist and not re-fetch when language has not changed.
> 
> Are you still talking about the way these two tickets interact? Because `changeLanguage` will now be called as part of the reload procedure from that ticket, but with these code changes that will trigger a re-fetch of data when that won't be necessary. That is a good point, I think we possibly need a way of distinguishing the two uses for changing language in that case. Again, that's something I could roll into this branch once the other is merged.

Sorry, @carbowizzy  I think I misinterpreted what you were saying here. I think you're just saying we don't need to refresh anything if the user 'changes' language in the drop-down by choosing the current language. I will make that change.

EmmaLRussell:
Now merged with the changes from mrc-2608, so on refresh page we see both UI text and datasets retain selected language.

src/app/static/src/app/store/root/actions.ts
```diff
@@ -51,11 +51,26 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
     },
 
     async changeLanguage(context, payload) {
-        const {dispatch, rootState} = context
-        await changeLanguage<RootState>(context, payload)
+        const {commit, dispatch, rootState} = context;
+
+        if (rootState.language === payload) {
+            return;
+        }
```
mwoodbri, line 58: I think this is sensible. Is the root cause that the language selector doesn't disable current language? If so then that's worthy of another ticket I think.

mwoodbri (approved):
Looks good.

mwoodbri:
> Also moved ADRClientTests so that they are now tested by the ADR test task in buildkite. Both ADR tasks currently failing because ADR dev is down, all other tests passing.

Now fixed

src/app/static/src/app/components/Stepper.vue
```diff
@@ -102,9 +104,12 @@
                 currentProject: state => state.currentProject,
                 projectLoading: state => state.loading
             }),
+            updatingLanguage: mapStateProp<RootState, boolean>(null,
+                (state: RootState) => state.updatingLanguage
+            ),
             ...mapGettersByNames<keyof ComputedGetters>(namespace, ["ready", "complete"]),
             loading: function () {
-                return this.loadingFromFile || !this.ready;
+                return this.loadingFromFile || this.updatingLanguage || !this.ready;
```
mwoodbri, line 112: At first I thought that `updatingLanguage` should be folded into `ready`, but your approach is more appropriate as `ready` relates more to progress through wizard than to global state

src/app/static/src/app/store/root/actions.ts
EmmaLRussell: I've made a ticket for this here: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2647
But with the change in this branch it is at least benign to select the current language, as nothing will happen.

carbowizzy (approved):

PR #559, opened 2021-10-05
Mrc 2640 - Endpoint for posting error report to Teams

carbowizzy:
* This PR creates an endpoint `/error-report` to accept JSON definition of error report, see comments on this ticket for structure https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2643. 
* PR sends body of the `/error-report` endpoint to office365 web hook which was created by this ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2646.
* When post to office365 is successful, a message or messages will appear in Naomi Support Teams channel.

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ErrorReportController.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ErrorReportController.kt
@@ -1,0 +1,29 @@
+package org.imperial.mrc.hint.controllers
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.imperial.mrc.hint.models.EmptySuccessResponse
+import org.imperial.mrc.hint.models.ErrorReport
+import org.imperial.mrc.hint.models.asResponseEntity
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.*
+
+@RestController
+class ErrorReportController(private val objectMapper: ObjectMapper)
+{
+    @PostMapping("/error-report")
+    @ResponseBody
+    fun sendErrorReport(@RequestBody errorReport: ErrorReport): ResponseEntity<String>
+    {
+        val errorReportJson = objectMapper.writeValueAsString(errorReport)
+
+        //post errorReportJson to webhook
+        val postErrorReport = ResponseEntity.ok().build<String>()
+
+        if(postErrorReport.statusCode == HttpStatus.OK){
+            return ResponseEntity.badRequest().build()
+        }
+
+        return EmptySuccessResponse.asResponseEntity()
+    }
+}
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ErrorReportController.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ErrorReportController.kt
@@ -4,26 +4,41 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import org.imperial.mrc.hint.models.EmptySuccessResponse
 import org.imperial.mrc.hint.models.ErrorReport
 import org.imperial.mrc.hint.models.asResponseEntity
-import org.springframework.http.HttpStatus
-import org.springframework.http.ResponseEntity
+import org.springframework.http.*
 import org.springframework.web.bind.annotation.*
+import org.springframework.web.client.RestTemplate
+import org.springframework.web.client.postForEntity
 
 @RestController
 class ErrorReportController(private val objectMapper: ObjectMapper)
 {
     @PostMapping("/error-report")
     @ResponseBody
-    fun sendErrorReport(@RequestBody errorReport: ErrorReport): ResponseEntity<String>
+    fun errorReport(@RequestBody errorReport: ErrorReport): ResponseEntity<String>
     {
-        val errorReportJson = objectMapper.writeValueAsString(errorReport)
+        val response = postErrorReport(errorReport)
 
-        //post errorReportJson to webhook
-        val postErrorReport = ResponseEntity.ok().build<String>()
-
-        if(postErrorReport.statusCode == HttpStatus.OK){
-            return ResponseEntity.badRequest().build()
+        if (response.statusCode != HttpStatus.OK)
+        {
+            return ResponseEntity.status(response.statusCode).build()
         }
 
         return EmptySuccessResponse.asResponseEntity()
+    }
+
+    fun postErrorReport(errorReport: ErrorReport): ResponseEntity<String>
+    {
+        val errorReportJson = objectMapper.writeValueAsString(errorReport)
+
+        val headers = HttpHeaders()
+        headers.contentType = MediaType.APPLICATION_JSON
+
+        val httpEntity = HttpEntity(errorReportJson, headers)
+        val url = "https://prod-58.westeurope.logic.azure.com:443/workflows/a13847fef9034c6099297b59490f8be2/triggers" +
+                "/manual/paths/invoke?api-version=2016-06-01&sp=%2Ftriggers%2Fmanual%2Frun&sv=1.0&sig" +
+                "=cVhgcE7b75SnaFfLsYZBYHaSvBRFHLk4enyZ2H7yMj4"
+
+        val restTemplate = RestTemplate()
+        return restTemplate.postForEntity(url, httpEntity)
     }
 }
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/models/ErrorReport.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/models/ErrorReport.kt
@@ -1,0 +1,17 @@
+package org.imperial.mrc.hint.models
+
+import java.time.Instant
+
+data class ErrorReport(
+        val email: String,
+        val country: String? = null,
+        val projectName: String,
+        val section: String,
+        val errors: List<Errors>?,
+        val description: String? = null,
+        val stepsToReproduce: String? = null,
+        val browserAgent: String,
+        val timeStamp: Instant? = Instant.now()
+)
+
+data class Errors(val jobId: String, val errorMessage: String)
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/models/ErrorReport.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/models/ErrorReport.kt
@@ -7,11 +7,11 @@ data class ErrorReport(
         val country: String? = null,
         val projectName: String,
         val section: String,
-        val errors: List<Errors>?,
+        val errors: List<Errors>,
         val description: String? = null,
         val stepsToReproduce: String? = null,
         val browserAgent: String,
         val timeStamp: Instant? = Instant.now()
 )
 
-data class Errors(val jobId: String, val errorMessage: String)
+data class Errors(val jobId: String? = null, val errorMessage: String? = null)


src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ErrorReportController.kt
```diff
@@ -0,0 +1,44 @@
...
+    fun postErrorReport(errorReport: ErrorReport): ResponseEntity<String>
+    {
+        val errorReportJson = objectMapper.writeValueAsString(errorReport)
+
+        val headers = HttpHeaders()
+        headers.contentType = MediaType.APPLICATION_JSON
+
+        val httpEntity = HttpEntity(errorReportJson, headers)
+        val url = "https://prod-58.westeurope.logic.azure.com:443/workflows/a13847fef9034c6099297b59490f8be2/triggers" +
+                "/manual/paths/invoke?api-version=2016-06-01&sp=%2Ftriggers%2Fmanual%2Frun&sv=1.0&sig" +
```
mwoodbri, line 38: "/manual/paths/invoke?api-version=2016-06-01&sp=/triggers/manual/run&sv=1.0&sig"

src/app/src/main/kotlin/org/imperial/mrc/hint/models/ErrorReport.kt
```diff
@@ -0,0 +1,18 @@
+package org.imperial.mrc.hint.models
+
+import java.time.Instant
+
+data class ErrorReport(
+        val email: String,
+        val country: String? = "",
+        val projectName: String,
```
mwoodbri, line 8: Is this field mandatory?

src/app/src/main/kotlin/org/imperial/mrc/hint/models/ErrorReport.kt
```diff
@@ -0,0 +1,18 @@
+package org.imperial.mrc.hint.models
+
+import java.time.Instant
+
+data class ErrorReport(
+        val email: String,
+        val country: String? = "",
```
mwoodbri, line 7: Normally I'd prefer to leave these as null and let the JSON serialiser elide them from the request sent to the Flow. That feels more natural than trying to represent an unset value as an empty string.

But I guess that sending empty strings makes it easier to write the Flow?

src/app/src/main/kotlin/org/imperial/mrc/hint/models/ErrorReport.kt
carbowizzy: We might have a situation where a user sends an error report without uploading baseline data, in that case it should be empty.

mwoodbri: Sorry, I meant why isn't it like this:
```suggestion
        val projectName: String? = "",
```

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/AppPropertiesTest.kt
```diff
@@ -158,4 +158,12 @@ class AppPropertiesTests
         val sut = ConfiguredAppProperties(props)
         assertThat(sut.adrOutputSummarySchema).isEqualTo("test-summary-schema")
     }
+
+    @Test
+    fun `can read issue report  url`()
```
Nicolas-Dolan, line 163:
```suggestion
    fun `can read issue report url`()
```

carbowizzy:
I don't see where the coverage issue is coming from as it doesn't show which lines.

src/app/src/main/kotlin/org/imperial/mrc/hint/clients/FuelClient.kt
```diff
@@ -10,7 +10,7 @@ import org.imperial.mrc.hint.asResponseEntity
 import org.springframework.http.ResponseEntity
 import java.io.File
 
-abstract class FuelClient(protected val baseUrl: String)
+abstract class FuelClient(protected val baseUrl: String? = null)
```
mwoodbri, line 13: `baseUrl` should never be null. `url` (the path/suffix) can be.

src/app/src/main/kotlin/org/imperial/mrc/hint/clients/FuelClient.kt
```diff
@@ -42,7 +42,13 @@ abstract class FuelClient(protected val baseUrl: String)
 
     protected fun postJson(url: String, json: String): ResponseEntity<String>
     {
-        return "$baseUrl/$url".httpPost()
+        val urlPath = when
+        {
+            baseUrl.isNullOrEmpty() -> url
```
mwoodbri, line 47: As above, this is the wrong way round... `url` can be optional, but not `baseUrl`. Otherwise we're inconsistent with the other, existing, clients.

src/app/src/main/kotlin/org/imperial/mrc/hint/clients/FlowClient.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Component
+
+interface Flow
+{
+    fun notifyTeams(url: String, data: ErrorReport): ResponseEntity<String>
+}
+
+@Component
+class FlowClient(val objectMapper: ObjectMapper) : FuelClient(), Flow
```
mwoodbri, line 14: As I mentioned in my previous review, and implied by Alex's suggested code, the FlowClient should configure itself:
```suggestion
class FuelFlowClient(val objectMapper: ObjectMapper) : FuelClient(appProperties.issueReportUrl), FlowClient
```
This way the Controller doesn't need to concern itself with the relevant application property (which is an implementation detail of the Flow client).

Also notice use of `FuelFlowClient` i.e. a Fuel-backed implementation of `FlowClient`.

src/app/src/main/kotlin/org/imperial/mrc/hint/clients/FlowClient.kt
```diff
@@ -0,0 +1,27 @@
+package org.imperial.mrc.hint.clients
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.imperial.mrc.hint.models.ErrorReport
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Component
+
+interface Flow
```
mwoodbri, line 8: For consistency:
```suggestion
interface FlowClient
```

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/clients/FlowClientTests.kt
```diff
@@ -0,0 +1,47 @@
...
+                "test steps",
+                "test agent",
+                Instant.now()
+        )
+
+        val url = "http://example.com"
+
+        val sut = FlowClient(ObjectMapper())
+
+        val result = sut.notifyTeams(url, data)
```
mwoodbri, line 38: What we don't have anywhere is a test that the data sent to Flow is actually in the expected format. That's probably ok at the moment as it's passed through verbatim, but there is still the potential of a field being accidentally omitted from the `ErrorReport` data class and therefore dropped on the way through.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/adr/ADRClientTests.kt
```diff
@@ -41,24 +41,24 @@ class ADRClientTests
...
         val response = sut.postFile("garbage", listOf(), Pair("garbage", File("/dev/null")))
-        assertThat(response.statusCodeValue).isEqualTo(500)
+        assertThat(response.statusCodeValue).isEqualTo(400)
         val errors = ObjectMapper().readValue<JsonNode>(response.body!!)["errors"]
-        assertThat(errors.isArray).isTrue()
+        assertThat(errors.isArray).isTrue
         assertThat(errors.count()).isEqualTo(1)
         assertThat(errors[0]["error"].textValue()).isEqualTo("OTHER_ERROR")
-        assertThat(errors[0]["detail"].textValue()).isEqualTo("Could not parse response.")
+        assertThat(errors[0]["detail"].textValue()).isEqualTo("Bad request - Action name not known: garbage")
```
hillalex, line 62: The change to `asResponseEntity` was really worthwhile even independently of the current ticket, it's a big improvement to be getting informative error messages back here!

hillalex (approved):
This is looking really good now, thanks.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/adr/ADRClientTests.kt
carbowizzy: Agreed ☝️

mwoodbri (commented):
Changes to `FlowClient` appreciated - thanks.

I have just one question remaining, in addition to a couple of minor comments.

src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
```diff
@@ -63,10 +63,10 @@ fun Response.asResponseEntity(logger: Log = LogFactory.getLog(HintExceptionHandl
 
         if (!json.has("status") && !json.has("success"))
         {
-            logger.error(json)
-            throw IOException("Response does not have status or success properties")
+            if (httpStatus < HttpStatus.BAD_REQUEST ) SuccessResponse(json).asResponseEntity()
+            else ErrorDetail(httpStatus, json.asText()).toResponseEntity<String>()
```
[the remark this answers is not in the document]
hillalex: But you're right about the logging, I missed that. Could log directly here, or rather than returning `ErrorDetail` throw some generic exception with the json response as the message...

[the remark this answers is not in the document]
mwoodbri: Ok, cool, thanks. In which case suggest something like this:
```suggestion
            if (!httpStatus.isError)
            {
                SuccessResponse(json).asResponseEntity()
            }
            else
            {
                val message = json.asText()
                logger.error(message)
                ErrorDetail(httpStatus, message).toResponseEntity<String>()
            }
```

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ErrorReportControllerTests.kt
```diff
@@ -1,18 +1,29 @@
...
+import org.imperial.mrc.hint.integration.SecureIntegrationTests
 import org.imperial.mrc.hint.models.ErrorReport
 import org.imperial.mrc.hint.models.Errors
 import org.junit.jupiter.api.Test
 import org.springframework.http.*
+import java.net.URL
 import java.time.Instant
 
-class ErrorReportControllerTests
+class ErrorReportControllerTests: SecureIntegrationTests()
```
mwoodbri, line 23: This class doesn't contain integration tests, so I don't think this is an appropriate superclass. If you're just wanting to use `readPropsFromTempFile()` then best to move that function to somewhere like existing `org.imperial.mrc.hint.helpers.Helpers`. Or make `can serialize request body` an actual integration test (see below).

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ErrorReportControllerTests.kt
```diff
@@ -47,6 +58,47 @@ class ErrorReportControllerTests
...
+            assertThat(response["errors"][0]["error"].asText()).isEqualTo("test error msg")
+            assertThat(response["errors"][0]["detail"].asText()).isEqualTo("#65ae0d095ea")
+            assertThat(response["errors"][1]["key"].asText()).isEqualTo("fomot-hasah-livid")
+            assertThat(response["errors"][1]["error"].asText()).isEqualTo("test error msg2")
+            assertThat(response["errors"][1]["detail"].asText()).isEqualTo("#25ae0d095e1")
+            assertThat(response["description"].asText()).isEqualTo("")
+            assertThat(response["stepsToReproduce"].asText()).isEqualTo("test steps")
+            assertThat(response["browserAgent"].asText()).isEqualTo("test agent")
+            assertThat(Instant.ofEpochSecond(response["timeStamp"]["epochSecond"].asLong()))
+                    .isEqualTo(Instant.ofEpochSecond(timeStamp.epochSecond))
```
mwoodbri, line 98: Adding this test seems to have revealed a problem: we don't want to send a field of the form `timestamp: {epochSecond: ... }` to Flow. I assume it's expecting an ISO date format string - not something specific to Java. Safer to use `String` rather than `Instant` in `ErrorReport.kt`? That way it will just be passed through. Or you'll need to ensure that it's (de)serialized correctly.

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ErrorReportControllerTests.kt
```diff
@@ -47,6 +58,47 @@ class ErrorReportControllerTests
         assertThat(result.statusCode).isEqualTo(HttpStatus.BAD_REQUEST)
     }
 
+    @Test
+    fun `can serialize request body`()
```
mwoodbri, line 62: I think this test would have more value as an integration test i.e. using `testRestTemplate.postForEntity()` rather than `sut.postErrorReport()`

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ErrorReportTests.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.imperial.mrc.hint.controllers.ErrorReportController
+import org.imperial.mrc.hint.models.ErrorReport
+import org.imperial.mrc.hint.models.Errors
+import org.junit.jupiter.api.Test
+import java.time.Instant
+
+class ErrorReportTests: SecureIntegrationTests()
+{
+    @Test
+    fun `can post error report`()
```
EmmaLRussell, line 18: This is testing the the FuelFlowClient and controller classes together, but I was suggesting that you add a test of the new endpoint so it would test both these things and also the ErrorReport deserialisation. So I'd expect it to look more like the other integration tests at this level, using `testRestTemplate.postForEntity` e.g. [ModelRunTests](https://github.com/mrc-ide/hint/blob/92af71ff8273ed38215efaf15601ddf9727e904c/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ModelRunTests.kt#L55)

src/app/tmp/fake.properties
```diff
@@ -0,0 +1 @@
+issue_report_url=https://mock.codes/400
```
EmmaLRussell, line 1: Was this included in error?

EmmaLRussell (commented):
Thanks for adding the integration test. Just a query about fake properties file.

EmmaLRussell (approved):

mwoodbri (approved):

PR #560, opened 2021-10-05
mrc-2653 modal dialog for reporting errors

hillalex:
## Description

Adds the modal dialog for a user to report issues, linked in the top right "Online support" menu. This changeset was large enough, so a further ticket will add the inputs that are conditional on the user being logged in [(project, email)](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2656). Note also further tickets for adding [French and Portuguese translations](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2656), and for [wiring up the component](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2643).

## Type of version change
None

## Checklist


src/app/static/src/app/components/ErrorReport.vue
```diff
@@ -0,0 +1,130 @@
...
+            }),
+            currentSection: {
+                get() {
+                    return this.section || i18next.t(this.currentSectionKey, {lng: this.currentLanguage});
+                },
+                set(newVal: string) {
+                    this.section = newVal
+                }
+            },
+            steps: mapStateProp<StepperState, StepDescription[]>("stepper", state => state.steps)
```
mwoodbri, line 107: I think we'll need to supplement this with a least an "Other" option (as shown in spec). This would include the Projects page. Spec also includes "Login".

src/app/static/src/app/components/ErrorReport.vue
```diff
@@ -0,0 +1,130 @@
...
+                <div class="col-4">
+                    <label for="section" v-translate="'section'"></label>
+                </div>
+                <div class="col-8">
+                    <select class="form-control"
+                            v-model="currentSection"
+                            id="section">
+                        <option v-for="step in steps"
+                                :key="step.number"
+                                v-translate:value="step.textKey"
```
mwoodbri, line 15: Nice. I wouldn't have thought of this, but it makes sense if we're assuming that if the interface is currently in (e.g.) French then the user will be writing the error report in French, and the step name should therefore also be submitted translated so that the whole report is in a single language.

src/app/static/src/app/components/ResetConfirmation.vue
```diff
@@ -1,10 +1,10 @@
 <template>
     <div>
         <modal :open="open">
-            <h4 v-if="isGuest" v-translate="'haveYouSaved'"></h4>
-            <h4 v-if="!isGuest" v-translate="'saveVersion'"></h4>
+                <h4 v-if="isGuest" v-translate="'haveYouSaved'"></h4>
+                <h4 v-if="!isGuest" v-translate="'saveVersion'"></h4>
 
-            <p v-translate="'discardWarning'"></p>
+                <p v-translate="'discardWarning'"></p>
```
mwoodbri, line 7: Hardly a big issue, but this change seems to make the slightly haphazard formatting of this file less consistent, rather than more. Also doesn't seem relevant to rest of PR?

src/app/static/src/app/components/ErrorReport.vue
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    interface Computed {
+        currentLanguage: Language
+        currentSectionKey: string
+        currentSection: string
+        steps: StepDescription[]
+    }
+
+    export default Vue.extend<Data, Methods, Computed, "open">({
```
mwoodbri, line 82: I had never seen a literal prop defined in the this way! At first glance I was rather wary of it introducing an untyped prop, but it's actually consistent with relevant prop of the Modal component.

src/app/static/src/app/components/ResetConfirmation.vue
hillalex: This was an accident, thanks for spotting!

src/app/static/src/app/components/ErrorReport.vue
hillalex: You're right, this is a little lazy and does lose the type info. I'll actually change it over to a regular typed prop.

src/app/static/src/app/components/ErrorReport.vue
EmmaLRussell: I guess it depends if were seeing this value as text or more of an enum. Might be easier to do analytics on the logged error reports by section if these values were the same across languages (as well as easier on anglophones dealing with the error reports). But consistency is good too!

EmmaLRussell (commented):
Looks good!

src/app/static/src/app/components/ErrorReport.vue
hillalex: Good shout, will use keys

mwoodbri (approved):
Looks merge-ready to me :)

As part of mrc-2655 we could consider making the form purely vertical rather than mix of horizontal and vertical. Might look a bit cleaner and would provide more space to accommodate long project names or email addresses, even if it's inconsistent with mock-up:

Screenshot from 2021-10-06 12-44-59

hillalex:
> As part of [mrc-2655](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2655) we could consider making the form purely vertical rather than mix of horizontal and vertical. Might look a bit cleaner and would provide more space to accommodate long project names or email addresses, even if it's inconsistent with mock-up:

Agree this would be nicer!

EmmaLRussell (approved):
LGTM

PR #562, opened 2021-10-06
mrc-2643 extract all errors from Vuex app state and combine with user provided data to generate an error report

hillalex:
## Description

~~Contains commits from https://github.com/mrc-ide/hint/pull/561 which should be reviewed/merged first~~

 This PR adds:
* a root getter that returns all errors in the app state, across all individual modules
* a root action that takes user provided data and combines with error data, and other app state as per spec to generate an error report object

## Type of version change
None

## Checklist


mwoodbri:
> "Section" should already be pre-populated with the current step! That was implemented in the first ticket: https://github.com/mrc-ide/hint/pull/560/files#diff-e6984caba3fbf8158d5c515645626b129fcc9ca5e5755f4657b76153cd8e2c03R112

Ah yes, sorry - might be possible to do a better job of automatically selecting projects and login options though? That's what I was testing. Maybe just having a catch-all "other" would be simpler to cover all non-stepper pages. But that's for another ticket.

src/app/static/src/app/store/root/actions.ts
```diff
@@ -6,8 +6,26 @@ import {LanguageActions} from "../language/language";
...
+    description: string,
+    stepsToReproduce: string,
+    email: string
+}
+
+export interface ErrorReport extends ErrorReportManualDetails {
+    country: string,
+    project: string | undefined,
+    browserAgent: string,
+    timeStamp: string,
```
carbowizzy, line 26: timeStamp is generated in the backend automatically, so you won't need to add it in frontend.

src/app/static/src/app/store/root/actions.ts
```diff
@@ -72,5 +90,21 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
...
+    },
+
+    generateErrorReport({state, getters}, payload) {
+        // TODO: instead of returning this object, POST it to the server
+        return {
+            email: payload.email || state.currentUser,
+            country: state.baseline.country,
+            project: state.projects.currentProject?.name,
+            browserAgent: navigator.userAgent,
+            timeStamp: new Date().toISOString(),
```
carbowizzy, line 102: likewise

src/app/static/src/app/store/root/actions.ts
```diff
@@ -72,5 +90,21 @@ export const actions: ActionTree<RootState, RootState> & RootActions = {
...
+            jobId: state.modelRun.modelRunId,
+            description: payload.description,
+            section: payload.section,
+            stepsToReproduce: payload.stepsToReproduce,
+            errors: getters.errors
+        }
```
carbowizzy, line 108: It would be nice to wrap this object in a type. JobId and errorMessage are type of array.

```
export interface ErrorReportMetadata {
    email: string,
    country: string,
    projectName: string,
    section: string,
    errors: ErrorReportMessage[],
    "description": string,
    "stepsToReproduce": string,
    "browserAgent": string
}
```

```
interface ErrorReportMessage {
    jobId: string,
    errorMessage: string
}
```

carbowizzy (changes requested):
Code looks brilliant, I would have loved for this bit to have it's own module. RootState is rather obscure. I have few inline suggestions below.

src/app/static/src/app/store/root/getters.ts
```diff
@@ -1,12 +1,63 @@
...
+    const errors = [] as Error[];
+    extractErrorsRecursively(state, errors);
+    return errors;
+}
+
+const isComplexObject = (state: any) => {
+    return typeof state === 'object' && !Array.isArray(state) && state !== null
+}
+
+const extractErrorsRecursively = (state: any, errors: Error[]) => {
```
EmmaLRussell, line 56: It's cool that this is recursive, but does it actually need to be at the moment or is it future-proofing? I can't think of any module that doesn't have its errors at the top level. Or is this handing errors properties which are arrays? I can't quite see how it would do that this since `isComplexObject` is ruling out arrays (and also you're handing the array types, `modelRun.errors` and `errors.errors` separately).

src/app/static/src/app/store/root/getters.ts
```diff
@@ -1,12 +1,63 @@
...
+            metadata,
+            modelCalibrate,
+            modelOptions,
+            modelOutput,
+            plottingSelections,
+            projects,
+            surveyAndProgram
+        } = state;
+
+        return ([] as Error[]).concat.apply([] as Error[], [extractErrors(adr),
```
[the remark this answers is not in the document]
hillalex: Looks like password state is not part of the main app? I think it has it's own app right?

src/app/static/src/app/store/root/getters.ts
hillalex: It's recursive to handle the download results objects, which contain nested errors, see this test: https://github.com/mrc-ide/hint/pull/562/files#diff-c80530016ccbe2b220a5f11a243bed39425fe75f6626fb4ee6b357281710b7e0R81

src/app/static/src/tests/root/getters.test.ts
```diff
@@ -0,0 +1,192 @@
...
+    mockMetadataState,
+    mockModelCalibrateState,
+    mockModelOptionsState,
+    mockModelRunState, mockProjectsState,
+    mockRootState, mockSurveyAndProgramState
+} from "../mocks";
+import {RootState} from "../../app/root";
+import {initialDownloadResults} from "../../app/store/downloadResults/downloadResults";
+
+describe("root getters", () => {
```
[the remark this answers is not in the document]
hillalex: https://github.com/mrc-ide/hint/pull/562/commits/3cee24b4ead6cff8dfc3bb4afd80d3ddf8630273

src/app/static/src/app/store/root/actions.ts
mwoodbri: Agreed - I think it's preferable to add timestamp here, even if backend does intend to fiddle with the JSON.

src/app/static/src/app/store/root/getters.ts
```diff
@@ -1,12 +1,63 @@
...
+            extractErrors(modelOutput),
+            extractErrors(plottingSelections),
+            extractErrors(projects),
+            extractErrors(surveyAndProgram),
+            state.modelRun.errors,
+            state.errors.errors]);
+    }
+}
+
+const extractErrors = (state: any) => {
```
mwoodbri, line 46: I'm afraid I would be keen to see a test specifically for this slightly gnarly function - not least as a means of documenting its intended behaviour. I know you're testing it indirectly already, but given it's standalone it should be quite easy to add a further test to ensure that things with and without `[Ee]rror` suffix, and without and without an `error` field get in/excluded appropriately.

hillalex: https://github.com/mrc-ide/hint/pull/562/commits/2d0cc813d499ae15a78a48842c9abf29b1b4b696

PR #563, opened 2021-10-07
mrc-2468 Re-fetch Input Time Series data when language changes 

EmmaLRussell:
## Description

One issue this has highlighted is that selected Review Inputs tab is not retained when the view is reloaded, as happens when the language is updated, but also if return to the step from another, or when refresh page. I've made a separate ticket to persist selected tab: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2659

Another issue is that not all Input Time series metadata is being translated. Ticket for this issue: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2660

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/app/store/genericChart/actions.ts
+++ b/src/app/static/src/app/store/genericChart/actions.ts
@@ -3,10 +3,12 @@ import {RootState} from "../../root";
 import {api} from "../../apiService";
 import {GenericChartState} from "./genericChart";
 import {GenericChartMutation} from "./mutations";
+import {DatasetConfig, Dict, GenericChartMetadata} from "../../types";
 
 export interface MetadataActions {
     getGenericChartMetadata: (store: ActionContext<GenericChartState, RootState>) => void
     getDataset: (store: ActionContext<GenericChartState, RootState>, payload: getDatasetPayload) => void
+    refreshDatasets: (store: ActionContext<GenericChartState, RootState>) => void
 }
 
 export interface getDatasetPayload {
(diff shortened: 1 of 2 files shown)


src/app/static/src/app/store/genericChart/actions.ts
```diff
@@ -40,5 +42,26 @@ export const actions: ActionTree<GenericChartState, RootState> & MetadataActions
...
+                                    urls[dataset.id] = dataset.url
+                                });
+                                return urls;
+                            }, {});
+
+        const getDatasetActions: Promise<unknown>[] = [];
+        Object.keys(state.datasets).forEach((datasetId: string) => {
+            const url = datasetUrls[datasetId];
+            getDatasetActions.push(dispatch("getDataset", {datasetId, url}));
+        });
```
mwoodbri, line 64: I would find something like this easier to read, but very subjective:
```suggestion
        const getDatasetActions = Object.keys(state.datasets).map(datasetId => dispatch("getDataset", {datasetId, url: datasetUrls[datasetId]}));
```

carbowizzy (approved):
LGTM

Nicolas-Dolan:
> > Maybe there is something wrong with my setup but the translations that are coming through are rather hit and miss for me
> 
> See above - not all translation is covered by this ticket:
> 
> > To test: if you change language after input time series data has been fetched, then view that data, you should see that filter values are now in the new language (apart from area level which is language-independent, and time step which is not yet being translated). Filter labels and descriptions (shown in help text) will be translated in mrc-2660.

Ok, so by process of elimination, at this point, it's just the plot type values being translated (not including the plot type label).

PR #565, opened 2021-10-07
mrc-2628 Include Back/Continue links at bottom of model options page

mwoodbri:
## Description

To test:
- Check that the Back/Continue links now appear at bottom of Step 3 (the model options page) in addition to being present and enabled/disabled at the top of all pages as previously
- Also check that clicking Back link has no effect when on Step 1 (this wasn't previously the case)
 
## Type of version change

Minor

## Checklist

### DIFF
--- a/src/app/static/src/tests/components/stepper.test.ts
+++ b/src/app/static/src/tests/components/stepper.test.ts
@@ -46,6 +46,7 @@ import {RootState} from "../../app/root";
 import ModelCalibrate from "../../app/components/modelCalibrate/ModelCalibrate.vue";
 import {getters as rootGetters} from "../../app/store/root/getters";
 import {expectTranslated} from "../testHelpers";
+import StepperNavigation from "../../app/components/StepperNavigation.vue";
 
 const localVue = createLocalVue();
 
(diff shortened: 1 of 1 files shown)


src/app/static/src/tests/components/stepperNavigation.test.ts
```diff
@@ -0,0 +1,49 @@
...
+        backLink.trigger("click");
+        expect(back.mock.calls.length).toBe(1);
+
+        const continueLink = wrapper.find("#continue");
+        expect(continueLink.classes("disabled")).toBe(false);
+        continueLink.trigger("click");
+        expect(next.mock.calls.length).toBe(1);
+    });
+
+    it("disables links", () => {
```
hillalex, line 32: So this is just adding a disabled style, not actually disabling the links. You'll want to test for a `disabled` attribute as well.

src/app/static/src/app/components/Stepper.vue
```diff
@@ -49,6 +36,10 @@
                 <download-results v-if="isActive(7)"></download-results>
             </div>
         </div>
+        <template v-if="activeStep === 3">
```
hillalex, line 39: might want to test this logic as well, i.e. that the extra widget is shown on the appropriate step

src/app/static/src/tests/components/stepperNavigation.test.ts
hillalex: Oh yeah, `disabled` only works for buttons :facepalm:

carbowizzy (approved):
LGTM

Thanks for making changes.

PR #567, opened 2021-10-11
mrc-2635 Show correct output table headers in French and Portuguese

hillalex:
## Description

Add a flag to the `Table` component to turn off filter label translations, and turn them off for the output tables.

## Type of version change
Patch

## Checklist


carbowizzy (approved):
LGTM

PR #569, opened 2021-10-13
Mrc 2641 Action for posting error report

LekanAnni:
This PR sends json data to /error-report endpoint and shares project with Naomi support after a successful post. 

Integration tests for frontend and backend returns `Could not parse API response.` instead of 500 error because the error format is unknown to hint. Hence frontend returns custom error msg.
 
Sorry about the mess created while resolving merge conflicts.
 
## Type of version change
_Delete as appropriate_

Minor

## Checklist

### DIFF
--- a/src/app/static/src/app/apiService.ts
+++ b/src/app/static/src/app/apiService.ts
@@ -20,6 +20,7 @@ export interface API<S, E> {
     ignoreSuccess: () => API<S, E>
 
     postAndReturn<T>(url: string, data: any): Promise<void | ResponseWithType<T>>
+
     get<T>(url: string): Promise<void | ResponseWithType<T>>
 
     delete(url: string): Promise<void | true>


src/app/static/src/app/components/ErrorReport.vue
```diff
@@ -108,7 +108,8 @@
                 description: "",
                 stepsToReproduce: "",
                 section: "",
-                email: ""
+                email: "",
+                timeStamp: ""
```
hillalex, line 112: `timeStamp` shouldn't be here

src/app/static/src/tests/integration/root.itest.ts
```diff
@@ -0,0 +1,48 @@
...
+            email: "test@example.com",
+            stepsToReproduce: "steps to reproduce",
+            section: "login",
+            description: "test"
+        }
+
+        await actions.generateErrorReport({commit, rootState: state, getters, dispatch} as any, payload);
+        expect(commit.mock.calls[0][0].payload).toStrictEqual(
+            {
+                detail: "Could not parse API response. Please contact support.",
```
hillalex, line 41: is this what we're expecting? makes it a bit of a redundant test as this error could come from anywhere!

PR #571, opened 2021-10-13
Update error report object in back end to include version info

mwoodbri:
## Description

## Type of version change

Patches

## Checklist


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ErrorReportTests.kt
```diff
@@ -16,27 +16,34 @@ class ErrorReportTests: SecureIntegrationTests()
     fun `can post error report`()
     {
         val data = """{
-            "email":"test.user@example.com",
```
hillalex, line 19: It'd be handy really if we had a json schema for this somewhere wouldn't it? Then we could include schema validation in back-end tests and auto-generate the ts type for the front-end to guarantee compliance, in the same way we do with hintr objects. 

But this looks correct in any case.

PR #572, opened 2021-10-15
Mrc 2669 Getter to filter warnings from hintr

carbowizzy:
This PR adds warnings returned from hintr to root getters. 

* warnings getter accepts stepName as parameter, aggregates and returns warnings relating to the given steps.
* warnings can be used in relevant warning UI components.

## Type of version change
_Delete as appropriate_

Minor

## Checklist

### DIFF
--- a/src/app/static/src/app/root.ts
+++ b/src/app/static/src/app/root.ts
@@ -80,7 +80,7 @@ export interface WarningsState {
 
 export enum STEPS  {
     modelOptions = "model_options",
-    modelFit = "model_fit",
+    modelRun = "model_fit",
     modelCalibrate = "model_calibrate",
     reviewOutput = "review_output",
     downloadResult = "download_results"
--- a/src/app/static/src/app/root.ts
+++ b/src/app/static/src/app/root.ts
@@ -78,6 +78,14 @@ export interface WarningsState {
     warnings: Warning[]
 }
 
+export enum STEPS  {
+    modelOptions = "model_options",
+    modelFit = "model_fit",
+    modelCalibrate = "model_calibrate",
+    reviewOutput = "review_output",
+    downloadResult = "download_results"
+}
+
 const persistState = (store: Store<RootState>): void => {
     store.subscribe((mutation: MutationPayload, state: RootState) => {
         console.log(mutation.type);
--- a/src/app/static/src/app/store/root/getters.ts
+++ b/src/app/static/src/app/store/root/getters.ts
@@ -1,12 +1,32 @@
 import {RootState} from "../../root";
 import {Getter, GetterTree} from "vuex";
+import {Warning} from "../../generated";
 
 interface RootGetters {
     isGuest: Getter<RootState, RootState>
+    warnings: Getter<RootState, RootState>
 }
 
 export const getters: RootGetters & GetterTree<RootState, RootState> = {
     isGuest: (state: RootState, getters: any) => {
         return state.currentUser == "guest";
+    },
+
+    warnings: (getters: any, rootGetters: any, rootState: RootState) => (stepName: string) => {
+
+        const warnings = {
+            modelOptions: rootState.modelOptions.warnings,
+            modelRun: rootState.modelRun.warnings,
+            modelCalibrate: rootState.modelCalibrate.warnings
+        }
+
+        const filterWarnings = (warnings: Warning[], stepName: string) =>
+            warnings.filter(warning => !!warning.locations.find(location => location === stepName))
+
+        return {
+            modelOptions: filterWarnings(warnings.modelOptions, stepName),
+            modelRun: filterWarnings(warnings.modelRun, stepName),
+            modelCalibrate: filterWarnings(warnings.modelCalibrate, stepName)
+        }
     }
 };
--- a/src/app/static/src/tests/root/getters.test.ts
+++ b/src/app/static/src/tests/root/getters.test.ts
@@ -1,0 +1,46 @@
+import {mockModelCalibrateState, mockModelOptionsState, mockModelRunState, mockRootState} from "../mocks";
+import {getters} from "../../app/store/root/getters";
+import {Warning} from "../../app/generated";
+import {STEPS} from "../../app/root";
+
+describe(`root getters`, () => {
+
+    const warning: Warning[] = [
+        {text: "model option test", locations: ["model_options"]},
+        {text: "model calibrate test", locations: ["model_calibrate"]},
+        {text: "model run test", locations: ["model_fit"]}
+    ]
+
+    const testState = () => {
+        return mockRootState({
+            modelOptions: mockModelOptionsState({warnings: warning}),
+            modelRun: mockModelRunState({warnings: warning}),
+            modelCalibrate: mockModelCalibrateState({warnings: warning})
+        })
+    }
+
+    it(`can get model run warnings`, () => {
+        const rootState = testState()
+
+        const warn = getters.warnings(rootState, null, testState() as any, null)
+        const result = warn(STEPS.modelRun).modelRun
+        expect(result).toEqual([{text: "model run test", locations: [STEPS.modelRun]}])
+    })
+
+    it(`can get model calibrate warnings`, () => {
+        const rootState = testState()
+
+        const warn = getters.warnings(rootState, null, testState() as any, null)
+        const result = warn(STEPS.modelCalibrate).modelCalibrate
+        expect(result).toEqual([{text: "model calibrate test", locations: [STEPS.modelCalibrate]}])
+    })
+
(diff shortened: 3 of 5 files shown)


src/app/static/src/app/root.ts
```diff
@@ -78,6 +78,14 @@ export interface WarningsState {
     warnings: Warning[]
 }
 
+export enum STEPS  {
```
EmmaLRussell, line 81: We generally use pascal case rather than all caps for enums e.g. Language, LanguageMutation.

I guess we need this enum because it matches the 'locations' in the responses from hintr, which don't quite match the textKeys of the stepper steps?

src/app/static/src/tests/root/getters.test.ts
```diff
@@ -0,0 +1,46 @@
+import {mockModelCalibrateState, mockModelOptionsState, mockModelRunState, mockRootState} from "../mocks";
+import {getters} from "../../app/store/root/getters";
+import {Warning} from "../../app/generated";
+import {STEPS} from "../../app/root";
+
+describe(`root getters`, () => {
+
+    const warnings: Warning[] = [
```
EmmaLRussell, line 8: I think these would be slightly better tests if the mock state modules did not have identical warning arrays, so that we could test that warnings from each of the origins is being sourced correctly. 

Could we also include tests for warnings to be displayed on a step other than the origin e.g. show model options warning on model_fit or on review_output.

src/app/static/src/app/hintVersion.ts
```diff
@@ -1 +1 @@
-export const currentHintVersion = "1.56.2";
+export const currentHintVersion = "1.57.0";
```
EmmaLRussell, line 1: I don't think we actually need to update the version where there is a non-behavioural change like this one, but it doesn't really matter.

src/app/static/src/app/store/root/getters.ts
```diff
@@ -1,12 +1,32 @@
...
     isGuest: Getter<RootState, RootState>
+    warnings: Getter<RootState, RootState>
 }
 
 export const getters: RootGetters & GetterTree<RootState, RootState> = {
     isGuest: (state: RootState, getters: any) => {
         return state.currentUser == "guest";
+    },
+
+    warnings: (getters: any, rootGetters: any, rootState: RootState) => (stepName: string) => {
```
mwoodbri, line 15: This function signature doesn't match definition `Getter<S, R> = (state: S, getters: any, rootState: R, rootGetters: any) => any`. You've gotten away with it due to the way you've structured the tests but suggest the following which is both correct, and consistent with `isGuest`:
```suggestion
    warnings: (state: RootState, getters: any) => (stepName: string) => {
```

src/app/static/src/app/store/root/getters.ts
```diff
@@ -1,12 +1,32 @@
...
+            modelOptions: rootState.modelOptions.warnings,
+            modelRun: rootState.modelRun.warnings,
+            modelCalibrate: rootState.modelCalibrate.warnings
+        }
+
+        const filterWarnings = (warnings: Warning[], stepName: string) =>
+            warnings.filter(warning => !!warning.locations.find(location => location === stepName))
+
+        return {
+            modelOptions: filterWarnings(warnings.modelOptions, stepName),
```
mwoodbri, line 27: You could do the following, avoiding the definition of the `warnings` object entirely (and therefore also the clash of variable names inside `filterWarnings`):
```suggestion
            modelOptions: filterWarnings(state.modelOptions.warnings, stepName),
```

src/app/static/src/tests/root/getters.test.ts
```diff
@@ -0,0 +1,78 @@
...
+            modelRun: mockModelRunState({warnings: modelRunWarnings}),
+            modelCalibrate: mockModelCalibrateState({warnings: calibrateWarnings})
+        })
+    }
+
+    it(`can get modelRun warnings when on model run step`, () => {
+        const rootState = testState()
+
+        const warn = getters.warnings(rootState, null, testState() as any, null)
+        const result = warn(Steps.modelRun).modelRun
```
mwoodbri, line 32: Personally I'd find the tests easier to read if variables where renamed as follows, but not a huge deal:
```suggestion
        const getter = getters.warnings(rootState, null, testState() as any, null)
        const warnings = warn(Steps.modelRun).modelRun
```

src/app/static/src/app/store/root/getters.ts
```diff
@@ -1,12 +1,32 @@
...
+    warnings: (getters: any, rootGetters: any, rootState: RootState) => (stepName: string) => {
+
+        const warnings = {
+            modelOptions: rootState.modelOptions.warnings,
+            modelRun: rootState.modelRun.warnings,
+            modelCalibrate: rootState.modelCalibrate.warnings
+        }
+
+        const filterWarnings = (warnings: Warning[], stepName: string) =>
+            warnings.filter(warning => !!warning.locations.find(location => location === stepName))
```
mwoodbri, line 24: Doesn't `.find()` always return a boolean?
```suggestion
            warnings.filter(warning => warning.locations.find(location => location === stepName))
```

src/app/static/src/app/store/root/getters.ts
```diff
@@ -1,12 +1,32 @@
...
+
+    warnings: (getters: any, rootGetters: any, rootState: RootState) => (stepName: string) => {
+
+        const warnings = {
+            modelOptions: rootState.modelOptions.warnings,
+            modelRun: rootState.modelRun.warnings,
+            modelCalibrate: rootState.modelCalibrate.warnings
+        }
+
+        const filterWarnings = (warnings: Warning[], stepName: string) =>
```
mwoodbri, line 23: This doesn't feel right to me - I think this function should accept a `stepName` (as implied by the argument's name!) rather than the name of a location. Then the `enum` could be private to this file/function.

Therefore, callers of this getter would specify the name of the step (e.g. `modelOptions`), the getter would look up the corresponding location (i.e. `model_options`), find all the warnings that specify that location, and return them.

src/app/static/src/app/root.ts
```diff
@@ -78,6 +78,14 @@ export interface WarningsState {
     warnings: Warning[]
 }
 
+export enum Steps  {
```
mwoodbri, line 81: Feels like this should be a map of location to step, not an enum mapping step to location (see comment below).

src/app/static/src/app/store/root/getters.ts
carbowizzy: I struggle to understand your preference here, the function works perfectly well as intended. 

Caller of the getter can make a call such as `getters.warnings(Steps.modelOptions)`

carbowizzy: The enum lives in root state, this implementation is for root module. That's not external. It is also reused in tests.

mwoodbri: I did look at that, but thought (maybe incorrectly) they were translation keys and therefore not the most appropriate. And I assumed (again incorrectly?) that the warnings component would be embedded in the relevant steps rather than the stepper itself, and the step would know its name rather than `textKey`!

EmmaLRussell: That sounds good to me.

EmmaLRussell:
> Per [this comment](https://github.com/mrc-ide/hint/pull/572/files#r732502356), suggestion is remove changes to `root.ts` and instead do mapping from step name to location within `getters.ts`. This will simplify invocation of getter from `Stepper.vue` - it will just pass in existing `textValue` value rather than location ([mrc-2671](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2671)).

I haven't tested this, but it might look a bit like this: 
```
const warningStepLocationMapping: Dict<string> = {
    modelOptions: "model_options",
    fitModel: "model_fit",
    calibrateModel: "model_calibrate",
    reviewOutput: "review_output",
    downloadResults: "download_results"
};

export const getters: RootGetters & GetterTree<RootState, RootState> = {
    isGuest: (state: RootState, getters: any) => {
        return state.currentUser == "guest";
    },

    warnings: (state: RootState) => (stepName: string) => {
        const filterWarnings = (warnings: Warning[], stepLocation?: string) =>
            stepLocation ?
                warnings.filter(warning => warning.locations.some(location => location === stepLocation)) :
                [];

        const location = warningStepLocationMapping[stepName];

        return {
            modelOptions: filterWarnings(state.modelOptions.warnings, location),
            modelRun: filterWarnings(state.modelRun.warnings, location),
            modelCalibrate: filterWarnings(state.modelCalibrate.warnings, location)
        }
    }
};
```

EmmaLRussell (approved):

PR #576, opened 2021-10-19
mrc-2444 accessibility issues on projects page

hillalex:
## Description

This adds aria labels to all the icon buttons on the projects page and labels or title attributes to all the form controls. Auto-formatting seems to have made quite a few whitespace changes, so might be worth looking at the diff with show whitespace turned off.

Note the report linked in the ticket contains other failures to do with the header bar, but since the ticket pertains to the projects page only, these aren't addressed here and a separate ticket has been opened for those.

## Type of version change

Patch

## Checklist


src/app/static/src/app/components/projects/ShareProject.vue
```diff
@@ -12,6 +13,7 @@
                 <div class="row mb-2" v-for="(email, index) in emailsToShareWith" :key="index">
                     <div class="col">
                         <input autocomplete="no"
+                               title="enter email address to share with"
```
mwoodbri, line 16: Any reason why this isn't `aria-label`? Should probably localise too...

PR #578, opened 2021-10-19
mrc-2668 component to display warnings from hintr

Nicolas-Dolan:
## Description

SEE COMMENT AT END OF PR FOR LATEST INSTRUCTIONS

This component deviates from the [mocks](https://docs.google.com/presentation/d/1FEQ6n-qw70V4U3Ue8D6Nx_L7zpYvo_QVP_NhX0tyTyI/edit#slide=id.gcb1e6717ba_0_3) aesthetically to reflect the comments  given in the mocks. It also displays where warnings originate from as per [comments](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2669) on mrc-2669. It does not receive a location prop as it will only receives appropriate warnings for its step in mrc-2671 and therefore it doesn't need to know what step it is on. It receives a warnings prop in the format:
```
interface Warning {
        text: string;
        locations: ("model_options" | "model_fit" | "model_calibrate" | "review_output" | "download_results")[];
 }

interface Warnings {
                modelOptions: Warning[],
                modelRun: Warning[],
                modelCalibrate: Warning[]
 }
```

The component is not currently implemented in the code base, which will be covered in mrc-2671. The easiest way to test manually is to add these lines of code to the stepper (the second component is to showcase what the warnings look like if there are less than or equal to the max number of lines in the warning message and the third when there are no warnings in the arrays):
```
<warning-alert :warnings="{ 
            modelOptions: [{ text: 'a warning', locations: []}, { text: 'another warning', locations: []}, { text: 'third warning', locations: []}, { text: 'final warning', locations: []}],
            modelRun: [],
            modelCalibrate: [{ text: 'calibrate warning', locations: []}]
            }"></warning-alert>
 <warning-alert :warnings="{ 
            modelOptions: [{ text: 'a warning', locations: []}, { text: 'another warning', locations: []}, { text: 'third warning', locations: []}],
            modelRun: [],
            modelCalibrate: []
            }"></warning-alert>
<warning-alert :warnings="{ 
            modelOptions: [],
            modelRun: [],
            modelCalibrate: []
            }"></warning-alert>
```

 image

## Type of version change
Minor

## Checklist


src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -0,0 +1,175 @@
...
+
+    export default Vue.extend<Data, Methods, Computed, Props>({
+        name: "WarningAlert",
+        props: {
+            warnings: Object,
+            maxLines: {
+                default: 3,
+                required: false,
+                type: Number
+            }
```
Nicolas-Dolan, line 90: The user has the option to configure the number of lines (after the first warning header) that will be displayed. If the height of the warnings are more than this number, the last line will be replaced with an ellipsis and the show more button will appear.

src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    interface Warning {
+        text: string;
+        locations: ("model_options" | "model_fit" | "model_calibrate" | "review_output" | "download_results")[];
+    }
+
+    interface Warnings {
+        [key: string]: Warning[];
+    }
```
Nicolas-Dolan, line 80: These can probably be imported once the component gets implemented in mrc-2671. Note, I made the Warnings interface very generic here as the interface with specific keys would not play nice with some of my functions in ts.

src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -0,0 +1,175 @@
...
+            }
+        },
+        watch: {
+            warnings(){
+                this.updateDimensions()
+            }
+        },
+        mounted(){
+            this.updateDimensions()
+        },
```
Nicolas-Dolan, line 167: Because these updates only occur on mount or when the warnings prop changes, there is an edge case where the user could adjust their screen size after rendering and the height of the warning box will not be appropriate.

src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -0,0 +1,175 @@
+<template>
+    <div v-if="showAlert">
+        <div class="content alert alert-warning pt-0">
+            <div :style="containerBox">
+                <div ref="warningBox" id="warningBox">
+                    <div v-for="(value, key) in filteredWarnings" :key="key">
```
hillalex, line 6: How about just:
```
<div v-for="(value, key) in warnings" v-if="value.length > 0" :key="key">
```
Then you don't need to computed `filteredWarnings` at all.

src/app/static/src/tests/components/warningAlert.test.ts
```diff
@@ -0,0 +1,128 @@
...
+        })
+        const showToggle = wrapper.find("#showToggle")
+        expect(showToggle.find("p").text()).toBe("...")
+        expectTranslated(showToggle.find("button"), 
+            "Show more", 
+            "Montre plus", 
+            "Mostre mais", 
+        store)
+        await showToggle.find("button").trigger("click")
+        expect(showToggle.find("p").exists()).toBe(false)
```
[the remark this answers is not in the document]
hillalex: The `showToggle` element was retrieved before the state change, so you need to check:
```
wrapper.find("#showToggle").find("p").exists()
```

I'm not sure why the `expectTranslated` helper doesn't have the same problem tbh

src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -0,0 +1,175 @@
...
+                        <h4 class="alert-heading pt-2">
+                            <alert-triangle-icon size="1.5x" class="custom-class mr-1 mb-1"></alert-triangle-icon>
+                            {{ headerText(key) }}
+                        </h4>
+                        <ul class="mb-0">
+                            <li v-for="warning in value" :key="warning.text">{{ warning.text }}</li>
+                        </ul>
+                    </div>
+                </div>
+                <div ref="incHeader" class="invisible">
```
hillalex, line 16: What is `incHeader` short for? Spent a while trying to work out what this meant, so maybe a different name would be clearer!

src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -0,0 +1,175 @@
...
+                    <h4 class="alert-heading pt-2">
+                        <alert-triangle-icon size="1.5x" class="custom-class mr-1 mb-1"></alert-triangle-icon>
+                        Hidden header
+                    </h4>
+                    <p ref="line" class="mb-0">...</p>
+                </div>
```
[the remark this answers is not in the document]
hillalex: This works really nicely. I think it's quite clever and I can't think of another way of doing it.

src/app/static/src/app/components/WarningAlert.vue
hillalex: What was the nature of the ts problem? But yes should be able to share these interfaces with the state/getter. In fact `Warning` (singular) should already exist in `generated.d.ts` and can be imported from there.

src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -0,0 +1,175 @@
...
+                required: false,
+                type: Number
+            }
+        },
+        data() {
+            return {
+                showFullBox: false,
+                fullBoxHeight: 0,
+                lineHeight: 0,
+                headerHeight: 0
```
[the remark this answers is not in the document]
hillalex: I think the current approach is correct

src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -0,0 +1,175 @@
...
+            };
+        },
+        computed: {
+            renderedBoxHeight(){
+                if (!this.warningsLengthy){
+                    return this.fullBoxHeight
+                } else {
+                    return this.showFullBox ? this.fullBoxHeight : this.maxBoxHeight - this.lineHeight
```
EmmaLRussell, line 105: I don't really understand why `this.lineHeight` is being subtracted from `this.maxBoxHeight` here. The maxBoxHeight is the full rendered height for a section with header (but no lines) plus the height of max number of lines, when truncating, isn't it? So subtracting a line here seems a bit artbitrary unless I've just misunderstood.

src/app/static/src/tests/components/warningAlert.test.ts
```diff
@@ -0,0 +1,128 @@
...
+        expect(hiddenWarning.find("p").text()).toBe("...");
+    });
+
+    it("renders ellipsis and show more/less buttons if many warnings", async () => {
+        const wrapper = createWrapper()
+        wrapper.setData({
+            fullBoxHeight: 232,
+            lineHeight: 24,
+            headerHeight: 56
+        })
```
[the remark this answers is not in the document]
EmmaLRussell: We worked around this sort of issue in comet tests by mocking an implementation of `clientWidth`, should work for height too: https://github.com/mrc-ide/comet/blob/c38838467e97d3a69c8059c6c03c3e574cb9ad8c/src/app/static/tests/unit/components/parameters/editPhases.test.ts#L31

hillalex:
Meant to add, shall we get confirmation of the aesthetic from someone in Jeff's team or UNAIDs? The header font looked a bit big to me given it's just generic text and the actual info is in the bullets beneath it.

src/app/static/src/app/components/WarningAlert.vue
Nicolas-Dolan: It meant 'including header' as it was a box containing a header and line plus any additional margin space in between. I'll just change it to `headerPlusLine` to be clearer.

src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -0,0 +1,175 @@
...
+</template>
+
+<script lang="ts">
+    import Vue from "vue";
+    import { AlertTriangleIcon } from "vue-feather-icons";
+    import i18next from "i18next";
+    import { mapStateProp } from "../utils";
+    import { RootState } from "../root";
+    import { Language } from "../store/translations/locales";
+    import { switches } from "../featureSwitches";
```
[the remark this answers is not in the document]
Nicolas-Dolan: I think my IDE most have auto-imported this when I was considering using a switch statement!

src/app/static/src/app/components/WarningAlert.vue
Nicolas-Dolan: Yeah, a couple of the functions iterate through the keys and it would say something like type string does match key of Warnings and I could not make it work.

src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -0,0 +1,175 @@
...
+                    this.fullBoxHeight = (this.$refs.warningBox as HTMLElement).clientHeight;
+                }
+            },
+            headerText(key){
+                const headers: { [key: string]: string } = {
+                    modelOptions: "warningsHeaderModelOptions",
+                    modelRun: "warningsHeaderModelRun",
+                    modelCalibrate: "warningsHeaderModelCalibrate"
+                }
+                return key in headers ? i18next.t(headers[key], { lng: this.currentLanguage }) : ""
```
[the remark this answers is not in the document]
Nicolas-Dolan: I'm not sure v-translate would work nicely if we have the triangle icon inside the element. I'll investigate.

src/app/static/src/app/components/WarningAlert.vue
Nicolas-Dolan: They would have to adjust their screen size to the extent that the number of warning lines increases or decreases due to wrapping. So it's an unlikely scenario but I would be remiss if I didn't mention it.

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -762,6 +772,8 @@ const fr: Partial<Translations> = {
         "avec lesquelles vous souhaitez partager ce projet. Appuyez sur Enter pour ajouter une autre adresse. Ces adresses e-mails doivent être " +
         "déjà enregistrées dans Naomi." +
         "</p>",
+    showLess: "Montrer moins",
+    showMore: "Montre plus",
```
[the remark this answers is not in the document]
Nicolas-Dolan: I did think that looked odd but I trusted Google translate over my own judgement on French conjugations! I'll change to Montrer.

src/app/static/src/app/components/Warning.vue
```diff
@@ -0,0 +1,173 @@
+<template>
+    <div v-if="warnings.length > 0">
```
Nicolas-Dolan, line 2: This isn't strictly necessary as the warning alert component filters out empty warnings. Just added for robustness and so that the component can stand alone theoretically.

src/app/static/src/app/components/Warning.vue
```diff
@@ -0,0 +1,173 @@
...
+        <h4 class="alert-heading pt-2">
+            <alert-triangle-icon size="1.5x" class="custom-class mr-1 mb-1"></alert-triangle-icon>
+            <span v-translate="headerText(origin)"></span>
+        </h4>
```
Nicolas-Dolan, line 6: One thing that wasn't clear after speaking with Jeff was whether the headers should be smaller. Easy to change if they need to be though.

src/app/static/src/app/components/Warning.vue
```diff
@@ -0,0 +1,173 @@
...
+        <!-- The outer styled divs are the visible window of warnings and inside them are the 
+        full window of warnings (the warningBox), which is always rendered as html but with 
+        its overflow hidden by the outer box. The size of the outer box dynamically adjusts 
+        depending on the size of the warningBox and whether the user has clicked show more or less. -->
+        <div :style="{ overflowY: 'hidden', height: `${this.renderedBoxHeight}px` }">
+            <ul class="mb-0" ref="warningBox" id="warningBox">
+                <li v-for="warning in warnings" :key="warning.text"><div :style="lineStyling">{{ warning.text }}</div></li>
```
Nicolas-Dolan, line 13: I've read mixed messages about whether it is ok to put a `div` inside a `li` but I needed to do this to prevent the bullets from disappearing after applying the style modifications.

src/app/static/src/app/components/Warning.vue
```diff
@@ -0,0 +1,173 @@
...
+
+    interface Computed  {
+        currentLanguage: Language;
+        renderedBoxHeight: number;
+        warningsLengthy: boolean;
+        buttonText: string;
+        maxBoxHeight: number;
+        lineStyling: {
+            height?: string
+        };
```
Nicolas-Dolan, line 59: This works for ts oddly enough but `Dict<string>` does not.

src/app/static/src/app/components/Warning.vue
```diff
@@ -0,0 +1,173 @@
...
+                let counter = 0
+                const id = setInterval(() => {
+                    counter++
+                    if (this.$refs.warningBox){
+                        this.lineHeight = (this.$refs.line as HTMLElement).clientHeight;
+                        this.fullBoxHeight = (this.$refs.warningBox as HTMLElement).clientHeight;
+                        clearInterval(id)
+                    } else if (counter > 20){
+                        clearInterval(id)
+                    }
```
Nicolas-Dolan, line 146: Perhaps unnecessary, but I added a counter to prevent this setInterval persisting.

src/app/static/src/tests/components/warning.test.ts
```diff
@@ -0,0 +1,153 @@
...
+        internal("line", l)
+        internal("warningBox", wb)
+    }
+
+    it("empty warnings are not rendered and setIntervals are cleared", async (done) => {
+        const wrapper = createWrapper({origin: "anything", warnings: []})
+        const clearIntervalSpy = jest.spyOn(window, "clearInterval");
+        expect(wrapper.find("div").exists()).toBe(false);
+        setTimeout(() => {
+            expect(clearIntervalSpy.mock.calls[0][0]).toBeTruthy();
```
Nicolas-Dolan, line 55: Does anyone have ideas on how to test the clearInterval counter?

Nicolas-Dolan:
This has now been adapted in light of Jeff's comments:
- All origin headers with warnings are shown at once with their warnings truncated
- Each set of warnings has its own show more/less button
- Each set of warnings will show up to two lines of warnings before truncating
- If truncated, each warning gets one line with ellipsis at end of line (rather than on line below)
- If only one warning but it is more than two lines, two lines of that warning will be shown and rest is truncated

To test manually, change hintr version to mrc-2664 to get the hintr warnings or mrc-2680 to get the dummy warnings.

Could also add these below the existing warning component in the stepper to test various configurations.

```
<warning-alert :warnings="{ 
            modelOptions: [{ text: 'a warning that is really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really long', locations: []}, { text: 'another warning', locations: []}, { text: 'third warning', locations: []}, { text: 'final warning', locations: []}],
            modelRun: [{ text: 'model run warning', locations: []}],
            modelCalibrate: [{ text: 'calibrate warning', locations: []}, { text: 'another warning', locations: []}, { text: 'third warning', locations: []}, { text: 'final warning', locations: []}]
            }"></warning-alert>
        <warning-alert :warnings="{ 
            modelOptions: [{ text: 'a warning that is really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really long', locations: []}],
            modelRun: [],
            modelCalibrate: []
            }"></warning-alert>
        <warning-alert :warnings="{ 
            modelOptions: [],
            modelRun: [],
            modelCalibrate: []
            }"></warning-alert>
```

src/app/static/src/app/components/Warning.vue
EmmaLRussell: Sorry, I'm confused about the need for this `setInterval` - what is it for? It looks like it's waiting for `$refs.warningBox` to exist, but shouldn't that always exist (as long as you have any warnings in the prop)?

Do you just need to make this async and add `await Vue.nextTick()` to let the DOM catch up following the toggle of `showFullBox`? I tried it that way locally and everything seemed as if it worked to me.

src/app/static/src/app/components/Warning.vue
EmmaLRussell: `Dict<string | number | undefined>` worked for me - number because of `maxLines` and I think it needs `undefined` because the keys are different depending on a conditional, though that seems a bit odd to me.

EmmaLRussell (commented):
This is looking very good. The webkit line clamp stuff is a neat way to achieve that use case. I'm hoping we can lose that `setInterval` though.

EmmaLRussell (commented):
Looks good. Probably don't need the `setTimeout`s in the tests anymore? e.g. try
```
it("renders show more/less buttons if warnings are lengthy", async () => {
        const wrapper = createWrapper(mockPropsModelRun)
        setDivHeights(wrapper, 24, 96)
        await Vue.nextTick();

        expectTranslated(wrapper.find("button"),
            "Show more",
            "Montrer plus",
            "Mostre mais",
            store);

        wrapper.find("button").trigger("click")

        expectTranslated(wrapper.find("button"),
            "Show less",
            "Montrer moins",
            "Mostre menos",
            store);
    });
``

**Review by EmmaLRussell — APPROVED** [review]

**Comment by Nicolas-Dolan** [comment]
> The instruction on testing does not work. I hope I am not missing anything. I tried both branches to get warnings from hintr and dummy then ran the app but no error was displayed throughout.

Did you run `./scripts/run-development-dependencies.sh` after changing the hintr version and saving the file? I think the first warning on mrc-2664 appears once model run has been run.

**Comment by LekanAnni** [comment]
> > The instruction on testing does not work. I hope I am not missing anything. I tried both branches to get warnings from hintr and dummy then ran the app but no error was displayed throughout.
> 
> Did you run `./scripts/run-development-dependencies.sh` after changing the hintr version and saving the file? I think the first warning on [mrc-2664](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2664) appears once model run has been run.

Thanks. I needed to re-run dependencies.sh my fault.

File: src/app/static/src/app/components/Warning.vue [hunk 20]
```diff
@@ -0,0 +1,164 @@
...
+                            "-webkit-box-orient": "vertical"
+                        }
+                    } else {
+                        return {
+                            height: `${this.lineHeight}px`,
+                            whiteSpace: "nowrap",
+                            overflow: "hidden",
+                            textOverflow: "ellipsis"
+                        }
+                    }
```
LekanAnni, line 120: I reckon the second if statement can be separated from the lineStyling().

LekanAnni (changes requested):
Looks good just a question and potential improvement.

- Warnings are not translated but headers. Can we ensure warnings are translated.
- To reproduce:
   run model, after warnings are displayed,  change language.

EmmaLRussell:
> Warnings are not translated but headers. Can we ensure warnings are translated.

That's a good point, but can we do that in another ticket please? It's outside the scope of the component

Nicolas-Dolan:
> > Warnings are not translated but headers. Can we ensure warnings are translated.
> 
> That's a good point, but can we do that in another ticket please? It's outside the scope of the component ticket. @Nicolas-Dolan could you maybe add that but possibly not to the sprint, as we have a lot of points to get through this sprint already. I don't think tickets related to changing language when user has progress far enough through a project to see warnings are likely to be high priority as won't come up much in practice for real users.

I don't think hintr returns translated warnings, although I could be wrong. The only reason the headers are translated is because they are generated in the frontend. I'll create a ticket for this.

src/app/static/src/app/components/Warning.vue
Nicolas-Dolan: Not quite sure what you mean. Could you elaborate?

LekanAnni: You can separate the highlighted snippet and call it from inside lineStyling() instead of having all the codes inside lineStyling()

Nicolas-Dolan: The code is very specific to this purpose and is unlikely to be reusable elsewhere so I think I'll keep it all together.

LekanAnni (approved):

PR #579, opened 2021-10-19
mrc-2688 accessibility improvements

hillalex:
## Description
* Fixes contrast in the header and the stepper 
* Adds some missing input labels
* Adds html lang attribute
 
There are still a lot of accessibility failures as you progress through the steps, in particular all uses of TreeSelect result in inputs with no label. And the tab/key navigation is all over the place. Is our aim that the app be usable by a screenreader? 

## Type of version change
Patch

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
-# hint 1.56.4
+# hint 1.57.1
 
 * accessibility fixes
+
+# hint 1.57.0
+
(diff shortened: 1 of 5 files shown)


mwoodbri (approved):
Looks good. A really significant reduction in warnings! Some low-hanging fruit remains on login page but already more than enough going on in this PR.

Could have used a mixin for setting lang attribute but not really justified given it's only a couple of pages.

PR #580, opened 2021-10-20
mrc-2618 Do not poll for ADR dataset updates when using a release

LekanAnni:
When a user imports fromADR data using a release, we should not need to poll for updates, since the data in a release will not change after it has been created.

* I added statement to stops polling when import is not from release dataset 
* I added stop and start polling to import dataset method, so, when users toggle between latest and release dataset, polling will work as expected.

## Type of version change
_Delete as appropriate_

Minor

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,11 @@
-# hint 1.59.0
+
+# hint 1.60.0
 
 * Do not poll for ADR dataset updates when using a release
+
+# hint 1.59.0
+
+* Do not auto-progress to calibrate step if there are model fit warnings
 
 # hint 1.58.0
 
--- a/src/app/static/src/app/components/Stepper.vue
+++ b/src/app/static/src/app/components/Stepper.vue
@@ -61,7 +61,7 @@
     import ModelOptions from "./modelOptions/ModelOptions.vue";
     import VersionStatus from "./projects/VersionStatus.vue";
     import {mapGettersByNames, mapStateProp, mapStateProps} from "../utils";
-    import {Project} from "../types";
+    import {Project, StepWarnings} from "../types";
     import {ProjectsState} from "../store/projects/projects";
     import {RootState} from "../root";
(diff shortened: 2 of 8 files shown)


src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -322,6 +322,8 @@
             importProgram: mapActionByName("surveyAndProgram", "importProgram"),
             importANC: mapActionByName("surveyAndProgram", "importANC"),
             async importDataset() {
+                this.stopPolling();
```
hillalex, line 325: I don't understand what this stop/start combination is achieving? To me it looks like polling should be started here conditional on there not being a release selected, and similarly the polling that is started on mount should be conditional on there not being a release selected.

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -982,6 +990,14 @@ describe("select dataset", () => {
         expect(spy.mock.calls[0][0]).toBe(pollingId);
     });
 
+    it("stops polling dataset metadata on newDatasetRelease", () => {
+        const rendered = shallowMount(SelectDataset, {store: getStore()});
+        const pollingId = (rendered.vm as any).pollingId;
+        const spy = jest.spyOn(window, "clearInterval");
+        rendered.setData({newDatasetRelease: {"test": "DATA"} as any})
```
hillalex, line 997: This doesn't reflect the way in which the widget is actually used - after choosing a release the user selects "import", at which point you start polling for changes again. In general tests should avoid using the `setData` method and use user actions (e.g.  `input.setValue("v1")`, `button.trigger("click")` etc) and replicate actual user behaviour

src/app/static/src/app/components/adr/SelectDataset.vue
carbowizzy: This ensures polling does not continue when switching from Latest to release dataset.

PR #581, opened 2021-10-21
mrc-2671 Include WarningAlert component in Stepper

EmmaLRussell:
## Description

This branch adds a dummy WarningAlert component (not yet merged with mrc-2668 branch which implements full WarningAlert) into the Stepper component, and provides it with the warnings for the active (current) step from the root state `warnings` getter. 

The dummy WarningAlert component shows a simple view of warnings messages. It renders nothing if there are no warnings for the current step. 

This is manually testable with the dummy warnings from mrc-ide/hintr#30 - run dev deps against this hintr branch (mrc-2680), which will return warnings for modelOptions, modelRun and modelCalibrate results as follows: 
- modelOptions: "You have chosen to fit model with estimating neighbouring ART attendance.." etc. This message should be visible in Model Options step after validation
- modelRun: "Zero population in put for 8 population groups.." etc. This message should be visible in Fit model step after model fit completes (auto-advance to model calibrate is already prevented by mrc-2672)
- modelCalibrate:
  - "ART coverage greater than 100% for 10 age groups" This message should be visible in Calibrate model step, after calibration completes
  -  "Prevalence greater than 40%" This message should be visible in Calibrate model step, after calibration completes, and in Review outputs step
  - "ART coverage greater than 100%" This message should be visible in Review outputs step

There are no tests for the dummy WarningAlert component as this will shortly be entirely replaced by mrc-2668

One possible issue - user now does not have to scroll up to continue to next step after model options validate, so may well not see warnings on this step. Perhaps WarningAlert should be shown at bottom of this step only?

This branch should be safe to merge with master while no warnings are being returned from master hintr. 

## Type of version change
_Delete as appropriate_

Minor

## Checklist

### DIFF
--- a/src/app/static/src/tests/root/getters.test.ts
+++ b/src/app/static/src/tests/root/getters.test.ts
@@ -74,4 +74,25 @@ describe(`root getters`, () => {
         expect(result).toEqual(modelOptionWarnings)
     })
 
+    it(`can get empty warnings when all warnings properties are undefined in state`, () => {
+        // We should be able to remove this test when warnings are fully implemented in hintr, as warnings
+        // array will always be provided. Can also remove defensive code for this case in the getter.
+        const rootState = mockRootState({
+            modelOptions: mockModelOptionsState({warnings: undefined}),
+            modelRun: mockModelRunState({warnings: undefined}),
+            modelCalibrate: mockModelCalibrateState({warnings: undefined})
+        });
+        const warnings = getters.warnings(rootState, null, rootState, null);
+        const expected = {
+            modelOptions: [],
+            modelRun: [],
+            modelCalibrate: []
+        };
+        expect(warnings("modelOptions")).toStrictEqual(expected);
+        expect(warnings("fitModel")).toStrictEqual(expected);
+        expect(warnings("calibrateModel")).toStrictEqual(expected);
+        expect(warnings("reviewOutput")).toStrictEqual(expected);
+        expect(warnings("downloadResults")).toStrictEqual(expected);
+    });
+
 })


src/app/static/src/tests/components/stepper.test.ts
```diff
@@ -759,4 +760,48 @@ describe("Stepper component", () => {
...
+            {
+                valid: true,
+                warnings: [{
+                    text: "Model Options warning",
+                    locations: ["model_options", "model_fit"]
+                }]
+            }
+        );
+
+        const warnings = wrapper.find(WarningAlert).props("warnings");
```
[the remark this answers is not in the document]
mwoodbri: This is fine as is... could do following but not pretty:
```suggestion
        const warnings = JSON.parse(JSON.stringify(wrapper.find(WarningAlert).props("warnings")));
```

src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -0,0 +1,38 @@
...
+        </div>
+    </div>
+</template>
+
+<script lang="ts">
+    import Vue from "vue";
+    import {StepWarnings} from "../types";
+
+    interface Props {
+        warnings: StepWarnings
```
mwoodbri, line 29: Could argue that this component (and successor) just needs a `Dict<String[]>` rather than locations etc. But re-using type is nice!

mwoodbri (approved):
This is all fitting together very nicely...

src/app/static/src/app/store/root/getters.ts
```diff
@@ -23,9 +23,11 @@ export const getters: RootGetters & GetterTree<RootState, RootState> = {
 
     warnings: (state: RootState) => (stepName: string): StepWarnings => {
         const filterWarnings = (warnings: Warning[], stepLocation: string) =>
-            (warnings || []).filter(warning => warning.locations.some(location => location === stepLocation))
+            stepLocation ?
+                (warnings || []).filter(warning => warning.locations.some(location => location === stepLocation)) :
```
carbowizzy, line 27: I am still unsure why `(warnings || [])` is needed 

Also this `stepLocation ?`

PR #586, opened 2021-10-26
mrc-2712 Improve appearance of Input time series x axis, remove Quarter column

EmmaLRussell:
## Description

This branch:
- removes Quarter column from Input Time Series for ART, as it is not required since quarter is obvious from Time period values e.g. 2019 Q4
- Shows only first and last time period values on time series plot x axes. This is because quarterly time period values are more verbose than years, and there will typically be more of them.

NB We only have Q4 data being shown for test Antarctica dataset

## Type of version change
_Delete as appropriate_

None

## Checklist


Nicolas-Dolan:
> > Update feedback is a general issue, should probably show the spinner while re-rendering the plot. Could get Plotly component to emit events to trigger that I think.
> 
> Likewise, not a huge issue - and likely not event a common use case - but wondering just how slow it would be on a slower machine.

For me using Chrome, this wasn't so much of a problem as the select input freezes momentarily while it does the processing and then completes the selection once the charts were re-rendered, so there was never an option to select something else in the meantime. I could tell something was happening by the noises my laptop suddenly start making ;). You could add some loading icon while it computes in the background but I think that could be very tricky to implement and might not be worth the effort.

Nicolas-Dolan (approved):
Looks good to me and charts render nicely.

carbowizzy (approved):
LGTM

EmmaLRussell:
I will address lack of feedback during data update in a separate ticket, as it doesn't seem to be a showstopper, and is not trivial to fix: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2730

PR #590, opened 2021-11-03
Mrc-2567-hotfixes  add default placeholder to some error report data

LekanAnni:
- Add default value to 
description
step to reproduce and
country

- Make properties of Error report data object mandatory.

- Update tests 

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/tests/root/actions.test.ts
+++ b/src/app/static/src/tests/root/actions.test.ts
@@ -390,33 +390,6 @@ describe("root actions", () => {
         const expectedError = {error: "OTHER_ERROR", detail: "TestError"};
         expect(commit.mock.calls[0][0]).toEqual({payload: expectedError, type: RootMutation.ErrorReportError});
         expect(dispatch.mock.calls.length).toBe(0)
-        expect(mockAxios.history.post.length).toEqual(1)
-        expect(mockAxios.history.post[0].url).toEqual(url)
-
-        const expected = {
-            email: "some.user@example.com",
-            country: "no associated country",
-            projectName: "no associated project",
-            timeStamp: new Date(),
-            jobId: "no associated jobId",
-            description: "",
-            section: "",
-            stepsToReproduce: "",
-            errors: []
-        };
-
-        const data = JSON.parse(mockAxios.history.post[0].data) as ErrorReport
-
-        expect(data.email).toStrictEqual(expected.email)
-        expect(data.country).toStrictEqual(expected.country)
-        expect(data.projectName).toBe(expected.projectName)
-        expect(data.jobId).toStrictEqual(expected.jobId)
-        expect(data.browserAgent).toContain("Mozilla")
-        expect(new Date(data.timeStamp).getDate()).toBe(expected.timeStamp.getDate());
-        expect(data.description).toStrictEqual(expected.description)
-        expect(data.section).toStrictEqual(expected.section)
-        expect(data.stepsToReproduce).toStrictEqual(expected.stepsToReproduce)
-        expect(data.errors).toStrictEqual(expected.errors)
     });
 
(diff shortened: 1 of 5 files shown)


src/app/static/src/tests/root/actions.test.ts
```diff
@@ -213,10 +212,10 @@ describe("root actions", () => {
 
     it("posts error report to teams", async () => {
 
-        const url = "error-report"
+        const url = `error-report`
```
hillalex, line 215: why this change to backticks?

src/app/static/src/tests/root/actions.test.ts
```diff
@@ -238,31 +237,32 @@ describe("root actions", () => {
             })
         });
 
-        const err = mockError("err")
         const getters = {
-            errors: [err]
+            errors: []
```
hillalex, line 241: why remove the error here?

src/app/static/src/tests/root/actions.test.ts
```diff
@@ -319,12 +320,14 @@ describe("root actions", () => {
...
         const expectedError = {error: "OTHER_ERROR", detail: "TestError"};
         expect(commit.mock.calls[0][0]).toEqual({payload: expectedError, type: RootMutation.ErrorReportError});
+        expect(dispatch.mock.calls.length).toBe(0)
         expect(mockAxios.history.post.length).toEqual(1)
         expect(mockAxios.history.post[0].url).toEqual(url)
 
         const expected = {
             email: "some.user@example.com",
             country: "",
+            projectName: "",
```
hillalex, line 330: Since this object is called "expected" this should be "no associated project", country and jobId should be set similarly and then below assert `expect(data.country).toBe(expected.country)` etc.

hillalex (commented):
Looks fine, just a bit of tidying up to do in the tests

src/app/static/src/tests/root/actions.test.ts
```diff
@@ -319,14 +321,16 @@ describe("root actions", () => {
...
         const expectedError = {error: "OTHER_ERROR", detail: "TestError"};
         expect(commit.mock.calls[0][0]).toEqual({payload: expectedError, type: RootMutation.ErrorReportError});
+        expect(dispatch.mock.calls.length).toBe(0)
         expect(mockAxios.history.post.length).toEqual(1)
         expect(mockAxios.history.post[0].url).toEqual(url)
 
         const expected = {
             email: "some.user@example.com",
-            country: "",
+            country: "no associated country",
```
EmmaLRussell, line 330: So this test is testing both that the behaviour when the POST fails is as expected, and that the error report defaults are populated? I'd prefer it if that was reflected in the test name or tested in separate tests.

EmmaLRussell (commented):
Doesn't look like the code matches the PR description, but it does match what I think needed to happen i.e. default values for country, projectName and jobId (not description and steps to reproduce, which must always be populated from the form)? Also updating `project` to `projectName`?

src/app/static/src/tests/root/actions.test.ts
hillalex: +1 for this being 2 tests please!
* "can return error when error report fails"
* "sends default values when country, project or jobId are missing"

EmmaLRussell (commented):
Test failure on buildkite, and it looks like Alex has voted for splitting up that test rather than just a rename (I'd agree that's better approach).

EmmaLRussell (approved):
Nice. :+1:
I'd say you don't actually need lines 393-419 any more since that's already tested in the new test so the error test can just be that it commits the error correctly - doesn't hurt for it to be there, but would make tests less verbose to remove it.

hillalex (approved):

PR #591, opened 2021-11-08
mrc-2735 Use default language in root state if undefined in localStorage

EmmaLRussell:
## Description

See ticket for description of bug on prod - language menu not available, and error logged in console, when loading a project. 

I think this is because language was not previously being stored in local storage, so when a user who has used HINT previously loads from local storage, that language is undefined, and the code was using that undefined language as long as the previous state existed at all. 

To reproduce the error locally: load the application running on master branch, edit the local storage value for hintAppState to remove the 'language' key, and reload the page. You should see the problems described in the bug. Try again in this branch, and the language menu should not disappear. Selected language should still persist on reload.

One thing that confuses me is that the hintAppState is now unique per version, so I'm a bit surprised than an old version without language could be reloaded at all. Any ideas?

## Type of version change
_Delete as appropriate_

Patches 

## Checklist

### DIFF
--- a/src/app/static/src/app/root.ts
+++ b/src/app/static/src/app/root.ts
@@ -164,7 +164,7 @@ export const emptyState = (): RootState => {
 const existingState = localStorageManager.getState();
 
 export const storeOptions: StoreOptions<RootState> = {
-    state: {...emptyState(), ...existingState && {language: existingState.language}},
+    state: {...emptyState(), ...existingState},
     modules: {
         adr,
         genericChart,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.64.3
+
+* Fix language bug on reload
+
 # hint 1.64.2
 
 * add default placeholder to some error report data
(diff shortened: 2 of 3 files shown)


src/app/static/src/app/root.ts
```diff
@@ -164,7 +164,7 @@ export const emptyState = (): RootState => {
 const existingState = localStorageManager.getState();
 
 export const storeOptions: StoreOptions<RootState> = {
-    state: {...emptyState(), ...existingState && {language: existingState.language}},
```
LekanAnni, line 167: I have always wondered why language was specifically set here when it's already part of ...existingState objects.

hillalex:
> One thing that confuses me is that the hintAppState is now unique per version, so I'm a bit surprised than an old version without language could be reloaded at all. Any ideas?

Only major version changes prevent re-loading. This was likely a minor version difference.

> To reproduce the error locally: load the application running on master branch, edit the local storage value for hintAppState to remove the 'language' key, and reload the page.

This is a little different from the bug described but I verified working as expected by downloading a project, removing the language, disabling checksum matching in the app and re-uploading the project.

hillalex (approved):

EmmaLRussell:
> > One thing that confuses me is that the hintAppState is now unique per version, so I'm a bit surprised than an old version without language could be reloaded at all. Any ideas?
> 
> Only major version changes prevent re-loading. This was likely a minor version difference.
> 
Ah of course!

PR #594, opened 2021-11-09
mrc-2752 Expect errors from calibrate endpoint because calibrate id is not valid

EmmaLRussell:
## Description

I think this build error was due to the dummy calibrate endpoint having been relaxed about accepting any calibrate id. The tests pass a fake calibrate id, and this now causes an error, as it does when fake id is passed for calibrate result. I have updated the tests to be consistent with the calibrate result tests .i.e. pass fake id, and expect an error. It would be good to test the success case if and when we get browser testing added to HINT. 

## Type of version change
_Delete as appropriate_

None

## Checklist


LekanAnni (approved):
LGTM

PR #595, opened 2021-11-09
mrc-2736 data exploration mode prototyping

hillalex:
Proof of concept for data exploration mode. To see working, run the app locally and navigate to `/explore`. Note there will be console errors if you navigate to step 2 without adding data.

Comments welcome.

I've added a few tickets and tidied up the existing ones:

House-keeping:
* Rename `baseline` and `surveyAndProgram` modules and components - I see that the content of these have diverged from the names and I would like to tidy this up before re-using them
* Move error report related mutations into own class, analogous to language mutations (or possibly into the existing `errors` module?)

Backend:
* Add new route and freemarker template. Include logic to store mode against session and clear version id if moving between modes. Include `var mode` as a js variable in the same way as we do current user and modify local storage manager so that state is not persisted between apps: https://github.com/mrc-ide/hint/pull/595/files#diff-8a4dde40541e3038e68c635e6c52b363a07a7447c06691ee04f8b7cd797e008bR51

Front-end:
* Add new app with shared state and modules
* Show `ResetConfirmation` components conditionally based on flag in state
* New navbar component for data exploration mode
* New error report component for data exploration mode as the existing one is tightly coupled to projects and stepper modules
* Add data exploration component itself, including back/next buttons (validation rule before allowing forward navigation may be different from existing app?)

### DIFF
--- a/src/app/static/src/app/components/adr/SelectDataset.vue
+++ b/src/app/static/src/app/components/adr/SelectDataset.vue
@@ -83,7 +83,7 @@
             </template>
         </modal>
         <reset-confirmation
-            v-if="showConfirmation"
+            v-if="!dataExplorationMode && showConfirmation"
             :continue-editing="continueEditing"
             :cancel-editing="cancelEditing"
             :open="showConfirmation">
@@ -99,7 +99,7 @@
         mapActionByName,
         mapMutationByName,
         mapStateProp,
-        mapGetterByName,
+        mapGetterByName, mapStatePropByName,
     } from "../../utils";
     import {RootState} from "../../root";
     import Modal from "../Modal.vue";
@@ -144,6 +144,7 @@
     }
 
     interface Computed {
+        dataExplorationMode: boolean;
         datasets: any[];
         selectedRelease: Release | null;
         releaseName: string | null;
@@ -205,10 +206,13 @@
         },
         directives: {tooltip: VTooltip},
         computed: {
-            editsRequireConfirmation: mapGetterByName(
-                "stepper",
-                "editsRequireConfirmation"
-            ),
+            dataExplorationMode: mapStatePropByName(null, "dataExplorationMode"),
+            editsRequireConfirmation() {
+                if (this.dataExplorationMode){
+                    return false
+                }
+                return this.$store.getters["stepper/editsRequireConfirmation"]
+            },
             hasShapeFile: mapStateProp<BaselineState, boolean>(
                 "baseline",
                 (state: BaselineState) => !!state.shape
--- a/src/app/static/src/app/localStorageManager.ts
+++ b/src/app/static/src/app/localStorageManager.ts
@@ -30,6 +30,7 @@ export const serialiseState = (rootState: RootState): Partial<RootState> => {
 };
 
 declare const currentUser: string;
+declare const currentMode: string;
 
 export class LocalStorageManager {
 
@@ -46,6 +47,10 @@ export class LocalStorageManager {
         if (currentUser != window.localStorage.getItem("user")) {
             localStorage.clear();
             localStorage.setItem("user", currentUser);
+        }
+        if (currentMode != window.localStorage.getItem("mode")) {
+            localStorage.clear();
+            localStorage.setItem("mode", currentMode);
         }
         const item = window.localStorage.getItem(appStateKey);
         if (item) {
--- a/src/app/static/src/app/store/adr/actions.ts
+++ b/src/app/static/src/app/store/adr/actions.ts
@@ -7,16 +7,17 @@ import {ADRMutation} from "./mutations";
 import {datasetFromMetadata} from "../../utils";
 import {Organization, Release} from "../../types";
 import {BaselineMutation} from "../baseline/mutations";
+import {DataExplorationState} from "../dataExploration/dataExploration";
 
 export interface ADRActions {
-    fetchKey: (store: ActionContext<ADRState, RootState>) => void;
-    saveKey: (store: ActionContext<ADRState, RootState>, key: string) => void;
-    deleteKey: (store: ActionContext<ADRState, RootState>) => void;
-    getDatasets: (store: ActionContext<ADRState, RootState>) => void;
-    getReleases: (store: ActionContext<ADRState, RootState>, id: string) => void;
-    getDataset: (store: ActionContext<ADRState, RootState>, payload: GetDatasetPayload) => void;
-    getSchemas: (store: ActionContext<ADRState, RootState>) => void;
-    getUserCanUpload: (store: ActionContext<ADRState, RootState>) => void;
+    fetchKey: (store: ActionContext<ADRState, DataExplorationState>) => void;
+    saveKey: (store: ActionContext<ADRState, DataExplorationState>, key: string) => void;
+    deleteKey: (store: ActionContext<ADRState, DataExplorationState>) => void;
+    getDatasets: (store: ActionContext<ADRState, DataExplorationState>) => void;
+    getReleases: (store: ActionContext<ADRState, DataExplorationState>, id: string) => void;
+    getDataset: (store: ActionContext<ADRState, DataExplorationState>, payload: GetDatasetPayload) => void;
+    getSchemas: (store: ActionContext<ADRState, DataExplorationState>) => void;
+    getUserCanUpload: (store: ActionContext<ADRState, DataExplorationState>) => void;
 }
 
 export interface GetDatasetPayload {
@@ -24,7 +25,7 @@ export interface GetDatasetPayload {
     release?: Release
 }
 
-export const actions: ActionTree<ADRState, RootState> & ADRActions = {
+export const actions: ActionTree<ADRState, DataExplorationState> & ADRActions = {
     async fetchKey(context) {
         await api<ADRMutation, ADRMutation>(context)
             .ignoreErrors()
--- a/src/app/static/src/app/store/dataExploration/actions.ts
+++ b/src/app/static/src/app/store/dataExploration/actions.ts
@@ -1,0 +1,61 @@
+import {ActionContext, ActionTree} from "vuex";
+import {RootState} from "../../root";
+import {LanguageActions} from "../language/language";
+import {changeLanguage} from "../language/actions";
+import {api} from "../../apiService";
+import {VersionInfo} from "../../generated";
+import {currentHintVersion} from "../../hintVersion";
+import {DataExplorationState} from "./dataExploration";
+import {ErrorReportManualDetails} from "../../types";
+import {LanguageMutation} from "../language/mutations";
+import {DataExplorationMutation} from "./mutations";
+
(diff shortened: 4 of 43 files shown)


src/app/static/src/app/store/root/mutations.ts
```diff
@@ -150,10 +150,6 @@ export const mutations: MutationTree<RootState> = {
         Object.assign(state.downloadResults, initialDownloadResultsState());
     },
 
-    [RootMutation.SetUpdatingLanguage](state: RootState, action: PayloadWithType<boolean>) {
```
hillalex, line 153: moved to shared language mutations

src/app/static/src/app/store/root/getters.ts
```diff
@@ -68,22 +69,3 @@ export const getters: RootGetters & GetterTree<RootState, RootState> = {
         }
     }
 }
-
-export const extractErrors = (state: any) => {
```
hillalex, line 72: moved to shared utils

src/app/static/src/app/store/root/actions.ts
```diff
@@ -16,23 +18,6 @@ export interface RootActions extends LanguageActions<RootState> {
                           payload: ErrorReportManualDetails) => void;
 }
 
-export interface ErrorReportManualDetails {
```
hillalex, line 19: moved to shared types

src/app/static/src/app/store/plottingSelections/getters.ts
```diff
@@ -16,7 +16,8 @@ export const getters = {
                 return {}
         }
     },
-    calibratePlotDefaultSelections: (state: PlottingSelectionsState, getters: any, rootState: RootState): BarchartSelections => {
-        return rootState.modelCalibrate.calibratePlotResult!.plottingMetadata.barchart.defaults;
+    calibratePlotDefaultSelections: (state: PlottingSelectionsState, getters: any, rootState: DataExplorationState): BarchartSelections => {
+        // TODO find solution to this hack
```
hillalex, line 20: this is the only place where a module we need for both modes spans multiple steps - options would be to leave something like this in, I guess using the flag:

```
if (!rootState.dataExplorationMode) {
    return (rootState as RootState).modelCalibrate.calibratePlotResult!.plottingMetadata.barchart.defaults;
}
```

 or to split this module up, maybe move plotting selection getters into the modules where they apply (i.e. this would live in `modelCalibrate`, the above in `surveyAndProgram` etc)

src/app/static/src/app/store/password/password.ts
```diff
@@ -11,6 +11,7 @@ export interface PasswordState extends TranslatableState {
 
 export const initialPasswordState: PasswordState = {
     language: Language.en,
+    updatingLanguage: false,
```
hillalex, line 14: this now lives on `LanguageState` which `PasswordState` inherits from

src/app/static/src/app/store/dataExploration/mutations.ts
```diff
@@ -0,0 +1,25 @@
...
+import {DataExplorationState} from "./dataExploration";
+
+export enum DataExplorationMutation {
+    ErrorReportError = "ErrorReportError",
+    ErrorReportSuccess = "ErrorReportSuccess"
+}
+
+export const mutations: MutationTree<DataExplorationState> = {
+
+    [DataExplorationMutation.ErrorReportError](state: DataExplorationState, action: PayloadWithType<Error>) {
```
hillalex, line 14: These two mutations are duplicated from root mutations, so probably want pulling out into `errorMutations` as we have done with `languageMutations` below.

src/app/static/src/app/store/dataExploration/getters.ts
```diff
@@ -0,0 +1,34 @@
...
+import {extractErrors} from "../../utils";
+import {DataExplorationState} from "./dataExploration";
+
+interface DataExplorationGetters {
+    isGuest: Getter<DataExplorationState, DataExplorationState>
+    errors: Getter<DataExplorationState, DataExplorationState>
+}
+
+export const getters: DataExplorationGetters & GetterTree<DataExplorationState, DataExplorationState> = {
+    isGuest: (state: DataExplorationState) => {
```
hillalex, line 12: this is duplicated from root getters, but is so small as to probably not be worth pulling into it's own shared module

src/app/static/src/app/store/dataExploration/actions.ts
```diff
@@ -0,0 +1,61 @@
...
+import {DataExplorationMutation} from "./mutations";
+
+export interface DataExplorationActions extends LanguageActions<DataExplorationState> {
+    generateErrorReport: (store: ActionContext<DataExplorationState, DataExplorationState>,
+                          payload: ErrorReportManualDetails) => void;
+}
+
+export const actions: ActionTree<DataExplorationState, DataExplorationState> & DataExplorationActions = {
+
+    async changeLanguage(context, payload) {
```
hillalex, line 20: This is subtly different from the root action because the latter references the `modelCalibrate` module and this does not. Possibly there is some nice way to share code, maybe not worth it.

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -83,7 +83,7 @@
             </template>
         </modal>
         <reset-confirmation
-            v-if="showConfirmation"
+            v-if="!dataExplorationMode && showConfirmation"
```
hillalex, line 86: The `ResetConfirmation` component appears here, in `ManageFile` and in `FileUpload`. These are the only 3 places I think we need this flag.

src/app/static/src/app/store/dataExploration/dataExploration.ts
```diff
@@ -0,0 +1,85 @@
...
+import {Module, StoreOptions} from "vuex";
+import {TranslatableState} from "../../root";
+import {Error} from "../../generated";
+import {actions} from "./actions";
+import {mutations} from "./mutations";
+import {getters} from "./getters";
+
+declare const currentUser: string;
+
+export interface DataExplorationState extends TranslatableState {
```
hillalex, line 30: Might want to call this something else, `SharedState` or `BaseRootState` or something to reflect the fact that the main app `RootState` inherits from this.

src/app/static/src/app/localStorageManager.ts
```diff
@@ -47,6 +48,10 @@ export class LocalStorageManager {
             localStorage.clear();
             localStorage.setItem("user", currentUser);
         }
+        if (currentMode != window.localStorage.getItem("mode")) {
```
hillalex, line 51: Or perhaps better to modify all this logic to save/retrieve app state under a different key for the data exploration app

EmmaLRussell (commented):
This looks good to me! It's a bit fiddly, but not as bad as I thought it might be!

So the intention is that this ticket/branch is now just for prototyping, and all functionality will be implemented in further tickets? I've added the new tickets to the Epic, and updated the dependencies (maybe a bit approximate). No need to use an Epic branch for this of course since changes won't be visible to users of the main app.

One thing it doesn't do from the mockups is redirect to the login page if user is not authenticated (and then have login redirect to the requested explore page). Need another ticket for that?

We should also hide the ADR component when move on to the second step, but that's one for the implementation ticket.

src/app/static/src/app/store/plottingSelections/getters.ts
hillalex: Yes, I wasn't proposing more modules, but moving these methods into existing modules. Probably removing this module completely is not worth the faff but this getter actually only references state from the `modelCalibrate` module, not from the `plottingSelections` module at all! So it would make sense to just move it to be a `modelCalibrate` getter. The casting is prettty hack-y because it involves contradicting the type declaration.

hillalex: Similarly, although not needed for this change, the first getter in this class only references state from the `surveyAndProgram` module, so should probably be a getter on that module instead.

src/app/static/src/app/components/adr/SelectDataset.vue
hillalex: Except the component itself references the possibly non-existent modules `project` and `stepper` via `mapAction` and `mapState` helpers, so even mounting it throws an error.

src/app/static/src/app/localStorageManager.ts
hillalex: Yeah, agree that's better. And yes we can.

hillalex:
> So the intention is that this ticket/branch is now just for prototyping, and all functionality will be implemented in further tickets? I've added the new tickets to the Epic, and updated the dependencies (maybe a bit approximate). No need to use an Epic branch for this of course since changes won't be visible to users of the main app.

I think so yes, I renamed this ticket and all proper implemention is now covered by other tickets. 

> One thing it doesn't do from the mockups is redirect to the login page if user is not authenticated (and then have login redirect to the requested explore page). Need another ticket for that?

Oh yes, I didn't even think about that! Ok, added here: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2761

src/app/static/src/app/store/adr/actions.ts
```diff
@@ -7,24 +7,25 @@ import {ADRMutation} from "./mutations";
...
 export interface ADRActions {
-    fetchKey: (store: ActionContext<ADRState, RootState>) => void;
-    saveKey: (store: ActionContext<ADRState, RootState>, key: string) => void;
-    deleteKey: (store: ActionContext<ADRState, RootState>) => void;
-    getDatasets: (store: ActionContext<ADRState, RootState>) => void;
-    getReleases: (store: ActionContext<ADRState, RootState>, id: string) => void;
-    getDataset: (store: ActionContext<ADRState, RootState>, payload: GetDatasetPayload) => void;
-    getSchemas: (store: ActionContext<ADRState, RootState>) => void;
-    getUserCanUpload: (store: ActionContext<ADRState, RootState>) => void;
+    fetchKey: (store: ActionContext<ADRState, DataExplorationState>) => void;
```
LekanAnni, line 13: is there a reason why you change RootState to DataExplorationState considering RootState already extends DataExplorationState ?

LekanAnni (commented):
Looking great, just a question below.

src/app/static/src/app/store/adr/actions.ts
hillalex: Yes, these actions will be invoked by both apps, so sometimes the state will be of the type `DataExplorationState`, and sometimes it will be the superset `RootState` (but will only invoke properties on the subset included in `DataExplorationState`)

src/app/static/src/app/components/adr/SelectDataset.vue
hillalex: But I suppose the conditional logic I put in here (https://github.com/mrc-ide/hint/pull/595/files#diff-6f2399eff5f0ba4c4c5783235ba412ffc39c74db05af0349ca5d986e30237312R210) could live in the computed vars of `ResetConfirmation` itself

PR #596, opened 2021-11-09
mrc-2498 Switch calibrate plot feature on

EmmaLRussell:
## Description

Switch the calibrate plot feature on. Leaving the switch in place in case we need to roll it back. 

Calibrate plot should be visible in the Calibrate Model page, after calibration completes.

## Type of version change
_Delete as appropriate_

Minor

## Checklist


Nicolas-Dolan (approved):
Works well. Just need to update with master.

LekanAnni (approved):

PR #598, opened 2021-11-10
mrc-2758 Input Time Series jsonata - cope with null values

EmmaLRussell:
## Description

Input Time Series was failing to load for Nigeria ART plots 'ART proportion < 15' and 'ART adult sex ratio'.

This was because the jsonata producing the red line sections (when ratios of consecutive values exceed expected values) was not able to deal with null values. I've fixed this by adding checks for nulls - if either the value under consideration, or the adjacent value being compared, is null, the expression will return null itself.

To test: manually upload the Nigeria files attached to the[ ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2758), and view the Input Time Series. You should be able to load all plots for both ART and ANC, including 'ART proportion < 15' and 'ART adult sex ratio'. You'll see that the states Enugu and Niger have missing values compared to the other states - these correspond to the null values.

## Type of version change
_Delete as appropriate_
Patches

## Checklist


Nicolas-Dolan (approved):
Worked well for me. I wonder what the best way to represent missing values in the table is? Currently we include the row but leave the key value blank. That's probably the clearest way to do it but it's worth discussing.

LekanAnni (changes requested):
Looks good but new code is missing test.

EmmaLRussell:
> Looks good but new code is missing test.

There aren't any tests for the jsonata currently. Could potentially add some, but outside the scope of this ticket

EmmaLRussell:
Added a ticket to add jsonata validation test: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2760

LekanAnni (approved):

PR #602, opened 2021-11-10
mrc-2607 Update review output plots to use metadata from modelCalibrate response

LekanAnni:
## Description

## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/tests/components/header/fileMenu.test.ts
```diff
@@ -90,7 +90,7 @@ describe("File menu", () => {
         input.trigger("change");
     };
 
-    it("downloads file", (done) => {
```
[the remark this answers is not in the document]
hillalex: Without the `done` callback this test assertion is never executed: https://github.com/mrc-ide/hint/pull/602/files#diff-78dd5a914f8678755091c5fbd21a6a208c64133bb6670c85754f71aba8ee4a61L137

Why would the changes here make this test obsolete?

I appreciate this test is hard to debug but try adding these lines:
```
console.log(result)
console.log(expectedJson)
```
inside the `loadend` callback to see why the test is failing.

hillalex (changes requested):
See inline comment

src/app/static/src/tests/components/header/fileMenu.test.ts
```diff
@@ -7,7 +7,7 @@ import {
     mockError,
     mockFile,
     mockLoadState,
-    mockMetadataState,
+    mockMetadataState, mockModelCalibrateState,
```
Nicolas-Dolan, line 10:
```suggestion
    mockMetadataState, 
    mockModelCalibrateState,
```

Nicolas-Dolan (approved):
Works for me

mwoodbri:
@LekanAnni Could you sync this branch with master? Will make it easier to review - currently seeing changes from mrc-2762 which have already been merged.

mwoodbri (commented):
Generally looks good - though please see inline for some specific questions/suggestions.

src/app/static/src/app/store/modelOutput/modelOutput.ts
```diff
@@ -14,19 +14,19 @@ export interface ModelOutputState {
...
+        return rootState.modelCalibrate.result!.plottingMetadata.choropleth.indicators;
     },
-    bubblePlotFilters: (state: ModelOutputState, getters: any, rootState: RootState, rootGetters: any): Filter[] => {
+    bubblePlotFilters: (state: ModelOutputState, getters: any, rootState: RootState): Filter[] => {
         return outputPlotFilters(rootState);
     },
-    choroplethIndicators: (state: ModelOutputState, getters: any, rootState: RootState, rootGetters: any): ChoroplethIndicatorMetadata[] => {
-        return rootGetters['metadata/outputIndicatorsMetadata'];
+    choroplethIndicators: (state: ModelOutputState, getters: any, rootState: RootState): ChoroplethIndicatorMetadata[] => {
+        return rootState.modelCalibrate.result!.plottingMetadata.choropleth.indicators;
```
hillalex, line 29: Just wondering about these changes - is it right that we no longer use the plotting metadata at all  but just use the indicators that come back on the result object? I know that @EmmaLRussell recently made changes to the equivalent getters for the inputs choropleth which use both the plotting metdata and the indiviual endpoints' metadata: https://github.com/mrc-ide/hint/pull/617 

@EmmaLRussell are you in a position to confirm the correct behaviour here?

EmmaLRussell: We do still need the indicators from the plotting metadata endpoint - it contains the full metadata we require to use the indicator (e.g. what its data columns, what value format to use) - what's returned by the data endpoint really just indicates which filters should be included.

hillalex: Actually looks like the full metadata is returned with the calibration result, so this change is correct. But this means that the `outputIndicatorsMetadata` getter in the `metadata` module is not longer used anywhere and can be removed.

hillalex (changes requested):
Can we remove the obsolete getter? And then news and version can be incremented (and this should really be a patch rather than a minor increment)

hillalex (approved):

PR #603, opened 2021-11-10
mrc-2753 Add /explore endpoint

EmmaLRussell:
## Description

This branch adds an `/explore` endpoint which returns a placeholder template.

The `index` and `explore` methods in the `HomeController` now set mode in the session, "naomi" or "explore", and when mode changes, the Session clear the version id so that the underlying project version state is never persisted between modes.

Note that one of the routes for the `index` method is `/accessibility` so this means loading then refreshing the accessibility page from data exploration would reset the mode to 'naomi'. We'll need some way of dealing with this when the accessibility page is added for data exploration - perhaps a separate route.

To test the endpoint: browse to `/explore` and see a page with placeholder content. The title in the browser tab should be 'Naomi Data Exploration'
To test the version clearing behaviour: while logged in, open any project in the normal naomi mode. Browse to `/explore` then back to `/` - the previously open project should not be retained, but the app should show the Projects page

## Type of version change
_Delete as appropriate_

Minor

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.66.0
+
+* Add /explore endpoint
+
 # hint 1.65.0
 
 * Show Input Time Series plots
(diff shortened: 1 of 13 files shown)


hillalex (approved):

LekanAnni (approved):
LGTM

PR #606, opened 2021-11-11
mrc-2762 move getter from plottingSelections to modelCalibrate

hillalex:
## Description

This is a small change to the `plottingSelections` module so that it no longer depends on the `modelCalibrate` module and can be re-used in the data exploration app (which won't include the latter module)

## Type of version change
Patch

## Checklist


LekanAnni (approved):
LGTM

Nicolas-Dolan (approved):

PR #608, opened 2021-11-11
mrc-2756 basic Vuex app for data exploration mode

hillalex:
## Description

This adds the Vuex app and required modules for data exploration mode. 

Note that there is a separate ticket for adding the component itself, which will involve making some minor changes to the included components to prevent errors when the `ResetConfirmation` component tries to reference modules that don't exist (see coversation [here](https://github.com/mrc-ide/hint/pull/595/files#diff-6f2399eff5f0ba4c4c5783235ba412ffc39c74db05af0349ca5d986e30237312) The loading spinner component is used as a placeholder here.

This does the most minimal set of changes required to get an app running, so it adds the language mutations which are called from the base vue instance, but does not add any further actions, mutations or getters that will be needed ultimately for some of the nav components (language toggling, error reporting, detecting whether user is a guest) to avoid the changeset being too big to review effectively.

~~Contains commits from https://github.com/mrc-ide/hint/pull/606/files which should be merged first.~~

## Type of version change
Minor

## Checklist


src/app/static/src/app/store/dataExploration/dataExploration.ts
```diff
@@ -0,0 +1,81 @@
...
+        surveyAndProgram: initialSurveyAndProgramState(),
+        plottingSelections: initialPlottingSelectionsState(),
+        errors: initialErrorsState(),
+        currentUser: currentUser,
+        dataExplorationMode: true
+    }
+};
+const existingState = localStorageManager.getState();
+
+export const storeOptions: StoreOptions<DataExplorationState> = {
```
LekanAnni, line 68: Out of curiosity, is there a reason why namespaced is always false for root state?

is it because we can specify root as `{root: true} `. I just wonder why null we prefer to use null.

LekanAnni (approved):
LGTM

just need update with master

src/app/static/src/tests/components/dataExplorationApp.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        mount({
+            beforeMount: c.beforeMount,
+            methods: c.methods,
+            render: c.render
+        }, {store});
+
+        setTimeout(() => {
+            expect(baselineActions.getBaselineData).toHaveBeenCalled();
+            expect(surveyAndProgramActions.getSurveyAndProgramData).toHaveBeenCalled();
+            expect(adrActions.getSchemas).toHaveBeenCalled();
```
EmmaLRussell, line 60: Should also be getting genericChartMetadata. This was probably a missing test anyway!

EmmaLRussell (approved):
LGTM

src/app/static/src/app/store/dataExploration/dataExploration.ts
hillalex: Namespacing as a concept only applies to modules. See https://vuex.vuejs.org/guide/modules.html#namespacing

src/app/static/src/tests/components/dataExplorationApp.test.ts
hillalex: https://github.com/mrc-ide/hint/pull/608/commits/3b654ef69f23f28a480ebfe35e6322cc4eb4bf9b

PR #611, opened 2021-11-12
mrc-2755 Move error reporting related mutations out of root mutations

Nicolas-Dolan:
## Description

Moves error reporting-related state. mutations, and actions from the root module to the errors module so they can be shared with the data exploration mode. To test, check app loads up correctly and no issues with error reporting form.

## Type of version change
Patches 

## Checklist


src/app/static/src/tests/integration/errors.itest.ts
```diff
@@ -1,9 +1,6 @@
...
-import {ErrorReportManualDetails} from "../../app/store/root/actions";
-import {initialProjectsState} from "../../app/store/projects/projects";
-import {RootState} from "../../app/root";
-import {Language} from "../../app/store/translations/locales";
-import {initialHintrVersionState} from "../../app/store/hintrVersion/hintrVersion";
+import {actions} from "../../app/store/errors/actions";
+import {ErrorReportManualDetails} from "../../app/store/errors/actions";
+import {emptyState} from "../../app/root";
 
 describe(`root actions`, () => {
```
EmmaLRussell, line 5: This should be `error actions` now

src/app/static/src/app/store/errors/mutations.ts
```diff
@@ -16,5 +18,15 @@ export const mutations: MutationTree<ErrorsState> = {
 
     [ErrorsMutation.DismissAll](state: ErrorsState, action: PayloadWithType<null>) {
```
LekanAnni, line 19:
```suggestion
    [ErrorsMutation.DismissAll](state: ErrorsState) {
```

LekanAnni (changes requested):
LGTM 

Just a minor improvement

EmmaLRussell (changes requested):
I'm getting this error when I submit an error report (from a project with no errors in state, at step 2):
```
{
  "data": {},
  "status": "failure",
  "errors": [
    {
      "detail": "An unexpected error occurred. If you see this message while you are using Naomi at a workshop, please contact your workshop technical support and show them this code: uro-ers-iqg. Otherwise please contact support at naomi-support@imperial.ac.uk and quote this code: uro-ers-iqg",
      "error": "OTHER_ERROR",
      "trace": [
        "JSON parse error: Instantiation of [simple type, class org.imperial.mrc.hint.models.ErrorReport] value failed for JSON property errors due to missing (therefore NULL) value for creator parameter errors which is a non-nullable type; nested exception is com.fasterxml.jackson.module.kotlin.MissingKotlinParameterException: Instantiation of [simple type, class org.imperial.mrc.hint.models.ErrorReport] value failed for JSON property errors due to missing (therefore NULL) value for creator parameter errors which is a non-nullable type\n at [Source: (PushbackInputStream); line: 1, column: 430] (through reference chain: org.imperial.mrc.hint.models.ErrorReport[\"errors\"])"
      ]
    }
  ]
}
```

Looking at the POSTed data, the `errors' field is indeed missing.

I think this is because the action is using `getters.errors` to generate the `errors` field - but this getter has not been moved out of root. So it either needs to use the getter in root, or move it into errors - I think it would probably make for it to stay in root. You should be able to use `rootGetters` in the context.  Should be a test for this too.

src/app/static/src/app/store/errors/actions.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export const actions: ActionTree<ErrorsState, RootState> & ErrorsActions = {
+
+    async generateErrorReport(context, payload) {
+        const {dispatch, rootState, getters} = context
+        const data = {
+            email: payload.email || rootState.currentUser,
+            country: rootState.baseline.country || "no associated country",
+            projectName: rootState.projects.currentProject?.name || "no associated project",
```
hillalex, line 39: Ah, I'm afraid this needs to remain in root actions, because it won't be re-usable - it references the projects module, which won't exist in the new app.

hillalex (changes requested):
I think it should only be the mutations and the `ErrorReport` interface moved out. The actions and getter should remain in root.

hillalex (approved):
Looks great, thanks

src/app/static/src/app/store/errors/actions.ts
EmmaLRussell: Right, so we'll need a separate action to generate the report for data exploration mode. I guess the alternative would be some ugly casting to any to check if state is actually a root state or not (has projects). But if we're going to have a separate dialog component for DE error reporting anyway, cleaner to have that as separate action.

EmmaLRussell (approved):

LekanAnni:
@Nicolas-Dolan It would be nice if this ticket can be merged after update version.

Nicolas-Dolan:
> @Nicolas-Dolan It would be nice if this ticket can be merged after update version.

@LekanAnni up to date now

LekanAnni (approved):

PR #612, opened 2021-11-12
Mrc 2757 Add DataExploration app component

LekanAnni:
This PR adds data exploration component

## Type of version change
_Delete as appropriate_

 Minor

## Checklist


src/app/static/src/app/components/files/ManageFile.vue
```diff
@@ -85,7 +87,13 @@
             }
         },
         computed: {
-            editsRequireConfirmation: mapGetterByName("stepper", "editsRequireConfirmation"),
+            dataExplorationMode: mapStatePropByName(null, "dataExplorationMode"),
```
hillalex, line 90: Can this be pulled out into a mixin?

src/app/static/src/tests/components/uploadInputs/fileUploads.ts
```diff
@@ -7,12 +7,13 @@ import {
...
     mockError,
     mockMetadataState,
-    mockPlottingSelections,
+    mockPlottingSelections, mockRootState,
     mockSurveyAndProgramState
 } from "../../mocks";
 import UploadInputs from "../../../app/components/uploadInputs/UploadInputs.vue";
 import {BaselineState} from "../../../app/store/baseline/baseline";
 import {BaselineActions} from "../../../app/store/baseline/actions";
+import {emptyState} from "../../../app/root";
```
hillalex, line 16: unused imports?

src/app/static/src/tests/components/files/manageFile.test.ts
```diff
@@ -14,17 +13,14 @@ import {expectTranslated} from "../../testHelpers";
 
 describe("Manage file component", () => {
 
-    const mockGetters = {
-        editsRequireConfirmation: () => false
-    };
 
-    const createStore = () => {
+    const createStore = (customStore: any = emptyState(), requireConfirmation = false) => {
```
hillalex, line 17: cant this type not be `DataExplorationState`? and similarly in other test suites above

hillalex (changes requested):
Works great. Small tidy ups to the code requested inline.

I notice that the ADR dataset is not persisted on a page refresh, I think this is because we are not actually persisting anything to local storage for this app yet (the ADR key and input files are saved to the db, hence why they persist) so have added a ticket for that: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2773. 
Perhaps tomorrow we can briefly estimate and decide whether to pull that one in to this sprint.

src/app/static/src/app/components/dataExploration/DataExploration.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+    <div class="content">
+        <div class="pt-4">
+            <adr-integration></adr-integration>
```
EmmaLRussell, line 4: I'm assuming we should hide ADR Integraton if we're on Step 2, same as for the main app? That's what's shown in the [mockups](https://docs.google.com/presentation/d/1bX7WnIy9Thlub_4Uo17uWlA6f07oicOc/edit#slide=id.p5)

src/app/static/src/app/components/dataExploration/DataExploration.vue
```diff
@@ -0,0 +1,46 @@
...
+            <button type="button"
+                    class="btn btn-white shadow-none"
+                    v-translate="'back'"
+                    @click="back">
+            </button>
+            <button type="button"
+                    class="btn btn-red shadow-none"
+                    v-translate="'next'"
+                    @click="next">
+            </button>
```
EmmaLRussell, line 16: I guess we should also disable back if step === 1 and disable next if step === 2

src/app/static/src/tests/dataExploration/getters.test.ts
```diff
@@ -0,0 +1,17 @@
+import {getters} from "../../app/store/dataExploration/getters";
+import {mockDataExplorationState} from "../mocks";
+
+describe(`data exploration getters`, () => {
+
+    it(`return true if use is guest`, () => {
```
EmmaLRussell, line 6:
```suggestion
    it(`isGuest returns true if user is guest`, () => {
```

src/app/static/src/tests/dataExploration/getters.test.ts
```diff
@@ -0,0 +1,17 @@
...
+        const state = mockDataExplorationState({currentUser: "guest"})
+        const getter = getters.isGuest(state, null, null as any, null)
+        expect(getter).toBe(true)
+    })
+
+    it(`return false if user is not guest`, () => {
```
EmmaLRussell, line 12:
```suggestion
    it(`isGuest returns false if user is not guest`, () => {
```

EmmaLRussell (commented):
This works great. But I'm not sure why we're not matching the mockups in terms of the appearance of the back/continue buttons? I know the prototype didn't, but I thought the assumption was that we'd match the main app here in the real implementation. Is this not the case?

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -209,10 +209,10 @@ describe("select dataset", () => {
 
     let currentVersion = {id: "version-id", created: "", updated: "", versionNumber: 1}
 
-    const getStore = (baselineProps: Partial<BaselineState> = {}, adrProps: Partial<ADRState> = {}, requireConfirmation: boolean = false) => {
+    const getStore = (baselineProps: Partial<BaselineState> = {}, adrProps: Partial<ADRState> = {}, requireConfirmation: boolean = false, customStore: any = mockRootState()) => {
```
hillalex, line 212: Suggest `any` -> `DataExplorationState` and `customStore` -> `state`

hillalex (commented):
Related to my previous comment about local storage, noticed another weird artefact of the current code - if you first log in to the regular app and select a dataset, then switch to explore mode, the selected dataset appears but no files, because the `baseline` module is reading from local storage but the backend session has been cleared. Should be resolved by checking for data exploration mode in the local storage manager `getState` function and only returning state for the correct app mode. Just noting here and will add to the existing ticket around local storage.

LekanAnni:
> Related to my previous comment about local storage, noticed another weird artefact of the current code - if you first log in to the regular app and select a dataset, then switch to explore mode, the selected dataset appears but no files, because the `baseline` module is reading from local storage but the backend session has been cleared. Should be resolved by checking for data exploration mode in the local storage manager `getState` function and only returning state for the correct app mode. Just noting here and will add to the existing ticket around local storage.

Yes of course , I thought you mentioned you would create a ticket for that....

EmmaLRussell (commented):
Looks good in terms of hiding ADR component in step 2, and disable button behaviour, but still think we should match the Back/Continue navigation from the main app, as shown in the mockups. The buttons at the bottom of the page mean you scroll all the way down to bottom of table to navigate back. You should just be able to re-use the `StepperNavigation` component I think?

LekanAnni:
> Looks good in terms of hiding ADR component in step 2, and disable button behaviour, but still think we should match the Back/Continue navigation from the main app, as shown in the mockups. The buttons at the bottom of the page mean you scroll all the way down to bottom of table to navigate back. You should just be able to re-use the `StepperNavigation` component I think?

I suppose we can also have back/continue at the bottom as well as the top, right ?

hillalex (approved):
Looks great, thanks

EmmaLRussell (approved):
:+1: Nice, just needs conflicts resolved.

PR #613, opened 2021-11-15
Mrc 2739 Data Exploration mode navbar component

LekanAnni:
This PR adds nav bar for data exploration

- no functionality for troubleshooting request
- FAQ is hidden for DE

Minor

## Checklist


EmmaLRussell (changes requested):
Looks good! We talked about using a secondary colour in the header in DE mode rather than the standard red. (Not sure what that secondary colour would be!)

Changing language fails

The accessibility page needs to show the exploration mode header when accessed in DE mode. Could do that in another ticket if you prefer.

src/app/static/src/app/components/header/DataExplorationHeader.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+    <header class="mb-5">
+        <nav class="navbar navbar-dark bg-secondary">
+            <div class="container-fluid">
+                <div class="navbar-header">
+                    {{ title }}
```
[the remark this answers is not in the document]
hillalex: Could argue it's a proper name?

src/app/static/src/app/components/header/OnlineSupportMenu.vue
```diff
@@ -10,13 +10,14 @@
...
-            </router-link>
+            <a class="dropdown-item"
+               id="accessibility-link"
+               :href="'/accessibility'"
+               target="_blank"
+               v-translate="'axe'">
+            </a>
         </drop-down>
-        <error-report :open="errorReportOpen" @close="toggleErrorReportModal"></error-report>
+        <error-report v-if="show" :open="errorReportOpen" @close="toggleErrorReportModal"></error-report>
```
hillalex, line 20: Have introduced the router [here](https://github.com/mrc-ide/hint/pull/623) so accessibility should remain a `router-link`.

But I don't think this is the right approach here anyway. This component should simply not be included in the new header, and a new component created for the data exploration mode, since it will need it's own error report component and presumably it's own FAQ at some point. This should be a separate ticket.

hillalex (changes requested):
I think leave out the `online-support-menu` altogether. The new one will be sufficiently different as to warrant a new component.

src/app/static/src/app/components/header/DataExplorationHeader.vue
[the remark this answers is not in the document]
LekanAnni: I reckon we don't need to translate.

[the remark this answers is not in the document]
EmmaLRussell: Fair enough, let's see if anyone complains!

LekanAnni:
https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2802

Change language does not work but will be addressed in above ticket

src/app/static/src/app/components/header/DataExplorationHeader.vue
```diff
@@ -2,24 +2,23 @@
...
                 <hintr-version-menu class="pr-2 mr-2 border-right"/>
-                <online-support-menu class="pr-2 mr-2 border-right"/>
                 <a :href="'public/resources/' + helpFilename"
                    id="helpFile"
                    target="_blank"
                    class="pr-2 mr-2 border-right"
                    v-translate="'dataExplorationHelp'">
                 </a>
-                <a v-if="!isGuest" href="/logout" class="pr-2 mr-2 border-right" v-translate="'logout'">
+                <a v-if="!isGuest" :href="'/logout'" class="pr-2 mr-2 border-right" v-translate="'logout'">
```
EmmaLRussell, line 19: Shouldn't need the colon notation for these hrefs, they don't need to be evaluated since they won't change

EmmaLRussell (commented):
Looks good - I like the teal!

So adding a DE mode OnlineSupportMenu needs to be another ticket- do we have that yet? Could you add it if not.

src/app/static/src/app/components/header/DataExplorationHeader.vue
LekanAnni: colon is not needed to get it working but required to resolve path

LekanAnni:
> Looks good - I like the teal!
> 
> So adding a DE mode OnlineSupportMenu needs to be another ticket- do we have that yet? Could you add it if not.

https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2803 ticket to add online support menu

src/app/static/src/app/components/header/DataExplorationHeader.vue
EmmaLRussell: Sorry, I don't know what you mean by that. I just tried making that change and it works fine manually, and the tests still pass, so not sure what the problem is?

LekanAnni: I meant it does work fine without colon.

The reason for adding colon is because intelliJ complains about the file reference as it is not getting resolved. Changing href to a computed href means it does not see file directory as plain html but computed value, hence issue resolved.

EmmaLRussell: Well, ok - but we do have other things in the templates that IntelliJ low-key complains about e.g 'typos' in class names, and it's easy enough to suppress that warning, which isn't really relevant for us.

hillalex (approved):

EmmaLRussell (approved):
Still not convinced that using that colon notation is necessary but otherwise good to merge when conflicts fixed.

PR #614, opened 2021-11-15
mrc-2771 BUG fix - handle case where indicator is null in choropleth

hillalex:
## Description

The bug can be replicated as follows:
* Select Antarctica 2022-1 owned by Naomi Development Team
* Navigate to step 2 and notice that  "household survey" is selected and there are 4 available indicators: HIV prevalence, ART coverage, Proportion recently infected and Viral load suppression
* Navigate back to step 1 and select Antarctica-2022-1-2 owned by Imperial College London
* Navigate to step 2 and notice  that "anc testing" is selected and there are 2 available indicators: ANC HIV prelanence, ANC prior ART coverage. At this point the selected indicator no longer exists in the list of available indicators, hence the error. 

This doesn't seem to happen every time. Appears as though sometimes the choropleth defaults to "household survey" and sometimes to "ANC testing" - is this due to order files are loaded perhaps? Try this process many times to replicate the error. 

There is logic in place to update the selected indicator when the metadata changes, but the component is trying to reference it before this happens. The fix involves removing an inappropriate `!` cast and handling the case where the selected indicator is null.

## Type of version change

Patch

## Checklist


src/app/static/src/app/hintVersion.ts
```diff
@@ -1 +1 @@
-export const currentHintVersion = "1.69.0";
+export const currentHintVersion = "1.70.1";
```
hillalex, line 1: looks like this didn't get updated in previous merge

EmmaLRussell (commented):
This works great! 

It feels a little bit weird to have the `| undefined` leaking out into the utils methods to calculate things that aren't really going to be available until the indicator is selected anyway - particularly for the bubble util, since the bubble plot shouldn't be affected by any of this. Could the choropleth selectively not call out to utils or include MapLegend if its indicator is undefined? 

Because this bug is related to the order in which files are loaded, the reliable way to reproduce the bug in master is to manually load a single file when you return to step 1, which doesn't correspond to the latest selected indicator. e.g. select a survey indicator then go back and manually load anc file. Then progress to step 2 - ANC Testing is selected and the error is logged to console. Rerunning with this branch fixes the problem. 

> This doesn't seem to happen every time. Appears as though sometimes the choropleth defaults to "household survey" and sometimes to "ANC testing" - is this due to order files are loaded perhaps? Try this process many times to replicate the error.

That's interesting - probably it's a legacy of when we used to update the selected tab in step 2 to show whichever file had just loaded. That might be obsolete now, we could probably always default to Survey since we know we need at least that -- or rather we will do until data exploration is implemented, which can move on to step 2 without survey being available. 

But either way we would still need this fix to ensure the alignment of selected data type and selected indicator can proceed without the error.

src/app/static/src/app/components/plots/bubble/utils.ts
```diff
@@ -31,11 +31,11 @@ export const getFeatureIndicators = function (data: any[],
 
     const result = {} as BubbleIndicatorValuesDict;
     iterateDataValues(data, [sizeIndicator, colorIndicator], selectedAreaIds, filters, selectedFilterValues,
-        (areaId: string, indicatorMeta: ChoroplethIndicatorMetadata, value: number, values: any) => {
+        (areaId: string, indicatorMeta: ChoroplethIndicatorMetadata | undefined, value: number, values: any) => {
```
LekanAnni, line 34: I am not sure undefined should be here. Should there be a need for computation if indicatorMeta is undefined ?

LekanAnni (changes requested):
Still can't tell if it does not work, so it works well. 

See inline comment, undefined being a valid parameter value stood out.

hillalex:
Thanks @LekanAnni and @EmmaLRussell - agree the use of `undefined` as a param is rather unpleasant, just seemed slightly less fiddly to only have to check it in one place (the `iterateDataValues` functions). But have changed to check for undefined values locally in the choropleth rather than push that out to the utils 

Wrt the `MapLegend` component, it already had checks in place for undefined metadata (e.g. https://github.com/mrc-ide/hint/pull/614/files#diff-44d66b7e90e2c4ea3a07ea59621b8af2c92f4187dab4d157187021b446a0d8f9L91) so the only change there is accurately reflecting in the prop types that the metadata prop can be undefined. A `v-if` clause that was frequently toggled wouldn't be the best for performance so I don't think that would be an improvement there.

src/app/static/src/app/components/plots/MapLegend.vue
```diff
@@ -103,7 +103,7 @@
                         val = roundToContext(val, [min, max]);
 
                         let valAsProportion = (max != min) ? (val - min) / (max - min) : 0;
-                        if (this.metadata.invert_scale) {
+                        if (this.metadata?.invert_scale) {
```
hillalex, line 106: For some reason even though this is [nested in a falsy check](https://github.com/mrc-ide/hint/pull/614/files#diff-44d66b7e90e2c4ea3a07ea59621b8af2c92f4187dab4d157187021b446a0d8f9L91) the compiler doesn't pick up that `metadata` must be defined here.

Nicolas-Dolan: Can you reference the `this` object within this method? I'm not sure if this counts as an anonymous function or not. You could try putting `this.metadata` into its own variable and referencing that instead.

Nicolas-Dolan (approved):
Works for me.

src/app/static/src/app/components/plots/MapLegend.vue
hillalex: Great shout, I will do this. Maybe this has never worked!?

hillalex: Oh actually, using the arrow function syntax means that `this` is preserved within `map` functions. So this does seem like a compiler mistake. But I'll still pull out the variable.

EmmaLRussell (commented):
LGTM, apart from codeciv issue

EmmaLRussell (approved):

LekanAnni (approved):

PR #617, opened 2021-11-16
mrc-2775 Filter Survey and Program choropleth indicators by those included in response metadata

EmmaLRussell:
## Description

This branch address addresses the bug whereby 'VLS tested` and 'VLS tests suppressed' indicators were still being displayed for ART data even when using the 'no vls' ART file (attached to ticket), which contains no VLS data. 

The problem here was that indicator metadata is spread across two sets of responses and we were only using one of them:
- the 'full' indicator metadata, including complete specification of column ids etc comes from the plotting metadata endpoint. This is only refreshed when a new PJNZ is uploaded
- the data response for survey, anc and programme (ART) contains a `filters` section which actually includes `indicators`,which lists the indicators which should be available for the current dataset

This branch combines these metadata sources to filter the full indicator metadata returned by the `sapIndicatorsMetadata` getter to include only those indicators listed in the relevant data response. 

To test:
 - import ADR Antarctica dataset as usual, and observe VLS indicators in Review Input's Map view for ART data
 - Go back to step 1 and replace the ART file with manual upload of `programme_no_vls.csv` attached to this ticket
 - Proceed to step 2, and observe that VLS indicators have been removed from the Map view for ART data 

## Type of version change
_Delete as appropriate_

Patches

## Checklist

### DIFF
--- a/src/app/static/src/app/store/metadata/metadata.ts
+++ b/src/app/static/src/app/store/metadata/metadata.ts
@@ -42,7 +42,7 @@ export const metadataGetters = {
         const selectedDataType = sap.selectedDataType;
 
         let metadataForType: Metadata | null = null;
-        let dataIndicators: FilterOption[];
+        let dataIndicators: FilterOption[] = [];
         switch (selectedDataType) {
             case (DataType.ANC):
                 metadataForType = plottingMetadata.anc;
@@ -58,10 +58,10 @@ export const metadataGetters = {
                 break;
         }
 
-        const unfiltered =  (metadataForType && metadataForType.choropleth) ? metadataForType.choropleth.indicators : [];
+        const unfiltered = (metadataForType && metadataForType.choropleth) ? metadataForType.choropleth.indicators : [];
         return (unfiltered as ChoroplethIndicatorMetadata[]).filter(
-            (metaIndicator: any) => dataIndicators!.some(
-                (dataIndicator: any) => metaIndicator.indicator === dataIndicator.id
+            (metaIndicator: ChoroplethIndicatorMetadata) => dataIndicators.some(
+                (dataIndicator: FilterOption) => metaIndicator.indicator === dataIndicator.id
             )
         )
     },
--- a/src/app/static/src/app/store/metadata/metadata.ts
+++ b/src/app/static/src/app/store/metadata/metadata.ts
@@ -58,7 +58,7 @@ export const metadataGetters = {
                 break;
         }
 
-        const unfiltered = (metadataForType && metadataForType.choropleth) ? metadataForType.choropleth.indicators : [];
+        const unfiltered = metadataForType?.choropleth ? metadataForType.choropleth.indicators : [];
         return (unfiltered as ChoroplethIndicatorMetadata[]).filter(
             (metaIndicator: ChoroplethIndicatorMetadata) => dataIndicators.some(
                 (dataIndicator: FilterOption) => metaIndicator.indicator === dataIndicator.id
--- a/src/app/static/src/app/store/metadata/metadata.ts
+++ b/src/app/static/src/app/store/metadata/metadata.ts
@@ -5,7 +5,7 @@ import {
     PlottingMetadataResponse,
     Error,
     Metadata,
-    AdrMetadataResponse, FilterOption
+    AdrMetadataResponse, FilterOption, ChoroplethIndicatorMetadata
 } from "../../generated";
 import {localStorageManager} from "../../localStorageManager";
 import {DataType} from "../surveyAndProgram/surveyAndProgram";
@@ -39,7 +39,7 @@ export const metadataGetters = {
         }
 
         const sap = rootState.surveyAndProgram;
-        const selectedDataType = rootState.surveyAndProgram.selectedDataType;
+        const selectedDataType = sap.selectedDataType;
 
         let metadataForType: Metadata | null = null;
         let dataIndicators: FilterOption[];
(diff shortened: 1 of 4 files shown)


src/app/static/src/app/store/metadata/metadata.ts
```diff
@@ -38,22 +38,32 @@ export const metadataGetters = {
...
             case (DataType.Survey):
                 metadataForType = plottingMetadata.survey;
+                dataIndicators = sap.survey?.filters.indicators ||[];
                 break;
         }
 
-        return (metadataForType && metadataForType.choropleth) ? metadataForType.choropleth.indicators : [];
+        const unfiltered =  (metadataForType && metadataForType.choropleth) ? metadataForType.choropleth.indicators : [];
+        return (unfiltered as ChoroplethIndicatorMetadata[]).filter(
+            (metaIndicator: any) => dataIndicators!.some(
```
hillalex, line 63:
```suggestion
            (metaIndicator: ChoroplethIndicatorMetadata) => dataIndicators.some(
```

src/app/static/src/app/store/metadata/metadata.ts
```diff
@@ -38,22 +38,32 @@ export const metadataGetters = {
...
+                (dataIndicator: any) => metaIndicator.indicator === dataIndicator.id
```
hillalex, line 64:
```suggestion
                (dataIndicator: FilterOption) => metaIndicator.indicator === dataIndicator.id
```

src/app/static/src/app/store/metadata/metadata.ts
```diff
@@ -38,22 +38,32 @@ export const metadataGetters = {
...
+                dataIndicators = sap.program?.filters.indicators || [];
                 break;
             case (DataType.Survey):
                 metadataForType = plottingMetadata.survey;
+                dataIndicators = sap.survey?.filters.indicators ||[];
                 break;
         }
 
-        return (metadataForType && metadataForType.choropleth) ? metadataForType.choropleth.indicators : [];
+        const unfiltered =  (metadataForType && metadataForType.choropleth) ? metadataForType.choropleth.indicators : [];
```
hillalex, line 61: Is the generated `Metadata` type wrong? It is:
```
export interface Metadata {
  choropleth?: {
    indicators?: ChoroplethMetadata;
    [k: string]: any;
  };
  [k: string]: any;
}
```
which would imply that `unfiltered` here would be of type singular `ChoroplethIndicatorMetadata`, rather than an array. And that `choropleth.indicators` could be undefined. If this is wrong then looks we need to update this hintr schema and regenerate types: https://github.com/mrc-ide/hintr/blob/master/inst/schema/PlottingMetadataResponse.schema.json Also suggests `hintr` is missing a test that validates the plotting metadata against the schema...

hillalex (commented):
Just a query around typings/schema

src/app/static/src/app/store/metadata/metadata.ts
```diff
@@ -38,22 +38,32 @@ export const metadataGetters = {
             return [];
         }
 
-        const selectedDataType = rootState.surveyAndProgram.selectedDataType;
+        const sap = rootState.surveyAndProgram;
+        const selectedDataType = sap.selectedDataType;
 
         let metadataForType: Metadata | null = null;
+        let dataIndicators: FilterOption[];
```
hillalex, line 45:
```suggestion
        let dataIndicators: FilterOption[] = [];
```

hillalex: Then can remove `!` cast below. Am minded to avoid these!

src/app/static/src/app/store/metadata/metadata.ts
EmmaLRussell: Oops, I actually thought I'd got rid of the `any`s. Obviously not! Thanks!

src/app/static/src/app/store/metadata/metadata.ts
EmmaLRussell: Yeah, the types are  complete mess. There are two `ChoroplethMetadata`s, which confuses the issue - one a type and one an interface. What we actually seem to get in `indicators` corresponds to the type definition, which is indeed an array of indicators.
I'm not really sure how this is happened, but as you say it needs to be fixed in hintr and types regenerated - I'll open a ticket for that

hillalex: Just looking at the schema again and it looks correct actually, I didn't read it carefully enough. But there's both an internal definition called "choropleth_metadata" and a separate file schema called ChoroplethMetadata, which I guess is causing the duplicate and the weird behaviour. Let me see if updating the hintr schema on a branch and regenerating fixes it, as then we can rely on those typings here and keep things a bit clearer.

EmmaLRussell: Yeah, I was just getting there myself. Is the schema obsolete then, can that go to avoid the duplicate? (which causes problems in Typescript)

EmmaLRussell: I've added a ticket to fix in hintr, though text might not be accurate now: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2787

hillalex:
Fixed up schema and regenerated types, patch proposal here: https://github.com/mrc-ide/hint/pull/621

src/app/static/src/app/store/metadata/metadata.ts
```diff
@@ -38,22 +38,32 @@ export const metadataGetters = {
...
+                dataIndicators = sap.program?.filters.indicators || [];
                 break;
             case (DataType.Survey):
                 metadataForType = plottingMetadata.survey;
+                dataIndicators = sap.survey?.filters.indicators ||[];
                 break;
         }
 
-        return (metadataForType && metadataForType.choropleth) ? metadataForType.choropleth.indicators : [];
+        const unfiltered = (metadataForType && metadataForType.choropleth) ? metadataForType.choropleth.indicators : [];
```
Nicolas-Dolan, line 61:
```suggestion
        const unfiltered = (metadataForType?.choropleth) ? metadataForType.choropleth.indicators : [];
```
I think this would be the same and a little shorter

Nicolas-Dolan (approved):
Works for me

hillalex (approved):

PR #625, opened 2021-11-17
mrc-2778 Fix issues with Input Time Series Plot for DRC data

EmmaLRussell:
## Description

Ian reported being unable to view Input Time Series chart for DRC data. The immediate cause of this was the existence of  null values in the data, but there were actually several issues in addition to this when viewing Input Time Series for DRC. This branch addresses each:

1.  Handle null values in data by checking for null in `filterData` util before attempting `toString()`
2.  Improve performance of jsonata by grouping all data by `area_name` at the start of evaluation instead of filtering to each `area_name` as required for each subplot. This cuts down jsonata evaluation time for the (very large) DRC dataset from ~20s to ~1s, and also has a noticeable improvement in update time for smaller Malawi dataset
3. Tab crash when attempt to show ANC prevalence chart - i'm not completely sure what the issue was here but I think with 517 (!) subplots to show, it was just all too much for Plotly to handle. Unsetting the 'responsive' flag in config has fixed this for me, and doesn't really lose us anything since the width of the component is essentially fixed anyway. However, see my comment in Naomi channel - still seeing sporadic tab crashes. I think implementing some sort of paging will help here, and have created a [new ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2799) for that. 
4. Slow Plotly loading without user feedback - DRC data takes ~10s to render in Plotly. I've added a spinner in the Plotly component itself which displays while Plotly is rendering. In order to get this to show, I needed to pull out the actual rendering from the main thread, in a `setTimeout`, otherwise the spinner was not being shown until the rendering had completed anyway. Even so, the spinning itself hangs after a second or so, not sure why!

To test: manually upload the files for the DRC data attached to [the ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2778)

## Type of version change
_Delete as appropriate_
Patches

## Checklist


src/app/static/src/app/components/genericChart/Plotly.vue
```diff
@@ -1,6 +1,10 @@
 <template>
     <div>
-        <div ref="chart" style="width:100%; height:100%;"></div>
+        <div v-if="rendering" class="text-center mt-4" style="position: absolute; top: 0; width: 100%;">
```
EmmaLRussell, line 3: Make the spinner absolute position so it shows even if user has scrolled down through the subplots. The positioned parent for this is actually the chartContainer's parent in GenericChart

LekanAnni (changes requested):
Looking good.

Chart is significantly slow, I didn't experience any crash though.

- It still takes almost 20s for the chart to display.
- toggling between tabs is noticeably slow 
- when ANC testing data is loaded, and then toggle to and back, it still takes 20s to reload chart. Perhaps you could look into caching data.
- quarterly control takes time to get populated, obviously waiting for data to be loaded.
- selecting data source takes a significant amount of time to respond.

After several attempts the app finally crashed.

With large dataset, I think it will be slower and unusable when it's deployed to production server.

EmmaLRussell:
> With large dataset, I think it will be slower and unusable when it's deployed to production server.

This is the full DRC dataset, which is one of the largest, so I don't think prod will be any larger. We do need to test with Nigeria data too though, by the sounds of it. 

It's not really a data caching issue, the actual loading and re-filtering of the data is pretty quick. The delay is in re-rendering the changed data in Plotly. We are a doing a full re-render, which we need to do because of the number of subplots changing between filter selections. 

All the issues you mention will, I think, be addressed by adding paging ([ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2799)). But the improvements in this branch should help with smaller datasets.

hillalex (approved):
No doubt these are improvements but clearly some way to go. Haven't benchmarked these against our use case but they look worth trying, I suspect especially the `autorange` and the `Plotly.react`: https://www.somesolvedproblems.com/2018/07/how-do-i-make-plotly-faster.html

Although rendering is the main bottleneck it does look as though the data manipulation is also taking a while, judging by the way the page hangs after changing a filter but before the chart loading spinner starts. I think worth digging into the code to see if further optimisations can be made there.

EmmaLRussell:
> No doubt these are improvements but clearly some way to go. Haven't benchmarked these against our use case but they look worth trying, I suspect especially the `autorange` and the `Plotly.react`: https://www.somesolvedproblems.com/2018/07/how-do-i-make-plotly-faster.html
> 
> Although rendering is the main bottleneck it does look as though the data manipulation is also taking a while, judging by the way the page hangs after changing a filter but before the chart loading spinner starts. I think worth digging into the code to see if further optimisations can be made there.

I don't think `Plotly.react` is an option -we were using that before but it doesn't cope well with new data with a different number of subplots. I'll have a look at `autorange`.

Yes, the filtering itself it taking ~1s. Might be able to find some improvements there.

LekanAnni (approved):

EmmaLRussell:
I've added another ticket for finding further performance improvements: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2801
..and am also implementing paging in mrc-2799

PR #626, opened 2021-11-18
mrc-2781 When importing from ADR remove all datasets currently uploaded

LekanAnni:
PR error removes upload input files before re-uploading from ADR.

Patches

## Checklist


src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -301,13 +303,23 @@
...
             importSurvey: mapActionByName("surveyAndProgram", "importSurvey"),
             importProgram: mapActionByName("surveyAndProgram", "importProgram"),
             importANC: mapActionByName("surveyAndProgram", "importANC"),
+            deleteReviewInputs: mapActionByName("surveyAndProgram", "deleteAll"),
             async importDataset() {
                 this.stopPolling();
 
                 this.loading = true;
+
+                if (this.hasShapeFile) {
```
LekanAnni, line 316: Trying to avoid sending unnecessary delete action here but I think it might be better to ignore this checking, your thoughts? 

Perhaps I could check if surveyFile exists too `if(shape || survey)...`

EmmaLRussell: See comment above, I think you only need to send action to delete baseline files. You could check if any exist before sending that action i.e. any of shape, pjnz or population

EmmaLRussell (approved):
Works great - tested it with the Zambia dataset on dev, which currently lacks Spectrum, population and survey files. 

If you wanted to you could reduce the verbosity of the `deletes previously imported files if x file exists` tests since they're all basically the same, by having a shared function they each call which sets up the store with a partial baseline state provided in parameter then does all the common checking of the resulting wrapper.

hillalex (approved):

PR #627, opened 2021-11-18
mrc-2782 prevent tooltip auto-hiding on mouseover

hillalex:
## Description

Does what it says on the tin. Haven't changed the event to a click, just made it a bit less eager to close. See what you think/whether this resolves usability issue.

## Type of version change
Patch

## Checklist


src/app/static/src/app/components/plots/FilterSelect.vue
```diff
@@ -1,10 +1,16 @@
...
+        <span v-if="labelTooltip"
+              v-tooltip="{
+                    content: `<dl>${labelTooltip}</dl>`,
+                    classes: 'filter-select',
+                    autoHide: false
+                }"
+              class="icon-small">
             <help-circle-icon></help-circle-icon>
         </span>
-        <treeselect id="survey-filters" :multiple=multiple
```
hillalex, line 7: also removed this erroneous id - created multiple elements on page with same id

LekanAnni (approved):
LGTM

EmmaLRussell (approved):
So this means that you can at least scroll with mouse wheel or arrow buttons, and keep the tooltip visible, but it will be hidden if the tooltip loses focus e.g. if move mouse to use scrollbar? It's definitely an improvement for most users I'd think.

hillalex:
> So this means that you can at least scroll with mouse wheel or arrow buttons, and keep the tooltip visible, but it will be hidden if the tooltip loses focus e.g. if move mouse to use scrollbar? It's definitely an improvement for most users I'd think.

Exactly. We could also change it to a click event if we think that's a better solution? Maybe we deploy this and see if people still struggle?

EmmaLRussell:
Click would be better for users who like to user the scrollbar to scroll, but I think that could cause confusion too if people aren't sure how to close it! Let's try it this way and see if people still have issues.

PR #628, opened 2021-11-19
mrc-2799 Paging in Input Time Series

EmmaLRussell:
## Description

Implements paging in the Generic Chart component to prevent Plotly being overwhelmed by rendering hundreds of subplots for district level Input Time Series for larger countries. 

The GenericChart component uses the knowledge it already has of the subplots configuration to defined page of values for the `distinctColumn` in the config (in practice for Input Time Series, this is the area_name column) and to further filter data provided to Plotly by the current page number. 

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/static/src/app/components/genericChart/GenericChart.vue
```diff
@@ -32,6 +32,28 @@
                          </div>
                      </div>
                 </div>
+                <div v-if="totalPages > 1" id="page-controls" class="text-center mt-2">
+                    <button id="previous-page"
+                            class="btn btn-sm btn-red mr-2"
+                            v-translate:aria-label="'previousPage'"
+                            v-tooltip="translate('previousPage')"
```
EmmaLRussell, line 39: The tooltips might be a little redundant/intrusive. What do you think?

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -161,7 +161,7 @@ export interface Translations {
     missingError: string,
     modelOptions: string,
     newPassword: string,
-    next: string,
+    nextPage: string,
```
EmmaLRussell, line 164: I don't believe we need `next` any more - I think this was added when the Data Exploration component was being delveloped, but then 'next' button was removed and replaced by StepperNavigation

src/app/static/src/app/components/genericChart/GenericChart.vue
```diff
@@ -252,7 +295,46 @@
...
+                if (this.pageDistinctColumnValues && this.chartData && this.chartMetadata.subplots) {
+                    const distinctColumn = this.chartMetadata.subplots.distinctColumn;
+                    const data = this.chartData["data"].filter((row: any) => this.pageDistinctColumnValues!.includes(row[distinctColumn]));
+                    return {...this.chartData, data}
+                } else {
+                    return this.chartData;
+                }
+            },
+            pageNumberText: function () {
+                return i18next.t("pageNumber", {
```
LekanAnni, line 332: Language works well but when changeLanguage is triggered selected tab "Time series" changes to Map

src/app/static/src/tests/components/genericChart/genericChart.test.ts
```diff
@@ -718,4 +731,325 @@ describe("GenericChart component", () => {
...
+    it("renders tooltips in Portuguese", (done) => {
+        const state = {datasets: pagedDatasets};
+        const wrapper = getWrapper(state, pagedMetadata);
+        wrapper.vm.$store.state.language = Language.pt;
+        setTimeout(() => {
+            expect(mockTooltip.mock.calls[0][1].value).toBe("Página anterior");
+            expect(mockTooltip.mock.calls[1][1].value).toBe("Próxima página");
+            done();
+        });
+    });
```
LekanAnni, line 945:
```suggestion
    it("renders tooltips in English, French and Portuguese", (done) => {
        const state = {datasets: pagedDatasets};
        const wrapper = getWrapper(state, pagedMetadata);
        setTimeout(() => {
             expectTranslated(mockTooltip.mock.calls[0][1],
                "Previous page",
                "Page précédente",
                "Página anterior",
                wrapper.vm.$store)

            expectTranslated(mockTooltip.mock.calls[0][1],
                "Next page",
                "Page suivante",
                "Próxima página",
                wrapper.vm.$store)
            done();
        });
```

LekanAnni (changes requested):
Looking really good, significant improvement in performance, it hides pagination nicely when not needed.

- there is a delay before loading spinner displays, ie. when changing data source.
- I think we should give users the option to display all chart in a page as well but default display should be paginated.
- The below image shows anomalies in x axis line, some bold some not.
Screenshot 2021-11-22 at 10 19 09

EmmaLRussell:
> Looking really good, significant improvement in performance, it hides pagination nicely when not needed.
> 
>     * there is a delay before loading spinner displays, ie. when changing data source.

Yep hoping to address that in mrc-2801, though it's actually a real puzzler that one. 

(The data filtering takes very little time (<50ms) but there's still a delay after it gets as far as invoking the `setTimeout` in Plotly before completing the mousedown event and showing the spinner. Not really sure why. Some overhead associated with kicking off the setTimeout? I've done some experiments using web workers, which seem to help, but I think I might need to move the spinner out of the Plotly component into GenericChart in order to respond quickly enough.)

> 
>     * I think we should give users the option to display all chart in a page as well but default display should be paginated.
> 

I'm not too keen on that idea tbh. I don't think we should throw any arbitrary number of subplots at Plotly - we know it will break - so I think it's reasonable to have a maximum number of plots per page that we will show. Maybe that could be tweaked upwards, but I think the current value is a reasonable compromise between speed of render and reducing number of pages.

An alternative might be to try multiple plotly components stacked on top of each other, each showing a page of data - but that would be complex to implement, and I'm not convinced it would perform any better. 

>     * The below image shows anomalies in x axis line, some bold some not.
> 

It's not that the x axis is bold, it's that you don't get any x axis values, but you do see a y axis line, in cases where a subplot has no non-null data points. I'll check with Jeff what he wants to see in that case, and see if I can tidy up the axes a bit if they should stay.

src/app/static/src/app/components/genericChart/GenericChart.vue
```diff
@@ -252,7 +295,46 @@
...
+                    return null;
+                }
+            },
+            totalPages() {
+                return this.allDistinctColumnValues ? Math.ceil(this.allDistinctColumnValues.length / this.chartMetadata.subplots!.subplotsPerPage) : 1;
+            },
+            chartDataPage() {
+                if (this.pageDistinctColumnValues && this.chartData && this.chartMetadata.subplots) {
+                    const distinctColumn = this.chartMetadata.subplots.distinctColumn;
+                    const data = this.chartData["data"].filter((row: any) => this.pageDistinctColumnValues!.includes(row[distinctColumn]));
```
hillalex, line 325: So this function is almost an order of magnitude slower than the filtering function (about 700ms in my benchmarking) and adds considerable overhead to the scripting on this page. I wonder if the paging can be pushed back to the server a bit - each subplot is one area right? And each data point only appears on one plot? So the server could group the areas at each geographic level into pages and return that information with the data, e.g.

```
{
    "area_id": "COD",
    "area_name": "Democratic Republic of The Congo",
    "area_level": 0,
    "age_group": "Y015_049",
    "time_period": "2019",
    "quarter": "Q4",
    "plot": "anc_clients",
    "value": 4026371,
    "page:" 1
}
```

You would also need the number of pages for each area level, am still not familiar with the structure of the metadata here so not sure where that might fit, but you could also work it out once in the front end if you knew the number of plots per page and the total number of areas.

(of course this would mean that you couldn't make num plots per page configurable by the user)

A less drastic solution would be to use this proposed data structure but construct it in the front-end - replace the `allDistinctColumnValues` function with one that adds page numbers to the data points. 

Basically if we can replace this nested loop with:

```
this.chartData["data"].filter((row: any) => row["page"] == this.currentPage)
```

we'll get a speed up of 1 order of magnitude  (as long as we don't introduce any more complexity in the function that adds the page numbers.)

The below attempt is of the same order of complexity as the existing `allDistinctColumnValues` function and looks like it returns the right collection of points but for some reason its rendering them all on the same axes so I've obviously not understood something, but maybe you will understand why and be able to implement this properly!

```
  const subplots = this.chartMetadata.subplots;
  let plotCount = 0;
  let pageNum = 1;
  if (subplots && this.chartData) {
      const distinct: any[] = [];
      const subplotsPerPage = this.chartMetadata.subplots!.subplotsPerPage;
      return this.chartData["data"].map((row: any) => {
          const distinctVal = row[subplots.distinctColumn];
          if (!distinct.includes(distinctVal)) {
              distinct.push(distinctVal);
              plotCount++;
              if (plotCount == subplotsPerPage * pageNum + 1) {
                  pageNum++;
                  this.totalPages = pageNum
              }
          }
          return {...row, page: pageNum}
      })
  }
  return this.chartData ? this.chartData["data"]: []
```

hillalex (changes requested):

src/app/static/src/app/components/genericChart/GenericChart.vue
EmmaLRussell: I guess it could go into the kotlin - it could catch the hintr data as it comes back and assign page numbers to the rows based on configured subplots per page and geo level. Seems quite complex, would be good to avoid if possible. 

So your function here that's appending the page numbers goes into a further computed prop `pagedChartData` or something and then it does the filtering to the page number in `chartDataPage`? That would avoid the nested loop for getting the individual pages by taking the hit to calculate the page for each row when chartData changes (when filters or dataSource change). 

I don't think this will eliminate the filter select lag, unfortuntately, since that's happening in the non-paged code too, but ti should improve it.

hillalex: I was imagining the server side approach would just be done in the R, presumably at hintr level.

The function here would replace `allDistinctColumnValues`  (although yes, would need a rename to `chartDataWithPageNums` or something) and is actually of the same order of complexity as that existing function and would be executed the same number of times, so this would only be a performance speed up. Yes, still need to look into that filtering logic and see if some optimisations can be made there! But this would avoid adding even more scripting time when paging.

hillalex: (`allDistinctColumnValues` has a `includes` nested in a `reduce` and this is an `includes` nested in a `map`)

src/app/static/src/app/components/genericChart/GenericChart.vue
```diff
@@ -32,6 +32,28 @@
                          </div>
                      </div>
                 </div>
+                <div v-if="totalPages > 1" id="page-controls" class="text-center mt-2">
+                    <button id="previous-page"
+                            class="btn btn-sm btn-red mr-2"
+                            v-translate:aria-label="'previousPage'"
+                            v-tooltip="translate('previousPage')"
+                            :disabled="currentPage <= 1"
```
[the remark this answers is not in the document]
EmmaLRussell: We're a bit inconsistent in the app, I think we tend to have pink disabled buttons in dialogs and grey on the page. So I'll go along with that, and change to grey here!

src/app/static/src/app/components/genericChart/GenericChart.vue
```diff
@@ -252,7 +295,46 @@
...
+            },
+            allDistinctColumnValues() {
+                const subplots = this.chartMetadata.subplots;
+                if (subplots && this.chartData) {
+                    return this.chartData["data"].reduce((distinct: string[], row: any) => {
+                        const distinctVal = row[subplots.distinctColumn];
+                        return distinct.includes(distinctVal) ? distinct : [...distinct, distinctVal]
+                    }, []);
+                } else {
+                    return null;
```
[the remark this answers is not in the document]
EmmaLRussell: This code now removed

src/app/static/src/app/components/genericChart/GenericChart.vue
```diff
@@ -252,7 +295,46 @@
                 return !this.chartData ||
                     !Object.values(this.chartData).some(e => e.length);
 
-            }
+            },
+            allDistinctColumnValues() {
```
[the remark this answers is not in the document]
EmmaLRussell: Hopefully it's a bit clearer now with the explicit addition of page numbers to data?

src/app/static/src/app/components/genericChart/GenericChart.vue
```diff
@@ -252,7 +295,46 @@
...
+                    return null;
+                }
+            },
+            pageDistinctColumnValues() {
+                if (this.chartData && this.chartMetadata.subplots && this.allDistinctColumnValues) {
+                    const subplotsPerPage = this.chartMetadata.subplots.subplotsPerPage;
+                    const start = (this.currentPage - 1) * subplotsPerPage;
+                    return this.allDistinctColumnValues.slice(start, start + subplotsPerPage);
+                } else {
+                    return null;
```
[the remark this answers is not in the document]
EmmaLRussell: This code now removed

src/app/static/src/tests/components/genericChart/genericChart.test.ts
EmmaLRussell: Tooltips are gone, so these tests now removed

src/app/static/src/app/components/genericChart/GenericChart.vue
EmmaLRussell: Outside the scope of this ticket, should be covered by: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2612

EmmaLRussell:
Thanks both. I have:

- Implemented Alex's suggestion for performance improvement by including page numbers in the data. I think the only significant change I made here was to allow for area rows to appear after an area which belongs to another page, hence replacing the `distinct` array with a dictionary, `distinctValuePageNums` to look up page numbers for distinct values.
- Made chart area scroll to top on page change - that seemed to make sense, that user would scroll to bottom of page, advance to next page and expect to be at the top of new page
- Hidden the y axis line when no chart values
- Made the paging buttons grey when disabled
- One reviewer was for keeping tooltips and one against, so I guess I have the casting vote - so I've taken them out.

hillalex (approved):
Definitely a bit more performant, and I've had another insight with respect to that. Looks like the data in the generic chart module is not frozen, so Vue is appending watchers to every row. It's because the freezing logic is only being applied [within the `onSuccess` callback in the api service](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/apiService.ts#L92) but the action which fetches the data is [not using the auto-committing callback](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/genericChart/actions.ts#L35). We might see as much as a halving of scripting time if we fix this (so applying the freezing logic in the api service in `_handleAxiosResponse` rather than `_onSucesss`)

Want to add this to the further ticket on performance that you're working on?

I suggest we also add a call to `freezeResponse` to the [action that fetches metadata](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/genericChart/actions.ts#L22). Not mega important but the metadata objects are iterated over in nested loops in `GenericChartTable.vue` so any shaving off of size might help. Looks like the `labelledData` computed value in that component is the main performance sink btw (aside from the plotly code itself), haven't looked at it properly but would be good to see if there's any way that part of the component can be re-engineered.

src/app/static/src/app/components/genericChart/GenericChart.vue
```diff
@@ -253,12 +295,39 @@
 
                     result[dataSourceId] = filterData(unfilteredData, filters, selectedFilterOptions);
                 }
-                return result;
+
+                if (result["data"]) {
```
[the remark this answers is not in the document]
EmmaLRussell: Yeah, that's because it was kind of a hack to get the subplots working. In theory, generic chart could support displaying multiple datasets - in practice this is in the only instance we have. Earlier incarnations of it didn't use subplot configuration, but once we needed that to add annotations etc into plotly, it was either maintain a dict of these things, or treat 'data' as a special case which can support subplots, and we went for the latter.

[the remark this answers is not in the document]
hillalex: Ok, I don't fully understand but I figured it was something like that!

LekanAnni (approved):
LGTM 👍

PR #629, opened 2021-11-19
Mrc 2784 - Add navigation buttons to bottom of every page

LekanAnni:
Adds navigation to steps except run model step.

Minor

## Checklist


src/app/static/src/app/components/Stepper.vue
```diff
@@ -37,7 +37,7 @@
                 <download-results v-if="isActive(7)"></download-results>
             </div>
         </div>
-        <template v-if="activeStep === 3">
+        <template v-if="activeStep !== 4">
```
EmmaLRussell, line 40: This isn't quite right, because it's bundling in that warning alert which does need to go with step3 only (model Options - where we want to show warnings at the bottom). 
So I think we need split this template in two, and include only the warnings component when activeStep === 3
and, under that, another template to include stepperNavigation (and the hr above it) if activeStep !== 4

EmmaLRussell (changes requested):

EmmaLRussell (approved):
LGTM

hillalex (approved):
perfect, just need to fix conflicts

PR #630, opened 2021-11-19
mrc-2797 Remove calibration plot when user changes calibration options

LekanAnni:
Displays calibrate plot only when calibration is complete

Patches 

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.73.2
+
+* Remove calibration plot when user changes calibration options
+
 # hint 1.73.1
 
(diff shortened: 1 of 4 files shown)


EmmaLRussell (changes requested):
Nice! It just needs to to the same thing for the 'Review results below' label too - since both of these use `calibrationPlotGenerated`, and nothing else does, it might be worth updating this computed property to check for `complete` as well.

EmmaLRussell (approved):
LGTM

hillalex (approved):
Great, just need to fix conflicts

PR #631, opened 2021-11-19
mrc-2783 Add explanatory description to input time series

EmmaLRussell:
## Description

Adds optional `description` to the generic chart config, and uses this to give guidance on chart line colour in the case of input time series. 

The description value in the config is a translation key (as are the other strings in config which are shown to users). 

Screenshot from 2021-11-19 17-06-53

## Type of version change
_Delete as appropriate_
Minor

## Checklist


src/app/static/src/app/components/genericChart/GenericChart.vue
```diff
@@ -15,6 +15,10 @@
                                 @update="updateSelectedFilterOptions(ds.config.id, $event)">
                     </filters>
                 </div>
+                <div id="chart-description"
+                     v-if="chartConfigValues.description"
```
LekanAnni, line 19: I don't think you need this if statement, description value is static and will always be true.

LekanAnni (changes requested):
Looks good.

When filtering the chart and no values are shown in red, the description still appears. I believe description should be shown based on available data in the displayed/filtered chart.

src/app/static/src/app/components/genericChart/GenericChart.vue
EmmaLRussell: It comes from the chart configuration metadata - in the case of the Input Time Series it will be true as currently configured, but the type is defined to have that value optional (see [here](https://github.com/mrc-ide/hint/pull/631/files#diff-b4e97f9053896e4bab011155c7ed2ad40cf453d0da07ccbb80ae5144a9424c23R238)) so it could be undefined.

EmmaLRussell:
> Looks good.
> 
> When filtering the chart and no values are shown in red, the description still appears. I believe description should be shown based on available data in the displayed/filtered chart.

I think we should still show the description if all values are black, so that the user knows that means no values are unexpectedly anomalous (also it would be a massive pain to filter based on that since that calculation is all done in plotly just now - not that that's a good reason not to do it!)  But I'll tweak the description a bit to include mention that black colour implies the opposite of red. 

Could also not show the description if empty dataset for current filters. But to be honest I think having the description appearing and disappearing might be more confusing to the user...

LekanAnni (approved):
LGTM 👍

hillalex (approved):
I think always showing the text is fine, I think preferable to err on the side of fewer moving parts!

Just needs conflicts fixing (what a pain this manual versioning is, perhaps we need to rethink?)

PR #633, opened 2021-11-22
mrc-2796 Bug: Output barchart x-axis should be ordered

Nicolas-Dolan:
## Description

Ensures that the order of data on the x axis of the barchart matches the order given in the filter (and therefore the Naomi sort order), even after the user toggles options. Nested x-axis filters, such as area, are displayed like this: `parent1 > parent1 children > parent2 > parent2 children > parent3`. This also preserves the groupings within the filter, such as keeping age increments of 10 years separate from age increments of 5 years (eg, `... , 25-34, 35-49, ... , 10-14, 15-19, ...`). 

Ordering the x-axis automatically causes the disaggregates to maintain the filter order in turn. However, for nest values, the dissagregates are being ordered `parent1 > parent2 > parent3 > parent1 children > parent2 children`. This still preserves the filter order but in a different manner to how the xaxis preserves it. The question is whether it is worth adding additional logic to ensure disaggregates are ordered in exactly the same fashion as the xaxis or whether it is fine as is and still makes sense to the user.

To test: proceed to review output and add/remove values from the xAxis filter. New values should be added in the order of the filter rather than at the end.

## Type of version change
Patches

## Checklist


src/app/static/src/app/store/plottingSelections/mutations.ts
```diff
@@ -26,7 +30,34 @@ export const mutations: MutationTree<PlottingSelectionsState> & PlottingSelectio
...
+                selectedFilterOptions[xAxisId]
+                    .some(option2 => option2.id === option.id) && updatedFilterOption.push(option)
+            })
+            const update = {...state.barchart, ...action.payload}
+            update.selectedFilterOptions[xAxisId] = updatedFilterOption
+            state.barchart = update
+        // if unable to do the above, just updates the barchart as normal
+        } else {
+            state.barchart = {...state.barchart, ...action.payload};
+        }
```
EmmaLRussell, line 60:
```suggestion
       // Get the list of filter option ids in their configured order, whether nested or not
        let originalFilterOptionsIds;
        if (originalFilterOptionsOrder && originalFilterOptionsOrder[0].children){
            const flattenedOptions = flattenOptions(originalFilterOptionsOrder)
            originalFilterOptionsIds = Object.keys(flattenedOptions)
        } else {
          originalFilterOptionsIds = originalFilterOptionsOrder.map((option: FilterOption) => o.id);
        }
        
        // Sort the selected filter values according to configured order
        if (originalFilterOptionsOrder && xAxisId && selectedFilterOptions && selectedFilterOptions[xAxisId]) {
           const updatedFilterOptions = selectedFilterOptions[xAxisId].sort((a: FilterOption, b: FilterOption) => {
             return originalFilterOptionsIds.indexOf(a.id) - originalFilterOptionsIds.indexOf(b.id);
           });
            
            const update = {...state.barchart, ...action.payload}
            update.selectedFilterOptions[xAxisId] = updatedFilterOption
            state.barchart = update
        // if unable to do the above, just updates the barchart as normal
        } else {
            state.barchart = {...state.barchart, ...action.payload};
        }
```

EmmaLRussell (commented):
This is the sort of thing I was talking about (haven't tested this!)

LekanAnni (changes requested):
Looking good.

The ordering is still not accurately achieved. For example, 50+, 60+ ,<1 , 1-4 to name a few are not ordered.

Nicolas-Dolan:
> Looking good.
> 
> The ordering is still not accurately achieved. For example, 50+, 60+ ,<1 , 1-4 to name a few are not ordered.

@LekanAnni the ordering isn't alpha-numeric but matches the order given in the filter. There are certain groupings (eg, 5 year vs 10 year increments), which mean it wouldn't be appropriate to order alpha-numerically.

EmmaLRussell:
> Ordering the x-axis automatically causes the disaggregates to maintain the filter order in turn. However, for nest values, the dissagregates are being ordered parent1 > parent2 > parent3 > parent1 children > parent2 children. This still preserves the filter order but in a different manner to how the xaxis preserves it. The question is whether it is worth adding additional logic to ensure disaggregates are ordered in exactly the same fashion as the xaxis or whether it is fine as is and still makes sense to the user.

I would say it would be better to be consistent here - I think the user would naturally expect bar groupings and indivudal bars to follow the same ordering rules. Not sure which would be the preferable order. I believe the data rows tend to be ordered by area id, hence all parents will come before any children - so the disaggregateBy style is more consistent with the order in the table. We should seek user input probably. But nested options are a special case, so I'd suggest leaving that for a separate ticket anyway.

Nicolas-Dolan:
> > Ordering the x-axis automatically causes the disaggregates to maintain the filter order in turn. However, for nest values, the dissagregates are being ordered parent1 > parent2 > parent3 > parent1 children > parent2 children. This still preserves the filter order but in a different manner to how the xaxis preserves it. The question is whether it is worth adding additional logic to ensure disaggregates are ordered in exactly the same fashion as the xaxis or whether it is fine as is and still makes sense to the user.
> 
> I would say it would be better to be consistent here - I think the user would naturally expect bar groupings and indivudal bars to follow the same ordering rules. Not sure which would be the preferable order. I believe the data rows tend to be ordered by area id, hence all parents will come before any children - so the disaggregateBy style is more consistent with the order in the table. We should seek user input probably. But nested options are a special case, so I'd suggest leaving that for a separate ticket anyway.

@EmmaLRussell These issues (the preferred ordering style and the discrepancy between xAxis and disaggregates) only concern nested options so do I need to make further changes in this ticket?

EmmaLRussell (commented):
Looks good to me. One thing that occurs is that it will have to do a lot of work to flatten those nested options on every update select, so it could be more efficient to make computed property 'flattenedFilterOptionIds' which contains that flattened list (or a dictionary of lists, keyed by filter id - in practice there will only be one such filter though, the area filter), which can be re-used on further updates.

src/app/static/src/tests/components/modelOutput/modelOutput.test.ts
```diff
@@ -304,8 +304,124 @@ describe("ModelOutput component", () => {
         expect(store.state.plottingSelections.colourScales.output).toBe(bubbleSizeScales);
     });
 
+    it("commits updated selections from barchart and orders them according to filter", () => {
```
EmmaLRussell, line 307: This test is a little hard to reason about. I think I see what you're doing, but it's a little counterintuitive, because you're setting up the expected selection order to the comit as a re-order of the order emitted from the barchart? I think it could be made more readable by just renaming 'updatedBarchartSelections' to 'expectedBarchartSelections' and either just including the expected options objects literally or else setting them as a copy of the order in `testBarchartFilters`  (which is closer to the logic of what's happening)

Similarly for the test of nested filters.

EmmaLRussell (commented):
Sorry, hit 'Submit' without meaning - just a comment on readability of tests.

EmmaLRussell:
> @EmmaLRussell These issues (the preferred ordering style and the discrepancy between xAxis and disaggregates) only concern nested options so do I need to make further changes in this ticket?

Not on this issue I would say, but see my other comments. Could you create a ticket for looking at that discrepancy for nested?

Nicolas-Dolan:
> > @EmmaLRussell These issues (the preferred ordering style and the discrepancy between xAxis and disaggregates) only concern nested options so do I need to make further changes in this ticket?
> 
> Not on this issue I would say, but see my other comments. Could you create a ticket for looking at that discrepancy for nested?

https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2851

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -223,6 +225,17 @@
...
+            },
+            flattenedFilterOptionIds(){
+                const ids: Dict<string[]> = {}
+                this.barchartFilters.forEach(filter => {
+                    if ((filter.options[0] as NestedFilterOption).children){
+                        ids[filter.id] = Object.keys(flattenOptions(filter.options))
+                    } else {
+                        ids[filter.id] = filter.options.map((option: FilterOption) => option.id)
+                    }
+                })
```
EmmaLRussell, line 237: Sorry, I really meant that this just needed to be for the nested filter options, not all filters. But, actually, thinking about it -if it's only ever required for the xAxis filter, it could just be `flattenedXAxisFilterOptionIds` and be recalculated only when x axis filter changes, whether that's nested or not. Might need to amend that later if we end up doing something similar with the disaggregateBy filter to make it consistent.

EmmaLRussell (approved):
Apologies, I didn't quite think through my last suggestion, so your new computed prop is doing more than it needs to I think (just needs to be for the x axis filter, right?) So you could scale that down to just that filter if you like.

LekanAnni (approved):
LGTM

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -226,15 +226,23 @@
...
+                        const filter = this.barchartFilters[index]
+                        if (filter.id === xAxisId){
+                            if ((filter.options[0] as NestedFilterOption).children){
+                                ids = Object.keys(flattenOptions(filter.options)) || []
+                                break
+                            } else {
+                                ids = filter.options.map((option: FilterOption) => option.id) || []
+                                break
+                            }
+                        }
```
EmmaLRussell, line 243: So it looks like here, you're looping through all the barchart filters in order to find the one which is the x axis and flatten that. It would be less verbose to ditch the for loop and just use find wouldn't it? Something like 
`const filter = barchartFilters.find((f: Filter) => f.id === xAxisId)`?

EmmaLRussell (approved):

PR #634, opened 2021-11-29
mrc-2815 validate send error report email field

LekanAnni:
This PR validates email field and some extra form validation to error report modal, the email address is used to send mail to users through Sharepoint flow. If valid email format is not provided, flow fails.

I have made the versioning a minor because of obvious changes in UI.

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/static/src/tests/components/errorReport.test.ts
```diff
@@ -19,6 +19,8 @@ describe("Error report component", () => {
 
     const generateErrorReport = jest.fn();
 
+    const pattern = RegExp("^\\w+([\\.-]?\\w+)*@\\w+([\\.-]?\\w+)*(\\.\\w{2,3})+$")
```
hillalex, line 22: This all looks great but can we have some tests that confirm that this regex does what we want? What sort of invalid emails are users adding? Because that might inform the regex, e.g. it might be enough to just check that the value provided has an "@" sign in it rather than worry too much about the exact structure of the email address. Technically there are some valid email address that would fail this regex e.g. "test+user@gmail.com", "test..user@gmail.com". 

Also might be useful to trim whitespace from the email field first because that's a really common typo and often hard for the user to spot what's wrong if they get the invalid feedback for " test.user@gmail.com" or "test.user@gmail.com ".

hillalex (changes requested):

src/app/static/src/tests/components/errorReport.test.ts
LekanAnni: > 

- I don't think these  "test+user@gmail.com", "test..user@gmail.com" are valid email prefixes. The format I adopted conforms with Microsoft email checks as we will be sending via sharePoint. I will write more tests.
- I will trim the email.

hillalex: They are definitely valid email address, (alex+hill@gmail.com and alex..hill@gmail.com both come to me actually, gmail is a bit funny like that!) Admittedly unlikely that anyone would have email addresses that look like that

hillalex (approved):
Great, thank you

Nicolas-Dolan (approved):
LGTM

PR #636, opened 2021-12-05
mrc-2801 Performance improvements for Input Time Series

EmmaLRussell:
## Description

This branch:
- freezes genericChart data and metadata so that Vue does not watch it for changes

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/app/store/genericChart/actions.ts
+++ b/src/app/static/src/app/store/genericChart/actions.ts
@@ -4,6 +4,7 @@ import {GenericChartState} from "./genericChart";
 import {GenericChartMutation} from "./mutations";
 import {DatasetConfig, Dict, GenericChartMetadata} from "../../types";
 import {DataExplorationState} from "../dataExploration/dataExploration";
+import {freezer} from "../../utils";
 
 export interface MetadataActions {
     getGenericChartMetadata: (store: ActionContext<GenericChartState, DataExplorationState>) => void
@@ -21,6 +22,7 @@ export const actions: ActionTree<GenericChartState, DataExplorationState> & Meta
         await api<GenericChartMutation, "">(context)
             .withSuccess(GenericChartMutation.GenericChartMetadataFetched)
             .ignoreErrors()
+            .freezeResponse()
             .get(`/meta/generic-chart`);
     },
     async getDataset(context, payload) {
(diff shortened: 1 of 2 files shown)


src/app/static/src/app/store/genericChart/actions.ts
```diff
@@ -29,15 +31,15 @@ export const actions: ActionTree<GenericChartState, DataExplorationState> & Meta
         await api<GenericChartMutation, GenericChartMutation>(context)
             .ignoreSuccess()
             .withError(GenericChartMutation.SetError)
-            .freezeResponse()
             .get(payload.url)
             .then((response) => {
                 if (response) {
+                    const frozenData = freezer.deepFreeze(response.data);
```
hillalex, line 37: I think it would be preferable to do this in the api client - would expect that calling `freezeResponse` means both that auto-comitted payloads are frozen but also that the returned `response` object is frozen (and fixing that would prevent this being missed again elsewhere.)

hillalex (changes requested):
I don't think using threads or timeouts are the way to go; like you say it doesn't actually change the wait time for results and make the code hard to reason about. 

The sluggish-ness is due to the `labelledData` computed method in the `GenericChartTable` component which has some serious nesting of loops. Like a recursive search within a `foreach` within a `map` and a `find` within a `foreach` within a `filter` within a `map`!

I suggest as a further ticket having a good look at whether that can be re-engineered using different data structures because ditching some of that nesting is going to be the only way to really improve the performance.

hillalex:
Oh and yes I had imagined that x axis ranges would come from hintr to get that performacne improvement, but appreciate that would also mean some changes to the structure of the metadata

Nicolas-Dolan (commented):
Works well and seems to be stable. Alex makes a good point about the loops within loops in the labelledData function so restructuring that would be worth tackling in another ticket. Perhaps these changes are a good start to build on though.

hillalex:
Am happy to have a go at re-writing that function, I quite like fiddling around with this stuff! In a separate branch I mean, I'm happy for these changes to be merged as is (pending requested change to freezing logic)

EmmaLRussell:
> Am happy to have a go at re-writing that function, I quite like fiddling around with this stuff! In a separate branch I mean, I'm happy for these changes to be merged as is (pending requested change to freezing logic)

I was going to have a play with it later today - I could do with the practice, but you could no doubt polish up my first attempt! Yeah, I'll do that in another branch.

src/app/static/src/app/apiService.ts
```diff
@@ -104,11 +103,12 @@ export class APIService<S extends string, E extends string> implements API<S, E>
     private _handleAxiosResponse(promise: Promise<AxiosResponse>) {
         return promise.then((axiosResponse: AxiosResponse) => {
             const success = axiosResponse && axiosResponse.data;
-            const data = success.data;
+            const finalResponse = this._freezeResponse ? freezer.deepFreeze(success) : success;
```
EmmaLRussell, line 106: I think this should suffice to freeze the entire response, which can then be returned, and also have its `data` passed to `_onSuccess`, which no longer needs to do its own freeze.

I've also removed manual freeze from the other actions which were doing it in modelRun and modelCalibrate.

LekanAnni (approved):
LGTM 👍

hillalex (approved):
Great, thanks

PR #637, opened 2021-12-07
mrc-2811 Use format from input time series metadata  in plots

EmmaLRussell:
## Description
This branch contains the HINT changes required to format values in the Input Time Series charts as percentages where appropriate (e.g. ART sex ratio, ANC prevalence) as shown below:
Screenshot from 2021-12-10 17-31-19

Format metadata  has been added to the data received from hintr (see mrc-2810) - thid is at the plot_type column option level, so there is now a 'format' property in the 'values' in the 'plot_type' column metadata, containing the Numeral.js format which should be used for the plot type.

This branch:
- adds 'valueFormatColumn' to the generic chart metadata for the input time series to specify the column to use for formats
- adds a util to convert Numeral.js format (standard in hintr metadata) to d3 format required for plotly. Currently we only support numeric and percentage formats. 
- in GenericChart, extracts `valueFormatColumn` from the metadata, uses it to fetch the format for the currently selected plot type, converts it to a d3 format and includes it in the layout data passed to plotly, as 'yAxisFormat' property
- in the input time series jsonata, sets `yAxisFormat` as tick format for the y axes of the subplots

As for subplots, while genericChart does in theory support multiple data sources, here we assume we only need to format y axis values for the first data source. 

## Type of version change
_Delete as appropriate_
Minor

## Checklist


src/app/static/jest.config.js
```diff
@@ -22,6 +22,9 @@ module.exports = {
         "^.+\\.ts?$": "ts-jest",
         "^.+\\.js$": "<rootDir>/node_modules/babel-jest"
     },
+    "moduleNameMapper": {
+        'd3-format': '<rootDir>/node_modules/d3-format/dist/d3-format.min.js',
+    },
```
EmmaLRussell, line 27: Needed to import d3-format for the tests

Nicolas-Dolan (approved):

LekanAnni (approved):
LGTM 👍

PR #638, opened 2021-12-07
Mrc-2802 Move change Language action to DE state

LekanAnni:
- I have added changeLanguage method to DE actions and left the changeLanguage in root because changeLanguage method refers to 'modelCalibrate' in root which does not exist in DE state so they can't share the action.
- I created language dropDown menu view to be reused by DE and RunModel States
- DE helpFile refers to old filename, so I have added an opportunistic fix to make helpFile location generic.

move changeLanguage action to DE state and ensure it works for both root and DE states.

## Type of version change
Minor 

## Checklist


src/app/static/src/app/components/header/DataExplorationLanguageMenu.vue
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    interface Computed {
+        currentLanguage: string
+    }
+
+    export default Vue.extend<unknown, Methods, Computed, unknown>({
+        computed: {
+            currentLanguage: mapStateProp<DataExplorationState, string>(null,
+                (state: DataExplorationState) => state.language.toUpperCase())
```
hillalex, line 24: So the only difference between this and the `LanguageMenu.vue` component is that this references `DataExplorationState` instead of `RootState`. But `RootState` inherits from `DataExplorationState`, so it would be fine to just use the same component for both. Because both Vuex stores now have an action named "changeLanguage" this component will work just fine for both apps!

src/app/static/src/app/components/header/DataExplorationHeader.vue
```diff
@@ -48,11 +48,10 @@
         computed: {
             helpFilename: mapStateProp<DataExplorationState, string>(null,
                 (state: DataExplorationState) => {
-                    let filename = "Naomi-basic-instructions.pdf";
                     if (state.language == Language.fr) {
-                        filename = "Naomi-instructions-de-base.pdf";
+                        return HelpFile.french;
                     }
-                    return filename;
+                    return HelpFile.english;
```
hillalex, line 54: this is a nice little util

hillalex (changes requested):
This all looks great except that I don't think we need a new component for the language menu itself...

src/app/static/src/app/components/header/LanguageMenu.vue
```diff
@@ -1,22 +1,14 @@
...
-            EN
-        </a>
-        <a class="dropdown-item" href="#" v-on:mousedown="() => changeLanguage('fr')">
-            FR
-        </a>
-        <a class="dropdown-item" href="#" v-on:mousedown="() => changeLanguage('pt')">
-            PT
-        </a>
-    </drop-down>
+    <language-menu-view :change-language="changeLanguage"
```
hillalex, line 2: Couple of small edits - is this inner component now needed? I'd revert this change to keep the component as it was originally.

And below when computing the current language, although this will work, we should technically use the `DataExplorationState`:

```
currentLanguage: mapStateProp<DataExplorationState, string>(null,
                (state: DataExplorationState) => state.language.toUpperCase())
```

hillalex (changes requested):

src/app/static/src/app/store/dataExploration/actions.ts
```diff
@@ -0,0 +1,34 @@
...
+        if (rootState.baseline?.iso3) {
+            actions.push(dispatch("metadata/getPlottingMetadata", rootState.baseline.iso3));
+        }
+
+        if (Object.keys(rootState.genericChart.datasets).length > 0) {
+            actions.push(dispatch("genericChart/refreshDatasets"));
+        }
+
+        await Promise.all(actions);
+        commit({type: LanguageMutation.SetUpdatingLanguage, payload: false});
```
EmmaLRussell, line 31: It is a pity to have so much shared code between this and the corresponding action in `root`- I can see that tripping us up if we ever need to change the shared logic. Could have a shared util which each calls, which takes the rootState, commit and dispatch as parameters, along with a flag to indicate whether all root actions need to be done? But maybe that's a bit convoluted.

src/app/static/src/app/utils.ts
```diff
@@ -281,4 +276,9 @@ export function getFilenameFromImportUrl(url: string) {
 export function getFilenameFromUploadFormData(formdata: FormData) {
     const file = formdata.get("file");
     return (file as File).name;
+}
+
+export enum HelpFile {
```
EmmaLRussell, line 281: Nice. I think we're expecting DE mode help files at some point, so we can expand this to include those then.

src/app/static/src/tests/components/header/dataExplorationHeader.test.ts
```diff
@@ -58,13 +60,13 @@ describe(`Data Exploration header`, () => {
...
         const loginInfo = wrapper.find("span");
-        expectTranslated(loginInfo, "Logged in as someone@email.com",
+        expectTranslatedWithStoreType(loginInfo, "Logged in as someone@email.com",
             "Connecté en tant que someone@email.com", "Sessão iniciada como someone@email.com", store);
     });
 
     it("renders language menu", () => {
         const wrapper = getWrapper()
-        expect(wrapper.findAll(LanguageMenu).length).toBe(1);
+        expect(wrapper.find("language-menu-stub").exists()).toBe(true)
```
EmmaLRussell, line 69: `exists` is neater than length check on findAll, but could you not use `find(LanguageMenu)` rather than `find("language-menu-stub")`?

EmmaLRussell (commented):
Looks good, works great.

src/app/static/src/app/store/dataExploration/actions.ts
LekanAnni: Well, better than duplication, I had prefer to add it to utils.

src/app/static/src/app/store/language/actions.ts
```diff
@@ -3,8 +3,39 @@ import i18next from "i18next";
...
 import {Language} from "../translations/locales";
 import {TranslatableState} from "../../types";
+import {DataExplorationState} from "../dataExploration/dataExploration";
 
 export async function changeLanguage<T extends TranslatableState>({commit}: ActionContext<T, T>, lang: Language) {
     await i18next.changeLanguage(lang);
     commit({type: LanguageMutation.ChangeLanguage, payload: lang})
 }
+
+export const ChangeLanguageAction = async (context: ActionContext<any, any>, payload: Language, root = false) => {
```
hillalex, line 13: There's actually a nice way to do this in Typescript without losing typings and without using a boolean argument here:

```
function isRoot(object: DataExplorationState): object is RootState {
    return !object.dataExplorationMode
}

export const ChangeLanguageAction = async (context: ActionContext<DataExplorationState, DataExplorationState>,
                                           payload: Language) => {

    const {commit, dispatch, rootState} = context;

    if (rootState.language === payload) {
        return;
    }

    commit({type: LanguageMutation.SetUpdatingLanguage, payload: true});

    await i18next.changeLanguage(payload);
    commit({type: LanguageMutation.ChangeLanguage, payload: payload})

    const actions: Promise<unknown>[] = [];

    if (rootState.baseline?.iso3) {
        actions.push(dispatch("metadata/getPlottingMetadata", rootState.baseline.iso3));
    }

    if (isRoot(rootState) && rootState.modelCalibrate.status.done) {
        actions.push(dispatch("modelCalibrate/getResult"));
    }

    if (Object.keys(rootState.genericChart.datasets).length > 0) {
        actions.push(dispatch("genericChart/refreshDatasets"));
    }

    await Promise.all(actions);
    commit({type: LanguageMutation.SetUpdatingLanguage, payload: false});
}
```

hillalex (changes requested):
Looks good - there's actually a nice way of checking whether an object implements an interface and getting Typescript to recognise that, see inline suggestion.

src/app/static/src/app/store/language/actions.ts
LekanAnni: > There's actually a nice way to do this in Typescript without losing typings and without using a boolean argument here:
> 
> ```
> function isRoot(object: DataExplorationState): object is RootState {
>     return !object.dataExplorationMode
> }
> 
> export const ChangeLanguageAction = async (context: ActionContext<DataExplorationState, DataExplorationState>,
>                                            payload: Language) => {
> 
>     const {commit, dispatch, rootState} = context;
> 
>     if (rootState.language === payload) {
>         return;
>     }
> 
>     commit({type: LanguageMutation.SetUpdatingLanguage, payload: true});
> 
>     await changeLanguage<DataExplorationState>(context, payload);
> 
>     const actions: Promise<unknown>[] = [];
> 
>     if (rootState.baseline?.iso3) {
>         actions.push(dispatch("metadata/getPlottingMetadata", rootState.baseline.iso3));
>     }
> 
>     if (isRoot(rootState) && rootState.modelCalibrate.status.done) {
>         actions.push(dispatch("modelCalibrate/getResult"));
>     }
> 
>     if (Object.keys(rootState.genericChart.datasets).length > 0) {
>         actions.push(dispatch("genericChart/refreshDatasets"));
>     }
> 
>     await Promise.all(actions);
>     commit({type: LanguageMutation.SetUpdatingLanguage, payload: false});
> }
> ```

Never seen that used before, really cool.

EmmaLRussell (approved):

LekanAnni:
Thanks 🙏

PR #639, opened 2021-12-08
pass query param to hintr

hillalex:
## Description

Adds back-end logic for accepting an optional "strict"  query parameter when validating data, as per https://github.com/mrc-ide/hintr/pull/317

Is backwards compatible with the front-end (as defaults to strict mode) but will require the above `hintr` branch to be merged, as `hintr` throws an error if an unknown query parameter is passed.

## Type of version change
None (no visible changes until the front-end is added)

## Checklist


hillalex:
Little bit stumped by the codecov complaint - it claims the `HintrController` constructor isn't covered, which seems impossible :woman_shrugging: I'm confident all code paths are in fact tested!

src/app/testdata/anc-lax.csv
```diff
@@ -0,0 +1,257 @@
+area_id,age_group,year,anc_clients,anc_known_pos,anc_already_art,anc_tested,anc_tested_pos
```
EmmaLRussell, line 1: How is this file lax? Maybe a clue in the filename would be helpful?

EmmaLRussell (commented):
LGTM

This should also work when importing files from ADR shouldn't it, since the ADRController also inherits from HintrController. It would be nice to have an integration test for that, but would be tedious to set up as it would require having another dataset with a lax file on the adr dev server. Could add a unit test at least I guess, thought it would really just be testing the same inherited behaviour as the DiseaseControllerTests...

EmmaLRussell:
> Little bit stumped by the codecov complaint - it claims the `HintrController` constructor isn't covered, which seems impossible woman_shrugging I'm confident all code paths are in fact tested!

Yeah me too, looked spurious to me

hillalex:
> LGTM
> 
> This should also work when importing files from ADR shouldn't it, since the ADRController also inherits from HintrController. It would be nice to have an integration test for that, but would be tedious to set up as it would require having another dataset with a lax file on the adr dev server. Could add a unit test at least I guess, thought it would really just be testing the same inherited behaviour as the DiseaseControllerTests...

Ah thaks for pointing that out, yes for completeness I'd like tests on the ADRController too, will add unit tests at least :+1:

EmmaLRussell (approved):

LekanAnni (approved):
LGTM

PR #640, opened 2021-12-08
mrc-2803 Add online-support-menu component to DE

LekanAnni:
Added online-support-menu component to DE

Menu  Items:
-Report issues
-Accessibility

Report issues not yet working but menu is added here.

## Type of version change

Minor

## Checklist


src/app/static/src/app/components/header/DataExplorationSupportMenu.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+    <div id="divclass">
+        <drop-down text="support" :right="true" :delay="true" style="flex: none">
+            <a class="dropdown-item"
+               v-translate="'reportIssues'">
```
hillalex, line 5: What was the reasoning behind having a different title for this to the other support menu (that one says "Troubleshooting request" rather than "Report issues")? I presume you're following the mockups here, but reckon we should keep this language consistent one way or the other - maybe one to run past Rachel or Rob to pin down the preferred terminology?

hillalex (commented):
Nice one. Just a query about terminology

src/app/static/src/app/components/header/DataExplorationSupportMenu.vue
LekanAnni: They are both changing to Report issues, should update RootState menu.

EmmaLRussell: Wasn't it the other way round? "Report issues" got renamed to "Troubleshooting request" in this ticket: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2731 from November.

LekanAnni: Confusing, I will use "Troubleshooting request" for both.

EmmaLRussell (approved):
LGTM. Not sure why codecov is complaining, since there is a test of the 'support' translation

hillalex (approved):
Looks good; I also can't make sense of that codecov complaint!

PR #641, opened 2021-12-08
mrc-2740 validate inputs in data exploration app

hillalex:
## Description

Builds on https://github.com/mrc-ide/hint/pull/639 which should be merged first and then the base of this PR set to `master`. I'll increment version pending review, as it will no doubt become out of date in the meantime!

This adds validation logic for the data exploration app. This differs in the following ways from primary app validation:

* `strict=false` is passed to all survey & program data endpoints
* the shape file is the only required baseline file
* baseline consistency is re-validated if a baseline file is deleted (this isn't needed in the main app because all baseline files are required and re-validation happens after an upload)
* deleting baseline files does not automatically remove survey & program files (side note - is this actually desirable in the main app?) 
* default plotting metadata is requested by the app on mount (country specific metadata will still be requested if/when a pjznz file is uploaded)

## Type of version change
Minor

## Checklist


LekanAnni (changes requested):
Looking good.

Data source should switch to available SAP data, at the moment if selected data source is deleted in baseline, it doe not auto select a different data source but errored with “1 unknown”.

When population file is removed, in the root app it does remove Survey file but this behaviour is not replicated here. does't the map need population ?

This branch will need to be merged with the branch where changeLanguage is implemented, we can then verify if metadata retrieves different languages when changeLanguage action is triggered.

hillalex:
> Data source should switch to available SAP data, at the moment if selected data source is deleted in baseline, it doe not auto select a different data source but errored with “1 unknown”.

Thanks, great catch, will fix here.

> When population file is removed, in the root app it does remove Survey file but this behaviour is not replicated here. does't the map need population ?

Apparently population is optional, so here we simply re-validate the files after a deletion. But you have made me realise that we should be re-validating the survey&Program files as well as just the baseline, will add that in!

hillalex:
Ok, ready for review again. Just a note to please also review this one (although merging both will have to wait for the hintr branch to be merged) https://github.com/mrc-ide/hint/pull/639

An unrelated test has started failing - to do with missing warnings from hintr. I'm guessing maybe a merge fail in the hintr branch I'm targetting here. In any case this shouldn't be merged until that hintr branch is merged.

EmmaLRussell (commented):
Looks good. I do think we need to support Survey file not being provided, as mentioned in Epic.

I did manage to confuse it by being a bit convoluted:
- upload DRC shape file
- upload DRC survey file
- Continue and look at map
- back 
- upload Mozambique survey - validation errors as expected, not compatibe with DRC shape
- upload Mozambique shape file. Errors cleared, nice. 
- Continue - signs of confusion (screenshot of this, and what it looks like if just upload those two files to start with). 
Screenshot from 2021-12-09 17-43-09
Screenshot from 2021-12-09 17-44-08

This might be an issue with the main app but possibly more likely to come up in DE mode where there are no projects to keep things separate. 

> deleting baseline files does not automatically remove survey & program files (side note - is this actually desirable in the main app?)

I guess that would be fine, since we revalidate SAP when a new baseline file is uploaded anyway.

hillalex:
> I did manage to confuse it by being a bit convoluted:
> This might be an issue with the main app but possibly more likely to come up in DE mode where there are no projects to keep things separate.

Actually there's an even quicker way to generate this bug:
* upload survey file
* upload shape file
* navigate and see bug

Looks like the *order* in which files are uploaded matters. I can in fact replicate this on prod but it requires a bit of an un-natural order of behaviour which is probably why it's never been reported - upload pjnz, population, survey, then shape file. The cause is that uploading a shape file triggers a re-validation of the inputs and non-existent program and anc files return 200 responses, upon which the selected data type is set to program/anc. Fixing here will also fix the bug for the main app.

EmmaLRussell (commented):
This is looking good for most of the issues raised, except that I can still trigger the 'No data are selected..' bug by going through the process of uploading survey then shape twice (first for DRC then for Mozambique). I'll have a look into this. 

It looks like language is refreshed as expected - or at least as much as the main app does (some untranslated labels still visible in data).

EmmaLRussell:
> This is looking good for most of the issues raised, except that I can still trigger the 'No data are selected..' bug by going through the process of uploading survey then shape twice (first for DRC then for Mozambique). I'll have a look into this.
> visible in data).

So I think that the reason this bug reappeared is actually for a quite different reason than the original survey-then-shape bug. What I'm seeing now happens quite specifically if you update from DRC data to Mozambique data, and I think it's because DRC has 4 area levels (with ids 0-3), whereas Mozambique only has 3. The default selection for DRC is 3, but that selection is not updated when we switch to Mozambique, which has no level 3 - hence no data displayed as filtered data is empty when filtering to a level which does not exist. I think we should be able to fix this by clearing the plottingSelections for the choropleth when change to a new shape file. Output selections would already be cleared (in the main app) when update shape as part of `ResetOutputs` so it should just be the input data which is affected. I'll check ITS data too (when you have ART/ANC as well as Survey).

EmmaLRussell:
I've made the following changes:
- clear SAPChoroplethSelections when update shape to avoid the area level bug (e.g. when load Mozambique data after DRC)
- removed the clear of genericChart data on survey delete - as generic chart data does not currently involve survey data, only ART and ANC
- updated version

LekanAnni (changes requested):
Looking great !

All works well but a tiny glitch.

when validating input files, continue button should be disabled.

EmmaLRussell:
> Looking great !
> 
> All works well but a tiny glitch.
> 
> when validating input files, continue button should be disabled.

Excellent point @LekanAnni  - and it actually applies to the main mode as well as DE, and it's not just during validation but during upload too (which probably takes longer). It's a bit more noticeable in DE mode since you're more likely to be valid before starting to upload an optional file and notice that Continue is enabled while doing so. 

Baseline state has a 'validating' flag (but not 'uploading'), surveyAndProgram has neither. They both extend ReadyState, but the ready flag is really just for initialisation. I'd suggest perhaps giving both an 'updating' flag which is set at the start of action both while uploading and validating, and Continue button disabled if either state is updating. We can implement that for DE mode in this branch - would suggest creating another ticket for using the new flags in main app. Does that sound reasonable? Happy for you to make those changes if you'd like, I think this ticket is a team effort now Alex is on hols, and I see you've just submitted your other ticket.

EmmaLRussell (approved):
I'd be happy to merge this now, with fix for disabling Continue button implemented in a separate ticket.

LekanAnni (approved):

LekanAnni:
> > Looking great !
> > All works well but a tiny glitch.
> > when validating input files, continue button should be disabled.
> 
> Excellent point @LekanAnni - and it actually applies to the main mode as well as DE, and it's not just during validation but during upload too (which probably takes longer). It's a bit more noticeable in DE mode since you're more likely to be valid before starting to upload an optional file and notice that Continue is enabled while doing so.
> 
> Baseline state has a 'validating' flag (but not 'uploading'), surveyAndProgram has neither. They both extend ReadyState, but the ready flag is really just for initialisation. I'd suggest perhaps giving both an 'updating' flag which is set at the start of action both while uploading and validating, and Continue button disabled if either state is updating. We can implement that for DE mode in this branch - would suggest creating another ticket for using the new flags in main app. Does that sound reasonable? Happy for you to make those changes if you'd like, I think this ticket is a team effort now Alex is on hols, and I see you've just submitted your other ticket.

https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2850

PR #642, opened 2021-12-08
mrc-2830 Improve performance of labelledData method in GenericChartTable

EmmaLRussell:
## Description

This makes two changes to improve performance:

- the biggest difference comes from not using `findPath` to get the area hierarchy string per row (which means traversing the hierarchy for every row). Instead, a new computed property saves a dictionary of column value ids to hierarchy strings for each 'hierarchy column' which is configured (in practice, we only have one, the area hierarchy). This is done at the level of the dataset, so does not need to be recalculated when filters change, but does when changing ART to ANC. This computed property is quick to calculate, ~5ms for me.
- uses a dictionary of ids to labels for each column which is configured to display labels, rather than having to find labels in the column metadata per row

Together, these have reduced the time taken for the labelledData method for DRC data at district level from ~1500ms to ~160ms.

Any ideas for additional improvements? We could potentially keep some of these intermediary lookup data structures in the state, to save components having to recreate them every time they are re-rendered or have their selected data change. But that feels like it could be part of a more general refactor of data to improve efficiency in the whole app. (We should schedule a discussion of our 2022 wishlist in the new year, to include that sort of thing.)

## Type of version change
_Delete as appropriate_

None

## Checklist

### DIFF
--- a/src/app/static/src/app/components/genericChart/GenericChartTable.vue
+++ b/src/app/static/src/app/components/genericChart/GenericChartTable.vue
@@ -101,33 +101,12 @@
                 // associated labels
                 const columnsDict = this.columns.reduce((dict, column) => ({...dict, [column.id]: column}), {} as Dict<GenericChartColumn>);
 
-                // Convert each label column's filter options to a dictionary of id to labels so we do not need to find values'
-                // labels in the options array per row
-                const configLabelColumns = this.tableConfig.columns.filter(column => column.data.labelColumn);
-                const reduceConfigLabelColumnsOptionsToDict = (dict: Dict<Dict<string>>, columnConfig: GenericChartTableColumnConfig) => {
-                    const key = columnConfig.data.labelColumn!;
-                    // If key is not already in dictionary, and options are an array (not always the case for hierarchies)
-                    if (!dict[key] && columnsDict[key].values.reduce) {
-                        const filterOptions = columnsDict[key].values;
-                        const labelDict = filterOptions.reduce((innerDict, option) => {
-                            innerDict[option.id] = option.label;
-                            return innerDict;
-                        }, {} as Dict<string>);
-                        dict[key] = labelDict;
-                    }
-                    return dict;
-                };
-                const columnValueLabels = configLabelColumns.reduce(reduceConfigLabelColumnsOptionsToDict, {} as Dict<Dict<string>>);
-
                 const result = this.filteredData.map(row => {
                     const friendlyRow = {...row};
-                    // Check the table configuration for each configured column, and replace data values
-                    // with friendly values (labels) taken from columnValueLabels where configured to do so - i.e.
-                    // when there is a 'labelColumn' value
-                    configLabelColumns.forEach(columnConfig => {
-                        const columnId = columnConfig.data.columnId;
-                        const rowValue = row[columnId];
-                        friendlyRow[columnId] = columnValueLabels[columnConfig.data.labelColumn!][rowValue];
+                    this.tableConfig.columns.filter(column => column.data.labelColumn).forEach(columnConfig => {
+                        const column = columnsDict[columnConfig.data.labelColumn!];
+                        const friendlyValue = column.values.find(value => value.id == row[columnConfig.data.columnId])?.label;
+                        friendlyRow[columnConfig.data.columnId] = friendlyValue;
                     });
 
                     // Include additional fields for any columns configured to include hierarchy values - use the
--- a/src/app/static/src/app/components/genericChart/GenericChartTable.vue
+++ b/src/app/static/src/app/components/genericChart/GenericChartTable.vue
@@ -96,17 +96,21 @@
                return result;
             },
             labelledData() {
-                console.log("started labelled data:" +  new Date().toISOString())
                 // Build a dictionary of the columns defined in the fetched dataset, which include all values and
                 // associated labels
                 const columnsDict = this.columns.reduce((dict, column) => ({...dict, [column.id]: column}), {} as Dict<GenericChartColumn>);
 
                 const result = this.filteredData.map(row => {
-                    const friendlyRow = {...row};
-                    this.tableConfig.columns.filter(column => column.data.labelColumn).forEach(columnConfig => {
-                        const column = columnsDict[columnConfig.data.labelColumn!];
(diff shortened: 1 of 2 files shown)


src/app/static/src/app/components/genericChart/GenericChartTable.vue
```diff
@@ -72,30 +69,80 @@
...
+               });
+               return result;
+            },
             labelledData() {
                 // Build a dictionary of the columns defined in the fetched dataset, which include all values and
                 // associated labels
                 const columnsDict = this.columns.reduce((dict, column) => ({...dict, [column.id]: column}), {} as Dict<GenericChartColumn>);
-                return this.filteredData.map(row => {
+
+                // Convert each label column's filter options to a dictionary of id to labels so we do not need to find values'
```
hillalex, line 103: Have you bench-marked this change?

hillalex (commented):
Nice, moving that `findPath` method out of the loop seems like a mssive win! The other change is less obvious to me; have you benchmarked it?

src/app/static/src/app/components/genericChart/GenericChartTable.vue
EmmaLRussell: Yes, it made a small difference ~20ms - maybe not worth it for the increased obfuscation?

src/app/static/src/app/components/genericChart/GenericChartTable.vue
```diff
@@ -72,30 +69,80 @@
...
+                // labels in the options array per row
+                const configLabelColumns = this.tableConfig.columns.filter(column => column.data.labelColumn);
+                const columnValueLabels = configLabelColumns.reduce((dict, columnConfig) => {
```
LekanAnni, line 106: For readability, can we have reduce function separated in such a way that it makes the code easy to read.

`function reducer(previousValue, currentValue) { doSomething }`
`configLabelColumns.reduce(reducer)`

LekanAnni (changes requested):
Looking good and complex.

See inline comment.

src/app/static/src/app/components/genericChart/GenericChartTable.vue
```diff
@@ -72,30 +69,80 @@
...
+                    // If key is not already in dictionary, and options are an array (not always the case for hierarchies)
+                    if (!dict[key] && columnsDict[key].values.reduce) {
+                        const filterOptions = columnsDict[key].values;
+                        const labelDict = filterOptions.reduce((dict, option) => ({
+                            ...dict,
+                            [option.id]: option.label
+                        }), {} as Dict<string>);
+
+                        return {
+                            ...dict,
```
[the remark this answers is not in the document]
hillalex: Just clocked that we clone every row in the inner loop (line 126) which will be slowing things down quite a bit but I can't really think of any way around that. Might be worth looking at the chart component again and just checking for any other instances of this slight anti-pattern.

src/app/static/src/app/components/genericChart/GenericChartTable.vue
hillalex: Hmm, this suprises me actually - array lookup is usually quicker when number of items is small. Might be worth just double checking as like you say, this is a bit hard to read! But 20ms is non trivial so if that's correct then I'm in favour of keeping it.

hillalex (commented):
A few queries

src/app/static/src/app/components/genericChart/GenericChartTable.vue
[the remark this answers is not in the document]
EmmaLRussell: Yeah, we can't really edit the original data which is in the state. The chart does not do this per row clone, as far as I can tell.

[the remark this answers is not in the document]
EmmaLRussell: Actually, this isn't true! The chart *does* copy and modify rows, in order to add page numbers. But we don't want the chart and table to be sharing the same copied rows since the chart may need the 'unfriendly' values. But I think I'll try making the table not copy the whole row, but only the columns it needs.

src/app/static/src/app/components/genericChart/GenericChartTable.vue
EmmaLRussell: You're completely right, I tried again with the latest version of everything and actually this only makes a marginal difference ~1ms or so. I think I must have run the two versions before at different times with different loads on my machine. OK, I'll take this out, but see if I can do anything about the row copy.

src/app/static/src/app/components/genericChart/GenericChartTable.vue
```diff
@@ -72,30 +69,60 @@
...
+                   if (option.children) {
+                       option.children.forEach((child) => addPathsToDict(child as NestedFilterOption, path, dict));
+                   }
+               };
+
+               const result = {} as Dict<Dict<string>>;
+               this.columnsWithHierarchy.forEach((columnConfig) => {
+                   const columnId = columnConfig.data.hierarchyColumn!;
+                   const pathsDict = {};
+                   const column = this.columns.find((column) => column.id == columnId)!;
```
LekanAnni, line 90:
```suggestion
                   const column = this.columns.find((column) => column.id === columnId)!;
```

src/app/static/src/app/components/genericChart/GenericChartTable.vue
```diff
@@ -72,30 +69,60 @@
...
-                        const friendlyValue = column.values.find(value => value.id == row[columnConfig.data.columnId])?.label;
-                        friendlyRow[columnConfig.data.columnId] = friendlyValue;
+
+                const result = this.filteredData.map(row => {
+                    const friendlyRow = {} as Dict<unknown>;
+                    this.tableConfig.columns.forEach(columnConfig => {
+                        const columnId = columnConfig.data.columnId;
+                        if (columnConfig.data.labelColumn) {
+                            const column = columnsDict[columnConfig.data.labelColumn!];
+                            const friendlyValue = column.values.find(value => value.id == row[columnConfig.data.columnId])?.label;
```
LekanAnni, line 109:
```suggestion
                            const friendlyValue = column.values.find(value => value.id === row[columnConfig.data.columnId])?.label;
```

LekanAnni (approved):
Looks good, works well. 👍 

see inline for some tiny suggestions.

hillalex (approved):

PR #643, opened 2021-12-09
mrc-2833 unfreeze default barchart selections so they can be mutated

EmmaLRussell:
## Description

This bug seems to have resulted resulted from the recent change to apiService freezeResponse, which means that the enitre response is frozen when returned from the Promise, and not just the data passed to `withSuccess`. 

Errors ("can't define array index property past the end of an array with non-writeable length") were being thrown when user changed barchart selections (adding an age group to default selections) since the entire selections object was frozen and could not be mutated. 

I'm surprised that this wasn't happening before. Since the modelCalibrate action was manually freezing the data returned it ought to have had the same problem. Previously, the spread operator was being used to set the plottingSelections selectedFilterOptions, but the arrays of options themselves were not being unfrozen. So in the barchart where it would (I think) push the new value to the options array, it's strange that this issue was not happening before. Maybe it was?

Also - this behaviour seems to imply that the original options arrays are being modified in the barchart component, which really shouldn't be happening. This is implied since, the mutation, which should be making the change to state, replaces the whole selections object - and that top level mutation should not cause this error. In FilterSelect, the `value` passed in is, I think that original array. That is set as `selected` in data and is then modified when selections change:
https://github.com/reside-ic/vue-charts/blob/master/src/bar/FilterSelect.vue#L68
(Here's that value as it is passed in from the barchart:
https://github.com/reside-ic/vue-charts/blob/master/src/bar/BarChartWithFilters.vue#L223 )
So to fix this properly I think will mean ensuring that the original selections array isn't edited in the barchart. Does that make sense?

The change I've made here is to accept the frozen response from api service and make a deep-thawed version of default selections to give to plotting selections. The alternative would be to not freezeResponse through the service, but to manually freeze the bits of response which did need to be frozen in the modelCalibrate action - I think it's probably better to specifically not freeze though. And actually we should be able to take out this stopgap fix when we decouple new selections arrays in the barchart. 

## Type of version change
_Delete as appropriate_

Patches

## Checklist


Nicolas-Dolan (commented):
Works well, but will this not need tests and version increments?

EmmaLRussell:
> Works well, but will this not need tests and version increments?

Yeah, should do a version increment actually, since it does change behaviour. Testing is kind of tricky, but yeah, I guess we could just test that the resulting selections can be modified.

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -4,9 +4,10 @@ import {DynamicFormData, DynamicFormMeta} from "@reside-ic/vue-dynamic-form";
 import {api} from "../../apiService";
 import {RootState} from "../../root";
 import {ModelCalibrateMutation} from "./mutations";
-import {ModelResultResponse, ModelStatusResponse, ModelSubmitResponse} from "../../generated";
+import {FilterOption, ModelResultResponse, ModelStatusResponse, ModelSubmitResponse} from "../../generated";
 import {freezer} from "../../utils";
```
LekanAnni, line 8: unused import

LekanAnni (changes requested):
I think it will be more natural to remove .freezeResponse() from the api call and freeze the bit that needs freezing.

Nicolas-Dolan (approved):
Works well for me. I think taking this whitelist approach with regard to properties to be unfrozen is the way to go.

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -75,13 +76,19 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
 
                 if (data && data.plottingMetadata && data.plottingMetadata.barchart.defaults) {
                     const defaults = data.plottingMetadata.barchart.defaults;
+                    const unfrozenDefaultOptions= Object.keys(defaults.selected_filter_options)
```
hillalex, line 79: So is it only the `response.data` object that we want frozen here? In which case I agree with Lekan that simply removing the `freezeResponse` call and doing `const data = freeze(response.data)` would be more parsimonious.

hillalex (changes requested):
Yes that makes sense re barchart, we ought not to be mutating props, oops! In the meantime this fix should do the job.

hillalex (approved):

EmmaLRussell:
Ticket for addressing the vue-charts bug: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2845

LekanAnni (approved):

PR #644, opened 2021-12-09
mrc-2742 Data exploration mode error reporting

LekanAnni:
This PR added error reporting to data exploration as described in mockup. 

- sections are  auto-populated as 'Data Exploration'
- project name is removed

I added project and section slots to error-report component so as it can be reused by Data Exploration and Root without having the need to create separate component, which will cause a lot of repetition. 

Testing:
I added real component to the slot for testing, this is really not needed but I will see what you guys think. 

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/static/src/app/components/header/DataExplorationSupportMenu.vue
```diff
@@ -10,31 +11,79 @@
...
                 (state: DataExplorationState) => state.language),
             support() {
                 return i18next.t("support", this.currentLanguage)
             }
         },
+        methods: {
+            generateErrorReport: mapActionByName(null, "generateErrorReport"),
+            toggleErrorReportModal() {
+                this.errorReportOpen = !this.errorReportOpen
+                this.errorReportOpen ? this.section = "dataExploration" : this.section = "";
```
EmmaLRussell, line 74: Not really sure why you need to set `this.section` to "" on the toggle? Whenever we actually want to use the section, to populate the error report, it's always going to be "dataExploration" so that could really just be a const or literal.

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -72,6 +72,7 @@ export interface Translations {
     createReleaseTooltip: string,
     currentProject: string,
     custom: string,
+    dataExploration: string,
```
EmmaLRussell, line 75: Again, I don't think we need to show this in a select, so shouldn't need to translate it.

src/app/static/src/app/components/header/DataExplorationSupportMenu.vue
```diff
@@ -10,31 +11,79 @@
...
+            },
+            async sendErrorReport(errorReport) {
+                await this.generateErrorReport({
+                    section: this.section,
```
hillalex, line 78: I think this can be simplified a little - no need to use a data property to track "section", since it's value will always be "dataExploration". So here
```suggestion
                    section: i18next.t("dataExploration", this.currentLanguage)
```

hillalex: then can delete lines 74 and 81 and simplify the `Data` interface to just 
```
 interface Data {
        errorReportOpen: boolean
        section: string
    }
```

src/app/static/src/app/components/header/DataExplorationSupportMenu.vue
```diff
@@ -10,31 +11,79 @@
                          v-translate="'axe'">
             </router-link>
         </drop-down>
+        <error-report :open="errorReportOpen"
+                      @send="sendErrorReport"
+                      @close="toggleErrorReportModal">
+            <template v-slot:sectionView>
+                <div>
+                    <label for="section" v-translate="'section'"></label>
+                    <select class="form-control" v-model="section" id="section">
```
hillalex, line 20: Can drop `v-model` here - since there is only one option in this dropdown, the value can only ever be "dataExploration"!

hillalex (changes requested):
Really nice use of slots! Can be simplified slightly. Mounting the full component for tests makes sense to me :+1:

EmmaLRussell (changes requested):
This works well, but we don't need to see the section drop-down in DE mode - see the [mockup](https://imperiallondon.sharepoint.com/:p:/r/sites/MRCGIDA-hint-dev/_layouts/15/doc2.aspx?sourcedoc=%7BB91CFC01-B73F-44E3-9A30-D8E9BA47F710%7D&file=Naomi%20Data%20Exploration%20Mode%20mockups%20v2.pptx&action=edit&mobileredirect=true&cid=2c0db408-c544-4492-93dc-be08ecd47246)

It's a nice idea to reduce repetition with slots, but I'm not too keen on the error reporting functionality leaking up in the menu components - I think it would be cleaner to keep those separate. Could do that by having two separate error report components for the modes which both contain the 'base error component' and slot contents. That would be similar to the two table implementations (GenericChartTable and AreaIndicatorsTable which both provide slot content to base Table component

hillalex:
> I'm not too keen on the error reporting functionality leaking up in the menu components - I think it would be cleaner to keep those separate. 

Personally I think having the support menu components handle the error reporting logic is fine! They are still pretty terse components that are easy to reason about, and error reporting is the bulk of their purpose...

EmmaLRussell:
> > I'm not too keen on the error reporting functionality leaking up in the menu components - I think it would be cleaner to keep those separate.
> 
> Personally I think having the support menu components handle the error reporting logic is fine! They are still pretty terse components that are easy to reason about, and error reporting is the bulk of their purpose...

Yeah, I don't think it's too unreadable, mainly because the parent component has so little else going on, it just feels a little messy. It's not where I'd naturally expect to find that logic. But then adding another layer of components is kind of clunky too I guess. So I'm prepared to be outvoted! 😄

hillalex:
> > > I'm not too keen on the error reporting functionality leaking up in the menu components - I think it would be cleaner to keep those separate.
> > 
> > 
> > Personally I think having the support menu components handle the error reporting logic is fine! They are still pretty terse components that are easy to reason about, and error reporting is the bulk of their purpose...
> 
> Yeah, I don't think it's too unreadable, mainly because the parent component has so little else going on, it just feels a little messy. It's not where I'd naturally expect to find that logic. But then adding another layer of components is kind of clunky too I guess. So I'm prepared to be outvoted! smile

Cool, let's leave to the discretion of the author then :)

LekanAnni:
> > > I'm not too keen on the error reporting functionality leaking up in the menu components - I think it would be cleaner to keep those separate.
> > 
> > 
> > Personally I think having the support menu components handle the error reporting logic is fine! They are still pretty terse components that are easy to reason about, and error reporting is the bulk of their purpose...
> 
> Yeah, I don't think it's too unreadable, mainly because the parent component has so little else going on, it just feels a little messy. It's not where I'd naturally expect to find that logic. But then adding another layer of components is kind of clunky too I guess. So I'm prepared to be outvoted! 😄

I left it because we have used v-select in that way before. It's not bad that bad. 😊

EmmaLRussell (approved):
:+1:

Nicolas-Dolan (approved):
LGTM

PR #645, opened 2021-12-13
mrc-2831 Bug fix in Input Time Series - some lines shown incorrectly in red

EmmaLRussell:
## Description

e.g. for Mozambique data, at Distrito level. ART adult plot - for Chifunde (MOZ_2_0715), the section between 2018 Q4 and 2019 Q4 was shown in red, despite the two consective values being 1184 and 1191.
WhatsApp Image 2021-12-08 at 2 03 54 PM

This was happening for the specific situation where a single black line should have been drawn between two threshold violating lines. Because the two markers at either end of the should-be-black line were also involved in red lines, they were being added to the 'highlight series', and hence that line in between them was also coloured red. Since there is no way I know of to colour individual lines within a series in Plotly, the only way I could find to break the series and ensure that the incorrect highlighted lines were not drawn, is to interpolate null values in betweeen (in both x and y arrays). 

In order to accommodate this I did a general reformat of the refactoring code - this makes it appear more complex, but should be more efficient. Rather than calculating for each point whether it participates in highlight line in either direction, instead, for each subplot we:
- check if each line segment should be highlighted, generating $highlightedLines
- calculate if there are any points in the series which null interpolation must occur to fix this bug
- for each point in series, first check if it participates in a highlighted line, and if it does add it to the series..
- ..and also check if nulls must be interpolated, in which case do so for both x and y arrays. 

Checking time to evaluate on this branch vs master branch suggests that this has indeed cut the evaluation time - from ~720ms on average for full page of Mozambique data to ~315ms. 

This can be tested with the Mozambique data attached to the ticket. You can see that the expected subplots are now shown for Chifunde and Chinde:
Screenshot from 2021-12-13 09-39-36

One thing this this has brought into focus is the need for some automated tests of the jsonata. Since the jsonata is delivered from the backend but evaluated in the front end, it might makes sense to do this as a front end integration test, to fetch the jsonata and then test that it evluates to expected json with given test data. Ticket for this is here, I've added it into the sprint: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2760

## Type of version change
_Delete as appropriate_

Patches 

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.80.1
+
+* Fix Input Time Series bug where some lines were shown incorrectly in red
+
 # hint 1.80.0
 
 * Add data exploration header
--- a/src/app/src/main/resources/metadata/input-time-series-config-jsonata.txt
+++ b/src/app/src/main/resources/metadata/input-time-series-config-jsonata.txt
@@ -4,6 +4,7 @@ $areaNames := $keys($dataByArea);
 
 $lineColor := "rgb(51, 51, 51)";
 $highlightColor := "rgb(255, 51, 51)";
+$colorscale := [[0, $lineColor], [0, $highlightColor]];
 $subplotHeight := 1/subplots.rows * 0.6;
 $timePeriods := $sort(data.time_period);
 $firstXAxisVal := $timePeriods[0];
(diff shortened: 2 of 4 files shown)


LekanAnni (commented):
A curious question, has Jsonata actually produced any performance improvements ?

EmmaLRussell:
> A curious question, has Jsonata actually produced any performance improvements compared to the former way of getting data into frontend?

Well, no, jsonata is always likely to be a performance sink compared to writing a specialist chart component, but the idea of it was to allow us to just have a single generic chart component which uses the jsonata as config to generate the input to Plotly. So we should in theory be able to use GenericChart for all the various scatter plots etc which were in Jeff and Rachel's full wishlist mockups. I think it's still a moot point whether that's really going to be worth it, though in fact the Input Time Series is likely the most complex of the possible charts from that set, so in theory we ought to have an easier time with the other (although there are some gotchas with those too). Full of discussion of those charts and their issues here: https://mrc-ide.myjetbrains.com/youtrack/articles/mrc-A-45/Implementation-approach-for-Charts#charts

LekanAnni:
> > A curious question, has Jsonata actually produced any performance improvements compared to the former way of getting data into frontend?
> 
> Well, no, jsonata is always likely to be a performance sink compared to writing a specialist chart component, but the idea of it was to allow us to just have a single generic chart component which uses the jsonata as config to generate the input to Plotly. So we should in theory be able to use GenericChart for all the various scatter plots etc which were in Jeff and Rachel's full wishlist mockups. I think it's still a moot point whether that's really going to be worth it, though in fact the Input Time Series is likely the most complex of the possible charts from that set, so in theory we ought to have an easier time with the others (although there are some gotchas with those too). Full of discussion of those charts and their issues here: https://mrc-ide.myjetbrains.com/youtrack/articles/mrc-A-45/Implementation-approach-for-Charts#charts
> 
> Perhaps with hindsight we would just have done a separate component for ITS since it is so different from the others.

Looks like the pains outweighs the gains.

LekanAnni (approved):
LGTM

Nicolas-Dolan (approved):

PR #647, opened 2021-12-14
mrc-2773 Persist Data Exploration Mode state to local storage

EmmaLRussell:
## Description

This branch persists state to local storage while in Data Exploration mode, to a different application key from the main app, so that state does not leak through from main Naomi to DE mode.

Previously, if you reloaded the page in DE mode, filenames appeared to be retained, however no data was visible in the second step, and also ADR dataset was not retained. Also, if you browsed from a project in main Naomi to DE mode (/explore), the dataset from the project was displayed in DE mode (although no data was present). 

Both these symptoms should now be fixed - refreshing the page in DE mode should retain all session data. Browsing from a Naomi project to DE mode should show a clean page. 

NB It is not possible to switch back and forth between Naomi main app and DE mode, and retain state from a mode when switching back to it. This is because, as far as the backend is concerned, switching mode forces a new session, so session file data is not returned when you switch back. Instead, switching back to Naomi main should take the user to the projects page (if logged in) or clean step 1 (if not), while switching back to DE mode takes the user to a clean page.

Implementation details:
- Saves data exploration mode state to a separate key in local storage from main Naomi state to prevent 'leakage'
- To facilitate this, modules no longer fetch from local storage themselves, but are provided with the saved state from their parent (Root or DataExploration) - since the parent knows which mode to use but the module does not. It feels like this should be a lot more efficient anyway, to only fetch from local storage once, rather than in every module. 
- Data Exploration state is persisted on each mutation, using a `persistState` plugin, similar to the one in `root` (and which also logs each mutation to console, to aid debugging, as the root plugin does). 
- On load, removes the local storage value for the other mode - as described above, we cannot go back to state in the previous mode anyway.  Without this clear step, the app will attempt to use the saved state when previous mode is returned to, giving the odd symptom of selected dataset being retained even though no files can be loaded from back end. 

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/static/src/tests/components/header/fileMenu.test.ts
```diff
@@ -127,14 +127,14 @@ describe("File menu", () => {
...
-        });
+        };
 
         const actualBlob = (mockCreateObjectUrl as jest.Mock).mock.calls[0][0];
         const reader = new FileReader();
         reader.addEventListener('loadend', function () {
             const text = reader.result as string;
             const result = JSON.parse(text)[1];
-            expect(result).toStrictEqual(expectedJson);
+            expect(JSON.parse(result)).toStrictEqual(expectedJson);
```
EmmaLRussell, line 137: This test started failing because it was acutally doing a string compare rather than JSON compare, and we have modified the order of the keys somewhat in saved state. This change is to do JSON compare instead.

LekanAnni (changes requested):
Looking great.

Please see my suggestions and observations below

- steppers, current step is not persisted in state 
- from exploration mode, when user clicks on run model after loading data, it should retain the loaded data. I understand you mentioned it is impossible to retain states for both app instances, I think we should be able to store root modules and explore modules in local storage separately, perhaps we will need to modify `pac4j` to allow this behaviour so that they can be retrieved. 
- I also noticed the main app reloads when change language is triggered, this is probably not a change from the branch but  a reactive behaviour instead of a reload should be expected.
- I think the approach you have used limits retaining states for both app instances, it would be doable to retain both states if we store app instances (exploration and root) separately.

EmmaLRussell:
> Looking great.
> 
> Please see my suggestions and observations below
> 
>     * steppers, current step is not persisted in state
> 
>     * from exploration mode, when user clicks on run model after loading data, it should retain the loaded data. I understand you mentioned it is impossible to retain states for both app instances, I think we should be able to store root modules and explore modules in local storage separately, perhaps we will need to modify `pac4j` to allow this behaviour so that they can be retrieved without session expiring.
> 
>     * I also noticed the main app reloads when change language is triggered, this is probably not a change from the branch but  a reactive behaviour instead of a reload should be expected.
> 
>     * I think the approach you have used limits retaining states for both app instances, it would be doable to retain both states if we store app instances (exploration and root) separately.

Thanks @LekanAnni  - as we discussed, I've created separate tickets for your suggestions:
Persist DE step: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2846
Allow 'promotion' of DE to Model run: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2847

LekanAnni (approved):
Works great 👍

Nicolas-Dolan (approved):
Seems like a sensible approach and works well.

PR #648, opened 2021-12-15
mrc-2832 Return calibrate_id in error report if it is available

LekanAnni:
## Description

This PR renamed jobId to modelRunId, adds calibrateId and downloadIds to error report payload. Once PR is merged, @r-ash  should be notified to make the corresponding changes in MS flow.

## Type of version change

Minor

## Checklist


EmmaLRussell (commented):
Looks good to me. I guess if we were being very fastidious we could avoid repeating string literals for things like "none" and "no associated calibrateId" by using consts kept in (probably) error report module. 

Could you give @r-ash a heads-up on the changes to expect if you haven't already?

LekanAnni:
> Looks good to me. I guess if we were being very fastidious we could avoid repeating string literals for things like "none" and "no associated calibrateId" by using consts kept in (probably) error report module.
> 
> Could you give @r-ash a heads-up on the changes to expect if you haven't already?

@r-ash is aware of the changes, I will make none value a reusable static string, others aren't used anywhere else really.

EmmaLRussell:
> > Looks good to me. I guess if we were being very fastidious we could avoid repeating string literals for things like "none" and "no associated calibrateId" by using consts kept in (probably) error report module.
> > Could you give @r-ash a heads-up on the changes to expect if you haven't already?
> 
> @r-ash is aware of the changes, I will make none value a reusable static string, others aren't used anywhere else really.

Thanks 👍 . The others are used in both dataExploration actions and root actions so there is a bit of repetition.

LekanAnni:
> > > Looks good to me. I guess if we were being very fastidious we could avoid repeating string literals for things like "none" and "no associated calibrateId" by using consts kept in (probably) error report module.
> > > Could you give @r-ash a heads-up on the changes to expect if you haven't already?
> > 
> > 
> > @r-ash is aware of the changes, I will make none value a reusable static string, others aren't used anywhere else really.
> 
> Thanks 👍 . The others are used in both dataExploration actions and root actions so there is a bit of repetition.

true, I will make those too.

EmmaLRussell (approved):
Looks good, thanks

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ErrorReportTests.kt
```diff
@@ -20,7 +20,13 @@ class ErrorReportTests: SecureIntegrationTests()
...
             "projectName": "South 2 Worldpop",
             "section": "Fit model",
-            "jobId": "job12",
+            "modelRunId": "job12",
+            "calibrateId": "1234",
+            "downloadIds": {
+            "spectrum": "spectrum123",
+             "summary": "summary123",
+              "coarse_output": "coarse123"
+            },
```
Nicolas-Dolan, line 29:
```suggestion
            "spectrum": "spectrum123",
            "summary": "summary123",
              "coarse_output": "coarse123"
            },
```

src/app/static/src/app/store/errors/errors.ts
```diff
@@ -10,6 +10,14 @@ export interface ErrorsState {
     sendingErrorReport: boolean
 }
 
+export enum ErrorReportDefaultValue {
+    calibrate = "no associated calibrateId",
+    model = "no associated modelRunId",
+    download = "none",
+    country = "no associated country",
+    project = "no associated project"
+}
```
Nicolas-Dolan, line 19: Could we use a locales key word instead so that they can be translated?

Nicolas-Dolan (changes requested):

src/app/static/src/app/store/errors/errors.ts
EmmaLRussell: They don't really need to be translated as these values in the error reports are only seen by the Naomi support team, not by the user.

Nicolas-Dolan (approved):

PR #650, opened 2021-12-16
mrc-2846 Persist Data Exploration step

EmmaLRussell:
## Description

This branch allows current step to be persisted for Data Exploration mode in local storage - this means that if you refresh the page while on the second step (review inputs) the page will reload with that step open (however it does not retain the tab - this will always be the Map tab if both Map and Input Time Series are available - to be addressed in a separate ticket). 

It seemed that the simplest way to do this was to include the `Stepper` state in DataExplorationState, since it is already used by `RootState` - which is an extension of DataExplorationState. DE State does not require everything in `StepperState`, but only the `activeStep` value. However, the alternative would have been to provide DE state with its own simplified stepper state which would not have been relevant for `RootState`, so it would have been necessary either to confusingly include that simplified step state in `Root` via its parent type, or to create a CommonRootState from which both RootState and DEState could derive, which seemed like a lot of upheaval!

In taking the `Stepper` approach, it was necessary to create two versions of the Stepper module, since DEState could not include the full Stepper getters, which refer to modules not in DE State (e.g. modelRun). 

## Type of version change
_Delete as appropriate_

Minor

## Checklist


Nicolas-Dolan (commented):
Seems like a sensible approach to persisting state. However, when I refresh on the second step, the step persists but the table fails to render and I get a bunch or error messages in the console. I'm not sure if handling this bug is explicitly the responsibility of this ticket but it's certainly exposed by it so should probably fix that bug in another ticket before merging this in.
image

EmmaLRussell:
> Seems like a sensible approach to persisting state. However, when I refresh on the second step, the step persists but the table fails to render and I get a bunch or error messages in the console. I'm not sure if handling this bug is explicitly the responsibility of this ticket but it's certainly exposed by it so should probably fix that bug in another ticket before merging this in. image

Ooh, interesting, I hadn't seen/noticed that. Is this just on the Map tab or Input Timer Series? Full Malawi data or something else?

Nicolas-Dolan:
> > Seems like a sensible approach to persisting state. However, when I refresh on the second step, the step persists but the table fails to render and I get a bunch or error messages in the console. I'm not sure if handling this bug is explicitly the responsibility of this ticket but it's certainly exposed by it so should probably fix that bug in another ticket before merging this in. 
> 
> Ooh, interesting, I hadn't seen/noticed that. Is this just on the Map tab or Input Time Series? Oh Map, presumably since it's AreaIndicatorsTabe. Full Malawi data or something else?

So it appears on the map tab when you reload. However, if you toggle time series tab its fine and if you toggle back to the map tab, it fixes the error in the map tab and the table loads normally. The problem probably exists on both tabs but you can only see it on the map tab as that's the tab that tries to render when you reload.

hillalex:
I've been holding off on reviewing this until the above bug is fixed, just fyi!

Nicolas-Dolan (approved):
The first time I checked this, I got the bug again (from a fresh build of the app). I ran build again to be safe and the bug went away and I wasn't able to reproduce it, so it looks fine. Maybe something to look out for by anyone else reviewing though.

src/app/static/src/app/store/stepper/stepper.ts
```diff
@@ -50,10 +51,16 @@ export const initialStepperState = (): StepperState => {
     }
 };
 
+export const initialDataExplorationStepperState = (): StepperState => {
+    return {
+        steps: [],
```
hillalex, line 56: Could see this empty array being confusing in the future - maybe just a comment to explain why there are apparently no steps?

src/app/static/src/app/store/stepper/stepper.ts
```diff
@@ -62,3 +69,12 @@ export const stepper = (existingState: Partial<RootState> | null): Module<Steppe
         mutations
     };
 };
+
+export const dataExplorationStepper = (existingState: Partial<DataExplorationState> | null): Module<StepperState, DataExplorationState> => {
+    return {
+        namespaced,
```
hillalex, line 75: Also maybe just a comment here highlighting that getters are omitted and would not be valid for this version of the module

hillalex (commented):
Logic seems good; since this is slightly more complicated than just re-using a module it might be worth a couple of comments as future-proofing.

hillalex (approved):

PR #652, opened 2022-01-05
mrc-2853 - ADR access level incorrectly displayed as 'Read only' on reload

LekanAnni:
When on baseline page and a user reloads the page, app incorrectly displays access level. This is because the method being watched does not trigger hence it can't getUserCanUpload on reload. I added another getUserCanUpload to watch where it is now being triggered.

## Type of version change

Patches

## Checklist


EmmaLRussell (commented):
It feels like `mounted` would be a more intuitive place to put this than a watch on `key`. Key being updated isn't obviously related to the need to check user dataset permission - typically when the user updates or adds their key they don't have a selected dataset yet to check permission for anyway. Whereas the bug here was about not checking for perms on reload, so `mounted` should cover that.

LekanAnni:
> It feels like `mounted` would be a more intuitive place to put this than a watch on `key`. Key being updated isn't obviously related to the need to check user dataset permission - typically when the user updates or adds their key they don't have a selected dataset yet to check permission for anyway. Whereas the bug here was about not checking for perms on reload, so `mounted` should cover that.

I intentionally added the method to watch() because I didn't want this.getUserCanUpload(); to be called when no key has been provided but you are right it does not belong to watch() and I could check for key before calling the method. 👍

EmmaLRussell:
> > It feels like `mounted` would be a more intuitive place to put this than a watch on `key`. Key being updated isn't obviously related to the need to check user dataset permission - typically when the user updates or adds their key they don't have a selected dataset yet to check permission for anyway. Whereas the bug here was about not checking for perms on reload, so `mounted` should cover that.
> 
> I intentionally added the method to watch() because I didn't want this.getUserCanUpload(); to be called when no key has been provided but you are right it does not belong to watch() and I could check for key before calling the method. 👍

Good call - could also check for `selectedDataset` before doing the perms check too.

LekanAnni:
> > > It feels like `mounted` would be a more intuitive place to put this than a watch on `key`. Key being updated isn't obviously related to the need to check user dataset permission - typically when the user updates or adds their key they don't have a selected dataset yet to check permission for anyway. Whereas the bug here was about not checking for perms on reload, so `mounted` should cover that.
> > 
> > 
> > I intentionally added the method to watch() because I didn't want this.getUserCanUpload(); to be called when no key has been provided but you are right it does not belong to watch() and I could check for key before calling the method. 👍
> 
> Good call - could also check for `selectedDataset` before doing the perms check too.

True, there would be no need to check key, checking selectedDataset instead.

EmmaLRussell (approved):
:+1:

PR #653, opened 2022-01-05
mrc-2741 Show (required) for required files in both modes

LekanAnni:
This PR adds asterisks to required fields for input uploads. 

- only area file is required for Data exploration mode 
- all baseline files and shape file required for run model mode

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/static/src/tests/components/uploadInputs/uploadInputs.test.ts
```diff
@@ -90,6 +91,66 @@ describe("UploadInputs upload component", () => {
...
+
+    it("ANC upload does not show asterisk when on data exploration mode", () => {
+        const store = createSut();
+        expectFileIsNotRequired(store, 4)
+    });
+
+    it("ANC upload does not show asterisk when not on data exploration mode", () => {
+        const store = createSut();
+        expectFileIsNotRequired(store, 4)
+    });
```
EmmaLRussell, line 152: Looks like you're testing both ART and ANC files with index 4. This suggests the expect methods taking an index param might be a bit prone to error e.g. files might get reordered potentially - maybe those helpers could find the ManageFile components by label attribute instead, and take that as a parameter?

EmmaLRussell (changes requested):
Looks good. Just need to add a test for the ManageFile changes, and there's a small bug in the UploadInputs tests. 

I wonder if, in data exploration mode we should show some label to indicate that you do need to upload one of ANC or ART, even though neither is required in all cases? Otherwise it seems a bit confusing for a user who has uploaded all asterisked files, but still can't proceed. 

We should probably also add a label explaining what the asterisk means - not every user is necessarily going to know this convention.

LekanAnni:
> I wonder if, in data exploration mode we should show some label to indicate that you do need to upload one of ANC or ART, even though neither is required in all cases? Otherwise it seems a bit confusing for a user who has uploaded all asterisked files, but still can't proceed.

There's no point having another form of required text, we can just show asterisk for ART label for DE mode.

LekanAnni:
> We should probably also add a label explaining what the asterisk means - not every user is necessarily going to know this convention.

We can do either show asterisk or (required) as we have it in `calibration form`. It's obfuscating to have both as suggested.

EmmaLRussell:
> > I wonder if, in data exploration mode we should show some label to indicate that you do need to upload one of ANC or ART, even though neither is required in all cases? Otherwise it seems a bit confusing for a user who has uploaded all asterisked files, but still can't proceed.
> 
> There's no point having another form of required text, we can just show asterisk for ART label for DE mode.

Not sure what you mean - ART isn't required in itself, and neither is ANC, so showing asterisks for either would be misleading, But you do need to upload one or both. So it is a different sort of 'required'...

EmmaLRussell:
> > We should probably also add a label explaining what the asterisk means - not every user is necessarily going to know this convention.
> 
> We can either show asterisk or (required) as we have it in `calibration form`. It is obfuscating to have both as suggested, also, since there are no corresponding validation behaviour for showing the required, I think it is not a bad way of showing required. Otherwise we can make asterisks red when no data input and black when data is uploaded.

There's is validation since you can't go on if you don't fill in the required fields. Sorry, maybe you meant something else?

You're right, we should really be consistent with the forms we already have if we're going to put this feature in at all. Which as you say would mean 'required' rather than asterisks, and ideally colour changing when provided, Sorry, the ticket wasn't very well thought through! (Arguably this isn't quite the same as the options forms since it's file uploads but as you say having two 'required' indicators is quite confusing'

EmmaLRussell (commented):
Looks good, but will need to translate 'required' - this already works for Model Options as shown
Could we make the 'required' label small. to match the Model Options/Calibrate form too? 

obligatoire

I think they still might end up asking for some indication that one of ART or ANC is required in DE mode - but that can be another ticket if they do, as it would need to be separate from these 'required' indicators anyway.

src/app/static/src/app/components/files/ManageFile.vue
```diff
@@ -1,7 +1,9 @@
 <template>
     <div class="form-group">
         <label class="font-weight-bold mb-0" v-translate="label"></label>
-        <span id="required" v-if="required" :class="existingFileName? '': 'text-danger'">(required)</span>
+        <span id="required" v-if="required"
+              class="small"
+              :class="existingFileName? '': 'text-danger'">({{requiredText}})</span>
```
EmmaLRussell, line 6: I think it would be simpler here to use the `v-translate` directive than to have to map currentLanguage and add the requiredText property - even though that would mean adding a nested span here so that you could still have the brackets in the outer one.

LekanAnni: I don't think a nested span is neat and readable.

EmmaLRussell: That's what `v-translate` is for though, we have nested spans in lots of other places because that's how it works. It adds less boilerplate than the additional map and property.

EmmaLRussell (approved):
Thanks!

Nicolas-Dolan (approved):
Works well. I think using 'required' instead of asterisks was the right call.

PR #654, opened 2022-01-07
mrc-2760 Add automated tests for input time series jsonata

EmmaLRussell:
## Description

This branch adds tests that the jsonata for the input time series (generic chart) evaluates to the expected json given simple test data. (This generated json is what is passed to plotly to render.) The jsonata is served from the backend but evaluated in the front end, so I've created these as front end integration tests. 

A new test file has been added for these tests, which are in a separate dir along with the other genericChart tests. 

Adding the tests uncovered a benign bug! Because the 'index' used when calculating the yaxis layout values starts from 1 rather than 0, the last yaxis in a subplot row was incorrectly being assigned to the previous row, and setting 'domain' values for the previous row. This didn't seem to matter though - perhaps plotly only uses the domain for the first yaxis in a row? I've fixed it anyway to be as expected and behaviour when run manually appears to be unaffected. 

No version update as no behaviour changes. 

## Type of version change
_Delete as appropriate_
None

## Checklist


src/app/static/src/tests/integration/genericChart/genericChart_jsonata.itest.ts
```diff
@@ -0,0 +1,248 @@
...
+        expect(layout.xaxis2).toStrictEqual(expectedXAxis);
+        expect(layout.xaxis3).toStrictEqual(expectedXAxis);
+
+        const expectYAxis = (domainStart: number, domainEnd: number, actual: any) => {
+            expect(Object.keys(actual)).toStrictEqual(["rangemode", "zeroline", "tickformat", "tickfont", "domain"]);
+            expect(actual.rangemode).toBe("tozero");
+            expect(actual.zeroline).toBe(false);
+            expect(actual.tickformat).toBe(".0f");
+            expect(actual.tickfont).toStrictEqual({"color": "grey"});
+            expect(actual.domain[0]).toBeCloseTo(domainStart, 8);
```
EmmaLRussell, line 240: The floating point values calculated for domain can be a little off (e.g. 0.199999999996) hence using `toBeCloseTo` on these values.

LekanAnni (approved):
Looks great 👍 

Some trivial suggestions below.

Nicolas-Dolan (approved):

PR #656, opened 2022-01-12
Add staging deploy triggered build

r-ash:
## Description

Triggers an automatic deploy on new staging server when a branch is merged into master

## Type of version change

None

## Checklist

### DIFF
--- a/buildkite/pipeline.yml
+++ b/buildkite/pipeline.yml
@@ -21,3 +21,20 @@ steps:
 
   - label: ":mag: Test front-end ADR Dataset integration"
     command: buildkite/test-adr-front-end-integration.sh
+
+      - wait
+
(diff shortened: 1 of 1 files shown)


LekanAnni (changes requested):
Looking good, Perhaps we could wait for master build to succeed before triggering deployment.

hillalex:
> Looking good, Perhaps we could wait for master build to succeed before triggering deployment.

I think the `wait` in the buildkite config ensures that, is that right @r-ash ?

LekanAnni:
> Yep that's right, will only trigger the build if all the tests pass

Oh ok, my understanding was deployment will trigger even when build failed as long as it the process is complete.

hillalex (approved):

LekanAnni (approved):
LGTM

PR #658, opened 2022-01-27
mrc-2851 Fix order of nested xAxis disaggregate options in barchart

Nicolas-Dolan:
## Description

This extends mrc-2796 (which maintained the barchart xAxis order after the selections were updated) and ensures that nested filter options follow the order `parent1 > parent2 > parent1 children > parent2 children` in the x-axis.

To test, set the xaxis of the barchart to area on the Review Output page and add/remove nested options in the area filter. See that parents (and their siblings at their level in the hierarchy) are always prioritised over children in the order. The order in the barchart should match the default order of the table below it.

## Type of version change
Patches 

## Checklist


src/app/static/src/app/utils.ts
```diff
@@ -162,6 +162,35 @@ const flattenOption = (filterOption: NestedFilterOption): NestedFilterOption =>
...
+    }
+    recursive(filterOptions)
+    
+    let output: string[] = []
+    result.forEach(r => {
+        output = [...output, ...r]
+    })
+    return output;
+};
+
```
EmmaLRussell, line 193: This works nicely, but I think it could be simplified a bit. What you're doing here is building up the `result` array where each item is a list of options at each layer, but I don't think this is necessary - we  should be able build up the output id array directly by recursing at each layer rather than for the children of each filterOption (so what gets passed to the next call of `recursive` is the concatenated list of filter options at a layer).  Something like this?

```suggestion
export const flattenOptionsIdsByHierarchy = (filterOptions: NestedFilterOption[]): string[] => {
    const result: string[] = [];
    const recursive = (filterOptions: NestedFilterOption[]) => {
        // 1. Push ids for the top level of options
        filterOptions.forEach((option: NestedFilterOption) => result.push(option.id));
        let nextLayer: NestedFilterOption[] = [];
        // 2. Get all options at next layer and recurse
        filterOptions.forEach((option: NestedFilterOption) => {
            if (option.children) {
                nextLayer = nextLayer.concat(option.children as NestedFilterOption[]);
            }
        });
        if (nextLayer.length > 0) {
            recursive(nextLayer);
        }
    };
    recursive(filterOptions);
    return result;
};

```

Nicolas-Dolan: > This works nicely, but I think it could be simplified a bit. What you're doing here is building up the `result` array where each item is a list of options at each layer, but I don't think this is necessary - we should be able build up the output id array directly by recursing at each layer rather than for the children of each filterOption (so what gets passed to the next call of `recursive` is the concatenated list of filter options at a layer). Something like this?

Ok great. Have you tested this? I had something similar to this originally, but I found as I went down to the deeper layers (great-grand children, etc), the recursive function would start to split, resulting in great-grandchildren coming before children.

EmmaLRussell: I tested it a bit, seemed to work ok. I think this should keep everything at a layer all together - but do try to break it!

Nicolas-Dolan: Yeah, seems to work fine. So as I understand, you are flattening out the children of layers below into a single array and then recursing, avoiding the splitting problem I mentioned.

EmmaLRussell (approved):

src/app/static/src/app/utils.ts
```diff
@@ -162,6 +162,26 @@ const flattenOption = (filterOption: NestedFilterOption): NestedFilterOption =>
...
 };
 
+export const flattenOptionsIdsByHierarchy = (filterOptions: NestedFilterOption[]): string[] => {
+    const result: string[] = [];
+    const recursive = (filterOptions: NestedFilterOption[]) => {
+        // 1. Push ids for the top level of options
+        filterOptions.forEach((option: NestedFilterOption) => result.push(option.id));
+        let nextLayer: NestedFilterOption[] = [];
+        // 2. Get all options at next layer and recurse
+        filterOptions.forEach((option: NestedFilterOption) => {
```
LekanAnni, line 172: Why are you looping twice ? you could perform the action in a loop, which is more efficient.

src/app/static/src/app/utils.ts
```diff
@@ -162,6 +162,26 @@ const flattenOption = (filterOption: NestedFilterOption): NestedFilterOption =>
     return result;
 };
 
+export const flattenOptionsIdsByHierarchy = (filterOptions: NestedFilterOption[]): string[] => {
+    const result: string[] = [];
+    const recursive = (filterOptions: NestedFilterOption[]) => {
```
LekanAnni, line 167: I am not sure recursive is the best name to give this function.

LekanAnni (changes requested):
Looking good, just some minor pointers.

LekanAnni (approved):
LGTM 👍

PR #659, opened 2022-01-27
mrc-2673 Make hintr warnings dismissable

Nicolas-Dolan:
## Description

Adds a small cross to top right of warning messages so that they can be dismissed. Clicking will remove the warnings/warning message for that page for the session. Warnings will not reappear if the user navigates away from the page and returns. Warnings will return on page refresh or if the relevant action is performed (validate, fit model, calibrate). Warnings appear on modeloptions, model run, model calibrate, and review output. Model calibrate and review output share warnings so dismissing the warnings in one will cause them to disappear in the other.

## Type of version change
Patches

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -35,7 +35,8 @@
 * Remove Projects state from localStorage
 
 # hint 1.88.3
-Use area_id to define unique areas in Input Time Series
+
+* Use area_id to define unique areas in Input Time Series
 
 # hint 1.88.2
 
@@ -56,7 +57,6 @@ Use area_id to define unique areas in Input Time Series
 # hint 1.87.0
 
 * Formatting Input Time Series table values
-
 
 # hint 1.86.0
 
@@ -113,7 +113,6 @@ Use area_id to define unique areas in Input Time Series
 # hint 1.75.1
 
 * Update review output plots to use metadata from modelCalibrate
-
 
 # hint 1.75.0
 
--- a/src/app/static/src/app/components/Stepper.vue
+++ b/src/app/static/src/app/components/Stepper.vue
@@ -164,6 +164,7 @@
                     4: this.clearModelRunWarnings,
                     5: this.clearModelCalibrateWarnings,
                     6: this.clearModelCalibrateWarnings,
+                    7: this.clearModelCalibrateWarnings
                 }
                 if (this.activeStep in mutationMethods){
                     mutationMethods[this.activeStep]()
--- a/src/app/static/src/app/hintVersion.ts
+++ b/src/app/static/src/app/hintVersion.ts
@@ -1,1 +1,2 @@
 export const currentHintVersion = "1.90.3";
+
--- a/src/app/static/src/tests/components/stepper.test.ts
+++ b/src/app/static/src/tests/components/stepper.test.ts
@@ -900,4 +900,10 @@ describe("Stepper component", () => {
         await wrapper.find(WarningAlert).vm.$emit("clear-warnings")
         expect(clearCalibrateWarnings.mock.calls.length).toBe(1);
     });
+
+    it("clear-warnings emit when in downloadResults triggers clear warnings mutation in modelCalibrate store", async () => {
+        const wrapper = createWarningAlertWrapper(7)
+        await wrapper.find(WarningAlert).vm.$emit("clear-warnings")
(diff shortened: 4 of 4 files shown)


src/app/static/src/app/components/Stepper.vue
```diff
@@ -19,7 +19,7 @@
         </div>
         <stepper-navigation v-bind="navigationProps"/>
         <hr/>
-        <warning-alert v-if="activeStep !== 3" :warnings="activeStepWarnings"></warning-alert>
+        <warning-alert v-if="activeStep !== 3" :warnings="activeStepWarnings" :active-step="activeStep"></warning-alert>
```
hillalex, line 22: It feels a bit inconsistent that we're passing active step warnings as a prop, but then also passing the active step itself to figure out which mutation to use. I think either the `WarningAlert` just receives the active step as a prop and figures out both which warnings to display and which clear handler to use, or both the clear handler and the warnings object get created here and passed as props. No strong feelings as to which way we resolve it.

hillalex (changes requested):
This works and the tests look good, but the architecture could benefit from a more consistent pattern.

src/app/static/src/app/components/Stepper.vue
EmmaLRussell: Agreed. My instinct would probably be to keep the child component simple and let the parent Stepper component know how to clear warnings for each step (so the child would emit a 'clearWarnings' event with a step parameter which the parent would handle).

src/app/static/src/app/components/WarningAlert.vue
```diff
@@ -1,6 +1,9 @@
 <template>
     <div v-if="showAlert">
         <div class="content alert alert-warning pt-0">
+            <button @click="dismissWarnings" type="button" class="close pt-2 pull-right" aria-label="Close">
```
EmmaLRussell, line 4: Aria label should be translated - we already have a translation key for 'Close'

src/app/static/src/tests/components/warningAlert.test.ts
```diff
@@ -15,26 +18,57 @@ describe("Warning alert component", () => {
...
+        jest.resetAllMocks();
+    });
+
     it("renders no warning alert if there are no warnings", () => {
         const wrapper = createWrapper({ modelOptions: [], modelRun: [], modelCalibrate: []})
         expect(wrapper.find(".alert-warning").exists()).toBe(false);
     });
 
-    it("renders warning messages", () => {
+    it("renders warning messages", async () => {
```
EmmaLRussell, line 71: This async wasn't needed?

EmmaLRussell (commented):
Looks good to me otherwise - couple of tiny comments

EmmaLRussell (commented):
Looks good to me, just a small comment on the translate

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# hint 1.90.31
```
hillalex, line 1:
```suggestion
# hint 1.90.1
```

hillalex (approved):
Great, thanks! Typo in the news but other than that good to go

src/app/static/src/app/components/Stepper.vue
```diff
@@ -154,7 +157,21 @@
...
                 }
                 return !this.isComplete(activeStep)
-            }
+            },
+            clearWarnings(){
+                const mutationMethods: { [key: number]: () => void; } = {
+                    3: this.clearModelOptionsWarnings,
+                    4: this.clearModelRunWarnings,
+                    5: this.clearModelCalibrateWarnings,
+                    6: this.clearModelCalibrateWarnings,
```
EmmaLRussell, line 166:
```suggestion
                    6: this.clearModelCalibrateWarnings,
                    7: this.clearModelCalibrateWarnings
```

EmmaLRussell: Sorry, should have spotted this little bug before - calibrate warnings couldn't be dismissed from the final step, think you just need to add this line.

EmmaLRussell (changes requested):

EmmaLRussell (commented):
You'll want to add  test too for that change I guess

Nicolas-Dolan:
> You'll want to add test too for that change I guess

Yep, just realised. It's up now.

EmmaLRussell (approved):

PR #660, opened 2022-01-31
mrc-2919 Prevent ModelRun from polling for a failed run on reload

EmmaLRussell:
## Description

See [ticket](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-2919) for Jeff's bug report. This was caused by ending up with the store polling for two model runs, the original one which had failed before reload as well as the new one. The branch makes 2 changes:

1. Do not start polling on reload for run which we know has failed (previously it would decide not to poll for runs which had successfully completed only.) This was usually benign since one poll would confirm that the run had finished, and polling would stop - however if user immediately started a new run before that first poll had completed (within 2 seconds), polling would continue for both run ids. 
2. `run` action clears any old polling before starting new run. Arguably it should do a full `cancelRun` action to also cancel the run on the server before commencing the run. This would add complexity however, and the clear ought to be belt-and-braces since the component should now prevent getting into this state at all, so this just adds additional avoidance of getting into the  state of having two simultaneous runs. 

## Type of version change
_Delete as appropriate_

Patches 

## Checklist


src/app/static/src/tests/integration/downloadResults.itest.ts
```diff
@@ -126,6 +126,6 @@ describe(`download results actions integration`, () => {
             expect(commit.mock.calls[1][0]["payload"].status).toBe("MISSING");
             done()
 
-        }, 2100)
+        }, 3100)
```
EmmaLRussell, line 129: Making this timeout consistent with the others - this was failing for me locally otherwise.

Nicolas-Dolan (approved):

PR #663, opened 2022-02-01
mrc-2903 restructure download logic to improve user experience

hillalex:
## Description

Nb the base of this PR right now is the `rename` branch which is waiting on a 2nd reviewer; will change the base to master once that is merged.

This is quite a big changeset but mostly deletions so hopefully not too bad to review. Should be clear that the `DownloadOutputs.vue` and `UploadModal.vue` components have been massively simplified, and all the logic (e.g. stopping/starting polling, triggering adr upload metadata on success, etc) moved into actions. This makes things much easier to reason about and to test! 

The flow has also changed slightly, so that the preparation of outputs happens when the user navigates to review outputs, or if they navigate straight to download results, so that the files are ready to be downloaded ahead of the user actually clicking on the download buttons. We only re-request preparation of the outputs as needed, so only if the model has been re-calibrated. 

## Type of version change
Minor

## Checklist

Will increment version pending review!


src/app/static/src/app/components/downloadResults/Download.vue
```diff
@@ -8,17 +8,17 @@
             <download-icon size="20" class="icon ml-2" style="margin-top: -4px;"></download-icon>
         </button>
         <div>
-            <download-progress v-if="!modalOpen" id="progress"
```
hillalex, line 11: This `v-if` condition was because previously, setting off an upload via the upload modal would first trigger the outputs to be prepared (even if already existent!) and so without this condition this progress bar would appear underneath the download buttons. Now that the logic has been simplified, this will not happen.

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -262,7 +271,11 @@
                 }
                 // if unable to do the above, just updates the barchart as normal
                 this.updateBarchartSelections({payload})
-            }
+            },
+            prepareOutputDownloads: mapActionByName("downloadResults", "prepareOutputs")
+        },
+        mounted() {
+            this.prepareOutputDownloads();
```
hillalex, line 278: This and line 275 are the only changes, everything else in this file is a whitespace change as it had not been properly formatted. The change here is that preparing the output files is triggered on mount, i.e. when the user navigates to the "review outputs" step. This should be late enough that they have already calibrated the model to their satisfaction, but early enough that the files are ready for download by the time they navigate to the last step.

src/app/static/src/app/store/downloadResults/actions.ts
```diff
@@ -1,61 +1,72 @@
...
-        const calibrateId = rootState.modelCalibrate.calibrateId
-
-        const response = await api<DownloadResultsMutation, DownloadResultsMutation>(context)
-            .withSuccess(DownloadResultsMutation.PreparingCoarseOutput)
-            .withError(DownloadResultsMutation.CoarseOutputError)
-            .get(`download/submit/coarse-output/${calibrateId}`)
-
-        if (response) {
-            await dispatch("poll", DOWNLOAD_TYPE.COARSE)
+        if (!state.coarseOutput.downloadId && !state.coarseOutput.preparing) {
```
hillalex, line 30: The idea here is that this can be triggered any time, but if there is already an output prepared or in preparation it will not re-request it. When the prepared outputs are invalidated by re-calibrating the model, we just clear the `downloadId` so that the next call to this method will request a new output as required.

src/app/static/src/app/store/downloadResults/actions.ts
```diff
@@ -77,28 +88,37 @@ export const actions: ActionTree<DownloadResultsState, RootState> & DownloadResu
...
-    const {state} = context;
+    const {state, dispatch} = context;
     const downloadId = state.summary.downloadId;
-    return api<DownloadResultsMutation, DownloadResultsMutation>(context)
+    const response = await api<DownloadResultsMutation, DownloadResultsMutation>(context)
         .withSuccess(DownloadResultsMutation.SummaryReportStatusUpdated)
         .withError(DownloadResultsMutation.SummaryError)
         .get<ModelStatusResponse>(`download/status/${downloadId}`)
+    if (response && response.data?.done) {
+        await dispatch("metadata/getUploadMetadata", response.data.id, {root: true});
```
hillalex, line 98: This logic was previously in the component.

src/app/static/src/app/store/downloadResults/mutations.ts
```diff
@@ -28,13 +28,17 @@ export const mutations: MutationTree<DownloadResultsState> = {
...
         }
         state.spectrum.error = null;
     },
 
     [DownloadResultsMutation.SpectrumError](state: DownloadResultsState, action: PayloadWithType<Error>) {
-        state.spectrum.error = action.payload
-        state.spectrum.preparing = false
+        state.spectrum.error = action.payload;
+        state.spectrum.preparing = false;
+        clearInterval(state.spectrum.statusPollId);
```
hillalex, line 40: Logic that was previously in the component(s)

src/app/static/src/app/store/downloadResults/mutations.ts
```diff
@@ -88,5 +100,11 @@ export const mutations: MutationTree<DownloadResultsState> = {
                 break
             }
         }
+    },
+
+    [DownloadResultsMutation.ResetIds](state: DownloadResultsState) {
+        state.summary.downloadId = "";
```
hillalex, line 106: For use after calibration, when any existing outputs necome invalid.

src/app/static/src/app/store/downloadResults/downloadResults.ts
```diff
@@ -18,12 +18,6 @@ export const initialDownloadResults = {
     error: null
 }
 
-export enum DOWNLOAD_TYPE {
```
hillalex, line 21: Moved to avoid a was a circular dependency, now here: https://github.com/mrc-ide/hint/pull/663/files#diff-b4e97f9053896e4bab011155c7ed2ad40cf453d0da07ccbb80ae5144a9424c23R185

src/app/static/src/app/components/downloadResults/DownloadResults.vue
```diff
@@ -5,31 +5,28 @@
...
                 </div>
             </div>
             <div id="upload" v-if="hasUploadPermission" class="col-sm">
                 <h4 v-translate="'uploadFileToAdr'"></h4>
                 <button @click.prevent="handleUploadModal"
                         class="btn btn-lg my-3"
-                        :class="uploading || isDownloading ? 'btn-secondary' : 'btn-red'"
-                        :disabled="uploading || isDownloading">
+                        :class="uploading || isPreparing ? 'btn-secondary' : 'btn-red'"
+                        :disabled="uploading || isPreparing">
```
EmmaLRussell, line 29: I guess technically we only need the spectrum and summary files to be available since coarse outputs aren't uploaded to ADR.

src/app/static/src/app/components/modelOutput/ModelOutput.vue
EmmaLRussell: Yeah, I think that's probably the case as they'd be expected to use the calibrate plot to check if calibration needed adjusting, thought I guess in some cases they may go back and re-calibrate, potentially before download files have finished preparing.

src/app/static/src/app/store/downloadResults/mutations.ts
EmmaLRussell: Should we also stop polling if that's still going on?

EmmaLRussell (commented):
Great, thank you, this is much tidier. Couple of small queries.

src/app/static/src/app/store/downloadResults/mutations.ts
hillalex: ah yes, good spot

src/app/static/src/app/components/modelOutput/ModelOutput.vue
hillalex: Good point; ok, so as long as when resetting ids we also stop polling, should be fine.

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -272,34 +198,22 @@
                 }
             },
             uploadDisabled() {
-                return !this.uploadFilesToAdr.length || this.downloadingFiles;
+                return this.uploadFilesToAdr.length == 0 || this.preparingFiles;
             },
-            downloadingFiles() {
+            preparingFiles() {
```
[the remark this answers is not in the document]
hillalex: You're totally right, can ditch this

EmmaLRussell (approved):
Works for me 👍

Nicolas-Dolan (approved):

src/app/static/src/app/components/downloadResults/DownloadResults.vue
```diff
@@ -133,42 +122,23 @@
...
+                (state) => state.coarseOutput),
+            currentFileUploading: mapStateProp<ADRUploadState, number | null>("adrUpload",
+                state => state.currentFileUploading),
+            totalFilesUploading: mapStateProp<ADRUploadState, number | null>("adrUpload",
+                state => state.totalFilesUploading),
+            uploading: mapStateProp<ADRUploadState, boolean>("adrUpload", state => state.uploading),
+            uploadComplete: mapStateProp<ADRUploadState, boolean>("adrUpload", state => state.uploadComplete),
+            releaseCreated: mapStateProp<ADRUploadState, boolean>("adrUpload", state => state.releaseCreated),
+            releaseFailed: mapStateProp<ADRUploadState, boolean>("adrUpload", state => state.releaseFailed),
+            uploadError: mapStateProp<ADRUploadState, Error | null>("adrUpload", state => state.uploadError),
```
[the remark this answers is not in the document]
hillalex: There is a way, to get good type checking involves splitting a `ComputedFromADRUpload`  interface out from the main `Computed` interface, pattern seen here: https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/components/modelRun/ModelRun.vue#L94

PR #664, opened 2022-02-01
mrc-2921 Remove calibrate plot result from local storage

EmmaLRussell:
## Description

Second attempt at slimming down saved state to local storage to prevent `setItem` quote exceeded error. Removes calibratePlotResult from state saved to local storage. No other changes appeared to be needed - plot result is re-fetched when project is reloaded. 

I'm struggling a bit to reproduce this bug in order to confirm that this will fix it (Jeff's issue was with full model run of latest Kenya data which I don't currently have access to). But it doesn't seem that this change should make things any worse at least!

## Type of version change
_Delete as appropriate_

Patches

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.88.5
+
+* Remove Calibrate Plot Result from localStorage
+
 # hint 1.88.4
 
 * Remove Projects state from localStorage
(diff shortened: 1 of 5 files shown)


Nicolas-Dolan (approved):

hillalex (approved):
Seems to work! I wonder what else in local storage is completely superfluous

PR #665, opened 2022-02-02
Ensure plotting metadata and set selected SAP data type on load 

EmmaLRussell:
## Description

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/app/store/baseline/actions.ts
+++ b/src/app/static/src/app/store/baseline/actions.ts
@@ -215,7 +215,7 @@ export const actions: ActionTree<BaselineState, DataExplorationState> & Baseline
     },
 
     async getBaselineData(context) {
-        const {commit, dispatch} = context;
+        const {commit, dispatch, state, rootState} = context;
         await Promise.all([
             api<BaselineMutation, BaselineMutation>(context)
                 .ignoreErrors()
@@ -233,6 +233,11 @@ export const actions: ActionTree<BaselineState, DataExplorationState> & Baseline
                 .freezeResponse()
                 .get<PjnzResponse>("/baseline/shape/")
         ]);
+
+        // Fetch plotting metadata if not present
+        if (state.iso3 && !rootState.metadata.plottingMetadata) {
+            await dispatch('metadata/getPlottingMetadata', state.iso3, {root: true});
+        }
 
         await dispatch('validate');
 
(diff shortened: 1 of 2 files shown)


hillalex:
I'm a little wary of pushing a code change for this rather than diagnosing the original bug; can we look back at recently merged PRs to look for likely cuplrits (seems like a very recent issue right?)

EmmaLRussell:
> I'm a little wary of pushing a code change for this rather than diagnosing the original bug; can we look back at recently merged PRs to look for likely cuplrits (seems like a very recent issue right?)

The one I'm most suspicious of is this: https://github.com/mrc-ide/hint/pull/647 since it changed the state saving code. It does treat metadata and surveyAndProgram in a slightly different way to the previous code, but it should be logically equivalent, and I can't really see how this issue could be arising for some projects but not others..

EmmaLRussell:
> > I'm a little wary of pushing a code change for this rather than diagnosing the original bug; can we look back at recently merged PRs to look for likely cuplrits (seems like a very recent issue right?)
> 
> The one I'm most suspicious of is this: #647 since it changed the state saving code. It does treat metadata and surveyAndProgram in a slightly different way to the previous code, but it should be logically equivalent, and I can't really see how this issue could be arising for some projects but not others..

This also does stuff with plottingMetadata, but at the modelCalibrate step - and the project this issue was observed for had not progress beyond Upload Inputs: https://github.com/mrc-ide/hint/pull/643/files

EmmaLRussell:
I think I understand a little more about what could be causing mrc-2992, but still not sure why we're seeing it on Production, 

I can reproduce the problem in dev if I 
- start a project, and take it beyond first step
- update the hintVersion and compile front end
- refresh

Similarly, if I delete the local storage key for the app and refresh. 

So what's happenening here is a mismatch between the front end and the back end. The front end queries for current project and current project files from the back end which uses the user's session to provide these, so the front end then knows the project id, name, note and input files. But it fails to load any state from localStorage because there's nothing in the expected key for the hint version. 

Worse than that, it then saves this mutilated state (with name, input files and nothing else) to the backend, so the project becomes permanently corrupted. 

So actually, this fix wouldn't help, as that doesn't refresh the whole state, just adds a couple of missing bits. This seems to be confirmed by report from Oli yesterday, who was seeing this same behaviour with a project that had calibrated. I think a better approach would be to identify when trying to load from an empty local storage state and in that case abort, clear current Project and redirect to the Projects page (unless you're a guest..) so it would load properly from the saved state from the back end when user re-selects the prject.

BUT I don't really understand how this could be happening on Production in the first place. If the problem is that the version is updated, that's because the whole app is redeployed, and the backend restarted. And in that case, all the existing sessions are blitzed, so people need re-login and select a project anyway. Aren't they..?

EmmaLRussell:
I've confirmed on dev that session is ended when app is redeployed, so I'm still confused about how this situation could occur outside a dev environment.

EmmaLRussell:
Closing this as it didn't address the core problem

PR #666, opened 2022-02-03
mrc-2903 Fix throttling of version state uploads

hillalex:
## Description

Prior to this change, multiple requests to upload version state were pending simultaneously, increasing load on the server and chance of timeouts/crashes while streaming large data downloads over a slow connection.

To see issue prior to this fix, set throttling (e.g. 450kb/s down, 150kb/s up) and watch console/network requests to see multiple pending requests to `/project/{pid}/version/{vid}/state` while downloading results.

After recompiling with this fix, repeat and see that only one pending request is in progress at any one time.

## Type of version change
Patch

## Checklist

### DIFF
--- a/src/app/static/src/app/store/projects/actions.ts
+++ b/src/app/static/src/app/store/projects/actions.ts
@@ -25,7 +25,7 @@ export interface ProjectsActions {
     createProject: (store: ActionContext<ProjectsState, RootState>, name: string) => void,
     getProjects: (store: ActionContext<ProjectsState, RootState>) => void
     getCurrentProject: (store: ActionContext<ProjectsState, RootState>) => void
-    uploadVersionState: (store: ActionContext<ProjectsState, RootState>) => void,
+    queueVersionStateUpload: (store: ActionContext<ProjectsState, RootState>) => void,
     newVersion: (store: ActionContext<ProjectsState, RootState>, note: string) => void,
     loadVersion: (store: ActionContext<ProjectsState, RootState>, version: VersionIds) => void
     deleteProject: (store: ActionContext<ProjectsState, RootState>, projectId: number) => void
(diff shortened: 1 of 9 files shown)


src/app/static/src/app/store/projects/actions.ts
```diff
@@ -145,7 +152,7 @@ export const actions: ActionTree<ProjectsState, RootState> & ProjectsActions = {
 
         const projectId = state.currentProject && state.currentProject.id;
         const versionId = state.currentVersion && state.currentVersion.id;
-        api<ProjectsMutations, ErrorsMutation>(context)
+        await api<ProjectsMutations, ErrorsMutation>(context)
```
hillalex, line 155: unrelated, just a missing `await`

src/app/static/src/app/store/projects/actions.ts
```diff
@@ -113,18 +113,25 @@ export const actions: ActionTree<ProjectsState, RootState> & ProjectsActions = {
...
         if (state.currentVersion) {
-            if (!state.versionUploadPending) {
-                commit({type: ProjectsMutations.SetVersionUploadPending, payload: true});
-                setTimeout(() => {
+            // remove any existing queued upload, as this request should supersede it
+            commit({type: ProjectsMutations.ClearQueuedVersionUpload});
+
+            const queuedId: number = window.setInterval(() => {
+                // wait for any ongoing uploads to finish before starting a new one
+                if (!state.versionUploadInProgress) {
```
hillalex, line 124: Note using `setTimeout` here would mean that if a request is already in progress at this point, the requested upload never happens. Using `setInterval` ensures that the most recent state change is persisted, even if it has to wait a little while for a previous request to finish uploading.

hillalex:
nb waiting to increment version after review

Nicolas-Dolan (approved):

EmmaLRussell (approved):

src/app/static/src/app/store/projects/actions.ts
hillalex: More or less, except that essentially the `setTimeout` we had in place was doing nothing other than delaying each request by 2 seconds! I think the original intention was that it would throttle requests, so if one was in flight, another would not be triggered... if that had been working however, it would have caused the first issue you reference - the missed uploads!

And yes, now there's always a queued request, but each incoming one supersedes the previous :+1:

PR #667, opened 2022-02-03
mrc-2903 rename downloading methods and properties for clarity

hillalex:
## Description

This PR just renames some methods and properties, fixes some type definitions, and adds a mock helper to the tests. This is ahead of some restructuring of the download logic (since the restructure is going to be a fairly large changeset, making these minor changes first to make the subsequent diff more reviewable.)

## Type of version change

None

## Checklist

### DIFF
--- a/src/app/static/src/app/components/downloadResults/DownloadResults.vue
+++ b/src/app/static/src/app/components/downloadResults/DownloadResults.vue
@@ -109,12 +109,12 @@
         getUserCanUpload: () => void
         getUploadFiles: () => void
         clearStatus: () => void;
-        downloadCoarseOutput: () => void
-        downloadSpectrum: () => void
+        prepareCoarseOutput: () => void
+        prepareSpectrumOutput: () => void
+        prepareSummaryReport: () => void
         downloadSummary: () => void
-        getSummaryDownload: () => void
-        getSpectrumDownload: () => void
-        getCoarseOutputDownload: () => void
+        downloadSpectrum: () => void
+        downloadCoarseOutput: () => void
         getUploadMetadata: (id: string) => void
         downloadUrl: (downloadId: string) => string
         stopPolling: (pollingId: number) => void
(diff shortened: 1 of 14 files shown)


src/app/static/src/app/types.ts
```diff
@@ -166,9 +166,8 @@ export interface UploadFile {
 
 export interface DownloadResultsDependency {
     downloadId: string
-    downloading: boolean
+    preparing: boolean
     statusPollId: number
-    status: DownloadStatusResponse
```
hillalex, line 171: This property was never referenced.

src/app/static/src/app/types.ts
```diff
@@ -166,9 +166,8 @@ export interface UploadFile {
 
 export interface DownloadResultsDependency {
     downloadId: string
-    downloading: boolean
```
hillalex, line 169: This caused me some confusion! Here and in the renamed methods, the outputs are actually being prepared for download.

EmmaLRussell (approved):
So the main idea here is to replace 'downloading' as a property of a downloadable file with 'preparing' to indicate that it's not actually being downloaded while this is true, but instead is being prepared on the server?

That makes a lot of sense. I wasn't sure if you were going to also want to rename the endpoints on the backend - I think it's probably a good thing to keep the submit/status/result pattern we've used elsewhere, and that pattern should make it clear where the prepare/download actions take place so the route being 'download/' hopefully isn't too confusing there. 

Coming back to this code, what confused me was how the state was moved on from the preparing stage to actually downloading the result. This happens, slightly weirdly, in a watch in the component, of summary/spectrum/coarseOutput, which then calls `handleDownloadResult` which checks if the `complete` flag is set and directly triggers the download from the `/result` url. I was going to suggest maybe renaming that handler to something more descriptive, but actually I guess that's all going to change in your next ticket, where the preparationg and download are going to be decoupled anyway.

All seems fine to me, tiny query about method naming.

src/app/static/src/app/components/downloadResults/DownloadResults.vue
```diff
@@ -109,12 +109,12 @@
...
-        downloadSpectrum: () => void
+        prepareCoarseOutput: () => void
+        prepareSpectrumOutput: () => void
+        prepareSummaryReport: () => void
         downloadSummary: () => void
-        getSummaryDownload: () => void
-        getSpectrumDownload: () => void
-        getCoarseOutputDownload: () => void
+        downloadSpectrum: () => void
+        downloadCoarseOutput: () => void
```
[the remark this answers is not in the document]
hillalex: Ah yeah, good spot

hillalex:
> Coming back to this code, what confused me was how the state was moved on from the preparing stage to actually downloading the result. This happens, slightly weirdly, in a watch in the component, of summary/spectrum/coarseOutput, which then calls `handleDownloadResult` which checks if the `complete` flag is set and directly triggers the download from the `/result` url. I was going to suggest maybe renaming that handler to something more descriptive, but actually I guess that's all going to change in your next ticket, where the preparationg and download are going to be decoupled anyway.

Yep, the current logic is all pretty weird to be honest, I think it will be a lot clearer once restructured!

Nicolas-Dolan (approved):

PR #668, opened 2022-02-04
Use help documents hosted by UNAIDS

r-ash:
## Description

Replace local help document with version hosted by UNAIDS. UNAIDS are now hosting this on their documentation website so we can link straight to that (I've mentioned to Ian that this URL cannot change otherwise things will break). This should hopefully mean we don't have to update this again and we can leave UNAIDS to update this document as and when.

Apologies if formatting doesn't fit the rules you guys use (I don't have autoformatting set up for ts - let me know if anything I need to update)

## Type of version change
_Delete as appropriate_

Minor

## Checklist


NEWS.md
```diff
@@ -1,3 +1,7 @@
+# hint 1.89.3
```
EmmaLRussell, line 1: Since this is a change in functionality rather than a bug fix, we'd generally make this a minor version rather than a patch, so would expect this to be 1.90.0

EmmaLRussell (commented):
This looks fine to me if this is how UNAIDS want the help files to be hosted. Would just do a tiny tweak to the version.

EmmaLRussell (approved):
LGTM 👍

hillalex (approved):

PR #669, opened 2022-02-07
mrc-2924 Prevent hiding 'Show more' button in warning when change step

EmmaLRussell:
## Description

The bug described in the ticket can be reprodued locally on master by changing the run-dependencies script to run with the real model (remove `-e USE_MOCK_MODEL=true \`). Create a new project, load Antarctica data and replace ART file with the file attached here. This will generate several ART warnings.
[demo_art_number_doctored.csv](https://github.com/mrc-ide/hint/files/8023599/demo_art_number_doctored.csv)

Run through to calibrate complete. You will see if you change steps between Calibrate/Review output/Save results that the "Show more" link is not visible on one or more of these steps so the truncated warnings cannot be expanded. 

This was caused by the component's computed method `warningsLengthy` returning false, which in turn was caused by the `fullBoxHeight` value, computed when warnings prop changes being smaller than expected. The `fullBoxHeight` values is taken from the element's clientHeight - however when the element is already being truncated by the css, that client height is not the elements 'natural' full height, but its truncated height. Thus, it is necessary to reset `fullBoxHeight` before recalculating it in order to find the true full height. 

It's difficult to unit test this, since the unit tests mock `clientHeight` anyway, so I think it's best just to manually test this fix. 

## Type of version change

_Delete as appropriate_

Patches

## Checklist


hillalex (approved):
Works for me :+1:

Nicolas-Dolan (approved):

PR #673, opened 2022-02-22
mrc-2930 add register link to login page

LekanAnni:
## Description

This PR adds registration link to login page.

## Type of version change
_Delete as appropriate_

Minor 

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.91.0
+
+* add register link to login page
+
 # hint 1.90.3
 
 * Make hintr warnings dismissable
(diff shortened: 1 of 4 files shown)


EmmaLRussell (commented):
LGTM. Could maybe avoid that awkward line break in the middle of the link just by adding a break before the link (see below)? Or alternatively could make the card a little wider. 
reqacc1
reqacc2

LekanAnni:
> I am not sure how you managed to reduce the page to that level, I tried to replicate it but the minimum screen size on Firefox, chrome and safari still fits well.

Okay, I think the second screen is what you had like, right ?

Nicolas-Dolan (approved):

EmmaLRussell:
> > I am not sure how you managed to reduce the page to that level, I tried to replicate it but the minimum screen size on Firefox, chrome and safari still fits well.
> 
> Okay, I think I misinterpreted your request, the second screen is what you had like, right ?

The second image was my suggestion to fix the issue with the first one. I was using Chrome on 100% zoom, nothing odd about settings I don't think.

EmmaLRussell (approved):
Thanks, looks good

PR #674, opened 2022-02-23
Mrc-2923 - Handle resource with missing download more gracefully

LekanAnni:
## Description
This PR checks if resource URL exists before importing its resource files. In some cases, resource will exists in ADR but without resource files, which causes the app to display errors.

## Type of version change
_Delete as appropriate_

Minor 

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# hint 1.91.0
+# hint 1.90.5
 
 * Handle resource with missing download gracefully
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.90.3
+
+* Make hintr warnings dismissable
+
 # hint 1.90.2
 
 * Fix warnings being truncated without 'Show more' visible
@@ -31,7 +35,8 @@
 * Remove Projects state from localStorage
 
 # hint 1.88.3
-Use area_id to define unique areas in Input Time Series
+
+* Use area_id to define unique areas in Input Time Series
 
 # hint 1.88.2
 
@@ -52,7 +57,6 @@ Use area_id to define unique areas in Input Time Series
 # hint 1.87.0
 
 * Formatting Input Time Series table values
-
 
 # hint 1.86.0
 
@@ -109,7 +113,6 @@ Use area_id to define unique areas in Input Time Series
 # hint 1.75.1
 
 * Update review output plots to use metadata from modelCalibrate
-
 
 # hint 1.75.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.90.4
+
+* Replace 'ART' with 'TARV' in French strings
+
 # hint 1.90.3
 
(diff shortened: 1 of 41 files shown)


EmmaLRussell (changes requested):
I can see how this would work but I think this approach adds quite a lot of boilerplate code to the component that's not ideal. I did think about changing this in the backend to just sidestep the error with trying to create a URL object with empty string - but then we would have to add code to return custom response in that situation. 

So I think the simplest thing to do would actually just be to change the relevant import actions eg [importPJNZ](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/baseline/actions.ts#L129) to do nothing if `url` parameter is falsy  e.g. 
```
async importPJNZ(context, url) {
       if (url) {
          await uploadOrImportPJNZ(context, {url: "/adr/pjnz/", payload: qs.stringify({url})},
              getFilenameFromImportUrl(url));
        }
    },
```

r-ash (approved):
I've deployed this to dev and checked it has fixed the issue, thanks!

LekanAnni:
> I can see how this would work but I think this approach adds quite a lot of boilerplate code to the component that's not ideal. I did think about changing this in the backend to just sidestep the error with trying to create a URL object with empty string - but then we would have to add code to return custom response in that situation.
> 
> So I think the simplest thing to do would actually just be to change the relevant import actions eg [importPJNZ](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/baseline/actions.ts#L129) to do nothing if `url` parameter is falsy e.g.
> 
> ```
> async importPJNZ(context, url) {
>        if (url) {
>           await uploadOrImportPJNZ(context, {url: "/adr/pjnz/", payload: qs.stringify({url})},
>               getFilenameFromImportUrl(url));
>         }
>     },
> ```

I actually considered this option, it was my first thought but I didn't think it's okay to handle this logic in action. We have always checked parameters before passing it down to action endpoint instead of checking in action.

EmmaLRussell:
>We have always checked parameters before passing it down to action endpoint instead of checking in action. It also makes testing easier.

It's actually a mixture - we have plenty of examples of actions which make decisions based on state/input parameters e.g. [here](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/baseline/actions.ts#L102) and [here](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/adrUpload/actions.ts#L23). I think in general it's better for the components to just be dealing with user interactions and leave any other decisions to do with state to the actions. 

Surely putting the check in the action would be much easier to test as the action test would just need to provide the truthy/falsy url parameter in the action parameter and then check if mockAxios has been called or not. The test you've added for the component check is doing masses of setup - it's almost 100 lines long! - plus you've had to add all those additional Vue.nextTicks...

EmmaLRussell (approved):
Could check mockAxios as well as commit, but I think as a minimal test this is fine since all the import functions do an initial commit.

Nicolas-Dolan (approved):

PR #676, opened 2022-02-25
Mrc-2929  Don't remove survey, ANC and ART data if user uploads a new baseline file

LekanAnni:
## Description

This PR removes the action to delete all SAP files when any Baseline file is deleted and run validation instead. 

**Problem statement**
We want to change this behaviour so that if a country say updates their PJNZ file but the other files remain the same they can easily remove it and upload the new one without also having to upload a new survey, anc and ART file.

## Type of version change
_Delete as appropriate_

Patches

## Checklist


EmmaLRussell (commented):
Looks good, just a tiny method rename suggestion

Nicolas-Dolan (approved):
LGTM although it's not clear to me how I should test the functionality.

EmmaLRussell (approved):

PR #683, opened 2022-03-02
Mrc 2917 Display warnings on review-inputs

LekanAnni:
## Description

This PR displays warning on `review inputs` step. Although, warnings are displayed on `review inputs` step but data was fetched from genericCharts data and not for other charts. @r-ash has confirmed that it's okay to display warnings generated from genericChart for `review inputs` step, ignoring other charts, so we do not need to worry about none GenericCharts.

Note that warnings are not displayed by default when a user reaches reviewInput step but when time-series is triggered. Science team asked for this behaviour as they had like to review warnings while chart is displayed. 

TO TEST
- load data
- navigate to step 2 (review inputs step)
- click on time-series chart
- see warning displayed

Please don't merge this branch to master yet. 
TO:DO 👍 
Change base hintr: mrc-2918 to master before merging

## Type of version change
_Delete as appropriate_

Minor

## Checklist


EmmaLRussell (approved):
Looks good 👍 

I think this approach is fine for the current app where the only GenericChart implementation is the Input Time Series, so we can assume that warnings for the Review Inputs step corresponds to the warnings stored in GenericCharts. If and when (maybe never!) any further generic charts are added elsewhere in the app we may need to distinguish between types of warnings in different generic  charts. But we don't need to worry about that for now.

Nicolas-Dolan (approved):
Works well. So science said that they only wanted the warning to appear when the user navigated to the time series tab, but did they mention what behaviour they would expect if the user navigates back to the Map tab? Currently, the warning persists, which seems fine, but you could argue that the tabs are being treated as separate warning states given their initial behaviour; so you might expect the warning to disappear if the user goes back to Map.

EmmaLRussell:
> 

Checked with Jeff on [Teams](https://teams.microsoft.com/l/message/19:83f74f91587b4848a35c45bca0c3a2b1@thread.tacv2/1647250600379?tenantId=2b897507-ee8c-4575-830b-4f8267c3d307&groupId=e24ac85e-dd9e-4505-8b9e-047b3a5fa6f0&parentMessageId=1647250600379&teamName=HIV%20Inference%20Group%20-%20WP&channelName=Naomi&createdTime=1647250600379), sounds like they're happy with current behaviour, so I think we're find to merge this.

PR #688, opened 2022-04-01
mrc-2800 Bug: hint tries to pull resources it does not have access to

Nicolas-Dolan:
## Description

To test, import [Togo HIV Estimates 2022](https://dev.adr.fjelltopp.org/country-estimates-22/togo-country-estimates-2022) from ADR. We are currently unauthorised to access the Geographic, Population, and Survey Data so hint should not try to download them and no error messages should appear. We can download the Spectrum file (ART and ANC files will not download because they require the Geographic aka shape files). Datasets that we have full permissions for should download as normal. Note: I decided against adding these changes to the reset method and kept the resource sources as from this.selectedDataset!.resources.

Unrelated backend tests failing on BK.

## Type of version change
Patches

## Checklist


LekanAnni (changes requested):
Looking good, not quite there yet.

I don't think this is the right approach because selectedDataset is heavily used within the app, where computed property listens to changes for resourceUpdates. But you replaced that with availableResources which is not reactive.

The right approach will be from here action or in BaselineMutation.SetDataset

EmmaLRussell (changes requested):
Works for me 👍 

I don't think there's a problem with the reactivity here as such, since this is computed in the component - it should be reactive as to changes in selected dataset. However, there is relevance to the refresh dataset metadata checking, since we not should not mark resources as out of date if user does not have access to them, and it should not attempt to pull them when user clicks refresh. 

Also, this component is getting pretty cluttered, so it would maybe be good anyway to make selectedDatasetAvailableResources a getter (in baseline module, I guess, since that's where selectedDataset lives). You could then also use it in  refreshDatasetMetadata action to not check for changes to resources which the user doesn't have access to.

Not sure what you mean by "decided against adding these changes to the reset method". Which method do you mean?

EmmaLRussell (commented):
This looks pretty good to me, a few small comments.

I'm a bit unclear about what would happen when we try to publish as a release to ADR with resources we don't have access to. I think it just wouldn't try to push unavailable resources, as they would probably have been set to null with your refresh mutation. Difficult to test with the Togo dataset though, as can't run without pjnz.

src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -112,15 +112,19 @@ export const actions: ActionTree<BaselineState, DataExplorationState> & Baseline
...
+                        }
+                        const resources: { [k in keyof DatasetResourceSet]?: DatasetResource | null } = {}
+                        Object.entries(avaliableResources).forEach(([key, value]) => {
+                            const schemaKey = key in exceptions ? exceptions[key as "pop" || "program"] : key
+                            resources[key as keyof typeof resources] = value ? findResource(metadata, schemas[schemaKey as keyof ADRSchemas]) : null
+                        })
 
                         commit(BaselineMutation.UpdateDatasetResources,
-                            {pjnz, pop, shape, survey, program, anc} as DatasetResourceSet)
+                            { ...resources } as DatasetResourceSet)
```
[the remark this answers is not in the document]
Nicolas-Dolan: I haven't been able to test publishing a release manually as I don't have sufficient files to reach that stage. It looks like you were using the Imperial College Togo dataset since that doesn't have pjnz. I was using the country's dataset, which does have the pjnz file but nothing else it can use. I tried combining the files from both datasets but there were validation issues. I'll try to find a dataset without the art and/or anc files to test.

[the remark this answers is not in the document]
Nicolas-Dolan: So Senegal matches the above criteria but it's read only so although I can reach the final step I don't have the option to do a release. I don't think it's going to be possible to test this manually as any dataset where we don't have permissions for some files is going to be read only.

src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -112,15 +112,19 @@ export const actions: ActionTree<BaselineState, DataExplorationState> & Baseline
...
                     if (response) {
                         const metadata = response.data;
-                        const pjnz = findResource(metadata, schemas.pjnz);
-                        const pop = findResource(metadata, schemas.population);
-                        const shape = findResource(metadata, schemas.shape);
-                        const survey = findResource(metadata, schemas.survey);
-                        const program = findResource(metadata, schemas.programme);
-                        const anc = findResource(metadata, schemas.anc);
+                        const avaliableResources = rootGetters ? rootGetters["baseline/selectedDatasetAvailableResources"] : {}
+                        const exceptions = { // where DatasetResource keys do not match ADRSchemas keys
```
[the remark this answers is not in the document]
Nicolas-Dolan: I don't think it's that simple as both ultimately come from adr. `DataResourceSet` is based on baseline's selectedDataset, which comes from getDataset (`/adr/datasets/${id}`) and `ADRSchemas `comes from getSchemas (`/adr/schemas/`).

src/app/static/src/app/store/baseline/baseline.ts
```diff
@@ -60,7 +60,33 @@ export const baselineGetters = {
...
+            const selectedDatasetFromDatasets = rooState.adr.datasets.find(dataset => dataset.id === selectedDataset.id) || null
+
+            const resourceTypes = {
+                pjnz: "inputs-unaids-spectrum-file",
+                pop: "inputs-unaids-population",
+                shape: "inputs-unaids-geographic",
+                survey: "inputs-unaids-survey",
+                program: "inputs-unaids-art",
+                anc: "inputs-unaids-anc"
+            }
```
LekanAnni, line 78: We do have these types in config.properties in backend, if you can't re-use that then you should make this reusable for frontend. You can add it to util and reuse in `baseline/getters.test.ts` and here

src/app/static/src/app/store/baseline/baseline.ts
```diff
@@ -60,7 +60,33 @@ export const baselineGetters = {
...
+
+            const checkResourceAvailable = (resourceType: string) => {
+                return selectedDatasetFromDatasets?.resources.some((resource: any) => resource.resource_type && resource.resource_type === resourceType)
```
LekanAnni, line 81: I am unsure why you need `resource.resource_type && resource.resource_type === resourceType`, having to access resource.resource_type twice.

src/app/static/src/app/store/baseline/baseline.ts
```diff
@@ -60,7 +60,33 @@ export const baselineGetters = {
...
+            }
+
+            Object.entries(resourceTypes).forEach(([key, value]) => {
+                resources[key as keyof typeof resources] = checkResourceAvailable(value) ? selectedDataset.resources[key as keyof typeof resources] : null
```
LekanAnni, line 85: You could format code to next line

src/app/static/src/app/store/baseline/baseline.ts
```diff
@@ -60,7 +60,33 @@ export const baselineGetters = {
         const validOrMissingPJNZ = !state.pjnzError
         const validOrMissingPop = !state.populationError
         return validOrMissingPJNZ && validOrMissingPop && state.validatedConsistent && !!state.shape
-    }
+    },
+    selectedDatasetAvailableResources: (state: BaselineState, getters: any, rooState: DataExplorationState) => {
```
LekanAnni, line 64:
```suggestion
    selectedDatasetAvailableResources: (state: BaselineState,  rooState: DataExplorationState): unknown => {
```

src/app/static/src/app/store/baseline/actions.ts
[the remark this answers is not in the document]
EmmaLRussell: OK, but looking at the logic, what does it seem to do? I guess it shouldn't give you the option to push files which aren't accessible and if that's the case for any file then it also shouldn't give you the option to do a release. I can't remember if it does any checks on this off the top of my head. Probably not? Or are we OK with them getting the option but it failing and showing an error message when user attempts to push? Probably need to check with Jeff/UNAIDS what desired behaviour is here. Jeff or Rachel might also be able to provide usable Togo pjnz to test with? Again, this could all go in another ticket, but I think we should try to get some clarity on it.

src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -112,15 +112,19 @@ export const actions: ActionTree<BaselineState, DataExplorationState> & Baseline
...
+                            pop: "population",
+                            program: "programme"
+                        }
+                        const resources: { [k in keyof DatasetResourceSet]?: DatasetResource | null } = {}
+                        Object.entries(avaliableResources).forEach(([key, value]) => {
+                            const schemaKey = key in exceptions ? exceptions[key as "pop" || "program"] : key
+                            resources[key as keyof typeof resources] = value ? findResource(metadata, schemas[schemaKey as keyof ADRSchemas]) : null
```
LekanAnni, line 123: I think you should revisit this code, I find it very hard to understand. You might want to have a separate method for separate computations.

src/app/static/src/app/store/baseline/actions.ts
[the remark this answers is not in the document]
LekanAnni: I think you should extend the schema to include new use case for better readability instead of mixing schemas within logic.

LekanAnni (changes requested):
Looking good, See inline comments.

src/app/static/src/app/store/baseline/baseline.ts
Nicolas-Dolan: I've added it to the utils, but I'm unsure where/how you intended it to be reused in the getters test.

src/app/static/src/app/store/baseline/baseline.ts
Nicolas-Dolan: I was being defensive, checking that resourceType existed on resource before comparing it.

src/app/static/src/app/store/baseline/actions.ts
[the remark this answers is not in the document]
Nicolas-Dolan: I've written up a ticket for this: mrc-3278

EmmaLRussell (approved):
This seems to fine to me, with the new tickets to address the remaining issues.

LekanAnni (approved):
LGTM 👍 

Thanks

PR #690, opened 2022-04-04
mrc-3111 Avoid mocking FuelManager instance client

EmmaLRussell:
## Description

Intermittent test failures appeared to be caused by the ErrorReportController unit test `can serialize request body`, because it mocks the FuelManager client instance, and this mock is not cleared for subsequent tests (see ticket mrc-3111). 

To avoid this, I moved the serialization test from ErrorReportControllerTests to a new unit test class for the FuelFlowClient, since that is where the serialization actually happens. Essentially, we just want to test that the client serializes the ErrorReport object to the expected json. The remaining ErrorReportControllerTests already implicitly check that `notifyTeams` is invoked from the controller, in the set-up of testFlowClient: https://github.com/mrc-ide/hint/blob/master/src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ErrorReportControllerTests.kt#L122 

The client's basic functionality is tested already in an integration test class, so this test is just checking the serialization part implemented within the `notifyTeams` method, not that the json is subsequently posted as expected. 

My first idea was to spy on the sut's `postJson` method and check that it called that method on itself with the expected json string. However, because the method is protected, spy will not work. So instead I created a subclass of the client class for testing purposes, which overrides the `postJson` method and captures the parameters which can then be verified. 

To enable this, I had to mark `postJson` in the client class as `open` - I think this is less of a code smell than opening it up from protected to public in order to test. 

## Type of version change
_Delete as appropriate_

None

## Checklist

### DIFF
--- a/src/app/src/test/kotlin/org/imperial/mrc/hint/unit/clients/FuelFlowClientTests.kt
+++ b/src/app/src/test/kotlin/org/imperial/mrc/hint/unit/clients/FuelFlowClientTests.kt
@@ -1,0 +1,98 @@
+package org.imperial.mrc.hint.unit.clients
+
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
+import com.nhaarman.mockito_kotlin.doReturn
+import com.nhaarman.mockito_kotlin.mock
+import org.assertj.core.api.Assertions
+import org.imperial.mrc.hint.AppProperties
+import org.imperial.mrc.hint.clients.FuelFlowClient
+import org.imperial.mrc.hint.models.ErrorReport
+import org.imperial.mrc.hint.models.Errors
+import org.junit.jupiter.api.Test
+import org.springframework.http.ResponseEntity
+
+class FuelFlowClientTests
+{
+    private val data = ErrorReport(
+            "test.user@example.com",
+            "Kenya",
+            "Kenya2022",
+            "Model",
+            "123",
+            "1234",
+            mapOf("spectrum" to "spectrum123", "summary" to "summary123", "coarse_output" to "coarse123" ),
+            listOf(
+                    Errors("#65ae0d095ea", "test error msg", "fomot-hasah-livad"),
+                    Errors("#25ae0d095e1", "test error msg2", "fomot-hasah-livid")
+            ),
+            "",
+            "test steps",
+            "test agent",
+            mapOf(
+                    "naomi" to "v1",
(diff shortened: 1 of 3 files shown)


src/app/src/test/kotlin/org/imperial/mrc/hint/unit/clients/FuelFlowClientTests.kt
```diff
@@ -0,0 +1,98 @@
...
+            ),
+            "2021-10-12T14:07:22.759Z"
+    )
+
+    @Test
+    fun `can serialize request body`()
+    {
+        class LocalFuelFlowClient(objectMapper: ObjectMapper,
+                                  appProperties: AppProperties
+        ): FuelFlowClient(objectMapper, appProperties)
```
LekanAnni, line 48: This looks suspicious but I have no comment.

LekanAnni (approved):
LGTM 👍 

Thanks for saving me

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/clients/FuelFlowClientTests.kt
EmmaLRussell: It is a bit smelly, as I say, but it's quite a tricky thing to mock.

LekanAnni: With your explanation, it could be something like below.

```
    @AfterEach
    fun tearDown() {
        FuelManager.instance.reset()
    }
```

LekanAnni: https://github.com/mrc-ide/hint/pull/691

EmmaLRussell: I did consider doing that , but to be honest, I still prefer this approach. Firstly because I think the controller test is wrong place to test the serialization of the ErrorReport class (it's not the controller that's doing that).  Secondly, because setting the FuelManager.instance makes even more assumptions about the internal guts of the Fuel clients than my subclass. Finally out of paranoia -  `tearDown` ought to be able to clear the mocked instance, but I think it's probably better not to mock it in the first place.

Nicolas-Dolan (approved):

Nicolas-Dolan:
Looks like Lekan would like to review again so I'll hold off merging for now.

LekanAnni (approved):

PR #691, opened 2022-04-05
Mrc 3110 - Reset Fuel client after each test

LekanAnni:
Alternative to solution provide her mrc-3111 

## Type of version change
_Delete as appropriate_

## Checklist


EmmaLRussell (changes requested):
Yeah, see my comment on mrc-3111 PR. This will fix the build, but I prefer the other approahc

LekanAnni:
> Yeah, see my comment on [mrc-3111](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3111) PR. This will fix the build, but I prefer the other approahc

I have seen the ticket but I don't see any reason why we should go with mrc-3111 approach, which is superfluous. 

For reference, the same approach is followed in their documentation. https://github.com/kittinunf/fuel/blob/105d3111d71623cb831af3f411ea253db766f369/fuel/src/test/kotlin/com/github/kittinunf/fuel/RequestHandlerTest.kt#L31

PR #692, opened 2022-04-06
mrc-3109 Display warnings from input validation

LekanAnni:
This PR displays warnings from SAP and GenericChart in Review input step, I merged the warnings from different data points such as SurveryResponse, ProgramResponse, AncResponse and TimeSeriesResponse. 

Even thou we currently don't expect data from GenericCharts, it will be populated once we do. 

Warnings will be displayed upon reaching review input step and once timeSeries is triggered, the additional chart warnings will be appended.

To test
* Start App
* Load data and navigate to review input
* Warnings should be displayed

If you want to test genericChart warnings are also appended, you can manually add test warnings here https://github.com/mrc-ide/hint/blob/a5ee5cf0dbe91b74ebbcfb9b86b9c8d99bed14ad/src/app/static/src/app/store/genericChart/actions.ts#L46

`const testWarnings = [{"text": "test genericChart", "locations": ['review_inputs']}]`

replace `data.warnings` with `testWarnings`

NB: DO NOT MERGE
depends on: https://github.com/mrc-ide/hintr/pull/370 to be merged
once hintr equivalent is merged then change hintr branch to master 

## Type of version change
_Delete as appropriate_

Minor 

## Checklist


Nicolas-Dolan (approved):

src/app/static/src/app/components/Stepper.vue
```diff
@@ -159,9 +160,13 @@
...
                 return !this.isComplete(activeStep)
             },
+            clearReviewInputWarnings: function() {
+                this.clearSurveyAndProgramWarnings()
+                this.clearGenericChartWarnings()
+            },
             clearWarnings(){
                 const mutationMethods: { [key: number]: () => void; } = {
-                    2: this.clearReviewInputsWarnings,
+                    2: this.clearReviewInputWarnings,
```
EmmaLRussell, line 169: I think `clearReviewInputsWarnings` is probably a better name since `Review Inputs` is the name of the step

src/app/static/src/app/store/root/getters.ts
```diff
@@ -67,8 +67,14 @@ export const getters: RootGetters & GetterTree<RootState, RootState> = {
             modelOptions: filterWarnings(state.modelOptions.warnings, location),
             modelRun: filterWarnings(state.modelRun.warnings, location),
             modelCalibrate: filterWarnings(state.modelCalibrate.warnings, location),
-            reviewInputs: filterWarnings(state.genericChart.warnings, location)
+            reviewInputs: filterWarnings(spreadReviewInputsWarnings(state), location)
         }
     }
 }
 
+const spreadReviewInputsWarnings = (state: RootState) => {
```
EmmaLRussell, line 75: Might be a little more obvious?

```suggestion
const allReviewInputsWarnings = (state: RootState) => {
```

src/app/static/src/app/store/surveyAndProgram/mutations.ts
```diff
@@ -73,5 +77,13 @@ export const mutations: MutationTree<SurveyAndProgramState> = {
 
     [SurveyAndProgramMutation.Ready](state: ReadyState) {
         state.ready = true;
+    },
+
+    [SurveyAndProgramMutation.WarningsFetched](state: SurveyAndProgramState, action: PayloadWithType<Warning[]>){
+        state.warnings = [...state.warnings, ...action.payload];
```
EmmaLRussell, line 83: This isn't quite going to work - if the user doesn't manually clear the warnings for a type, but then re-uploads or pulls a file, the warnings for the previous upload will remain behind and just get the new ones appended (so won't be cleared if the warnings are fixed) - unlike, say, in the modelRun's `WarningsFetched` where re-running causes the previous run's warnings to be overwritten. 

So unfortunately, I think you're going to have to keep the warnings per SAP type separately in the state, and pass the type as a parameter to the mutation (or have three separate mutations) so it can overwrite the correct set of warnings,

src/app/static/src/tests/root/getters.test.ts
```diff
@@ -349,7 +349,8 @@ describe("root getters", () => {
             modelOptions: mockModelOptionsState({warnings: undefined}),
             modelRun: mockModelRunState({warnings: undefined}),
             modelCalibrate: mockModelCalibrateState({warnings: undefined}),
-            genericChart: mockGenericChartState({warnings: undefined})
+            genericChart: mockGenericChartState({warnings: undefined}),
+            surveyAndProgram: mockSurveyAndProgramState({warnings: undefined})
```
EmmaLRussell, line 353: There should also be a test in this file that it gets all Review Inputs warnings when there are any?

EmmaLRussell (changes requested):
Looks really good in general, but it needs to clear warnings per sap file when revalidating, rather than always appending in the WarningsFetched mutation - see comment below.

To demonstrate this: 
- pull Antarctica HIV Estimates 2022 release Naomi testing2 v12
- go to Review Inputs, click Show more and see that you have three warnings
- go back to Upload Inputs and pull version 1.0 release of the same dataset
- go to Review Input, click Show more on the warnings and observe that you now have six warnings...

Otherwise, just some minor comments and suggestions. 

Also, I think there's a missing test of the changes to the Stepper.

EmmaLRussell (approved):
Looks good 👍 
I was going to suggest using a getter for warnings instead of storing the warnings twice - but of course we can't do that when SurveyAndProgramState inherits from WarningsState, so I can't think of a way of avoiding that..

PR #694, opened 2022-04-11
mrc-2849 Add area hierarchy to Input Time Series tooltips

EmmaLRussell:
## Description

As requested by UNAIDS, adds indication of area hierarchy to the Input Time Series charts. I've included area hierarchy in the tooltip rather than in a label on the subplot, as they can be quite long and it's difficult not to exceed the width of the subplot.

ITS_Plot_district

Targets related hintr branch so PLEASE DO NOT MERGE AS IS. 

Related ticket concerning area hierarchy in ITS map: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3138

## Type of version change
_Delete as appropriate_

Minor

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,11 @@
+# hint 1.97.0
+
+* Input Time Series area hierarchy tooltips
+
+# hint 1.96.0
+
+* Display warnings from input validation 
+
 # hint 1.95.1
 
 * Don't remove survey, ANC and ART data if user uploads a new baseline file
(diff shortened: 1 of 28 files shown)


LekanAnni (approved):
LGTM 👍 

Still can't wrap my head around jsonata

Nicolas-Dolan (approved):
Works well. In my opinion, we could leave off the last item in the hierarchy as it's given in the chart title and the tooltip can get a bit long. Could end with a trailing slash to make that clear. Happy to approve as is though.

PR #697, opened 2022-04-25
mrc-3170 clear warning before import/upload

LekanAnni:
A bug was discovered, Added a fix to clear current warning before import/upload of SAP files. See YT ticket for details. https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3170

## Type of version change

Patches

## Checklist


EmmaLRussell (commented):
Yup, that does the trick 👍 
While you've got this branch open, it would be good to fix the related bug, where if, instead of updating a SAP file, you delete it, and the warnings also are not removed (e.g. pull Niger dataset, then remove ANC file - ANC warnings are still visible on Review Inputs). That should just be a case of adding the same commit of empty warnings in the deleteXXX actions.

EmmaLRussell (approved):
Great, thanks for that. :slightly_smiling_face:

Nicolas-Dolan (approved):

PR #700, opened 2022-04-27
mrc-3074 Trigger generation of comparison report from front end

LekanAnni:
This PR triggered comparison report generation upon getting to download step. 

PLEASE DO NOT MERGE

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 1.97.0
+
+* Trigger generation of comparison report from front end
+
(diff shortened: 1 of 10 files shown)


Nicolas-Dolan (approved):

EmmaLRussell (approved):
Looks good. If we add another download type, we should definitely try to avoid replication of existing code by making some common methods.

PR #703, opened 2022-04-28
Mrc 3179 Display and upload output files to ADR if they are available for Upload

LekanAnni:
## Description

An error was reported in which some output files were not available because of meta/adr request failure but upload button was enabled, then allowed the user to proceeded upload to ADR which caused an error.  I thought about a simpler option where upload button gets disabled when an uploadable file failed to prepare but this will result in bad user experience. 

I also raised this issue in our weekly meeting and Jeff believes we should allow users to upload available outputFile but display a warning message. I have also adopted Emma suggestions, not allowing a release when all files are not being uploaded. 

TO TEST CASE 1
* navigate to download step
* restart hintr-redis container `docker restart hintr_redis`
* then click on upload modal

TO TEST CASE 2
* navigate to download step
* wait for all downloads to be complete
* go back to calibration page and recalibrate
* Console should show meta/adr request error, if not find a way to sabotage meta/adr request for output download
* then click on upload modal

You will see only the output files that succeeded in upload modal checklist, these are the files that will be uploaded to ADR. Error message has also been added to the modal.

## Type of version change

Patches

## Checklist

### DIFF
--- a/src/app/static/src/app/store/downloadResults/actions.ts
+++ b/src/app/static/src/app/store/downloadResults/actions.ts
@@ -117,6 +117,8 @@ export const getSummaryReportStatus = async function (context: ActionContext<Dow
 
     if (response && response.data?.done) {
         await getADRUploadMetadata(response, dispatch).then(() => {
+            // commit is neccessary for dispatching action
+            // to retrieve metadata for summary report
             commit({
                 type: DownloadResultsMutation.SummaryMetadataError,
                 payload: rootState.metadata.adrUploadMetadataError
@@ -135,6 +137,8 @@ export const getSpectrumOutputStatus = async function (context: ActionContext<Do
 
     if (response && response.data?.done) {
         await getADRUploadMetadata(response, dispatch).then(() => {
+            // commit is neccessary for dispatching action
+            // to retrieve metadata for spectrum report
             commit({
                 type: DownloadResultsMutation.SpectrumMetadataError,
                 payload: rootState.metadata.adrUploadMetadataError
@@ -153,6 +157,8 @@ export const getCoarseOutputStatus = async function (context: ActionContext<Down
 
     if (response && response.data?.done) {
         await getADRUploadMetadata(response, dispatch).then(() => {
+            // commit is neccessary for dispatching action
+            // to retrieve metadata for coarseOutput report
             commit({
                 type: DownloadResultsMutation.CoarseOutputMetadataError,
                 payload: rootState.metadata.adrUploadMetadataError
@@ -171,6 +177,8 @@ export const getComparisonOutputStatus = async function (context: ActionContext<
 
     if (response && response.data?.done) {
         await getADRUploadMetadata(response, dispatch).then(() => {
+            // commit is neccessary for dispatching action
+            // to retrieve metadata for comparison report
             commit({
                 type: DownloadResultsMutation.ComparisonOutputMetadataError,
                 payload: rootState.metadata.adrUploadMetadataError


src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -166,7 +193,9 @@
                 if (this.uploadableFiles) {
                     return Object.keys(this.uploadableFiles).reduce((sections, key) => {
                         if (outputFileTypes.includes(key)) {
-                            sections[0][key] = this.uploadableFiles[key];
+                            if (Object.keys(this.getAvailableOutputFiles(key)).length > 0) {
```
EmmaLRussell, line 196: This seems like quite a confusing way of expressing this - the name `getAvailableOutputFiles` implies it might return more than one, which never happens, and it's not idea to have to call it twice. How about replacing `getAvailableOutputFiles` with `outputFileIsAvailable` which just returns a bool, then use `this.uploadableFiles[key]` as before?

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -166,7 +193,9 @@
...
+                                sections[0][key] = this.getAvailableOutputFiles(key)
+                            }
```
EmmaLRussell, line 198:
```suggestion
                            if (outputFileIsAvailable) {
                                sections[0][key] = this.uploadableFiles(key)
                            }
```

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -147,9 +155,28 @@
...
+                    availableOutputFiles = this.uploadableFiles[key]
+                } else if (key === "outputSummary" && !this.summaryIsAvailableForUpload) {
+                    availableOutputFiles = this.uploadableFiles[key]
+                }
+
+                return availableOutputFiles
+            },
+            getSectionHeadings: function (sectionIndex) {
+                const hasOutputFile = Object.keys(this.uploadFileSections[0]).length > 0
+                return sectionIndex === 0 ? hasOutputFile ? 'outputFiles' : '' : 'inputFiles';
```
EmmaLRussell, line 172: Nested ternaries are pretty nasty to read - could add some formatting or parentheses to make it easier?

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -147,9 +155,28 @@
...
+                let availableOutputFiles = {} as UploadFile
+                if (key === "outputZip" && !this.spectrumIsAvailableForUpload) {
+                    availableOutputFiles = this.uploadableFiles[key]
+                } else if (key === "outputSummary" && !this.summaryIsAvailableForUpload) {
+                    availableOutputFiles = this.uploadableFiles[key]
+                }
+
+                return availableOutputFiles
+            },
+            getSectionHeadings: function (sectionIndex) {
```
EmmaLRussell, line 170: Could you rename this to `getSectionHeading` - it only returns one heading.

EmmaLRussell:
Sorry, meant to include that last comment in a full review, which is on its way!

EmmaLRussell (changes requested):
So if I understand it, this branch is intended only to address the issue where an output file cannot be uploaded to ADR because there was an error with generating it. (So it's not related to the discussion on https://github.com/mrc-ide/hint/pull/688 about not making ADR resources available to upload if user does not have permissions over the ADR resource?)

Is it likely to be the case that users want to upload anything to ADR if there has been an error in generating one of the files? I guess we can give them the option but are we definitely ok with letting them do that?

I think 'create a new ADR release' should also be unavailable if any output files failed, since that requires upload of all files. 

I wonder if we should make the situation even clearer to the user by including text in the upload modal explaining why files are not available to upload. There will be errors visible in the page, but user may not have noticed those or connected them with the files being gone from the modal. (And release not being available either)

LekanAnni:
> So if I understand it, this branch is intended only to address the issue where an output file cannot be uploaded to ADR because there was an error with generating it. (So it's not related to the discussion on #688 about not making ADR resources available to upload if user does not have permissions over the ADR resource?)
> 
You are right, it's not related to mrc-2800

> Is it likely to be the case that users want to upload anything to ADR if there has been an error in generating one of the files? I guess we can give them the option but are we definitely ok with letting them do that?
> 

This seems to be a valid question, I think we might need to ask science team.

> I think 'create a new ADR release' should also be unavailable if any output files failed, since that requires upload of all files.
> 
> I wonder if we should make the situation even clearer to the user by including text in the upload modal explaining why files are not available to upload. There will be errors visible in the page, but user may not have noticed those or connected them with the files being gone from the modal. (And release not being available either)

That's right, we could display a warning/info message on the modal and a retry button to pull resources again.

src/app/static/src/app/store/downloadResults/actions.ts
```diff
@@ -106,48 +106,78 @@ export const actions: ActionTree<DownloadResultsState, RootState> & DownloadResu
...
     const downloadId = state.spectrum.downloadId;
     const response = await api<DownloadResultsMutation, DownloadResultsMutation>(context)
         .withSuccess(DownloadResultsMutation.SpectrumOutputStatusUpdated)
         .withError(DownloadResultsMutation.SpectrumError)
         .get<ModelStatusResponse>(`download/status/${downloadId}`);
-    await getADRUploadMetadata(response, dispatch);
+
+    if (response && response.data?.done) {
+        await getADRUploadMetadata(response, dispatch).then(() => {
+            commit({
```
EmmaLRussell, line 137: Committing this mutation really ought to be the `withError` clause of the `getADRUploadMetadata` action - is there any reason to separate them here? In fact, why do we even need to do a separate mutation here to add it to the download state if we're already going to have the error available in the metadata state? So that it's available in 'summary' and 'spectrum' in the component? I think it would be better to just to have the values in the state once rather than having to manage duplication here. You could add a computed into the component to arrange them in a nice way if youlike.

EmmaLRussell (commented):
This is looking good. I don't think we should keep the metadata error in the state twice though.

src/app/static/src/app/store/downloadResults/actions.ts
EmmaLRussell: Sorry, yes, didn't appreciate that you'd be overwriting the error for each type so you do need to commit into a different bit of the state. I still don't like this commit happening here, though - it's associated with getting ADR metadata so it feels like it should be part of that action to make any necessary commits. 

Maybe we should just save separate errors by upload type in the adrUploadMetadataError rather than a single one, if we need to keep them by type now, rather than also adding to the DownloadResults state? As far as I can tell we don't seem to do anything with that one in the metadata state anyway so it's not doing anything useful for us. getADRUploadMetadata action could take type (summary/spectrum) as parameter so would know which upload type to commit for,

EmmaLRussell: OK, I think you're probably right that this is the least bad way. But would you mind putting a brief comment here about the need for this commit as this is certainly not the standard way to do commits following actions

EmmaLRussell (approved):

src/app/static/src/app/components/downloadResults/UploadModal.vue
```diff
@@ -6,9 +6,11 @@
                 <div id="dataset-id" class="mt-4">
                     <span v-translate="'uploadFileDataset'"></span>
                     <span>{{ dataset }}</span></div>
+                <div class="pt-3 text-danger" id="output-file-error" v-if="outputFileError" v-translate="outputFileError"></div>
```
Nicolas-Dolan, line 9: No change needed here, but we should probably think about standardising these basic error messages. Would a component be overkill?

Nicolas-Dolan (approved):

src/app/static/src/app/components/downloadResults/UploadModal.vue
LekanAnni: You are right, since it's not a standard api error and no repetition as such. I don't see a point creating a component for it.

PR #704, opened 2022-04-29
Mrc 3076 Upload Comparison report to ADR

LekanAnni:
## Description

## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/app/components/downloadResults/DownloadResults.vue
```diff
@@ -112,27 +120,31 @@
...
+        comparisonSwitch: boolean
     }
 
     export default Vue.extend<Data, Methods, Computed>({
         name: "downloadResults",
         data() {
             return {
-                uploadModalOpen: false
+                uploadModalOpen: false,
+                comparisonSwitch: !switches.comparisonOutput
```
EmmaLRussell, line 138: Why is this doing a negation? Show comparison view should be true if the switch is true, shouldn't it?

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -572,6 +575,7 @@ const en: Translations = {
     uploadingStatus: "Uploading {{ fileNumber }} of {{ totalFiles }} (this may take a while)",
     uploadInputs: "Upload inputs",
     uploadComplete: "Upload complete",
+    uploadFileOutputComparison: "Comparison report",
```
EmmaLRussell, line 578: I suspect these might change to something wordier, pending what UNAIDS say about the mockups, but good to have a placeholder.

EmmaLRussell (commented):
Will hold off doing a full review until previous branch is merged for clean diff, and agreed interface with UNAIDS/Fjeltopp. But looks good for now (except switch seems to be the wrong way round - seeing the elements when switch is off).

Might be best to add the schema to config.properties now so that you can properly test that it's getting returned by the schemas endpoint. The value might change after Fjeltopp add it to the datasets, but still better to test with some value, rather than null, for now I think.

EmmaLRussell:
One other small thing - with the switch (actually) off, there is a very noticeable delay between all visible files completing generation and the Upload to ADR button becoming enabled while the invisible comparison report is generated.  So might be best to switch on the files being awaited by feature switch.

EmmaLRussell (approved):
Thanks!

Nicolas-Dolan (commented):
I'm not sure if I'm testing this properly, but I'm still getting the summary report rather than a comparison report when I download. Also, if I select specific files and try to upload the comparison report alone I get an undefined POST.
image

Nicolas-Dolan:
> > I'm not sure if I'm testing this properly, but I'm still getting the summary report rather than a comparison report when I download. Also, if I select specific files and try to upload the comparison report alone I get an undefined POST. image
> 
> Comparison report still isn't implemented in the back end so what gets downloaded is still the summary report (though it should be called 'comparison report' now). Actually uploading to ADR won't work as the resource type hasn't been added at their end - still some uncertainty about whether UNAIDS will actually want to put these reports on ADR, tbc when they get a look at what's actually in them (though they probably will). Hence this lot being behind a feature switch for now.. so we will need to test all this for real later.

Ah ok, but should the post request have an undefined value in it, even if ADR are not ready to accept it?

EmmaLRussell:
> Ah ok, but should the post request have an undefined value in it, even if ADR are not ready to accept it?

Ah, well spotted - probably not!

r-ash (approved):
This looks great, it uploads the comparison report to ADR nicely. As mentioned in chat, the app is failing to create a release for me. So could be a new issue or me not understanding the intended behaviour. However ADR guys have said their dev system may be a bit flaky today so let's review again tomorrow. Either way I am having the same issue creating a release on staging with current master so I think unrelated to this change.

EmmaLRussell:
Ha, you merged just as I was testing it! Looks like they've fixed creating releases on the dev system, worked for me.

PR #707, opened 2022-05-13
mrc-3194 Prevent spamming download and meta endpoints after download generate completes

Nicolas-Dolan:
## Description

To reproduce the bug, progress to save results step. While download files are still generating (reloading the page will trigger a regeneration), navigate to the the project history page and create a new project. Observe in the console that `metadata` and `downloadResults` endpoints are being spammed:
image
This fix prevents this and returns normal behaviour, although it does not fix the bug outlined in mrc-3239.
image

**Tip:** to avoid having to go through the stepper every time you want to reproduce the bug, proceed to the final step once and then clone the project multiple times before triggering the bug. The clones should not be affected by the state bug outlined in mrc-3239.

**Update:** PR now also includes fix for state wiping bug outlined in mrc-3239

## Type of version change
Patches

## Checklist


LekanAnni:
https://user-images.githubusercontent.com/4240748/169860311-376af78e-acec-4973-9f2f-a31db8dc35ae.mov

LekanAnni:
Download pollings are not showing but the state us still destroyed.

Nicolas-Dolan:
> As discussed, a better solution would be to call `DownloadResultsMutation.ResetIds` from createProject action.

This pr was already doing that just without the `preparing` checking if statement beforehand. I'm still getting the bug though. Were there any other changes? Had we changed the resetId mutation?

Nicolas-Dolan:
PR now also includes fix for state wiping bug outlined in mrc-3239

LekanAnni (changes requested):
I can't seem to reproduce the bug anymore, looks like the fix works. 

I however noticed that when a tester follows the steps to reproduce the bug, upon creating a new project, the state from an old project such as current step and warnings appear for the new project before resetting itself to the actual project state. It would be nice if you could ensure a new project does not initially display old state.

Nicolas-Dolan:
> I can't seem to reproduce the bug anymore, looks like the fix works.
> 
> I however noticed that when a tester follows the steps to reproduce the bug, upon creating a new project, the state from an old project such as current step and warnings appear for the new project before resetting itself to the actual project state. It would be nice if you could ensure a new project does not initially display an old state.

I haven't been able to observe that testing it myself. So when you created a new project, it initially opens on a later step and visible warnings or are you just observing the state in the console?

LekanAnni:
> > I can't seem to reproduce the bug anymore, looks like the fix works.
> > I however noticed that when a tester follows the steps to reproduce the bug, upon creating a new project, the state from an old project such as current step and warnings appear for the new project before resetting itself to the actual project state. It would be nice if you could ensure a new project does not initially display an old state.
> 
> I haven't been able to observe that testing it myself. So when you created a new project, it initially opens on a later step and visible warnings or are you just observing the state in the console?

Warnings are displayed for a second or 2 on the page not in console.

Nicolas-Dolan:
> > > I can't seem to reproduce the bug anymore, looks like the fix works.
> > > I however noticed that when a tester follows the steps to reproduce the bug, upon creating a new project, the state from an old project such as current step and warnings appear for the new project before resetting itself to the actual project state. It would be nice if you could ensure a new project does not initially display an old state.
> > 
> > 
> > I haven't been able to observe that testing it myself. So when you created a new project, it initially opens on a later step and visible warnings or are you just observing the state in the console?
> 
> Warnings are displayed for a second or 2 on the page not in console.

Weird, I can't see that at all. Maybe it's just happening a lot faster for me. I'll take a look

EmmaLRussell:
> Warnings are displayed for a second or 2 on the page not in console.

I'm seeing this behaviour, but not on create project - I can see it when you reload an empty pre-existing project having been in a project on a step with warnings. That's happening for me on master, so I don't think it's just an issue for this branch. Would suggest fixing that in a separate ticket.

EmmaLRussell (approved):
Nice, this works for me.

However, we've still got the same issue with calibrate - i.e. if you browse to Projects and create w new project while calibration is still going on, you'll see that the calibrate status keeps getting spammed. It would be really good to get the state wipe fix in asap though, so I'd suggest addressing that in a separate ticket. And same for the stale warnings issue @LekanAnni mentioned. 

You can't seem get into the same pickle with Model Fit, because the model fit has a modal dialog that prevents you browsing to Projects while the fit it going on, and if you browse to /projects manually in the same tab that has the effect of clearing the intervals anyway.

LekanAnni (approved):
LGTM 👍 

If you could create a ticket to fix the stale state in my previous comment, I am happy for this to be merged.

PR #710, opened 2022-06-09
Mrc 2580 Include project notes in output zip

LekanAnni:
## Description

This PR adds project and version notes payload to download spectrum endpoint.

## Type of version change
_Delete as appropriate_

Minor 

## Checklist


src/app/src/main/kotlin/org/imperial/mrc/hint/clients/HintrAPIClient.kt
```diff
@@ -184,7 +184,7 @@ class HintrFuelAPIClient(
             return postEmpty("download/submit/${type}/${id}")
         }
 
-        val json = objectMapper.writeValueAsString(mapOf("state" to projectState))
+        val json = objectMapper.writeValueAsString(projectState)
```
EmmaLRussell, line 187: For clarity, might be worth calling this parameter something other than 'projectState' since it now contains both state and notes. "projectInfo" or "projectPayload"?

src/app/static/src/app/utils.ts
```diff
@@ -191,6 +191,10 @@ export const formatDateTime = (isoUTCString: string) => {
     return moment.utc(isoUTCString).local().format('DD/MM/YYYY HH:mm:ss');
 };
 
+export const formatToISODateTime = (isoUTCString: string) => {
+    return moment.utc(isoUTCString).local().format('YYYY/MM/DD HH:mm:ss');
```
EmmaLRussell, line 195: So we want to convert our UTC ISO strings to local ISO strings for the note outputs? Do we definitely want ISO format? Just our standard format as shown in the app might be friendlier if these things are basically human-readable

src/app/static/src/app/utils.ts
```diff
@@ -191,6 +191,10 @@ export const formatDateTime = (isoUTCString: string) => {
     return moment.utc(isoUTCString).local().format('DD/MM/YYYY HH:mm:ss');
 };
 
+export const formatToISODateTime = (isoUTCString: string) => {
```
EmmaLRussell, line 194:
```suggestion
export const formatToLocalISODateTime = (isoUTCString: string) => {
```

src/app/static/src/app/store/root/getters.ts
```diff
@@ -119,3 +133,11 @@ const allReviewInputsWarnings = (state: RootState) => {
     return [...sapWarnings, ...genericChartWarnings]
 }
 
+const getVersionNotes = (versions: Version[], name: string) => {
+    return versions.reduce((result: Note[], versionNote) => {
+        const note = versionNote.note || ""
+        const updated = versions && formatToISODateTime(versionNote.updated)
+
+        return result.concat({note, updated, name})
```
EmmaLRussell, line 141: I think "name" here should probably be the version marker, or project name plus version marker e.g. My Project v2. Otherwise all the note sections just get labelled with the project name.

EmmaLRussell (changes requested):
Looks like a good approach. A couple of issues in addition to the comments below:

1. In the ticket, it suggests that we should only be sending notes for this version plus those older than this one (by updated date). There isn't any code dealing with this here, but even if there was, I don't think it would give us exactly what we intended. For example, if we take version 1 to calibrate stage, then go off and make a version 2, then come back to version 1 - it looks like just loading the project gets recorded as an update so whichever project you have open is considered to be the most recent... @rob is that an issue? I think it was borderline whether we thought it was ok to always include all versions, including more recent ones. 

2. I have a project named 'p1' with 2 versions. The project note is "project 1 note", v1's version note is "version 1 note", v2's version note is "version 2 note". I can see all these notes correctly in the Project History page. However, the notes that get sent to the backend (when creating the output zip for version 2) get generated into notes.txt like this:
```
Project notes:

p1
2022/06/14 14:37:40
project 1 note

Version notes:

p1
2022/06/14 14:37:40
project 1 note

p1
2022/06/14 14:38:02
version 2 note
```

So the version 1 note appears to have been replaced by the project note (see comment below on separate issue where the notes are all being named just with the project name). I tried opening version 1 and generating the download for that - I checked network traffic to confirm that this is an issue coming from the front end, and this time I could see that both version notes were now replaced with "project 1 note"!  So it seems that there must be a bug on setting versions and their notes on `currentProject` when a version is re-opened (since this is where those version notes are coming from, as opposed to the full projects list which ProjectHistory is using) which will need fixing to enable this code to work.

src/app/static/src/app/utils.ts
LekanAnni: @r-ash has advocated for moving the standard format  as well. For now, rehydrate endpoint expects this format  `YYYY/MM/DD HH:mm:ss`, so will leave as is.

LekanAnni: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3302

EmmaLRussell (commented):
Great to get that version note fix in! Amazed we haven't spotted that before! 

This looks good, as long as @r-ash is happy with including all version notes. 

One small issue with the notes as output to notes.txt - not seeing a "v" prefix for the version number, which I was expecting. I think you just need to include that in your version identifier string format:
```
Project notes:

p1
2022/06/16 10:17:49
project 1

Version notes:

p1-1
2022/06/16 10:17:49
version 1

p1-2
2022/06/16 10:17:53
version 2
```

r-ash:
> 1. In the ticket, it suggests that we should only be sending notes for this version plus those older than this one (by updated date). There isn't any code dealing with this here, but even if there was, I don't think it would give us exactly what we intended. For example, if we take version 1 to calibrate stage, then go off and make a version 2, then come back to version 1 - it looks like just loading the project gets recorded as an update so whichever project you have open is considered to be the most recent... @r-ash  is that an issue? I think it was borderline whether we thought it was ok to always include all versions, including more recent ones.

Noted! Didn't realise the version updated every time it was opened. If that is the case then makes sense just to skip this requirement and include all the notes. If that becomes too noisy we can deal with it as an enhancement in the future.

src/app/static/src/app/store/projects/actions.ts
```diff
@@ -79,7 +79,7 @@ export const actions: ActionTree<ProjectsState, RootState> & ProjectsActions = {
 
         //Ensure we have saved the current version
         if (state.currentVersion) {
-            immediateUploadVersionState(context);
+            await immediateUploadVersionState(context);
```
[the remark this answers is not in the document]
LekanAnni: I am not sure the kind of issue that cam up in past, is it one that go against loading data asynchronously ? I am happy to remove it for now but I don't think we should not be waiting for immediateUploadVersionState when the call itself is asynchronous. Nevertheless, I will yank await out.

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ProjectsControllerTests.kt
```diff
@@ -163,7 +163,6 @@ class ProjectsControllerTests
         assertThat(projects[0]["sharedBy"].asText()).isEqualTo("")
         val versions = projects[0]["versions"] as ArrayNode
         assertThat(versions.count()).isEqualTo(1)
-        assertThat(versions[0]["note"].asText()).isEqualTo("version notes")
```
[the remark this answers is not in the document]
LekanAnni: Its duplicated in verifyZeroInteractions(mockLogger) below it.

LekanAnni:
> Great to get that version note fix in! Amazed we haven't spotted that before!
> 
> This looks good, as long as @r-ash is happy with including all version notes.
> 
> One small issue with the notes as output to notes.txt - not seeing a "v" prefix for the version number, which I was expecting. I think you just need to include that in your version identifier string format:
> 
> ```
> Project notes:
> 
> p1
> 2022/06/16 10:17:49
> project 1
> 
> Version notes:
> 
> p1-1
> 2022/06/16 10:17:49
> version 1
> 
> p1-2
> 2022/06/16 10:17:53
> version 2
> ```

version number was added from `hintr`, so @r-ash , could you add `v` to the version number ?

I will add `v` to the `projectName-v2`

LekanAnni:
> So I can prefix `v` on the version `name` if that would get the desired effect but in the example above it looks like the version `name` is coming through as `p1-2` which would then look weird to put a `v` on the front of?
> 
> I think `hint` part should just manage this as `hint` is setting the version name and project name for display so should be responsible for sending that through in the same format as we show in the web app.

The change has to be driven from hintr. For example, I need `DownloadSubmitRequest` schema to be updated and for hintr to expect versionNumber. So I can send `{name, updated, note, versionNumber}`.

EmmaLRussell:
> > So I can prefix `v` on the version `name` if that would get the desired effect but in the example above it looks like the version `name` is coming through as `p1-2` which would then look weird to put a `v` on the front of?
> > I think `hint` part should just manage this as `hint` is setting the version name and project name for display so should be responsible for sending that through in the same format as we show in the web app.
> 
> The change has to be driven from hintr. For example, I need `DownloadSubmitRequest` schema to be updated and for hintr to expect versionNumber. So I can send `{name, updated, note, versionNumber}`.

@LekanAnni I think you just need to tweak the string you're already formatting in the front end - that will get written out by hintr, without hintr caring what the actual version number is.

src/app/src/test/kotlin/org/imperial/mrc/hint/unit/controllers/ProjectsControllerTests.kt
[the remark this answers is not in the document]
EmmaLRussell: Sorry, I don't understand what you mean, `verifyZeroInteractions` just tests that the logger hasn't been called, it's not checking the notes in the version. Oh, do you mean it's duplicated in `assertExpectedVersion`? If so - yes, you're right! 😆

src/app/static/src/app/store/projects/actions.ts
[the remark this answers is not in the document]
EmmaLRussell: Fine to leave it,

EmmaLRussell (approved):
Thanks, LGTM. Notes with 'v' prefix on version numbers now being output correctly.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/DownloadTests.kt
```diff
@@ -70,11 +70,22 @@ class DownloadTests : SecureIntegrationTests()
...
+        val state = mapOf("state" to mapOf(
                 "datasets" to datasets,
                 "model_fit" to mapOf("id" to "", "options" to mapOf("" to "")),
                 "calibrate" to mapOf("id" to "", "options" to mapOf("" to "")),
                 "version" to mapOf("hintr" to "1", "naomi" to "1", "rrq" to "1")
+        ),
+                "notes" to mapOf(
+                        "project_notes" to note,
+                        "version_notes" to listOf(note, note)
+                )
```
Nicolas-Dolan, line 88:
```suggestion
        val state = mapOf(
                "state" to mapOf(
                        "datasets" to datasets,
                        "model_fit" to mapOf("id" to "", "options" to mapOf("" to "")),
                        "calibrate" to mapOf("id" to "", "options" to mapOf("" to "")),
                        "version" to mapOf("hintr" to "1", "naomi" to "1", "rrq" to "1")
                ),
                "notes" to mapOf(
                        "project_notes" to note,
                        "version_notes" to listOf(note, note)
                )
```

Nicolas-Dolan (approved):
LGTM

PR #711, opened 2022-06-09
mrc-3118 Upload output zip action which calls hintr endpoint via backend

LekanAnni:
## Description

This PR loads the whole outputZip and prints response in console. I have added endpoints to the backend and call these endpoints in front end. 

TO TEST
* Use the old load menu, it does accept json and zip file.
* Once loaded see console for response from rehydrate.

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/static/src/app/store/load/mutations.ts
```diff
@@ -27,5 +36,38 @@ export const mutations: MutationTree<LoadState> & LoadMutations = {
...
+        state.rehydrateResult = action.payload;
+        state.preparing = false;
+        state.loadingState = LoadingState.NotLoading;
+    },
+    RehydrateResultError(state: LoadState, action: PayloadWithType<Error>) {
+        state.loadError = action.payload;
+        state.preparing = false;
+        state.loadingState = LoadingState.LoadFailed;
+        if (state.statusPollId > -1) {
+            stopPolling(state);
```
EmmaLRussell, line 50: codecov warning claiming this line isn't tested.

src/app/static/src/app/store/load/mutations.ts
```diff
@@ -27,5 +36,38 @@ export const mutations: MutationTree<LoadState> & LoadMutations = {
...
+        }
+    },
+    PreparingModelOutput(state: LoadState, action: PayloadWithType<ProjectRehydrateSubmitResponse>) {
+        state.downloadId = action.payload.id;
```
EmmaLRussell, line 54: Can we call this `rehydrateId`? Download Id is used when generating the downloadable files.

src/app/static/src/app/store/load/mutations.ts
```diff
@@ -1,15 +1,24 @@
...
     SettingFiles: LoadMutation,
     UpdatingState: LoadMutation,
     LoadFailed: LoadMutation,
-    LoadStateCleared: LoadMutation
+    LoadStateCleared: LoadMutation,
+    PreparingModelOutput: LoadMutation,
+    ModelOutputStatusUpdated: LoadMutation,
+    PollingStatusStarted: LoadMutation,
+    RehydrateResult: LoadMutation,
+    RehydrateResultError: LoadMutation
```
EmmaLRussell, line 21: These mutations seem a bit inconsistently named - can we stick with 'Rehydrate..` here?
```suggestion
    LoadStateCleared: LoadMutation,
    PreparingRehydrate: LoadMutation,
    RehydrateStatusUpdated: LoadMutation,
    RehydratePollingStarted: LoadMutation,
    RehydrateResult: LoadMutation,
    RehydrateResultError: LoadMutation
```

EmmaLRussell (changes requested):
Works great -  I can see the state logged to console as described. 👍 

Rather than just repurposing the existing menu item to load both zip and JSON, I though we were going to add a new menu item specifically for loading zip files, and renaming the old ones to 'Load/Save JSON'? As shown in @Nicolas-Dolan 's latest mockups:
image
Did I miss a decision there?

src/app/static/src/app/store/load/actions.ts
```diff
@@ -8,44 +8,62 @@ import {localStorageManager} from "../../localStorageManager";
...
-        });
-        reader.readAsText(file);
+
+    async load({dispatch}, payload) {
+        const {file, projectName, source} = payload;
+
+        if (FileSource.ModelOutput === source) {
+            const formData = new FormData()
+            formData.append("file", file)
+            await dispatch("prepareModelOutput", {file: formData, projectName});
```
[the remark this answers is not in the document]
LekanAnni: Not sure what you mean, I don't see a problem using dispatch here.

src/app/static/package.json
```diff
@@ -86,13 +86,18 @@
...
     "chartjs-plugin-error-bars": "^1.2.2",
     "jsonata": "^1.8.5",
     "plotly.js-basic-dist": "^2.4.0",
     "vue-chartjs": "^3.5.0",
     "vue-router": "^3.3.4"
+  },
+  "browser": {
+    "fs": false,
+    "path": false,
+    "os": false
```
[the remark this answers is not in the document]
LekanAnni: rogue

LekanAnni:
> Works great - I can see the state logged to console as described. 👍
> 
> Rather than just repurposing the existing menu item to load both zip and JSON, I though we were going to add a new menu item specifically for loading zip files, and renaming the old ones to 'Load/Save JSON'? As shown in @Nicolas-Dolan 's latest mockups: image Did I miss a decision there?

Yes, this was just to simplify testing for reviewers. there's a different ticket for handling the interface here https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3117

src/app/static/src/app/store/load/actions.ts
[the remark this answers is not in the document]
EmmaLRussell: There's not a particular problem, but what's the point? Actions are dispatched from outside Vuex to get it to do something. Sometimes we also chain actions together to trigger common action code when we're already in an action. But here we're dispatching an action that isn't used anywhere else, so why shouldn't that just be included synchronously in this action? If you just want to split up the code then can call a non-action method.

src/app/static/src/app/store/load/actions.ts
```diff
@@ -41,14 +43,14 @@ export interface modelOutputPayload {
...
+    async load(context, payload) {
+        const {dispatch} = context
         const {file, projectName, source} = payload;
 
         if (FileSource.ModelOutput === source) {
             const formData = new FormData()
             formData.append("file", file)
-            await dispatch("prepareModelOutput", {file: formData, projectName});
-
+            await dispatch("preparingRehydrate", {file: formData, projectName});
```
EmmaLRussell, line 53: This action dispatch still seems unnecessary to me. Not a showstopper but I don't think we should get into this habit, of spawning actions from actions instead of just calling methods, .

EmmaLRussell (approved):
Looks good, thanks. Great to get that weird getter bug fixed. Still unconvinced by the need to chain actions from actions so much..

src/app/static/src/app/store/load/actions.ts
LekanAnni: I tried moving but introduced a different issue but don't worry II will sort this out in the UI ticket.

Nicolas-Dolan (approved):

PR #712, opened 2022-06-13
mrc-2993 Use commas as thousands separators for plain numbers in the input time series table

Nicolas-Dolan:
## Description

Formatting options "0,0" were not being passed to the input time series table, which needs to be in D3 format. These changes also cover mrc-2834 "Add commas to numbers in plot tooltips to improve readability" and mrc-3131 "Add ability to format numbers in input time series with ,"

I identified a few issues/bugs that were outside the scope of this ticket and have written up new tickets:
1) I ran into a state wiping bug when going between data exploration mode and an open project. I haven't since been able to replicate the bug and suspect it may be due to the projects being old, created in older versions of hint/hintr. I haven't written up a ticket for this as I cannot replicate it but it's something to look out for.
2) AFAIK we do not format numbers according to the selected language anywhere in the app. The write up for this ticket mentions taking into account these language differences, but I felt we should discuss our approach to this first and then apply any changes throughout the app rather than in just this specific instance. mrc-3287
3) I noticed that adding commas to the table values disrupts the table's ability to order the values accurately when ordering by highest/lowest. This isn't an issue specific to this ticket and occurs in all existing places where commas are used in the values. The effect of commas on the ordering should also inform how we approach the second issue mentioned regarding translating the numbers. mrc-3288

## Type of version change
Minor

## Checklist


EmmaLRussell (commented):
Works for me! I'm not really how this could have affected the barchart - was that already showing the separators, and it just needed fixing in the ITS?

src/app/static/src/app/components/genericChart/utils.ts
```diff
@@ -29,8 +29,13 @@ export function genericChartColumnsToFilters(columns: GenericChartColumn[], filt
...
-    // We currently support only numeric and percentage formats, and will return empty string for any other
-    // formats received, for default formatting in Plotly.
+    // We currently support only numeric, large number, and percentage formats, and will return 
+    // empty string for any other formats received, for default formatting in Plotly.
+
+    if (numeralJsFormat === "0,0") {
+        return ","
+    }
+
     const regex = /^0(\.0+)?(%)?$/; //This will always return three matches
```
LekanAnni, line 39:
```suggestion
    const regex = /^0(\.0+)?(%)?$|^0(,0+)/; //This will always return four matches
```

LekanAnni: then add below snippet after the nullable check

```
    if (match[3] !== undefined) {
        return ",";
    }
```

src/app/static/src/app/components/genericChart/utils.ts
```diff
@@ -29,8 +29,13 @@ export function genericChartColumnsToFilters(columns: GenericChartColumn[], filt
 
 export function numeralJsToD3format(numeralJsFormat: string) {
     // Convert hintr metadata format (which are numeralJs style) to d3 format to be used by Plotly
-    // We currently support only numeric and percentage formats, and will return empty string for any other
-    // formats received, for default formatting in Plotly.
+    // We currently support only numeric, large number, and percentage formats, and will return 
+    // empty string for any other formats received, for default formatting in Plotly.
+
+    if (numeralJsFormat === "0,0") {
```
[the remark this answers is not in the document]
LekanAnni: This won't be needed, see below comment.

LekanAnni (changes requested):
Looking good, 

Please see my below suggestion, you should be using regex.

src/app/static/src/app/components/genericChart/utils.ts
EmmaLRussell: Could do this - but I think at that point it would be good to comment the regex a little to explain the two parts of it. Regexes are pretty horrible to read..

Nicolas-Dolan:
> Works for me! I'm not really how this could have affected the barchart - was that already showing the separators, and it just needed fixing in the ITS?

I think the barchart may have already had the thousand separators on master. The ticket was 6 months old

LekanAnni (approved):
LGTM 👍 

Thanks

src/app/static/src/app/components/genericChart/utils.ts
```diff
@@ -32,17 +32,17 @@ export function numeralJsToD3format(numeralJsFormat: string) {
     // We currently support only numeric, large number, and percentage formats, and will return 
     // empty string for any other formats received, for default formatting in Plotly.
 
-    if (numeralJsFormat === "0,0") {
-        return ","
-    }
-
-    const regex = /^0(\.0+)?(%)?$/; //This will always return three matches
+    const regex = /^0(\.0+)?(%)?$|^0(,0+)/; //This will always return four matches
```
EmmaLRussell, line 35: I think there should probably be a $ at the end of the second part of the regex, if only to make it consistent, and as I say it would be worth putting in an explanatory comment (as regexes are horrible to read).
And I'm not sure we want the + after ,0 do we? That implies we could have 0,00 which I don't think is valid numeralJS.  
```suggestion
     // The first part of this regex (before the |) captures formats with decimal precision and optional percentage
     // The second part captures format with thousands separator for large integers
    const regex = /^0(\.0+)?(%)?$|^0(,0)$/; //This will always return four matches
```

Nicolas-Dolan:
This ticket also covers [mrc-3131](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3131) "Add ability to format numbers in input time series with ,"

PR #714, opened 2022-06-17
mrc-1264 Add language toggle to login page

Nicolas-Dolan:
## Description

Adds the logged out header to the login page so language can be changed. Only affects the login page. All other functionality on the login page should work as before.

## Type of version change
Minor

## Checklist


EmmaLRussell (commented):
I can completely see why you've taken this approach! But there's a couple of things I don't like about it:
1. difficult to test "floating" js which shoul really be in a component
2. keeping language in two separate places in localstorage and having to bodge between them, 

We actually already have precedent for 1. in the way that the forgotPassword/resetPassword templates works. This has a component and its own Vue app (configured in webpack.config.js). It has language toggle built in, and can use locales to do translations. I think we should follow this same pattern for the login page.  

This does not solve 2. The language selected in the reset password page does not survive into other pages. So rather than trying to fix this here just for the login case, I think we should consider that outside the scope of the ticket and solve that for everything in one go. I think probably the answer is to keep language outside the project state, and keep it as a separate value (as you have done here, but ending the use of the language in the state - it's always seemed weird to me that loading a project can change your language preference). 

We should probably discuss the generic language issue, with Rob and Lekan too, at Monday's planning session. But for this ticket I'd definitly advocate for the resetPassword approach.

src/app/static/src/tests/components/login.test.ts
```diff
@@ -0,0 +1,148 @@
...
+
+        expect(wrapper.findAll("a").at(3).attributes("href")).toBe("explore")
+    });
+
+    it("email is updated from username prop", () => {
+        const store = createStore();
+
+        const wrapper = createSut(store, "Naomi", "/", "test");
+
+        expect(wrapper.vm.$data.email).toBe("test")
```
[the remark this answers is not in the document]
EmmaLRussell: You can usually get this with `(wrapper.find("whatever").element as HTMLInputElement).value` - more generally casting `.element` to a particular HTMLElement subtype is a really useful way of getting properties that aren't available on the wrapper interface. Here's an example: https://github.com/mrc-ide/hint/blob/1645dc1761898800313a566407195c3b96efc974/src/app/static/src/tests/directives/translate.test.ts#L52

src/app/static/src/tests/components/login.test.ts
```diff
@@ -0,0 +1,148 @@
...
+        const wrapper = createSut(store);
+
+        expect(wrapper.find("#login-form").attributes("class")).toBe("needs-validation")
+
+        const loginBtn = wrapper.find(".btn-red")
+        await loginBtn.trigger("click")
+        expect(wrapper.find("#login-form").attributes("class")).toBe("needs-validation was-validated")
+    });
+
+    // it("continue as guest is set in session storage", async () => {
```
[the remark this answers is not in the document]
EmmaLRussell: The localStorageManager tests use `spyOn`: https://github.com/mrc-ide/hint/blob/master/src/app/static/src/tests/localStorageManager.test.ts#L189

src/app/static/templates/login.ftl
```diff
@@ -6,64 +6,8 @@
...
-        <a class="btn btn-red" onclick="continueAsGuest()" type="submit" href="${continueTo}">Continue as guest</a>
-    </div>
-    <div id="partner-logos" class="logos mx-auto mt-5">
-      <a href="https://www.fjelltopp.org"><img src="public/images/fjelltopp_logo.png" class="small-logo"></a>
-      <a href="https://www.imperial.ac.uk"><img src="public/images/imperial_logo.png" class="small-logo"></a>
-      <a href="https://github.com/reside-ic"><img src="public/images/reside_logo.png" class="small-logo"></a>
-      <a href="https://www.avenirhealth.org"><img src="public/images/avenir_logo.png" class="small-logo"></a>
-      <a href="https://www.washington.edu"><img src="public/images/uw_logo.png" class="small-logo"></a>
-    </div>
+<div id="app" title="${title}" app-title="${appTitle}" continue-to="${continueTo}" username="${username}" error="${error}"></div>
```
EmmaLRussell, line 11: Our usual pattern here, would be to put the "app" id on a div containing a <login> element with all the props on. That way you shoudn't need your render method in login,ts.
```suggestion
<div id="app" >
    <login title="${title}" app-title="${appTitle}" continue-to="${continueTo}" username="${username}" error="${error}"></login>
</div>
```

src/app/static/src/app/login.ts
```diff
@@ -0,0 +1,39 @@
...
+    render: function (h) {
+        return h(Login, {
+            props: {
+                "title": this.$el.getAttribute("title"),
+                "appTitle": this.$el.getAttribute("app-title"),
+                "username": this.$el.getAttribute("username"),
+                "continueTo": this.$el.getAttribute("continue-to"),
+                "error": this.$el.getAttribute("error"),
+            }
+        })
```
EmmaLRussell, line 37: See comment on the template - shouldn't need this

src/app/static/src/app/login.ts
```diff
@@ -0,0 +1,39 @@
...
+    mutations
+};
+
+const store = new Vuex.Store<TranslatableState>(loginStoreOptions);
+registerTranslations(store);
+
+export const loginApp = new Vue({
+    el: "#app",
+    store,
+    props: ["title", "appTitle", "username", "continueTo", "error"],
```
EmmaLRussell, line 24: Shouldn't need these props if you make suggested template change

EmmaLRussell (changes requested):
This looks good to me, but I think you can save yourself a bit of pain around the template/render function.

EmmaLRussell (approved):
Thanks!

LekanAnni (changes requested):
Looks great, I think you have one more translation to do as seen in the screenshot below, including `Username or password is incorrect` text

Screenshot 2022-07-07 at 16 53 08

src/app/static/src/tests/components/login.test.ts
```diff
@@ -0,0 +1,148 @@
...
+        expect(links.at(7).find("img").attributes("src")).toBe("public/images/avenir_logo.png")
+        expect(links.at(8).attributes("href")).toBe("https://www.washington.edu")
+        expect(links.at(8).find("img").attributes("src")).toBe("public/images/uw_logo.png")
+    });
+
+    it("renders form with error", () => {
+        const store = createStore();
+
+        const wrapper = createSut(store, "Naomi", "/", "", "test error");
+        expect(wrapper.find("#error").text()).toBe("test error")
```
Nicolas-Dolan, line 101: For some reason, whenever I try to test the translated errors, only the translation key comes through (eg, "badUsernamePassword"). It works as expected when testing manually though. To test, enter a correct password and change the language.

LekanAnni (changes requested):
I think you want to translate from backend in ErrorMessageBundles

see as other data coming from backend are translated https://github.com/mrc-ide/hint/tree/master/src/app/src/main/resources

Nicolas-Dolan:
> I think you want to translate from backend in ErrorMessageBundles
> 
> see as other data coming from backend are translated https://github.com/mrc-ide/hint/tree/master/src/app/src/main/resources

Not sure this would work as the backend has no knowledge of the language set on the login page.

LekanAnni:
> > I think you want to translate from backend in ErrorMessageBundles
> > see as other data coming from backend are translated https://github.com/mrc-ide/hint/tree/master/src/app/src/main/resources
> 
> Not sure this would work as the backend has no knowledge of the language set on the login page. I may be wrong, I'll investigate.

You may have to reuse below code in LoginController 

```
    private fun getBundle(request: WebRequest): ResourceBundle
    {
        val language = request.getHeader("Accept-Language") ?: "en"
        return ResourceBundle.getBundle("ErrorMessageBundle", Locale(language))
    }
```

see file https://github.com/mrc-ide/hint/blob/dd6bac39074f8b781dd214bc065e40d01619354f/src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt#L77

Nicolas-Dolan:
> > > I think you want to translate from backend in ErrorMessageBundles
> > > see as other data coming from backend are translated https://github.com/mrc-ide/hint/tree/master/src/app/src/main/resources
> > 
> > 
> > Not sure this would work as the backend has no knowledge of the language set on the login page. I may be wrong, I'll investigate.
> 
> You may have to reuse below code in LoginController
> 
> ```
>     private fun getBundle(request: WebRequest): ResourceBundle
>     {
>         val language = request.getHeader("Accept-Language") ?: "en"
>         return ResourceBundle.getBundle("ErrorMessageBundle", Locale(language))
>     }
> ```
> 
> see file
> 
> https://github.com/mrc-ide/hint/blob/dd6bac39074f8b781dd214bc065e40d01619354f/src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt#L77

I played around with it and might be possible to get that to work, but I actually think it's better to translate the errors on the frontend like we do with everything else. If we send the errors pretranslated from the backend, we'd have to request them again if the user changes their language, which I'm not sure is possible. The issue I was having with testing earlier was due to a typo, which I've now fixed.

LekanAnni:
> > > > I think you want to translate from backend in ErrorMessageBundles
> > > > see as other data coming from backend are translated https://github.com/mrc-ide/hint/tree/master/src/app/src/main/resources
> > > 
> > > 
> > > Not sure this would work as the backend has no knowledge of the language set on the login page. I may be wrong, I'll investigate.
> > 
> > 
> > You may have to reuse below code in LoginController
> > ```
> >     private fun getBundle(request: WebRequest): ResourceBundle
> >     {
> >         val language = request.getHeader("Accept-Language") ?: "en"
> >         return ResourceBundle.getBundle("ErrorMessageBundle", Locale(language))
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > see file
> > https://github.com/mrc-ide/hint/blob/dd6bac39074f8b781dd214bc065e40d01619354f/src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt#L77
> 
> I played around with it and might be possible to get that to work, but I actually think it's better to translate the errors on the frontend like we do with everything else. If we send the errors pretranslated from the backend, we'd have to request them again if the user changes their language, which I'm not sure is possible. The issue I was having with testing earlier was due to a typo, which I've now fixed.

I am not keen with translating backend behaviour in frontend. I strongly believe this should be translated from backend just like we translate other exceptions from backend. 

Your concern about language changes does not really affect the behaviour as login requests always go to the backend, which will then retrieve current language from headers.

Nicolas-Dolan:
> > > > > I think you want to translate from backend in ErrorMessageBundles
> > > > > see as other data coming from backend are translated https://github.com/mrc-ide/hint/tree/master/src/app/src/main/resources
> > > > 
> > > > 
> > > > Not sure this would work as the backend has no knowledge of the language set on the login page. I may be wrong, I'll investigate.
> > > 
> > > 
> > > You may have to reuse below code in LoginController
> > > ```
> > >     private fun getBundle(request: WebRequest): ResourceBundle
> > >     {
> > >         val language = request.getHeader("Accept-Language") ?: "en"
> > >         return ResourceBundle.getBundle("ErrorMessageBundle", Locale(language))
> > >     }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > see file
> > > https://github.com/mrc-ide/hint/blob/dd6bac39074f8b781dd214bc065e40d01619354f/src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt#L77
> > 
> > 
> > I played around with it and might be possible to get that to work, but I actually think it's better to translate the errors on the frontend like we do with everything else. If we send the errors pretranslated from the backend, we'd have to request them again if the user changes their language, which I'm not sure is possible. The issue I was having with testing earlier was due to a typo, which I've now fixed.
> 
> I am not keen with translating backend behaviour in frontend. I strongly believe this should be translated from backend just like we translate other exceptions from backend.
> 
> Your concern about language changes does not really affect the behaviour as login requests always go to the backend, which will then retrieve current language from headers.

The scenario I was thinking about is: there is an error message on the page in English. The user changes the language to French. Everything translates except that error message.

LekanAnni:
> > > > > > I think you want to translate from backend in ErrorMessageBundles
> > > > > > see as other data coming from backend are translated https://github.com/mrc-ide/hint/tree/master/src/app/src/main/resources
> > > > > 
> > > > > 
> > > > > Not sure this would work as the backend has no knowledge of the language set on the login page. I may be wrong, I'll investigate.
> > > > 
> > > > 
> > > > You may have to reuse below code in LoginController
> > > > ```
> > > >     private fun getBundle(request: WebRequest): ResourceBundle
> > > >     {
> > > >         val language = request.getHeader("Accept-Language") ?: "en"
> > > >         return ResourceBundle.getBundle("ErrorMessageBundle", Locale(language))
> > > >     }
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > see file
> > > > https://github.com/mrc-ide/hint/blob/dd6bac39074f8b781dd214bc065e40d01619354f/src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt#L77
> > > 
> > > 
> > > I played around with it and might be possible to get that to work, but I actually think it's better to translate the errors on the frontend like we do with everything else. If we send the errors pretranslated from the backend, we'd have to request them again if the user changes their language, which I'm not sure is possible. The issue I was having with testing earlier was due to a typo, which I've now fixed.
> > 
> > 
> > I am not keen with translating backend behaviour in frontend. I strongly believe this should be translated from backend just like we translate other exceptions from backend.
> > Your concern about language changes does not really affect the behaviour as login requests always go to the backend, which will then retrieve current language from headers.
> 
> The scenario I was thinking about is: there is an error message on the page in English. The user changes the language to French. Everything translates except that error message.

We can refresh login page when language changes. I think we should keep backend translation in backend and vice versa. @EmmaLRussell  what's your opinion ?

EmmaLRussell:
> We can refresh login page when language changes. I think we should keep backend translation in backend and vice versa. @EmmaLRussell what's your opinion ?

Yes, agreed, in general - if the error is coming from the backend originally it ought to be getting translated in the back end too, as we do for other errors in the errors bundle. The front end should send a language header,to get the translated message, and if the user updates language, while there is an error present, the error code should be available in the qs so refreshing the page to get the translated message for the new language should be possible (may required some backend changes to check for the qs).

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/LoginController.kt
```diff
@@ -7,6 +7,21 @@ import org.springframework.ui.set
...
+                .toByteArray(Charsets.ISO_8859_1)
+                .toString(Charsets.UTF_8)
+    }
+
+fun getErrorMessageTranslation(key: String, request: HttpServletRequest): String
+{
+    val language = request.getHeader("Accept-Language") ?: "en"
+    val resourceBundle = ResourceBundle.getBundle("ErrorMessageBundle", Locale(language))
+    return resourceBundle.getUTF8String(key)
+}
```
Nicolas-Dolan, line 24: A bit of could repetition here. Could be put in a util perhaps but I didn't feel it was worth it.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/Controllers.kt
```diff
@@ -0,0 +1,14 @@
+package org.imperial.mrc.hint.controllers
+
+import java.util.*
+import javax.servlet.http.HttpServletRequest
+
+open class Controllers(val request: HttpServletRequest)
```
LekanAnni, line 6:
```suggestion
abstract class ControllerExtension(val request: HttpServletRequest)
```

LekanAnni (changes requested):
Looking good, 

Login page does not display error message in other languages. When I select French and use wrong login details, it returns error message in English. The page literarily refreshes every time and re-renders in English.

EmmaLRussell:
Given that there seem to be tricky outstanding issues with this ticket, and that the primary login method is about to be completely overhauled so that this original login page won't be used any more *and* that Nick is leaving anyway, it feels like maybe we should put this PR on hold for now as other things are higher priority. It might be that we need to apply some of the lessons learned here to the login page which we're going to have to add for Auth0 login (supporting guest user etc) - we should clarify localisation requirements for that too! Thoughts @r-ash ?

r-ash:
> 

Yep agreed, let's put it on hold until switch to Auth0 has is completed

r-ash:
Closing this for now to tidy up the open PRs as it is quite out of date. It's linked up to the ticket so we can find it again when this comes back around as highest priority.

PR #716, opened 2022-06-24
mrc-3117 Implement upload output zip UI

LekanAnni:
## Description

This PR adds upload outputZip components to project history and on file menu.

*  I have tried to break most component down to achieve reusability. 
* I renamed modal header for creating new project to use new text. 
* I removed irrelevant header. 
* I have also refactored load/actions to remove use of dispatch within the same action
* Added validation for new project upload
* Added progress bar, I have made it striped animated pending the time we get progress status from hintr.

Lastly, I apologise for making a large PR, the ticket has a point of 3 so I didn't think it would ever get this large.

## Type of version change
_Delete as appropriate_

## Checklist


EmmaLRussell (changes requested):
Wow, this was loads of work! Yeah, we really messed up the estimate there, we should have considered each bit of the UI individually and probably made them separate tickets. 

Looks great in general. 👍 

I thought we were going to do without the new button on the Projects page, but looks like it is still in the mockup. Didn't Jeff say he was OK just to have the menu item though? 
If we are going to keep that button, the text is wrong, should be "Upload project from zip" not "Upload*ing*..." .. oh, also "from zip" is inconsistent with the file menu item. 
Also, the placement looks weird to me - I think it's far too close to the text box (for typing a new project name), so it isn't obvious which button is associated with the text box. It would be better if there was a gap, and probably an 'OR' between them to make clear you can choose to do one thing  or the other. Or some other way of keeping these separate. 
image
But really I think it would be better without it, since most users aren't going to be using this functionality and will probably just be confused by the option being so obvious on the main page.  @Nicolas-Dolan and @r-ash  what do you remember being agreed around the button on the mockups?

Cancelling a load shows 'Load Error' (with an empty message)which doesn't seem quite right - it should probably just close the dialog.

Nicolas-Dolan:
> But really I think it would be better without it, since most users aren't going to be using this functionality and will probably just be confused by the option being so obvious on the main page. @Nicolas-Dolan and @r-ash what do you remember being agreed around the button on the mockups?
As I recall, we agreed to remove the button from the history page. It looks like I removed it on slide 2 but missed deleting it on the last slide. Sorry about that!

Nicolas-Dolan (changes requested):
Phew that's a lot of code! A few suggested changes below. Taking the ProjectUploadButton out of the Projects component is probably going to be a fairly big refactor, so will look at this again when that is done.

src/app/static/src/app/components/header/FileMenu.vue
```diff
@@ -168,22 +138,35 @@
                     }
                 }
             },
+            loadModel() {
+                const input = this.$refs.loadModel as HTMLInputElement;
+                if (input.files && input.files.length > 0) {
+                    const file = input.files[0];
+                    if (this.isGuest) {
+                        const formData = new FormData()
```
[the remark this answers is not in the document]
LekanAnni: I have refactored this to use custom getForm. Using file directly from here causes Axios to fail in integration test. I believe it was expecting certain type of format. I tried fixing it but no avail so I ended up using getFormData.

src/app/static/src/app/components/load/UploadProgress.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+    <div id="upload-progress">
+        <modal :open="openModal">
+            <h4 class="pb-4">
+                <span id="upload-header" class="pr-2" v-translate="'uploadFromZip'"></span>
+                <span><loading-spinner size="xs"/></span>
```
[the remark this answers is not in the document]
LekanAnni: Screenshot 2022-07-14 at 10 13 31


Done

src/app/static/src/app/components/header/FileMenu.vue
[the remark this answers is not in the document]
EmmaLRussell: Using file directly from the action caused integration test failures? I didn't initially understand why that's a problem since we already had this pattern in the last branch, but doesn't look like we had any load integration tests! Well fair enough, but it seems like a shame to have to structure the code in this way to keep the axios adapter happy. Seems like it ought to mockable somehow. Don't think it's worth holding up this branch over, but could you maybe make a maintenance ticket to get to the bottom of this?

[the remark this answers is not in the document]
LekanAnni: https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-6646

EmmaLRussell (commented):
I think some bugs have crept in now - when I load a freshly created output zip, it appears to have empty state - loads on step 1 with no input files shown. 

Also, if I return to the Projects page and attempt to reload the same file, it intermittently fails to show the Project name dialog after file is selected. 

Are others seeing this behaviour?

Also - just a note on the UI, which is a little inconsistent. Menu item says 'Load Zip Outputs' whereas label next to loaded project says 'loaded form model outputs'. Should use one or the other. 
Screenshot from 2022-07-14 11-12-47

Nicolas-Dolan:
> I think some bugs have crept in now - when I load a freshly created output zip, it appears to have empty state - loads on step 1 with no input files shown.
> 
> Also, if I return to the Projects page and attempt to reload the same file, it intermittently fails to show the Project name dialog after file is selected.
> 
> Are others seeing this behaviour?
> 
> Also - just a note on the UI, which is a little inconsistent. Menu item says 'Load Zip Outputs' whereas label next to loaded project says 'loaded form model outputs'. Should use one or the other. Screenshot from 2022-07-14 11-12-47

I'm seeing the same issues.

Nicolas-Dolan (changes requested):
Seeing same issues as Emma

LekanAnni:
> I think some bugs have crept in now - when I load a freshly created output zip, it appears to have empty state - loads on step 1 with no input files shown.

Input files have not been populated, its in this ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3119

EmmaLRussell:
> Input files have not been populated, its in this ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3119

Oh sorry, of course! :facepalm:

This seems like a real issue though: 
> Also, if I return to the Projects page and attempt to reload the same file, it intermittently fails to show the Project name dialog after file is selected.

Have you seen that before?

EmmaLRussell (commented):
Apologies for yesterday's review brain fog. This looks good to me. 

However, I am still seeing the issue where, after creating an uploaded project, if I go back and try to create another from the same file, the project name dialogs is not shown and project is not created. No errors are logged and the network tab does not show any requests being made. If I reload the application, I can then load from zip again. 

Cancel button behaviour looking good now 👍 

Suggestion on making UI consistent still stands - I'd favour changing the menu item text to 'Load Model Outputs' since that's what's in the [mockups](https://docs.google.com/presentation/d/1bzdNRLaEWVZgdGhpJmEeOboBIDRZWz2H5qIxF3fKzmc/edit#slide=id.g131903c6d13_0_0), and would then be consistent with the 'Loaded from model outputs' notes in the project list.

src/app/static/src/app/components/header/FileMenu.vue
```diff
@@ -162,6 +166,16 @@
                     }
                 }
             },
+            clearLoadZipInput() {
+                // clearing value because browser does not
+                // allow selection of the same file twice
+                const input = this.$refs.loadZip as HTMLInputElement
+                input.value = ""
+            },
```
LekanAnni, line 174: I have added this fix to the bug discovered by @EmmaLRussell which originated from the existing code. Browsers have implemented guard against uploading the same input file more than once. This could have been resolved easily with `onclick() => setValue to null` but we do not directly interact with this input element as we trigger the input via fileMenu link.

src/app/static/src/app/store/projects/actions.ts
```diff
@@ -92,7 +92,10 @@ export const actions: ActionTree<ProjectsState, RootState> & ProjectsActions = {
         await api<RootMutation, ProjectsMutations>(context)
             .withSuccess(RootMutation.SetProject, true)
             .withError(ProjectsMutations.ProjectError)
-            .postAndReturn<string>("/project/", qs.stringify(payload));
+            .postAndReturn<string>("/project/", qs.stringify(payload))
+            .then(() => {
+                dispatch("getProjects")
+            })
```
LekanAnni, line 98: I also discovered that validation wasn't working when a project is newly created and users try to upload again from zip. Dispatching getProjects updates the previous projects hence validation works.

src/app/static/src/app/components/load/UploadNewProject.vue
```diff
@@ -5,7 +5,6 @@
             <input id="project-name-input"
                    type="text"
                    class="form-control"
-                   @keyup.enter="submitLoad"
```
LekanAnni: This was removed because I have not managed to synchronised it with valid entry. A user can just press enter and continue, even if project name is invalid. I have created a ticket to handle this here https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-6651

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -428,7 +432,7 @@ const en: Translations = {
     enterValidEmail: "Please enter a valid email address",
     enterKey: "Enter key",
     enterPassword: "Enter a new password",
-    enterProjectName: "Please enter a name for the new project",
+    enterProjectName: "What would you like to name the project?",
```
r-ash, line 435: I think I prefer the previous text. I think it it makes clearer that it is an action the user has to take.

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -579,6 +585,8 @@ const en: Translations = {
     uniqueProjectName: "Please choose a unique name",
     updatePassword: "Update password",
     upload: "Upload",
+    uploadFromZip: "Uploading project from zip",
```
r-ash, line 588: Can we keep the name we use to refer to the file consistent, looks like using `model outputs` elsewhere. So `Uploading project from model outputs`?

r-ash (commented):
Just saw this in my github notifications, sorry for drive by review but take or leave as you wish

src/app/static/src/app/store/translations/locales.ts
LekanAnni: see above.

src/app/static/src/app/store/translations/locales.ts
LekanAnni: I have added text according to the mockup. We can add another ticket to tackle this suggestion.

LekanAnni: https://github.com/mrc-ide/hint/pull/718 
done here

EmmaLRussell (approved):
Thanks, looks good to me. 👍

PR #719, opened 2022-07-15
Mrc 3119 Populate project input file data

LekanAnni:
## Description

This PR populates upload input step with data uploaded from output zip. I have also adde getPlottingMetadata on mount for upload input, data fetched from the endpoint will not just validate the step but also available for input review charts. 

## Type of version change
_Delete as appropriate_

Minor 

## Checklist


src/app/static/src/app/utils.ts
```diff
@@ -348,3 +350,81 @@ export const getFormData = (file: File) => {
...
+    const modelRun = {
+        ...rootState.modelRun,
+        modelRunId: data.state.model_fit.id,
+        ready: true
+    }
+
+    const modelOptions = {
+        ...rootState.modelOptions,
+        options: data.state.model_fit.options,
+    }
```
EmmaLRussell, line 383: I think these are outside the scope of this ticket - this was just about loading input files?

src/app/static/src/app/utils.ts
```diff
@@ -348,3 +350,81 @@ export const getFormData = (file: File) => {
...
+        }
+    })
+    return dataset
+}
+
+export const constructRehydrateProjectState = (rootState: RootState, data: ProjectRehydrateResultResponse) => {
+    const files = transformPathToHash({...data.state.datasets});
+
+    const modelCalibrate = {
+        ...rootState.modelCalibrate,
```
EmmaLRussell, line 368: I don't think we actually want to retain rootState modules' data here, do we? This root state originated with current project's root state, Using the initial state for each module might be safer. Maybe this might be part of why I'm seeing quota exceeded error? (Not sure!)

EmmaLRussell (changes requested):
This isn't working for me in Chrome - get a local storage exceeded quota error when loading file for Malawi data. Have you not been seeing this issue?
Screenshot from 2022-07-20 09-33-41

I wondered if this issue might be related to loading while another project was current (see comment below), so I tried again, after logging out and deleting localstorage project before logging back in. This time, I didn't get quota exceeded error when loading from model outputs, but review inputs step was empty:
 
Screenshot from 2022-07-20 10-27-29

EmmaLRussell:
> How are you getting this error ?
> 
> I can't reproduce it.

To get quota error: run full project (Antarctica) to final step. save out zip file. Go to Project History page and Load that zip file from model outputs. 

To get the empty Review Inputs error: log out, delete project state in local storage. Re-attempt load of zip file.

I've updated Chrome, and still seeing the issue.

src/app/static/src/app/store/load/actions.ts
```diff
@@ -161,7 +170,9 @@ const getRehydrateStatus = async (context: ActionContext<LoadState, RootState>)
     }
 }
 
-async function getFilesAndLoad(context: ActionContext<LoadState, RootState>, files: any, savedState: any) {
+async function getFilesAndLoad(context: ActionContext<LoadState, RootState>,
+                               files: any,
+                               savedState: any) {
```
EmmaLRussell, line 175: Could this be `Partial<RootState>`?

EmmaLRussell (commented):
These changes have fixed the issue I was having with storage quota exceeded (yay!).
However, I'm still having the problem where, if I clear my local storage for this application and login and attempt to load the file, I see empty data on the Review Inputs step (Map). No error is logged.  Clicking on the Time series tab does load the time series data ok. This doesn't happen every time I load in this way, but very frequently.  

The reason this is happening is because the construct rehydrate code is saving the project state to local storage, but then that is being overwritten by  the current project state during the following unload process (because of the persistState plugin that happens on every mutation). The input files get loaded ok because they've been saved to the session in the backend, but this issue would affect all other state properties - including if we were loading from having a different current project rather than no current project.  

I think the simplest solution is probably to set the partial rehydrate state as the current project state (overwriting the existing current state - similar to what happens in RootMutation.Reset when it does the Object.assign). However this could cause some weird behaviour - probably need to try it and see. Will probably have to use a full empty state with the rehydrate values populated rather than just the rehydrate values. One way of doing that might be to use the `initial..` helper methods in each module.

Nicolas-Dolan (changes requested):
This isn't working for me. I upload a file, get the name project prompt and then nothing happens.

LekanAnni:
> This isn't working for me. I upload a file, get the name project prompt and then nothing happens.

You need to download a new outputZip for current session to work with rehydrate states once you have reload docker containers via `clear-docker.sh`

src/app/static/src/app/store/load/actions.ts
```diff
@@ -144,14 +144,17 @@ const getRehydrateResult = async (context: ActionContext<LoadState, RootState>)
...
-                }, {root: true}));
+                }, {root: true});
 
-            savedState.projects.currentProject = rootState.projects.currentProject
-            savedState.projects.currentVersion = rootState.projects.currentVersion
+            const project = {
+                currentProject: rootState.projects.currentProject,
+                currentVersion: rootState.projects.currentVersion
+            }
+            Object.assign(savedState.projects, project)
```
EmmaLRussell, line 157: Isn't this the wrong way round? It's `Object.assign(target, source)` so I was expecting you to do something like. `Object.assign(rootState, {...emptyState(), ...savedState})` having first constructed an empty root state. (And the problem isn't just about the  projects state, it's all of it.) Remember, the probem is that the current root state gets constantly saved to local storage, so that's overwriting your savedState.

src/app/static/src/app/store/load/actions.ts
```diff
@@ -144,14 +144,17 @@ const getRehydrateResult = async (context: ActionContext<LoadState, RootState>)
...
         }
```
EmmaLRussell, line 158: I tried this and it fixed the problem for me. `emptyState` is imported from `root.ts`
```suggestion
            savedState.projects!.currentProject = rootState.projects.currentProject
            savedState.projects!.currentVersion = rootState.projects.currentVersion
            const newRootState = {
                ...emptyState(),
                ...savedState
            }
            Object.assign(rootState, newRootState);
        }
```

EmmaLRussell (commented):
This didn't quite fix the probem for me, but see below for a change which I think does work.

src/app/static/src/app/store/load/actions.ts
```diff
@@ -172,16 +175,18 @@ const getRehydrateStatus = async (context: ActionContext<LoadState, RootState>)
 
 async function getFilesAndLoad(context: ActionContext<LoadState, RootState>,
                                files: any,
-                               savedState: any) {
-    savedState.stepper.steps = initialStepperState().steps;
+                               savedState: Partial<RootState>) {
+    Object.assign(savedState.stepper, {steps: initialStepperState().steps})
```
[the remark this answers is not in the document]
LekanAnni: I am avoiding use of ! NotNull assertion, it wasn't necessary prior to making arg type Partial<RootState>.

src/app/static/src/app/store/load/actions.ts
LekanAnni: Unfortunately I can't test the behaviour. I has added fixes but I am unable to reproduce the state bug, hoping all is fixed as suggested.

src/app/static/src/app/store/load/actions.ts
```diff
@@ -133,19 +133,34 @@ export const actions: ActionTree<LoadState, RootState> & LoadActions = {
...
+        const {files, savedState} = constructRehydrateProjectState(rootState, response.data)
+
+        if (!rootGetters.isGuest) {
+            await dispatch("projects/createProject",
+                {
+                    name: state.projectName,
+                    isUploaded: true
+                }, {root: true});
+
+            Object.assign(
```
EmmaLRussell, line 153: Again, I would prefer not to use Object.assign here, and I don't think using Object.assign for local objects to avoid null checks or other Typescript issues is an ideal approach. I think it's actually preferable to use `any` since at least makes it clear what we're doing. 

In this case, I guess the Typescript error is that this project state  is missing some properties? Could avoid that by creating our emptyState first then using emptyState.projects as the basis for the project state we assign to savedState?

src/app/static/src/app/store/load/state.ts
```diff
@@ -0,0 +1,27 @@
+import {Error, ProjectRehydrateResultResponse} from "../../generated";
+
```
EmmaLRussell, line 2:
```suggestion

// NB We define state and initialLoadState in separate file to the load module to 
// avoid circular dependency issues when calling root emptyState from 
// load/actions

```

src/app/static/src/app/store/load/actions.ts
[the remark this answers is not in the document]
EmmaLRussell: A not null assertion would be preferable to Object.assign I think. We're only doing that for rootState in order to replace the state in the store, but it's a pretty weird way of doing a local assignment.

EmmaLRussell (commented):
This works perfectly to fix the bug I was seeing, thank you. Great to sort that circular dependency issue. I'm actually surprised you didn't need to move `emptyState` as well but whatever works!

Might be good to add a note to `load/state` about why this module is differently arranged to the others. I'll make a suggestion for that below.

src/app/static/src/app/store/load/actions.ts
LekanAnni: I don't see anything wrong in using Object.assign when updating for multiple properties of an object. I would go for equals assignment for its speed for single property assignment over object.assign.

You have also suggested Object.assign usage above with the spread operator `const newRootState = {...emptyState(), ...savedState}`, this is just a synthetic sagar for Object.assign.  However, I have updated the PR to effect your preference.

EmmaLRussell: It's more about the way it reads than the performance, and using it as a technique for avoiding Typescript errors, which I think is obfuscating.  Thanks for making the change.

Nicolas-Dolan (approved):
Nice! Works great

EmmaLRussell (approved):
👍 Thanks!

PR #723, opened 2022-07-20
Mrc 3121 Populate project model options from project summary

LekanAnni:
## Description

This PR populate model options from rehydrated outputZip. I constructed the DynamicFormMeta from the DynamicFormData which is being returned from rehydrate outputZip. Ideally we should be sending DynamicFormMeta to hintr so we don't have to construct the data back to meta. However, I have added this fix tentatively, we might end up adding DynamicFormMeta at some point but this works. 

I also blocked options validation from happening on empty options data. This is particularly necessary for combating racing issues with options validation.

## Type of version change
_Delete as appropriate_

Minor

## Checklist


src/app/static/src/app/utils.ts
```diff
@@ -411,8 +421,35 @@ export const constructRehydrateProjectState = (rootState: RootState, data: Proje
...
+        }
     }
 
     return {files, savedState}
 }
+
+export const constructOptionsFormMetaFromData = (state: ModelOptionsState, meta: DynamicFormMeta) => {
+    const payloadContainsMeta = meta.controlSections.length > 0
+    const stateContainsOptions = Object.keys(state.options).length > 0
+    const emptyStateMeta = state.optionsFormMeta.controlSections.length === 0
```
EmmaLRussell, line 437: You'll be able to take this check out here if you only call this when this is the case, as suggested above

src/app/static/src/tests/modelOptions/mutations.test.ts
```diff
@@ -140,6 +140,18 @@ describe("Model run options mutations", () => {
         expect(state.valid).toBe(false);
     });
 
+    it("sets valid to true if optionsFormMeta is empty and options exist when payload has data", () => {
```
EmmaLRussell, line 143: Should really check here that it populates the metadata in the form correctly too (puts the right values in the controls). 
Could also check the invalid case too, where some required options are provided and are populated from the options, but not all (could happen in case where new option has been added since the zip file was saved).

EmmaLRussell (changes requested):
Works for me - non default options set in the original project are retained after rehydration. I think the tests could be a bit more detailed.

EmmaLRussell (approved):
I think this is good to go. I'm a little bit unsure about `modelOptions.valid` always ending up false when you rehydrate - it makes sense when you go back to the options step that you should need to revalidate since the options could have changed. But I wonder if it's going to confuse the stepper/state management that options are not valid, but subsequent steps are completed. I guess we'll find out! (Actually, it's probably fine as I think this will only ever apply when return to options step..)

Nicolas-Dolan (approved):
LGTM

PR #728, opened 2022-07-29
mrc-3080 Display comparison barchart in Review output step

Nicolas-Dolan:
## Description

Adds a comparison barchart to Review output step as per [mockups](https://docs.google.com/presentation/d/10vVaIIKbRFlRBRxP2PCw1U7ExMydklVC3HqH02Ji884/edit#slide=id.g136a0a5e73b_0_2). Table and area level filter to be added in separate tickets.

## Type of version change
Minor

## Checklist


src/app/static/src/app/utils.ts
```diff
@@ -185,6 +186,35 @@ export const flattenOptionsIdsByHierarchy = (filterOptions: NestedFilterOption[]
...
+            // Sort the selected filter values according to the order given the barchart filters
+            const updatedFilterOptions = [...selectedFilterOptions[xAxisId]].sort((a: FilterOption, b: FilterOption) => {
+                return flattenedXAxisFilterOptionIds.indexOf(a.id) - flattenedXAxisFilterOptionIds.indexOf(b.id);
+            });
+            payload.selectedFilterOptions[xAxisId] = updatedFilterOptions
+        }
+    }
+    // if unable to do the above, just updates the barchart as normal
+    updateSelections({payload})
+}
```
Nicolas-Dolan, line 216: Not sure how necessary it is to test these in the utils file if they are being tested in the component

LekanAnni:
@Nicolas-Dolan could you merge with master, I currently can't run branch on jdk11?

src/app/static/src/app/utils.ts
EmmaLRussell: Yeah, we're not very consistent with that sort of thing. I usually do some minimal integrated testing via the component tests, but if there are any edge cases (eg what if `data.xAxisId` is null) then it's usually easier to test those with a direct unit test - though you might want to see how the component would react in those circumstances too. 
If we were doing true unit tests we'd mock the util in the component tests - and that might be worthwhile for some utils which are tricky to set up (eg if they handle file i/o). But it's often useful to test the integration too.

src/app/static/src/app/utils.ts
```diff
@@ -190,6 +191,35 @@ export const flattenOptionsIdsByHierarchy = (filterOptions: NestedFilterOption[]
     return result;
 };
 
+export const flattenXAxisFilterOptionIds = (selections: BarchartSelections, filters: Filter[]) => {
```
EmmaLRussell, line 194: I guess it's a bit arbitrary, but this and `updateSelectionsAndXAxisOrder` could go in plot utils since they relate to barchart plots, just because there's already so much stuff in here. Or could even just be methods in the ModelOutput component (although that's also getting pretty bloated).

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -76,25 +76,7 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
...
-                            payload: {
-                                indicatorId: defaults.indicator_id,
-                                xAxisId: defaults.x_axis_id,
-                                disaggregateById: defaults.disaggregate_by_id,
-                                selectedFilterOptions: unfrozenDefaultOptions
-                            }
-                        },
-                        {root: true});
-                }
+                selectFilterDefaults(data, commit, "updateBarchartSelections")
```
EmmaLRussell, line 79:
```suggestion
                selectFilterDefaults(data, commit, PlottingSelectionsMutations.updateBarchartSelections)
```

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -128,11 +110,18 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
...
+        const response = await api<ModelCalibrateMutation, ModelCalibrateMutation>(context)
             .withError(ModelCalibrateMutation.SetError)
-            .withSuccess(ModelCalibrateMutation.SetComparisonPlotData)
+            .ignoreSuccess()
             .freezeResponse()
             .get<ModelResultResponse>(`model/comparison/plot/${calibrateId}`);
+
+        if (response) {
+            if (response.data){
+                selectFilterDefaults(response.data, commit, "updateComparisonPlotSelections")
```
EmmaLRussell, line 121:
```suggestion
                selectFilterDefaults(response.data, commit, PlottingSelectionsMutations.updateComparisonPlotSelections)
```

EmmaLRussell (commented):
Looks good to me. Nice organisation of re-used code.

It seems that the test data is very sparse - there seem to be lots of indicators with few values to be found. I guess that's expected (?), and hence the issue about whether we should evenutally offer some hints to the user about which value combinations contain data. 

'Data type' appears as an option for showing on x-axis, which I don't think is intended as that's fixed as the disaggregation column. Do we not have the ability in the barchart to remove that from the x-axis list when show disaggregate is false? Might need to add that (in another ticket). 

Filter selections are correctly saved between steps, but aren't retained when project is re-opened - though the same goes for the existing barchart so I guess that's ok.

LekanAnni (changes requested):
Looks good, nice naming convention.

Just a few comments.

* Plot has high disproportionate y-axis compared to it's data as you can see in the screenshot attached. 
Screenshot 2022-08-15 at 10 19 16

* Most indicator items do not have data and yet they are being rendered, which displays blank chart. For all filters, I think items should be displayed when they have data to render.
* Refreshing the page adds error to the console.
* `Area level` appears on the mockup but not in implementation.

EmmaLRussell:
> Looks good, nice naming convention.
> 
> Just a few comments.
> 
>     * Plot has high disproportionate y-axis compared to it's data as you can see in the screenshot attached.
> 
The red 35-39 age group has a very high error bar which is actually filling the 'gap' so I think this is valid!

>     * Most indicator items do not have data and yet they are being rendered, which displays blank chart. For all filters, I think items should be displayed when they have data to render.
> 

We've discussed this before and it's easier said than done e.g. some filter combinations involving a filter value will have data but others won't depending on how they're combined with other values, so how to show that? There's a ticket outstanding for this, I don't think it's in scope for this ticket. 

>     * `Area level` appears on the mockup but not in implementation.

As I understand it, area level was removed from scope for the first iteration of this feature. It will be implemented for both barcharts together.

src/app/static/src/app/store/modelCalibrate/actions.ts
Nicolas-Dolan: To do this I would have to change `type PlottingSelectionsMutation = Mutation<PlottingSelectionsState>` to an enum

src/app/static/src/app/store/modelCalibrate/actions.ts
Nicolas-Dolan: See above

Nicolas-Dolan:
> * Refreshing the page adds error to the console.
Fixed

src/app/static/src/app/store/modelCalibrate/actions.ts
EmmaLRussell: Ah, sorry I assumed it already was an enum, e.g. like ProjectMutations etc. Most of the other mutations types follow the enum pattern, so I think it would be good to make that change unless there's some good reason why we have it the other way in this case.

Nicolas-Dolan:
> 'Data type' appears as an option for showing on x-axis, which I don't think is intended as that's fixed as the disaggregation column. Do we not have the ability in the barchart to remove that from the x-axis list when show disaggregate is false? Might need to add that (in another ticket).

That doesn't seem to be an option as of yet in the vue-chart config. Just this:
image
I'm not sure if this is something we want to configure from vue-charts or something in hint?

EmmaLRussell:
> > 'Data type' appears as an option for showing on x-axis, which I don't think is intended as that's fixed as the disaggregation column. Do we not have the ability in the barchart to remove that from the x-axis list when show disaggregate is false? Might need to add that (in another ticket).
> 
> That doesn't seem to be an option as of yet in the vue-chart config. Just this: image I'm not sure if this is something we want to configure from vue-charts or something in hint?

That's a configuration option we'd need to add to the barchart in vue-charts and then use from hint. I think we probably do need to do that as it's probably going to be confusing to the user - but I guess there's the same issue in the calibration plot, which doesn't seem to have been raised, so maybe it's ok! Maybe you could demo it at the hint lunchtime catchup tomorrow, and see what the other think?

Implementing should be pretty easy in the barchart, would need a `hideFixedDisaggregateByFilter` in the xAxisConfig, and remove the relevant option when true.

EmmaLRussell:
> That's a configuration option we'd need to add to the barchart in vue-charts and then use from hint. I think we probably do need to do that as it's probably going to be confusing to the user - but I guess there's the same issue in the calibration plot, which doesn't seem to have been raised, so maybe it's ok!  

Actually, I've just realised that the calibrate plot is different, since both disaggregateBy and xAxis are fixed and hidden there.

EmmaLRussell (commented):
This looks good, but you'll need to negate the feature switch so that comparison plot isn't shown by default since we still don't have the real data afaik. 

Also - do you have a test that Comparison tab is/isn't rendered depending on switch value? (I think I'm seeing a slightly odd diff, apologies if you've already done this.) You can just modify the values in the `switches` object as required in the test.

LekanAnni (changes requested):
Looks like a different errors now persists in console. Error using Antarctica and Zambia dataset.

Screenshot 2022-08-18 at 11 21 56


Screenshot 2022-08-18 at 11 23 49

Nicolas-Dolan:
> Looks like a different errors now persists in console. Error using Antarctica and Zambia dataset.

I've fixed the first error, but I believe the second error will need to be fixed in vue-charts as the error is occurring within BarChartsWithFilters. @EmmaLRussell I'll fix that while I make the change to remove the disaggregate value from the xAxis options.

Nicolas-Dolan:
> Implementing should be pretty easy in the barchart, would need a `hideFixedDisaggregateByFilter` in the xAxisConfig, and remove the relevant option when true. Or could even do it by default, without needing extra configuration, since it's always going to be quite a weird option to have.

I've written up a ticket for this https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3547/Remove-fixed-disaggregate-value-from-xAxis-options-in-vue-charts

EmmaLRussell:
> I've written up a ticket for this https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3547/Remove-fixed-disaggregate-value-from-xAxis-options-in-vue-charts

Thanks, looks good

EmmaLRussell:
> I've fixed the first error, but I believe the second error will need to be fixed in vue-charts as the error is occurring within BarChartsWithFilters. @EmmaLRussell I'll fix that while I make the change to remove the disaggregate value from the xAxis options.

Possibly, but this might be fixable in the metadata, it seems to be complaining about missing values. Needs investigation into what the problem values actually are rather than just swallowing any error in the component. 

One useful trick to to capture the metadata/data that's being fed into the barchart then use that in the demo app in the barchart repo to debug what's going wrong.

EmmaLRussell:
I'm now sesing local storage quota exceeded messages in the console, and could not parse API response displayed (this is probably related, and it's misinterpreting the storage error as an API problem). Is anyone else getting this? I think it's unrelated to the branch, but probably triggered by reloading the page with the feature switch query string. @LekanAnni You mentioned seeing something similar quite recently?

Screenshot from 2022-08-24 11-05-50

I think the hintr branch is pretty much ready to merged, in which case we should be able to remove the feature switch and target master hintr but sounds like the local storage issue needs looking at anyway (is it happening on master, on reload?)

Nicolas-Dolan:
> I'm now sesing local storage quota exceeded messages in the console, and could not parse API response displayed (this is probably related, and it's misinterpreting the storage error as an API problem). Is anyone else getting this? I think it's unrelated to the branch, but probably triggered by reloading the page with the feature switch query string. @LekanAnni You mentioned seeing something similar quite recently?
> 
> Screenshot from 2022-08-24 11-05-50
> 
> I think the hintr branch is pretty much ready to merged, in which case we should be able to remove the feature switch and target master hintr but sounds like the local storage issue needs looking at anyway (is it happening on master, on reload? Or is it just this branch, and there's state being saved to storage which shouldn't be?)

I've not seen this message myself but I'm looking at the hintr branch now.

LekanAnni (changes requested):
Looks great so far,

I have a concern about the time it takes to render comparison plot. Compared to barChart on tab 2, it renders without any delay but comparison plot delays rendering of the plot. It would be great if you can optimise it ?

LekanAnni:
> > I'm now sesing local storage quota exceeded messages in the console, and could not parse API response displayed (this is probably related, and it's misinterpreting the storage error as an API problem). Is anyone else getting this? I think it's unrelated to the branch, but probably triggered by reloading the page with the feature switch query string. @LekanAnni You mentioned seeing something similar quite recently?
> > Screenshot from 2022-08-24 11-05-50
> > I think the hintr branch is pretty much ready to merged, in which case we should be able to remove the feature switch and target master hintr but sounds like the local storage issue needs looking at anyway (is it happening on master, on reload? Or is it just this branch, and there's state being saved to storage which shouldn't be?)
> 
> I've not seen this message myself but I'm looking at the hintr branch now.

Oh yes a while back, I have not seen it in awhile though.

EmmaLRussell (approved):
I think this one should be ready to merge in now, shouldn't it? 
All remaining work to do seems to be happening in separate branches now, so it would make sense to merge this with the feature switdh in place I think. That would mean we wouldn't need to be reviewing all the other branches relative to this one.

Nicolas-Dolan:
> rendering of the

@LekanAnni I've written up a separate ticket for this issue: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3588/Fix-delay-when-rendering-comparison-barchart

LekanAnni (approved):
LGTM 👍

PR #734, opened 2022-08-15
mrc-3520 get projects only when user is logged in

LekanAnni:
## Description

This PR fixes login issue. Issues caused by calling getProjects when user is not logged in.

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/tests/components/header/fileMenu.test.ts
+++ b/src/app/static/src/tests/components/header/fileMenu.test.ts
@@ -31,6 +31,10 @@ describe("File menu", () => {
     const testProjects = [{id: 2, name: "proj1", versions: []}];
     const mockGetProjects = jest.fn();
 
+    beforeEach(() => {
+        jest.clearAllMocks()
+    })
+
     const storeModules = {
         baseline: {
             namespaced: true,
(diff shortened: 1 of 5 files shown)


src/app/static/src/tests/components/header/fileMenu.test.ts
```diff
@@ -331,11 +335,22 @@ describe("File menu", () => {
         expect(projectModal.props().openModal).toBe(false)
         projectModal.find(".btn").trigger("click");
         expect(mockLoadAction.mock.calls.length).toBe(1);
-        expect(mockGetProjects).toHaveBeenCalled()
         expect(projectModal.props().openModal).toBe(false)
     });
 
-    it("can open upload project modal as guest when file is uploaded", () => {
+    it("can get projects when user is logged in when file is uploaded", () => {
```
EmmaLRussell, line 341: Shouldn't this test, and the subsequent one, for guest user, be in `uploadNewProject.tests.ts`? that's where the code is that we're testing for! Even if we're rolling the tests for the two component together, won't you find that `getProjects` is called as soon as the parent component is mounted (without needing to upload file) - as we found the bug was happening as soon as the projects page was opened, without any user interaction

EmmaLRussell (commented):
Code looks good to me, works great. Just a query about the testing approach.

Nicolas-Dolan (approved):

src/app/static/src/tests/components/header/fileMenu.test.ts
LekanAnni: `uploadNewProject.tests.ts` are tested already, I know it's an over kill but since we can't test this behaviour directly via a browser test, I felt it would be better to add tests in all cases where this behaviour is expected.

EmmaLRussell: OK - I think maybe I was looking at a slightly earlier commit, before you added the uploadNewProject test, so I'm happy that's in. And I'm still not sure this one tells us much, particularly since we don't need to open the modal to trigger (or not) the getProjects. But no harm to have this too!

EmmaLRussell (approved):

PR #735, opened 2022-08-18
mrc-3486 Table for comparison barchart

Nicolas-Dolan:
## Description

Adds a table for the comparison plot on the review output step. To view visit http://localhost:8080/?comparisonPlot=1 Data type column header is coming through oddly (lower case and only translated for French), likely due to still pulling from the dummy endpoint currently.

## Type of version change
Minor

## Checklist

### DIFF
--- a/src/app/static/src/app/components/modelOutput/ModelOutput.vue
+++ b/src/app/static/src/app/components/modelOutput/ModelOutput.vue
@@ -99,6 +99,19 @@
                     :formatFunction="formatBarchartValue"
                     :showRangesInTooltips="true"
                     @update="updateComparisonPlotSelectionsAndXAxisOrder"></bar-chart-with-filters>
+                <div class="row mt-2">
+                    <div class="col-md-3"></div>
+                    <area-indicators-table class="col-md-9"
+                                        :table-data="comparisonPlotData"
+                                        :area-filter-id="areaFilterId"
+                                        :filters="comparisonPlotFilters"
+                                        :countryAreaFilterOption="countryAreaFilterOption"
+                                        :indicators="filteredComparisonPlotIndicators"
+                                        :selections="comparisonPlotSelections"
+
+                                        :selectedFilterOptions="comparisonPlotSelections.selectedFilterOptions"
+                    ></area-indicators-table>
+                </div>
             </div>
         </div>
     </div>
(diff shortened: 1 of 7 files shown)


EmmaLRussell (commented):
The table looks great, but you're right, there is an issue with the column headers. The `AreaIndicatorsTable` is expecting the labels of filters to be translation keys rather than displayable strings, but it looks like the backend is sending through those strings already translated. And actually, the table is "helpfully" lower-casing any label that isn't already lower-cased so it can treat it like a translation key, and is just displaying that if it can't translate it:
https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/components/plots/table/AreaIndicatorsTable.vue#L175

I'm not quite sure how we've ended up with this discrepancy between the responses for the two barcharts, but I'm guessing it's because the comparison plot response is modelled on the calibration plot response. It's interesting that the filters look fine in the barchart.. Modifying the response at this point might have knock-on effects, so the simplest solution might be to put a boolean prop on the AreaIndicatorsTable components, `translateFilterLabels` and don't attempt to do so for the comparison plot. (Although really it would be better to be consistent...)

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -100,6 +100,19 @@
...
+                                        :area-filter-id="areaFilterId"
+                                        :filters="comparisonPlotFilters"
+                                        :countryAreaFilterOption="countryAreaFilterOption"
+                                        :indicators="filteredComparisonPlotIndicators"
+                                        :selections="comparisonPlotSelections"
+                                        :translate-filter-labels="false"
+                                        :selectedFilterOptions="comparisonPlotSelections.selectedFilterOptions"
+                    ></area-indicators-table>
+                </div>
                 <error-alert v-if="!!comparisonPlotError" :error="comparisonPlotError"></error-alert>
```
Nicolas-Dolan, line 116: Alternatively, the error alert could go above the table?

EmmaLRussell: Presumably if there was an error there wouldn't be any data in the table, in which case it wouldn't really matter since the table wouldn't show anything. Or could it be that we have data hanging around from previous fetch which doesn't get cleared when error on next fetch is committed? Should either clear old data or put error somewhere definitely visible.

EmmaLRussell (approved):
Translation fix looks good

LekanAnni (approved):
LGTM 👍

PR #736, opened 2022-08-23
mrc-3549 Bug: HINT login is broken on dev and staging

LekanAnni:
## Description

This PR handles exception with keys that are unknown. Hint-deploy was failing due to exception thrown from userCLI, therefore unable to serialise user profile.

## Type of version change
_Delete as appropriate_

## Checklist


scripts/add-test-user.sh
```diff
@@ -4,7 +4,7 @@ NETWORK=hint_nw
 HERE=$(realpath "$(dirname $0)")
 TEST_CONFIG=$HERE/../src/app/static/scripts/test.properties
 
-image=mrcide/hint-user-cli:master
+image=mrcide/hint-user-cli:mrc-3549
```
LekanAnni, line 7: This will use changes made in this branch userCLI

src/app/build.gradle
```diff
@@ -44,10 +44,10 @@ dependencies {
...
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.2.1'
     implementation "org.jooq:jooq:3.14.16"
-    implementation 'org.pac4j:spring-webmvc-pac4j:6.0.2'
-    implementation 'org.pac4j:pac4j-http:5.4.3'
-    implementation 'org.pac4j:pac4j-jwt:5.4.3'
-    implementation 'org.pac4j:pac4j-sql:5.4.3'
+    implementation 'org.pac4j:spring-webmvc-pac4j:6.0.3'
+    implementation 'org.pac4j:pac4j-http:5.4.5'
+    implementation 'org.pac4j:pac4j-jwt:5.4.5'
+    implementation 'org.pac4j:pac4j-sql:5.4.5'
```
LekanAnni, line 50: Spring boot 2.7.3 support these versions.

src/build.gradle
```diff
@@ -16,7 +16,7 @@ buildscript {
 }
 
 plugins {
-    id 'org.springframework.boot' version '2.7.2' apply false
+    id 'org.springframework.boot' version '2.7.3' apply false
```
LekanAnni, line 19: Upgraded to the stable version

src/userCLI/src/main/kotlin/org/imperial/mrc/hint/userCLI/App.kt
```diff
@@ -36,6 +36,19 @@ fun main(args: Array<String>) {
...
+        val resources = ResourceBundle.getBundle("ErrorMessageBundle", Locale("en"))
+        try
+        {
+            val message = resources.getString(msg.takeLastWhile { it.isLetter() })
+            System.err.println(message)
+        } catch (e: MissingResourceException)
+        {
+            System.err.println("Could not load ErrorMessageBundle: $msg")
+        }
+    }
```
LekanAnni, line 50: Unknown error causes an issue so I am handling exception here. I am also taking the last word in the initial thrown exception.

src/userCLI/build.gradle
```diff
@@ -20,10 +20,9 @@ dependencies {
     implementation project(":app")
     implementation "com.offbytwo:docopt:0.6.0.20150202"
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
-    implementation "org.slf4j:slf4j-nop:1.7.26"
```
LekanAnni, line 23: Remove multiple binding as binding already exist in hint `build.gradle`.

src/userCLI/src/main/kotlin/org/imperial/mrc/hint/userCLI/App.kt
EmmaLRussell: Why was it trying to take the whole message as the resource key before , but now just the last word? What difference does that make? Are there some errors which do match based on only their last word (as oppsed to only word)?

I think it would be great to add an explanatory comment here, referring to the bug ticket and explaining the last-word usage.

EmmaLRussell (commented):
OK, so as I understand it, when an unknown error is thrown, it was bombing out because it was trying to cleanly exit process with an error state, but a further error was being thrown because it wasn't able to find a resources it was expecting to the in error bundle? So here it's handling that error and dealing with it gracefully. 

And it was this error-within-an-error that was causing the corrupted user profiles somehow? So now that doesn't happen with this error handling?

To be honest, I really don't understand enough about how the problem with the user profiles was being caused or why this should fix it, but it seems like a benign change so fine by me if it fixes it! I think it would be good to put in a comment about why errorOutputStream is required, maybe with a reference to the ticket number.

src/userCLI/src/main/kotlin/org/imperial/mrc/hint/userCLI/App.kt
```diff
@@ -36,6 +36,19 @@ fun main(args: Array<String>) {
 
     val dataSource = DbConfig().dataSource(ConfiguredAppProperties())
 
+    fun errorOutputStream(msg: String)
+    {
+        try
+        {
+            val resources = ResourceBundle.getBundle("ErrorMessageBundle", Locale("en"))
+            val message = resources.getString(msg.takeLastWhile { it.isLetter() })
```
Nicolas-Dolan, line 44: So is this the key line that is fixing the bug? Is the error just being masked here, and if so is it ok to do that?

LekanAnni:
> OK, so as I understand it, when an unknown error is thrown, it was bombing out because it was trying to cleanly exit process with an error state, but a further error was being thrown because it wasn't able to find a resources it was expecting to the in error bundle? So here it's handling that error and dealing with it gracefully.
> 
> And it was this error-within-an-error that was causing the corrupted user profiles somehow? So now that doesn't happen with this error handling?
> 
> To be honest, I really don't understand enough about how the problem with the user profiles was being caused or why this should fix it, but it seems like a benign change so fine by me if it fixes it! I think it would be good to put in a comment about why errorOutputStream is required, maybe with a reference to the ticket number.


When an exception is thrown in https://github.com/mrc-ide/hint/blob/1c82d6fadcd9762de86bd6c46f180963d87c9ec9/src/app/src/main/kotlin/org/imperial/mrc/hint/logic/UserLogic.kt#L39 UserLogic for a key `userExists`, userCLI expects the key, which it then uses to get resource bundle. However, UserException handler for UserLogic extends HintException where additional text is appended `HintException with key $key`. 

Apparently this was working with no issues but since pac4j upgrade, it causes profile issues. I hope that's clear enough.

src/userCLI/src/main/kotlin/org/imperial/mrc/hint/userCLI/App.kt
LekanAnni: Yes that's the issue, there many ways to fix the issues such as not inheriting HintException but this will cause other issues, see below comment.

src/userCLI/build.gradle
LekanAnni: I have restore this library, it is needed for getting logs from console in `hint-deploy` thus ignoring multiple slf4j binding warning.

src/userCLI/src/main/kotlin/org/imperial/mrc/hint/userCLI/App.kt
```diff
@@ -36,6 +36,19 @@ fun main(args: Array<String>) {
 
     val dataSource = DbConfig().dataSource(ConfiguredAppProperties())
 
+    fun errorOutputStream(msg: String)
+    {
+        try
+        {
+            val resources = ResourceBundle.getBundle("ErrorMessageBundle", Locale("en"))
```
EmmaLRussell, line 43: Thanks for the explanation @LekanAnni , that's really useful. Still not sure why pac4j burns down the profile, but at least we can work around it! I still think it would really good to include a comment that effect in the code, just so that this issue doesn't get lost to history. Something like this, just using what you've said?
```suggestion
   // When an exception is thrown in UserLogic with key userExists, userCLI 
   // expects the key, which it then uses to get resource bundle. However,      
   //  the UserException handler for UserLogic extends HintException where   
   // additional text is appended: "HintException with key $key" so the resource
   // would not be found in the bundle.
   // After pac4j upgrade, this issue caused corruption of the profile if not dealt
   // with gracefully, hence extracting final word here. See associated ticket 
   // mrc-3549  

            val resources = ResourceBundle.getBundle("ErrorMessageBundle", Locale("en"))
```

PR #741, opened 2022-09-08
mrc-3612 Remove feature switch for comparison plot

Nicolas-Dolan:
## Description

Does what it says on the tin. I haven't incremented the version number for this but let me know if you think I should.

## Type of version change
None

## Checklist


LekanAnni (approved):
Looks good, works well.

PR #743, opened 2022-09-09
Mrc-3240 Include system info in logs

LekanAnni:
## Description

PR uses application url in structured log to differentiate different profiles such as staging, production or dev. 

## Type of version change

Minor 

## Checklist


Nicolas-Dolan (approved):

EmmaLRussell (commented):
Code looks fine to me. `profile` seems like a slightly weird prop name now it's logging the whole base url (not 'dev' or 'prod' as was suggested in the ticket). Would just `url` be more accurate?

LekanAnni:
> Code looks fine to me. `profile` seems like a slightly weird prop name now it's logging the whole base url (not 'dev' or 'prod' as was suggested in the ticket). Would just `url` be more accurate?

it could be profileUrl, the whole idea is to differentiate deployments, which we can achieve using this approach. Using dev, prod, staging or whatever would mean we have to resolve all deployments manually.

EmmaLRussell:
> > Code looks fine to me. `profile` seems like a slightly weird prop name now it's logging the whole base url (not 'dev' or 'prod' as was suggested in the ticket). Would just `url` be more accurate?
> 
> it could be profileUrl, the whole idea is to differentiate deployments, which we can achieve using this approach. Using dev, prod, staging or whatever would mean we have to resolve all deployments manually.

`profileUrl` sounds good

NEWS.md
```diff
@@ -1,7 +1,16 @@
...
 * Include system info in logs
 
+# hint 2.14.0
+
+* Add Auth0 login method
+
+# hint 2.13.0
+
+* Update hint to work with options schema changes
+>>>>>>> master
```
EmmaLRussell, line 12: rogue merge artefact

EmmaLRussell (approved):
Thanks - just needs leftover merge artefact removing,..

PR #744, opened 2022-09-12
Mrc 2236 - Log elapsed time for ADR responses

LekanAnni:
PR adds time elapsed to ADR response, Adding log directly to HttpClient would pollute the output, so I decided to override required methods, call super and log time elapsed. I am not 100% sure this is the best solution, I am open to other ideas.

## Type of version change
_Delete as appropriate_

## Checklist


src/app/src/main/kotlin/org/imperial/mrc/hint/clients/ADRFuelClient.kt
```diff
@@ -37,9 +41,39 @@ interface ADRClient
...
+                    private val logger: GenericLogger = GenericLoggerImpl())
     : FuelClient(appProperties.adrUrl + "api/3/action"), ADRClient
 {
+    override fun get(url: String): ResponseEntity<String>
+    {
+        val start = Instant.now()
+        val response = super.get(url)
+        val end = Instant.now()
+        val timeElapsed = Duration.between(start, end).toMillis()
+        logger.info("ADR request time elapsed: $timeElapsed")
```
EmmaLRussell, line 54: It feels like the time should go into the structured JSON logging as a property valus, rather than just included in the log string like this. That would make it easier to do analytics queries against ADR durations e.g. show me the average ADR fetch duration.

EmmaLRussell (commented):
This looks like it will do the job, but could it be more DRY? e.g. could have a 'logDurationOf' method which takes function parameter along with the parameters to give it and records the start time, calls the function and logs the duration. 

Also, I think we need to log the duration as a separate value, not just rolled into a string. That's what the structured JSON logging should be good at, isn't it? Might also want to log the type of request to ADR, e.g. might find POSTs take much longer than GETs - though I guess that will be derivable from the verb/url of the HINT endpoint, which we should be logging too?

Nicolas-Dolan (changes requested):
Seems like you need to add some additional unit tests for `ADRFuelClient.kt`

LekanAnni:
> This looks like it will do the job, but could it be more DRY? e.g. could have a 'logDurationOf' method which takes function parameter along with the parameters to give it and records the start time, calls the function and logs the duration.
> 
> Also, I think we need to log the duration as a separate value, not just rolled into a string. That's what the structured JSON logging should be good at, isn't it? Might also want to log the type of request to ADR, e.g. might find POSTs take much longer than GETs - though I guess that will be derivable from the verb/url of the HINT endpoint, which we should be logging too?

Good idea 👍

LekanAnni:
> Also, I think we need to log the duration as a separate value, not just rolled into a string. That's what the structured JSON logging should be good at, isn't it? Might also want to log the type of request to ADR, e.g. might find POSTs take much longer than GETs - though I guess that will be derivable from the verb/url of the HINT endpoint, which we should be logging too?

I didn't change this part as I would further obfuscate the whole log, I think we should await @r-ash 's design and then update implementation.

EmmaLRussell:
> > Also, I think we need to log the duration as a separate value, not just rolled into a string. That's what the structured JSON logging should be good at, isn't it? Might also want to log the type of request to ADR, e.g. might find POSTs take much longer than GETs - though I guess that will be derivable from the verb/url of the HINT endpoint, which we should be logging too?
> 
> I didn't change this part as I would further obfuscate the whole log, I think we should await @r-ash 's design and then update implementation.

OK, I've added a note to that effect on the planning ticket: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3619/Plan-adding-additional-backend-logging

src/app/src/main/kotlin/org/imperial/mrc/hint/logging/LogElapsedTimeCallback.kt
```diff
@@ -0,0 +1,17 @@
...
+    logger: GenericLogger,
+): T
+{
+    val start = Instant.now()
+    val response = callback()
+    val end = Instant.now()
+    val timeElapsed = Duration.between(start, end).toMillis()
+    logger.info("ADR request time elapsed: $timeElapsed")
+    return response
+}
```
EmmaLRussell, line 17: Strange that codecov thinks this isn't tested, when it clearly is, in `LogElapsedTimeCallbackTests`!

src/app/src/main/kotlin/org/imperial/mrc/hint/logging/LogElapsedTimeCallback.kt
```diff
@@ -0,0 +1,17 @@
...
+inline fun <T> logDurationOf(
+    callback: () -> T,
+    logger: GenericLogger,
+): T
+{
+    val start = Instant.now()
+    val response = callback()
+    val end = Instant.now()
+    val timeElapsed = Duration.between(start, end).toMillis()
+    logger.info("ADR request time elapsed: $timeElapsed")
```
EmmaLRussell, line 15: Feels like either the log message prefix should be a parameter too, or the method name should be less generic, to indicate that this should only be used for ADR calls e.g. `logADRRequestDuration`

EmmaLRussell (commented):
Nice!

EmmaLRussell (approved):
Great, thanks!

PR #745, opened 2022-09-13
mrc-3629 Implement no data message in comparison chart

Nicolas-Dolan:
## Description

Upgrades vue-charts version and submits prop to it for the translated no data message. The easiest way to test is to go to the comparison tab on the review output step, set the x axis to sex, and then unselect the sex value from the filter. This should mean the chart is getting no data and prompt the no data message.

Note: you will need to run `npm install --prefix=src/app/static/` to get the latest chart.

## Type of version change
Minor

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 2.22.0
+
+* Bug: Fix download status errors
+
 # hint 2.21.0
 
 * Speed up display of project rehydrate dialog
--- a/src/app/static/src/app/hintVersion.ts
+++ b/src/app/static/src/app/hintVersion.ts
@@ -1,1 +1,1 @@
-export const currentHintVersion = "2.21.0";
+export const currentHintVersion = "2.22.0";
--- a/src/app/static/src/app/root.ts
+++ b/src/app/static/src/app/root.ts
@@ -8,7 +8,7 @@ import {
     SurveyAndProgramState,
 } from "./store/surveyAndProgram/surveyAndProgram";
 import {initialModelRunState, modelRun, ModelRunState} from "./store/modelRun/modelRun";
-import {initialStepperState, stepper, StepperState} from "./store/stepper/stepper";
+import {initialStepperState, stepper} from "./store/stepper/stepper";
 import {initialLoadState, LoadState} from "./store/load/state";
 import {load} from "./store/load/load";
 import {initialModelOutputState, modelOutput, ModelOutputState} from "./store/modelOutput/modelOutput";
@@ -44,6 +44,7 @@ import {ModelCalibrateMutation, ModelCalibrateUpdates} from "./store/modelCalibr
 import {GenericChartState, initialGenericChartState, genericChart} from "./store/genericChart/genericChart";
 import {Warning} from "./generated";
 import {DataExplorationState} from "./store/dataExploration/dataExploration";
+import {DownloadResultsMutation} from "./store/downloadResults/mutations";
 
 export interface RootState extends DataExplorationState {
     version: string,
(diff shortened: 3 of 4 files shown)


src/app/static/src/scss/style.scss
```diff
@@ -162,3 +162,13 @@ a.down::after {
   text-align: left !important;
   min-width: 32em;
 }
+
+.noDataMessage {
```
Nicolas-Dolan, line 166: For some reason, the class from the stylesheet in the vue-chart repo isn't being applied properly in hint so I've added a copy here as well, which fixes the problem.

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -99,6 +99,7 @@
                     :selections="comparisonPlotSelections"
                     :formatFunction="formatBarchartValue"
                     :showRangesInTooltips="true"
+                    :no-data-message="noChartData"
```
Nicolas-Dolan, line 102: I tried using `v-translate:no-data-message="'noChartData'"` but this wouldn't work properly as a prop. I suspect it had something to do with the order of operations.

LekanAnni:
Screenshot 2022-09-23 at 12 21 00

The message is meant to appear on the chart just like the below, so we can maintain consistency.

Screenshot 2022-09-23 at 12 22 55

LekanAnni (changes requested):
See my comment/screenshot above, no data message should appear on the chart.

Nicolas-Dolan:
> The message is meant to appear on the chart just like the below, so we can maintain consistency.

@LekanAnni It should do. Did you install the latest version of vue-charts?

src/app/static/src/scss/style.scss
LekanAnni: You don't need a new css, there's a class you can use, see usage https://github.com/mrc-ide/hint/blob/5b4730fa002956a8da762a74cb3f5f1eb9b69e66/src/app/static/src/app/components/plots/MapEmptyFeature.vue#L2.

src/app/static/src/app/components/modelOutput/ModelOutput.vue
LekanAnni, line 102: I don't think this is the right place for the message... again see usage https://github.com/mrc-ide/hint/blob/2443d5dccb18440fbdb9dd153b21a28f8c4de91c/src/app/static/src/app/components/genericChart/GenericChart.vue#L34. Should be added in `vue-charts`. If there's no data, vue-charts should display a placeholder with noCharData message. This means we won't have to add the same message everywhere the chart is reused.

LekanAnni (changes requested):
Sorry I didn't pull up to date branch. It does works as expected but not the right approach. See below comments.

src/app/static/src/app/components/modelOutput/ModelOutput.vue
Nicolas-Dolan: I disagree because you are assuming that vue-charts is only going to be used within hint. What if it is used in another application and they don't want a no data message, and if they do, why would we assume what sort of text the message should contain? Besides, this approach was agreed when the vue-charts changes were approved

EmmaLRussell: Agreed @Nicolas-Dolan - getting the client to specify a message is an approach we already take elsewhere. Also, there's the language issue - we can't expect vue-charts be able to translate the no data message appropriately.

src/app/static/src/scss/style.scss
Nicolas-Dolan: I don't think it is possible to apply these classes as the no data message html lives inside the vue-charts repo. Ideally, the repo would provide its own css styling, but that doesn't seem to be getting applied whenever it is used within hint (although the repo's classes are applied when the example html in the repo are run directly in the browser). The vue-chart's stylesheet was suspiciously empty of classes prior to me adding my class to it, which suggests this might have been a known issue with the repo.

Nicolas-Dolan: See this pr for context: https://github.com/reside-ic/vue-charts/pull/30

PR #747, opened 2022-09-21
mrc-3649 Update hint to work with options schema changes

LekanAnni:
This PR updates calibrate option endpoint, updated with changes made in hintr https://github.com/mrc-ide/hintr/pull/405#issuecomment-1253592332. 

New changes now corresponds with options schema as seen here https://github.com/mrc-ide/hintr/pull/405/files#diff-6bd8a7b93ff7c312277151e42ef21432985c55b46dc4c46d27046e3bef94246fR40-R48

## Type of version change

Minor

## Checklist

### DIFF
--- a/src/config/hintr_version
+++ b/src/config/hintr_version
@@ -1,1 +1,1 @@
-master
+mrc-3215-redo
--- a/src/config/hintr_version
+++ b/src/config/hintr_version
@@ -1,1 +1,1 @@
-mrc-3215-redo
+master
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 2.13.0
+
+* Update hint to work with options schema changes
+
 # hint 2.12.0
 
 * Update backend logging for new log structure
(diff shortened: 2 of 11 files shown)


r-ash (commented):
I've deployed this on dev and it is looking good. Couple of strings I missed in naomi.options but I have created a PR to add those. 

Do we need to do anything to work with a `multiselect` with multiple values? See https://github.com/mrc-ide/hintr/pull/405/files#diff-6bd8a7b93ff7c312277151e42ef21432985c55b46dc4c46d27046e3bef94246fR40-R48 multiselect value can now be an array or a single string.

r-ash (approved):
I double checked the multiselect default and it looks great. Works with Cameroon which has 2 values for prevalence default
Screenshot_20220921_173210

src/config/hintr_version
```diff
@@ -1 +1 @@
-master
+mrc-3215-redo
```
Nicolas-Dolan, line 1: Presumably this can't be merged if it's pointing at an old version of hintr

Nicolas-Dolan (approved):

PR #749, opened 2022-09-28
Mrc 3654 Verify auth0 state for protection against threat of cross-site request forgery

LekanAnni:
## Description

This PR adds protection against threat of cross-site request.

#Testing
* Please replace `oauth2_**` in config.properties with real config, please let me know if you don't have the real config.
* Alternatively, you can deployed an instance on dev `https://naomi-dev.dide.ic.ac.uk/login`, where oauth2_login_method is enabled.

## Type of version change

Minor 

## Checklist


src/app/src/main/kotlin/org/imperial/mrc/hint/security/oauth2/OAuth2StateGenerator.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    fun encodedState(): String
+    {
+        return Base64.getEncoder().encodeToString(randomString.toByteArray())
+    }
+
+    fun decodedState(code: String): String
+    {
+        return Base64.getDecoder().decode(code).decodeToString()
+    }
```
EmmaLRussell, line 22: Why is the base64 encoding needed? Is it because otherwise the hyphens would get url encoded in the query string so the comparison would fail? A comment here would be appreciated too! :)

EmmaLRussell (changes requested):
Looks like it works well to me, but you'd better modify the state generator to give a unique code per request rather than having a single value for the duration of the Application. 

We are eventually going to need some decent integration tests for these. No way to fully test the error case currently, but that should be possible with hoverfly or something similar. Do we have a ticket to add those?

EmmaLRussell (commented):
Looks good to me, and have confirmed it now works perfectly to generate new id on each login. It feels like pac4j ought to be able to help out more with this, but their docs are pretty shocking so probably wouldn't be able to tell even if they could..

I've added a ticket for adding integration tests for Auth0 login: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3660

Where did you get to with investigating the need for base64 encoding? I don't really like the way that both the Redirection class and the CredentialExtraction class need to know about the encoding, with one implementing the encoding and one the decoding - feels like it would be cleaner if only one thing knew how to encode+decode and they could both use that. Or - if we really do need to encode the UUID, could `generateStateParameter` in SecurityConfig just encode it as soon as it's generated, and use the encoded value only after that (save encoded to the session and compare that to returned value from Auth0).

LekanAnni:
> Looks good to me, and have confirmed it now works perfectly to generate new id on each login. It feels like pac4j ought to be able to help out more with this, but their docs are pretty shocking so probably wouldn't be able to tell even if they could..
> 
> I've added a ticket for adding integration tests for Auth0 login: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3660
> 
> Where did you get to with investigating the need for base64 encoding? I don't really like the way that both the Redirection class and the CredentialExtraction class need to know about the encoding, with one implementing the encoding and one the decoding - feels like it would be cleaner if only one thing knew how to encode+decode and they could both use that. Or - if we really do need to encode the UUID, could `generateStateParameter` in SecurityConfig just encode it as soon as it's generated, and use the encoded value only after that (save encoded to the session and compare that to returned value from Auth0).

encoded state is needed in login Redirection url, it should be unguessable hence we are encoding it. see guide  https://community.auth0.com/t/how-to-read-auth0-state-parameter/57316. 

Once a user is authenticated, CredentialExtraction class decodes and verifies the state. I could create a class to handle encoding and decoding but not really necessary as we are encoding in a known library `Base64`.

src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
```diff
@@ -63,6 +64,13 @@ class Session(
         private const val MODE = "mode"
     }
 
+    fun generateStateParameter(): String
+    {
+        val state = UUID.randomUUID().toString()
+        sessionStore.set(webContext, STATE_REQUEST_PARAMETER, state)
+        return Base64.getEncoder().encodeToString(state.toByteArray())
```
EmmaLRussell, line 71: So it's now saving the encoded id to the session...

EmmaLRussell:
>Once a user is authenticated, CredentialExtraction class decodes and verifies the state. I could create a class to handle encoding and decoding but not really necessary as we are encoding in a known library Base64.

Sure, but it would be nice if only one thing had to know about that.

> I could move encoding line to generateStateParameter method of course

Which is fine, but still means you've got the encoding/decoding happening in two places - now in `SecurityConfig` and `OAuth2CredentialExtractor`. Why not just save the encoded string to the session rather than the UUID, then you don't need to decode it in the extractor at all, but just compare the returned paramater with what's in the session?

LekanAnni:
> > Once a user is authenticated, CredentialExtraction class decodes and verifies the state. I could create a class to handle encoding and decoding but not really necessary as we are encoding in a known library Base64.
> 
> Sure, but it would be nice if only one thing had to know about that.
> 
> > I could move encoding line to generateStateParameter method of course
> 
> Which is fine, but still means you've got the encoding/decoding happening in two places - now in `SecurityConfig` and `OAuth2CredentialExtractor`. Why not just save the encoded string to the session rather than the UUID, then you don't need to decode it in the extractor at all, but just compare the returned paramater with what's in the session?

Not storing encoded state in session, I just returned an encoded state for the url redirection.

EmmaLRussell:
> Not storing encoded state in session, I just returned an encoded state for the url redirection.

Yes, you're storing the unencoded UUID in the session: , but returning the encoded value to the redirection class to use - and then Auth0 passes back that value, which you then decode in the extractor class and compare to what's in the session. My suggestion is to just save the encoded UUID in the session, so you only have to deal with encoded values after that. The encoded value should be unique per UUID so you could just compare those, and not have to do any decoding. With the advantage that the all the knowledge of what encoding you're using stays in the config class, since it's only happening one way.

EmmaLRussell (approved):
Looks good, thanks!

PR #754, opened 2022-10-04
mrc-3650 Speed up display of project rehydrate dialog

LekanAnni:
## Description

PR is self explanatory, speeds up display of project rehydrate dialog.

## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/tests/load/mutations.test.ts
```diff
@@ -87,4 +87,10 @@ describe("Load mutations", () => {
         expect(testState.statusPollId).toEqual(-1);
     });
 
+    it("StartPreparingRehydrate sets preparing state", () => {
+        const testState = mockLoadState();
+        mutations.StartPreparingRehydrate(testState, true);
+        expect(testState.loadingState).toBe(LoadingState.SettingFiles);
```
EmmaLRussell, line 93: Shouldn't this test the change to `preparing` too?

src/app/static/src/app/store/load/mutations.ts
```diff
@@ -66,10 +67,13 @@ export const mutations: MutationTree<LoadState> & LoadMutations = {
     },
     PreparingRehydrate(state: LoadState, action: PayloadWithType<ProjectRehydrateSubmitResponse>) {
         state.rehydrateId = action.payload.id;
-        state.preparing = true;
         state.complete = false;
         state.loadError = null;
     },
+    StartPreparingRehydrate(state: LoadState, action: PayloadWithType<boolean>) {
```
EmmaLRussell, line 73: Does this actually need a payload? It never gets called with `false` does it, so can we not just assume that it we're starting preparation we set `preparing` to `true`?

EmmaLRussell (commented):
Works for me!

EmmaLRussell (approved):
Looks good. I think you should be able to leave out the payload altogether rather than having to set it to null? But it doesn't really matter.

PR #756, opened 2022-10-11
Mrc-3629 new Add no data message

LekanAnni:
## Description

This PR is from Nick.

Upgrades vue-charts version and submits prop to it for the translated no data message. The easiest way to test is to go to the comparison tab on the review output step, set the x axis to sex, and then unselect the sex value from the filter. This should mean the chart is getting no data and prompt the no data message.

Note: you will need to run npm install --prefix=src/app/static/ to get the latest chart.

Something strange has just happened. Please see previous review comments https://github.com/mrc-ide/hint/pull/745

I assigned all Nick's tickets to myself and then pushed, which caused the branch to be deleted.

## Type of version change
_Delete as appropriate_

## Checklist


LekanAnni:
@EmmaLRussell What did you say was wrong with this branch ?

EmmaLRussell:
> @EmmaLRussell What did you say was wrong with this branch ?

I don't remember thinking there was anything much wrong with it - I thought you had issues with the css, and with the idea of the client providing the no data message, but Nick and I both argued with you!

EmmaLRussell (commented):
This works great for the Comparison chart, but no data message is not applied to the main output barchart. I assume it should be on both? (Can't remember if this was an issue in the previous incarnation of this branch or not!)

LekanAnni:
> This works great for the Comparison chart, but no data message is not applied to the main output barchart. I assume it should be on both? (Can't remember if this was an issue in the previous incarnation of this branch or not!)

The prop was not added to bar chart,  I have added. The problem is testing translated text, still remains.

EmmaLRussell (approved):

PR #757, opened 2022-10-11
mrc-3557 Barchart uncertainty ranges (error bars) are greyed out on first rend…

LekanAnni:
## Description

This PR Fixes greyed out error bars and awaits comparison data before attempting to plot bar chart. The image below shows the grey out and the expected error bars.

* The issue was caused because display class property was used to display chart. This affected the reactivity of the chart props and could not listen to change events on chart data.  There was also a delay in rendering the chart with both bar and comparison vue-charts. Upon a page refresh, chart config and chart were not displayed immediately, took a few seconds to display.

Screenshot 2022-10-11 at 17 18 45
Screenshot 2022-10-11 at 17 18 37

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 2.23.0
+
+* Barchart uncertainty ranges (error bars) are greyed out on first rendering
+
 # hint 2.22.0
 
 * Bug: Fix download status errors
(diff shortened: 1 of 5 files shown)


EmmaLRussell (approved):
Looks good. I remember talking to Nick about this as a potential approach, and he mentioned that the only issue with it is that it means you see the initialising animation for the barchart (with the bars rising up) whenever you switch tab - but I think that's ok since you see it the first time you open the tab anyway, and it's certainly better that the broken error bars!

PR #758, opened 2022-10-13
mrc-3081 Persist user selections for comparison barchart

LekanAnni:
## Description

This PR ensure persisted selections take precedence over default selections. I am unsure why the whole thing is setup this way, default not being persisted or updated but `comparisonPlotSelection` hence the confusion.

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/app/store/modelCalibrate/actions.ts
+++ b/src/app/static/src/app/store/modelCalibrate/actions.ts
@@ -77,7 +77,11 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
                 commit({type: ModelCalibrateMutation.CalibrateResultFetched, payload: data});
                 commit({type: ModelCalibrateMutation.WarningsFetched, payload: data.warnings});
 
-                selectFilterDefaults(data, commit, PlottingSelectionsMutations.updateBarchartSelections)
+                const barChartPlotSelections = context.rootState.plottingSelections.barchart
+                // barchart reverts to default selections if selection does not exist in the state.
+                if (!barChartPlotSelections) {
+                    selectFilterDefaults(data, commit, PlottingSelectionsMutations.updateBarchartSelections)
+                }
                 commit(ModelCalibrateMutation.Calibrated);
                 if (switches.modelCalibratePlot) {
                     dispatch("getCalibratePlot");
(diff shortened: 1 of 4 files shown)


src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -77,7 +77,11 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
                 commit({type: ModelCalibrateMutation.CalibrateResultFetched, payload: data});
                 commit({type: ModelCalibrateMutation.WarningsFetched, payload: data.warnings});
 
-                selectFilterDefaults(data, commit, PlottingSelectionsMutations.updateBarchartSelections)
+                const barChartPlotSelections = context.rootState.plottingSelections.barchart
+                // barchart reverts to default selections if selection does not exist in the state.
+                if (!barChartPlotSelections) {
```
EmmaLRussell, line 82: I don't think we should include this conditional  (though I understand why you added it!) - we shouldn't retain user plotting selections when the model has been recalibrated because those selections may no longer be valid for the new results. The point of this ticket was to retain user selections for the comparison plot when the user changes tab or step, not when there are new results

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -116,7 +120,9 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
             .get<ModelResultResponse>(`model/comparison/plot/${calibrateId}`);
 
         if (response) {
-            if (response.data){
+            const comparisonPlotSelections = context.rootState.plottingSelections.comparisonPlot
```
EmmaLRussell, line 123: Again, this is not going to help with the requirements of the ticket, and we shouldn't do this as new result set may have entirely different selections.

EmmaLRussell (commented):
Sorry, Lekan, it looks to me like this ticket is obsolete as the peristence of comparison plot selections is already working and peristed to the Vuex state in master. Maybe there was a duplicate ticket?

I don't think we should add the changes you've made here - peristing user selections when new results are fetched - as there is a chance the available filter options etc may have changed, so if we were going to provide that functionality we'd need to be more careful about making sure all the selected options still exist in the new result set. Probably not worth it, so I think we should just mark this ticket as obsolete and close this PR>

PR #759, opened 2022-10-13
Mrc 3679 - Login to data exploration throws an error on redirect

LekanAnni:
## Description

PR fixes exploration model login retired issue.  Adding a regression test here https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3682/Setup-Selenium-and-add-Browser-test-for-data-exploration-login

I will be setting up a selenium test framework and testing exploration login. 

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 2.23.0
+
+* Login to data exploration throws an error on redirect
+
 # hint 2.22.0
 
 * Bug: Fix download status errors
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,14 @@
-# hint 2.23.0
+# hint 2.25.0
 
 * Login to data exploration throws an error on redirect
+
+# hint 2.24.0
+
+* Barchart uncertainty ranges (error bars) are greyed out on first rendering
+
+# hint 2.23.0
+
+* Implement no data message in comparison chart
 
 # hint 2.22.0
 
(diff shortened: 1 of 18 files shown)


EmmaLRussell (changes requested):
So `http://localhost8080/explore` doesn't work at all anymore in this branch. although I can successfully navigate to Data Explore at `http://localhost8080/callback/explore` and login from there. Is that what was intended? I think we do need to continue to support `../explore` as the link which users can navigate to to access data exploration, don't we? Having to change that to `.../callback/explore` would be pretty ugly. Thoughts @r-ash ? @LekanAnni can we just redirect from `/explore` to `/callback/explore`?

I'm a little confused about why this bug  arose in the first place. It was because of the changes made to support SSO?

LekanAnni:
> So `http://localhost8080/explore` doesn't work at all anymore in this branch. although I can successfully navigate to Data Explore at `http://localhost8080/callback/explore` and login from there. Is that what was intended? I think we do need to continue to support `../explore` as the link which users can navigate to to access data exploration, don't we? Having to change that to `.../callback/explore` would be pretty ugly. Thoughts @r-ash ? @LekanAnni can we just redirect from `/explore` to `/callback/explore`?
> 
> I'm a little confused about why this bug arose in the first place. It was because of the changes made to support SSO?

explore is obsolete, I can add an endpoint to redirect requests from /explore but users should not really navigate the website via memoization of uri. As @r-ash suggested yesterday, we should have a link with can take users to exploration mode. 

So I am not inclined to add a redirect url as pac4j listens to callback/ for redirections now that we have different auth methods.

EmmaLRussell:
> users should not really navigate the website via memoization of uri

Why not? That's what we're going to do! Isn't it better to make the link more friendly rather than less, which this branch currently does? It's also going to break any current usage, which we'd need to warn then about.

EmmaLRussell:
Also - the `callback` path reflects an implementation detail of the backend. This shouldn't impact links available to users. (Would we have to change it again if we change our auth implementation again?)

EmmaLRussell (approved):
Thanks!

PR #760, opened 2022-10-14
mrc-3547 Remove fixed disaggregate value from xAxis options in vue-charts

LekanAnni:
## Description

This PR upgrade vue-charts version include new releases. New release removes disaggregate value from xAxis options when aggregates is fixed and vice versa.

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 2.23.0 
+
+* Remove fixed disaggregate value from xAxis options in vue-charts 
+
 # hint 2.22.0
 
 * Bug: Fix download status errors
--- a/src/app/static/package-lock.json
+++ b/src/app/static/package-lock.json
@@ -3707,9 +3707,9 @@
       }
     },
     "@reside-ic/vue-charts": {
-      "version": "0.3.0",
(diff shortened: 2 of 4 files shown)


EmmaLRussell (commented):
This isn't quite working how I expected. In the comparison chart, I was expecting the disaggregate by filter (which I think is Data type) to be removed from the X Axis filter drop-down, but this is not the case, and I can set X Axis to be Data Type too so we see the weird configuration we were trying to avoid where the same filter is both x axis and disaggregate by. 
Screenshot from 2022-10-19 10-14-47
@LekanAnni are you seeing that too?
@rob is that your understanding of what ought to be happening here?

r-ash:
> @rob is that your understanding of what ought to be happening here?

Yeah I was expecting that data type is not shown as an option for the x axis any more

EmmaLRussell:
So the problem here was that, although the source had been updated in the latest version of the vue-charts packages pushed to npm, the dist hadn't been. I've pushed a new version with latest built dist (version 0.4.1) and committed a change here to use that version. Small vue-charts PR with incremented version number accompanies this.

PR #763, opened 2022-10-24

LekanAnni:
## Description

This PR adds browser test setup and a basic data exploration test.

## Type of version change
_Delete as appropriate_

## Checklist


buildkite/shared-build-env.dockerfile
```diff
@@ -1,6 +1,8 @@
-FROM vimc/node-docker:master
+FROM vimc/node-docker:mrc-3682
```
LekanAnni, line 1: Remember to change before merging

r-ash (commented):
This looks good but tbh I think setting up the [projects](https://playwright.dev/docs/test-advanced#projects) was better for a few reasons:
* It keeps URL switch in one place (your change has that in 2 different files)
* We're going to have to use projects when we want to run some tests against deployed instances as we can use projects to run separate set of tests in that instance. We could just punt on it until the point that we need it
* I think also having the two modes of running the tests i.e. one running in docker and one running out of docker will be useful. With projects setup we could run the buildkite tests locally and it should work the same out of the box, with this setup we will have to set an env var on our machines. It also followed the pattern used elsewhere with our integration tests (one npm command to run in docker and one to run locally)

package-lock.json
```diff
@@ -1,3 +1,11 @@
 {
-  "lockfileVersion": 1
+  "requires": true,
```
EmmaLRussell, line 2: Doesn't seem like this should be here, or associated node_modules files.

EmmaLRussell (commented):
Great to get these in at last. 

Something is a bit weird about dotenv here - you've committed the dotenv node-module, and a package lock file at the top level. Was this done in error?

I think it would be good to include a brief section in the README on the Playwright tests, including how to run them, and an explanation of the use and need for the env variable, and storageState.json, as these are not obvious.

I think I've missed some discussions on Playwright projects, but Rob's comment on keeping switch values in one place sounds sensible. Maybe that could be done in a separate ticket?

LekanAnni:
> This looks good but tbh I think setting up the [projects](https://playwright.dev/docs/test-advanced#projects) was better for a few reasons:
> 
> * It keeps URL switch in one place (your change has that in 2 different files)
> * We're going to have to use projects when we want to run some tests against deployed instances as we can use projects to run separate set of tests in that instance. We could just punt on it until the point that we need it
> * I think also having the two modes of running the tests i.e. one running in docker and one running out of docker will be useful. With projects setup we could run the buildkite tests locally and it should work the same out of the box, with this setup we will have to set an env var on our machines. It also followed the pattern used elsewhere with our integration tests (one npm command to run in docker and one to run locally)

Projects should not be used for setting server path, it's mainly for testing different dir or ignoring. We can still use projects for testing multiple tests without altering the env or changing anything particular. 

```
  projects: [
    {
      name: 'Nigeria',
      testMatch: /.*nigeria.etest.ts/,
    },
    {
      name: 'Ghana',
      testIgnore: /.*ghana.etest.ts/,
    },
  ]
```

`npx playwright test --project= Nigeria`
`npx playwright test --project= Malawi`

The above setup does not need to worry about server, we can run these tests locally and on buildkite.
We shouldn't be setting server path in projects.

LekanAnni:
> * It keeps URL switch in one place (your change has that in 2 different files)

I can definitely export the path to be reused.

package-lock.json
LekanAnni: You are right, it's a mess, I will try to fix the modules.

LekanAnni:
> I think it would be good to include a brief section in the README on the Playwright tests, including how to run them, and an explanation of the use and need for the env variable, and storageState.json, as these are not obvious.

Absolutely, I will add test desc to readme. I don't think env or storageState should be explained in readme. I may write some comments but these are standard playwright usage as per documentation.

EmmaLRussell:
> > I think it would be good to include a brief section in the README on the Playwright tests, including how to run them, and an explanation of the use and need for the env variable, and storageState.json, as these are not obvious.
> 
> Absolutely, I will add test desc to readme. I don't think env or storageState should be explained in readme. I may write some comments but these are standard playwright usage as per documentation.

Yes, please write some comments. Yes, this is documented usage, but it definitely counts as advanced compared to many of our test scenarios, and it's good to add context for the benefit of other developers.

buildkite/test-browser-e2e.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Create an image based on the shared build env that compiles and tests the front-end ADR integration
+docker build --tag hint-test-browser-e2e \
+  --build-arg GIT_ID=$GIT_ID \
+  -f $HERE/test-browser-e2e.dockerfile \
+  .
+
+# Run the created image
+docker run --rm \
+  -e CI=$CI \
```
r-ash, line 26: What do you think about just setting the base URL here?

Thinking about the case that we want to debug the docker build, if we try and run this locally we will need to remember to set CI env or it won't work locally. If we just set the base URL here then it will avoid that issue?

A small point as it might not be a common way we run the tests but it will be useful if the tests are failing on CI but passing locally.

r-ash (commented):
> Projects should not be used for setting server path

Fair enough, is that in the docs somewhere? Struggling to find that and wondering if you've found some useful resource I haven't seen.

I also think some docs would be useful, maybe as a knowledge base article? Especially as this is new technology for most of us (and is something we probably want to be using more going forward)

buildkite/test-browser-e2e.sh
LekanAnni: `process.env.CI ` is used for other things, i.e setting the playwright workers for build kite and local ?

When debugging playwright docker container, it will display baseUrl in the console. If we have need for further things we will definitely extend it, for now this is fine.

buildkite/shared-build-env.dockerfile
LekanAnni: I will update other git repo using this changes before merging. I initially tried to merge changes into this docker file but it increased the build time massively, which I am not v. happy about. I have elected to update others repos.

EmmaLRussell (approved):

PR #764, opened 2022-10-31
mrc-3746 -  enable comparison report download

LekanAnni:
## Description

This PR shows comparison reports download in final step. Also handles download endpoint within apiService so as to handle errors gracefully. Generating comparison report is error prone because some old models would need to re-fit model to be able to generated comparison report.

I have added a ticket to update all download endpoints https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3814/Change-download-approach-for-remaining-download-files

## Type of version change
_Delete as appropriate_

## Checklist


LekanAnni:
> OUTPUT_GENERATION_FAILED

Download comparison requests for a file, hintr put the request in a queue, frontend polls for result, hintr sends result, frontend use resultId to download.

I think, the issue is meant to be resolved in hintr. Frontend should not receive a result id if result is not valid.

r-ash:
> > OUTPUT_GENERATION_FAILED
> 
> Download comparison requests for a file, hintr put the request in a queue, frontend polls for result, hintr sends result, frontend use resultId to download requested file.
> 
> I think the issue is meant to be resolved in hintr. Frontend should not receive a result id if result is not valid.

So this is failing in naomi whilst generating the result, so hintr doesn't know about the failure is as it coming from Naomi. So I could work around it in hintr and add some checks to fail early before it gets put on a queue but then I am duplicating logic over hintr and naomi. I think current failure is fine and more generally speaking there is always the potential for failures when generating the output. We should show the error message returned from hintr nicely.

src/app/static/src/tests/integration/downloadResults.itest.ts
```diff
@@ -244,4 +242,25 @@ describe(`download results actions integration`, () => {
...
+        await actions.downloadComparisonReport({
+            commit, dispatch, state, rootState: root
+        } as any);
+
+        expect(commit.mock.calls.length).toBe(1);
+        expect(commit.mock.calls[0][0]["type"]).toBe("ComparisonError");
+        expect(commit.mock.calls[0][0]["payload"].detail).toEqual("Missing some results")
+        expect(commit.mock.calls[0][0]["payload"].error).toEqual("FAILED_TO_RETRIEVE_RESULT")
+    })
 })
```
LekanAnni, line 266: skipping this test for now because hintr return `No handler found for GET /download/status`, btw this works locally but returns no handler on build kite.

Response for download result endpoint needs to be altered in `hintr`

EmmaLRussell (changes requested):
This works great for doing the actual download, but I think there is a real problem with showing errors, as revealed by the test failure.

When I tried sabotaging the endpoint, I did indeed see an error displayed, however it was our old friend "Could not parse API response" despite the fact that I could see, in Chrome, that the actual response body from the backend was a nicely formatted JSON error, with 'errors', 'data' etc. This accounts for the test getting an unexpected mutation, since unparseable errors are committed with the generic errors module mutation rather than the mutation requested by `withError`. 

The issue with the response seemed to be that the AxiosError did not contain any `data` content (where you'd expect to see the response body) - it is just an empty object {}. This is the `failure` object which is passed into `isHINTResponse` by `_handleError` in apiService. So of course it cannot parse it or extract the error details.

I assume that this must be because we've requested "blob" type response from Axios, as I've never seen this behaviour for any other request. There might be a way to parse the error out of what's in the AxiosResponse. Or we might be forced to do a second non-blob request when there's an error and then use that response to get error details from (ugh).  
(Update: it is definitely the `responseType: "blob" which is causing the issue - I tried taking it out, knowing that the endpoint would error, and the error was then correctly parsed and displayed, and ComparisonError mutation was logged.)

Incidentally, along the way I came across what seems like another issue, albeit a benign one. I initially thought that it wasn't parsing the response because the errors in the response I can see in Chrome have the `detail` prop, whereas in `isHINTError` we're checking for `details` - but I don't think `details` is used anyway, everything else uses `detail` (without the "s"). I think this has been a silent bug along because it allows 'details' to be undefined - which of course it always is!) We should fix this in another maintenance ticket.

LekanAnni:
Well, well, I could only recreate the error by restarting redis container. Finally found some axis limitation, when responseType is specified to be of `blob`, it applies to both success and failure response. We were receiving errors in blob of type json. 

There are few options to fix this issue, upgrading Axios to recent version which will then respond with json failure message but with a specific ```message: "Request failed with status code 400"```. I thought this might be our best solution, but we will be unable to display hint specific messages except we read the blob. So I didn't upgrade axios because I would have to do more stuffs to make it work with our current setup anyway.

Reading error message from blob, I managed to extract errors which displays on the page as expected. However, `Axios mock  adapter` can not read blob response, both failure and success. I am unable to test using this library. I have created a ticket to add a browser test and/or investigate ways of making axios mock work for blob here https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3821/Investigate-axios-mock-adapter-handling-blob-response.

I upgraded axios to `1.20.0` thinking bug might have been fixed according the release not https://github.com/ctimmerm/axios-mock-adapter/releases/tag/v1.20.0 but not luck.

EmmaLRussell (commented):
Great, this works really nicely! You're right, blobs *are* painful to work with!

The only thing is, I'm seeing the displayed error twice, any idea why that would be?
Screenshot from 2022-11-18 10-55-55

If mockAxios is causing problems, how about just mocking axios manually for this test? You'd probably want to do that in a separate test module so the mocks don't interfere with each other.

LekanAnni:
> If mockAxios is causing problems, how about just mocking axios manually for this test? You'd probably want to do that in a separate test module so the mocks don't interfere with each other.

This will be in another ticket.

LekanAnni:
> Great, this works really nicely! You're right, blobs _are_ painful to work with!
> 
> The only thing is, I'm seeing the displayed error twice, any idea why that would be?

Yes I saw that, I was meant to fix it. I have separated the error state, now downloadErro.

r-ash (changes requested):
Thanks Lekan this is looking great. I deployed this onto dev before most recent change and all was looking good except for the duplicate message Emma mentioned. I've just redeployed and behaviour is a little strange,

First time you click the button to download it shows a "Could not parse API response error", on second click it shows the actual error.

LekanAnni:
> Thanks Lekan this is looking great. I deployed this onto dev before most recent change and all was looking good except for the duplicate message Emma mentioned. I've just redeployed and behaviour is a little strange,
> 
> First time you click the button to download it shows a "Could not parse API response error", on second click it shows the actual error.

This looks like a case of stale state, can you redeploy with the updates and refresh ?

LekanAnni:
I have added extra commit to clear download error and additional tests.

r-ash:
I think there is still some issue here, I've redeployed this to dev and it is taking multiple clicks to see the message. But it is inconsistent, sometimes it works first try. See gif for an example, any idea what is going on? This is on dev, logging in as james@example.com using the project "Download comparison error"
comparison-download

LekanAnni:
> I think there is still some issue here, I've redeployed this to dev and it is taking multiple clicks to see the message. But it is inconsistent, sometimes it works first try. See gif for an example, any idea what is going on? This is on dev, logging in as [james@example.com](mailto:james@example.com) using the project "Download comparison error"
> comparison-download

Problem didn't happen locally for me. Typically happens when downloading large files. I have increase server connection timeouts.

EmmaLRussell (commented):
So it looks like the recent changes, with the new mutation, and the committing of errors from the `readStream` util are aimed at avoiding the duplicate error, and also reporting invalid blobs rather than the generic could not parse message.  I'm concerned about what this now does for separation of concerns - i don't think we want the util doing commits when that should be handled by apiService, and needing to add the `_commitType` member doesn't seem ideal. I appreciate that the 'INVALID_BLOB' error is more accurate than simply saying that the response could not be parsed, but I don't think we really need to provide that detail. 

Rob and I talked this over, and we think it is most likely that HINT is not to blame for these empty responses, which should not occur, and the 'Could not parse API response' message is appropriate behaviour for the HINT front end. So we're going to concentrate effort on looking into possible issues in the proxy. Given the above concerns, that Lekan is unwell, and that we need to get the comparison downloads available asap, I'd suggest merging this branch from commit df351de - I don't think we need to worry about the duplicate case too much, and can potentially handle that in a separate branch.

EmmaLRussell (approved):
Looks good. some code cov issues, but those could go into a separate maintenance ticket, so we can merge this asap.

LekanAnni:
> Looks good. some code cov issues, but those could go into a separate maintenance ticket, so we can merge this asap.

Coverage is due to axios adapter not able handling blob response, ticket added already https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3821/Investigate-axios-mock-adapter-handling-blob-response.

PR #765, opened 2022-11-04
Mrc-3275 - Bug: prevent spamming when creating a new project during model calibrate

LekanAnni:
## Description

This PR fixes issue continuous polling for results when polling in progress while a new project is created. I added commits to stop polling for modelRun and modelCalibrate if they exists. This method has stopped the known issue.

## Type of version change
_Delete as appropriate_

Minor 

## Checklist


src/app/static/src/app/store/modelCalibrate/mutations.ts
```diff
@@ -138,6 +139,10 @@ export const mutations: MutationTree<ModelCalibrateState> = {
 
     [ModelCalibrateMutation.CalibrateResultFetched](state: ModelCalibrateState, action: PayloadWithType<CalibrateResultResponse>) {
         state.result = action.payload
+    },
+
+    [ModelCalibrateMutation.ResetIds](state: ModelCalibrateState) {
```
EmmaLRussell, line 144: `ResetIds` seems a bit ambiguous (we have lots of ids!) - how about just `StopPolling` or `StopStatusPolling`?

src/app/static/src/app/store/modelRun/mutations.ts
```diff
@@ -88,6 +89,10 @@ export const mutations: MutationTree<ModelRunState> = {
     [ModelRunMutation.ClearWarnings](state: ModelRunState) {
         state.warnings = [];
     },
+
+    [ModelRunMutation.ResetIds](state: ModelRunState) {
```
EmmaLRussell, line 93: ...similarly here

EmmaLRussell (approved):
Works great 👍 I guess this is less necessary for ModelRun, since running happens with a modal shown, so you can't create a new project when it's running anyway - but it's good to be consistent, and we might change the UI some day. 

I think the name of the mutation could be clearer, but I've just noticed that the Download module has that mutation name too, so it's following precedent! Up to you whether you want to change that or not - if you do, it should be changed for the Downloads too..

LekanAnni:
> Works great 👍 I guess this is less necessary for ModelRun, since running happens with a modal shown, so you can't create a new project when it's running anyway - but it's good to be consistent, and we might change the UI some day.
> 
> I think the name of the mutation could be clearer, but I've just noticed that the Download module has that mutation name too, so it's following precedent! Up to you whether you want to change that or not - if you do, it should be changed for the Downloads too..

Looks good to me, I don't mind it.

PR #767, opened 2022-11-09
mrc-3767 Return more information for Could not parse API response

LekanAnni:
## Description

PR logs response detail and changed display text to human readable format.

## Type of version change
_Delete as appropriate_

## Checklist


EmmaLRussell (commented):
Sorry, didn't get around to looking at this yesterday, and don't have too much time to run it up now, but just from a quick look, it seems like this only does some of the things mentioned in the ticket (@r-ash 's initial write up and my follow-up comment). Those could be done in separate tickets as long as they don't get lost. 

This PR does:
- console logs full error (this is helpful for use when reproducing error, but not really for users themselves)
- reports status code

It doesn't appear to:
- report time of error - useful when reported by user who would only have vague idea of time themselves
- report content type
- report full response content (possibly in expandable panel - or at least show first n characters)
- distinguish genuine parsing errors from errors thrown by mutations

r-ash (approved):
From what we discussed on Friday let's get this merged as is and review any additional requirements in a separate PR

PR #769, opened 2022-11-09
Mrc-3765 YAxis scale is not automatically generating the correct range at low values for some indicators

LekanAnni:
## Description

This PR fixes the YAxis scale ranges issues as seen in the screenshot here https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3765/y-axis-scale-is-not-automatically-generating-the-correct-range-at-low-values-for-some-indicators

Rachel had suggested making accuracy 10 when value is in the low range but this does not solve the yAxis issue for most selections. However, I chose to rely on chartJs numeric algorithm to calculate scale range when data is too small for a given accuracy.

I have added inline comment explaining my approach, we can now view charts as seen below.

Screenshot 2022-11-10 at 10 11 25

## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/app/components/plots/utils.ts
```diff
@@ -193,7 +193,22 @@ export const formatOutput = function (value: number | string, format: string, sc
...
+         * rounding using the given accuracy value.
+         */
+
+        if (accuracy === 100) {
+            if (ans > 500) {
+                ans = Math.round(ans / accuracy) * accuracy
+            }
+        } else {
+            ans = Math.round(ans / accuracy) * accuracy
+        }
```
r-ash, line 211: I think I had a different understand from our discussion with Jeff. What he wants is for accuracy of 100
* If value in data < 200 then show to nearest 10
* If value in data < 5 then just show text "<5"

but we should also make this work for other accuracy values. I think logic could be something like,
```
if (value < 5) {
  show <5
} else if (value < accuracy * 2 && accuracy > 1) {
  round to accuracy / 10
} else {
  round to specified accuracy
}
```

r-ash: He suggested we check this with Mary though. He did also say that if this is a faff then just show the raw number.

r-ash (changes requested):

src/app/static/src/app/components/plots/utils.ts
LekanAnni: > I think I had a different understand from our discussion with Jeff. What he wants is for accuracy of 100
> 
> * If value in data < 200 then show to nearest 10

I suggested 200 to Jeff because yAxis displayed nicely when I tested but on testing again with different age groups, it disarranged YAxis. applying 100 accuracy when value is greater than 500 fixes issues with YAxis pretty well. I will inform him I updated to 500. 

I think you should try testing with different age groups to see how it displays.

> * If value in data < 5 then just show text "<5"
Only numeric value is applicable here
> 
> but we should also make this work for other accuracy values. I think logic could be something like,
> 
> ```
> if (value < 5) {
>   show <5
> } else if (value < accuracy * 2 && accuracy > 1) {
>   round to accuracy / 10
> } else {
>   round to specified accuracy
> }
> ```

I have allowed ChartJs to calculate the range using min and max value. What I am trying to avoid is exactly your suggestions where we round value based on unknown range. Your suggested solution disarranges YAxis for value for a lot age groups. Also, we can not show <5, only able to return numeric values.

EmmaLRussell: I'm coming to this late and may not understand all the issues. I think the problem is that the accuracy value is sometimes being set quite high relative to the highest value in the data (e.g. accuracy 100 when highest value is e.g. 200). However chartJs wants to show about 10 ticks on the chart, and we're asking it to do it with an accuracy to the nearest 100, so it's doing the best it can and rounding what are basically ticks per 10 to the nearest 100. Is that the problem?

It feels like the problem is greater than the scope of this method, which I think is formatting a single value - wouldn't a high accuracy value be ok for low values, as long as there were some high values in the data? (Although the filtering is handled in the barchart component so maybe we do need to get the accuracy right for the lowest possible values too.) Also, dealing with the issue for accuracy == 100 seems very specific. Don't we also support e.g. accuracy == 1000 potentially?

Would it make more sense to have a separate method which checks the appropriateness of the accuracy value across all data values. e.g. something like 'if there are any non-zero values which are  < 10 * accuracy, adjust accuracy to greatest power of 10 which is < 10 * accuracy`.

LekanAnni: We mostly support accuracy 100, where accuracy is greater or less than 100, CharJs will calculate the scale range and display YAxis nicely. Alternatively, if we do not want chartJS to handle scale range, we will then have to manually calculate for all expected accuracy. 

The point of this ticket is to show YAxis nicely for all data points. I think you should spin the app and play with different formatting.

EmmaLRussell (commented):
Ok, so there's two issues being discussed here:
1. Avoiding multiple ticks with the same displayed value - this is what Lekan is addressing in this branch. 
2. Showing appropriate values (in the tooltip) as described by Jeff, with pseudo-code suggested by Rob above. 

Lekan, you've suggested we shouldn't address 2. this branch, only 1. In that case, do we have a ticket to deal with 2.?

The reason issue 1. has been arising is because this method `formatOutput` is being passed into ChartJs via vue-charts, and is being used to format both ticks and tooltip values. So e.g. with accuracy 100, ChartJS is trying to format tick values of 20, 40 etc to nearest 100. 

What's implemented here is not a generic solution because it only applies to accuracy 100. Here's an example of bug showing up for indicator with accuracy = 1000:
Screenshot from 2022-11-15 12-31-48

There are also issues for very low values, but I think this might be inherent to ChartJS, and probably not worth worrying about now:
Screenshot from 2022-11-15 12-30-29

(We'll have to  deal with that when we look at issue 2. and the <5 business.)

Your approach of modifying for values < 500 does seem to work - not sure if there would be race conditions near that value? How about just generalising that for other accuracy values?
```
	if (accuracy > 1) {
            if (ans > 5*accuracy) {
                ans = Math.round(ans / accuracy) * accuracy
            }
        } else {
            ans = Math.round(ans / accuracy) * accuracy
        }
```

.. and dealing with issue 2. separately?

Incidentally, the format callback function chartJS takes can return a string, so in fact we *could* return '<5' - however, with the current way vue-charts is configured, that would be shown both in tooltips and on the ticks. Is that what we want? Should clarify that in separate ticket (?) for issue 2.

r-ash (approved):
This looks like a big improvement to what is currently there.

I've made a follow on ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3815/Show-value-5-for-indicators-with-accuracy-1 with the <5 stuff with a comment about the issue still persisting for very small values, we can review later (post workshops) how we want to fix it.

EmmaLRussell (commented):
This seems fine to me, to fix the axis issue, but do bear in mind that it means that unrounded values will now be shown for all values < 5 * accuracy. So, for population where accuracy = 1000, all values < 5000 are now shown unrounded. This seems to be quite rare though (could only find example for Likoma in the Malawi dataset).

PR #771, opened 2022-11-11
mrc-3800 Add spring actuator

LekanAnni:
## Description

Use spring actuator to see metrics.

`/actuator/httptrace` to see elapsed time for each request

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/README.md
+++ b/src/app/static/README.md
@@ -31,3 +31,7 @@ is as simple as running IntelliJ's auto-formatting command (<kbd>ctrl + alt + l<
 - js is bundled using webpack - this task can be triggered by running `npm run js`
 
 Or to compile both at once, `npm run build`
+
+### Management
+
+- `/actuator/health` to see running status of apps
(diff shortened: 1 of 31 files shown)


src/app/static/README.md
```diff
@@ -31,3 +31,7 @@ is as simple as running IntelliJ's auto-formatting command (<kbd>ctrl + alt + l<
 - js is bundled using webpack - this task can be triggered by running `npm run js`
 
 Or to compile both at once, `npm run build`
+
+### Management
+
+- `/actuator/health` to see running status of apps
```
EmmaLRussell, line 37: I think this might make more sense in the top level readme, with a mention of the httptrace endpoint too, and any others enabled by the spring config.

EmmaLRussell (commented):
This looks great, but I'm not sure what the wider plan is for getting this data into ELK. How is it shipped, and is there a ticket for that? This is using an in-memory store so the logs will currently disappear on redeploy?

The endpoint is (I think) currently supposed to be secured so that only authorized users can view it (though I seem to still be able to access it when I'm logged out). Are we happy for that to be an available undocumented endpoint, or are we going to lock it down in the proxy?

PR #772, opened 2022-11-14
Add google analytics to main page and data exploration

r-ash:
## Description

Add google analytics

This is a very basic impl, it will send info about users accessing main naomi site and the data exploration. These show up in the dashboard as
Screenshot_20221114_164232

I've given admin access to the dashboard to you @EmmaLRussell, @LekanAnni if you pm me your google account (or set one up for your imperial email) I will share with you too.

My plan with this is I will give Ian admin access and he can configure the reports how he wants. 

One potential issue is this will report back for all instances (prod, staging, preview, and dev). I have ignored localhost but all other systems will upload to google analytics, I think that is probably fine but let me know if you guys think otherwise.

## Type of version change

Patches

## Checklist

### DIFF
--- a/src/app/static/templates/data-exploration.ftl
+++ b/src/app/static/templates/data-exploration.ftl
@@ -6,12 +6,18 @@
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
     <!-- inject:css -->
     <!-- endinject -->
-    <script async src="https://www.googletagmanager.com/gtag/js?id=G-JPYDD02750"></script>
     <script>
-      window.dataLayer = window.dataLayer || [];
-      function gtag(){dataLayer.push(arguments);}
-      gtag('js', new Date());
-      gtag('config', 'G-JPYDD02750');
+    var host = window.location.hostname;
+    if(host != "localhost")
+    {
+        <script async src="https://www.googletagmanager.com/gtag/js?id=G-JPYDD02750"></script>
+        <script>
+        window.dataLayer = window.dataLayer || [];
+        function gtag(){dataLayer.push(arguments);}
+        gtag('js', new Date());
+        gtag('config', 'G-JPYDD02750');
+        </script>
+    }
     </script>
 </head>
 <body>
--- a/src/app/static/templates/data-exploration.ftl
+++ b/src/app/static/templates/data-exploration.ftl
@@ -6,6 +6,13 @@
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
     <!-- inject:css -->
     <!-- endinject -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=G-JPYDD02750"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', 'G-JPYDD02750');
+    </script>
 </head>
 <body>
 <div id="app" :class="language">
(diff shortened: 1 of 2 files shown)


LekanAnni:
> ## Description
> Add google analytics
> 
> This is a very basic impl, it will send info about users accessing main naomi site and the data exploration. These show up in the dashboard as Screenshot_20221114_164232
> 
> I've given admin access to the dashboard to you @EmmaLRussell, @LekanAnni if you pm me your google account (or set one up for your imperial email) I will share with you too.
> 
> My plan with this is I will give Ian admin access and he can configure the reports how he wants.
> 
> One potential issue is this will report back for all instances (prod, staging, preview, and dev). I have ignored localhost but all other systems will upload to google analytics, I think that is probably fine but let me know if you guys think otherwise.
> 
> ## Type of version change
> Patches
> 
> ## Checklist
> * [ ]  I have incremented version number, or version needs no increment
> * [ ]  The build passed successfully, or failed because of ADR tests

What's the path to the stats ?

LekanAnni (commented):
How do I see the stats ?

src/app/static/templates/data-exploration.ftl
```diff
@@ -6,6 +6,19 @@
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
     <!-- inject:css -->
     <!-- endinject -->
+    <script>
+    var host = window.location.hostname;
+    if(host != "localhost")
```
LekanAnni, line 11: Looks good to me, I guess we don't really want to show Ian the activities on preview, dev and etc. What is the implication of this tracking as regards GDPR ? Do we have to ask for users consent ?

LekanAnni (changes requested):
Looks good, just a few question.

src/app/static/templates/data-exploration.ftl
r-ash: Could just have this if `host == naomi.unaids.org` but I wonder if we can get that out of config somewhere?

GDPR wise, looks not - and looks like even if we asked for consent it still wouldn't be. Basing this off https://piwik.pro/blog/is-google-analytics-gdpr-compliant/#does-collecting-visitors%E2%80%99-consent-solve-compliance-issues-with-google-analytics?

r-ash: Let's leave this PR open for now and discuss with UNAIDS & science

EmmaLRussell (changes requested):
This actually isn't syntactically valid as it's just declaring script tags within javascript. You get this logged to console: `Uncaught SyntaxError: Unexpected token '<' (at (index):14:9)`

You *could* do a `document.write` to add the tag to the DOM but I think the better approach would be to use the template mechanism to include this scripts tags or not depending on the host, assuming that we can get this information (and provide it to the template) on the backend, 

It would also be nice to keep this DRY by putting it in a partial which both `data-exploration.ftl` and `index.ftl` can include. We don't use this anywhere in HINT just now, but we do in orderly-web, e.g. metadata-tab which is a partial template, and itself includes another partial, report-title.ftl:
https://github.com/vimc/orderly-web/blob/master/src/app/templates/partials/metadata-tab.ftl

We do also have template tests so you can unit test all this! :)
https://github.com/mrc-ide/hint/tree/master/src/app/src/test/kotlin/org/imperial/mrc/hint/unit/templates

EmmaLRussell:
Proof of concept branch for above suggested approach here: https://github.com/mrc-ide/hint/pull/775

r-ash:
See #775 for better impl

PR #774, opened 2022-11-16
Add delay to closing the dialog so link can be opened

r-ash:
## Description

Noticed the new News link doesn't work! :facepalm: this fixes it by adding a delay to dropdown closing so that the link can actually be opened

## Type of version change

Patches

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ app/static/coverage/**
 app/static/public/js/**
 app/static/public/css/**
 .DS_Store
+.vscode
--- a/src/app/static/src/app/components/header/HintrVersionMenu.vue
+++ b/src/app/static/src/app/components/header/HintrVersionMenu.vue
@@ -1,5 +1,5 @@
 <template>
-    <drop-down :text="`v${hintrVersions.naomi}`" :right="true" style="flex: none">
+    <drop-down :text="`v${hintrVersions.naomi}`" :right="true" :delay="true" style="flex: none">
         <a class="dropdown-item"
            href="https://naomi.unaids.org/news"
            target="_blank"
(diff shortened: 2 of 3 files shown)


EmmaLRussell (commented):
Works for me :+1:
Could add a unit test for this new attribute, like we already have for href:
https://github.com/mrc-ide/hint/blob/master/src/app/static/src/tests/components/header/hintrVersionMenu.test.ts#L50
The 1s delay in closing (defined in DropDown.vue) feels quite slow if you click away from the menu rather than opening up the link. Could probably drop that down,reduce that, though would need to check that whatever else uses that delay still works.

r-ash:
Thanks, yep I've added a test. I agree it is v v slow to close, I think my preference would be to get this merged and then address that in another ticket as atm master has a broken button. I bet this is also why the online support drop down doesn't close for mac users.

src/app/static/src/tests/components/header/hintrVersionMenu.test.ts
```diff
@@ -49,6 +59,7 @@ describe("Hintr Menu Version", () => {
         expectTranslated(link, "News", "Nouvelles", "Notícias", store);
         expect(link.attributes("href")).toBe("https://naomi.unaids.org/news");
         expect(link.attributes("target")).toBe("_blank");
+        expect(link.attributes("href")).toBe("https://naomi.unaids.org/news");
```
EmmaLRussell, line 62: Looks like you've added a duplicate line here?

EmmaLRussell (approved):

src/app/static/src/tests/components/header/hintrVersionMenu.test.ts
r-ash: Thanks! Removed that now

PR #777, opened 2022-11-22
mrc-3803 As a user I cannot start multiple model fits in the same session

LekanAnni:
## Description

This PR manually sets running state in modelRun so when a user runs a model button can be disables and also progress modal is shown without a delay.

## Type of version change
_Delete as appropriate_

## Checklist


EmmaLRussell (approved):
Works for me - in master branch I was consistently able to double-click the model run button and kick off two builds (with network throttling), but not in this branch.

PR #782, opened 2022-11-28
mrc-3842 Put save and load json behind feature switch

LekanAnni:
## Description

PR is self explanatory. The reason for placing load json behind feature switch is to avoid confusing users with load zip.

## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/app/featureSwitches.ts
```diff
@@ -2,9 +2,11 @@ const urlParams = new URLSearchParams(window.location.search);
 const modelBugReport = !urlParams.get('modelBugReport');
 const modelCalibratePlot = !urlParams.get('modelCalibratePlot');
 const comparisonOutput = !urlParams.get('comparisonOutput');
+const loadJson = !!urlParams.get('comparisonOutput');
```
EmmaLRussell, line 5:
```suggestion
const loadJson = !!urlParams.get('loadJson');
```

EmmaLRussell (changes requested):
Typo in the feature switch - it's using the same param as comparisonOutput just now. 

Could do with a test in `fileMenu.test.ts` that the JSON controls are not rendered when the switch is off. 

I don't think it's anything to do with this PR, but I'm sporadically seeing beheaviour where clicking on the File menu causes the whole page to refresh. It only happens sometimes though. Anyone else seen that?

EmmaLRussell (approved):
LGTM. Sorry, forgot to say last time, but could indent the nested controls inside the new span if you like.

PR #784, opened 2022-11-30
mrc-3843 - Improve comparison plot usability

LekanAnni:
## Description

PR uses default selection when indicator changes. If the same indicator is selected then default value will not be used.
## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -309,7 +312,37 @@
...
+                            xAxisId: selections.x_axis_id,
+                            disaggregateById: selections.disaggregate_by_id,
+                            selectedFilterOptions: selections.selected_filter_options
+                        }
+                        updateSelectionsAndXAxisOrder(
+                            data,
+                            comparisonSelections,
+                            this.comparisonPlotFlattenedXAxisFilterOptionIds,
+                            this.updateComparisonPlotSelections)
+                    }
```
EmmaLRussell, line 337: Is this really necessary? `updateSelectionsAndXAxisOrder` is about re-ordering selections to conform with those in the metadata. But you are already (I think) trusting default selections to be in the right order when the first arrive (for the default indicator) so can't we also trust the order when change indicator.

Also, we don't really want to retain `data` do we, other than determining that the indicator has changed? We always want to switch back to the default selections for the new indicator, regardless of what new filter values were selected by the user for the old indicator?

In which case, I think here you'd just need to do:
`this.updateComparisonPlotSelections({payload: selections})`

LekanAnni: You are right! I could just mutate default selections.

EmmaLRussell:
Seeing some odd behaviour with the Malawi dataset. Try the following steps:
- run through model fit and calibrate and open Comparison barchart tab. Things look good for the default indicator, 'HIV prevalence'
- change indicator to 'ANC prevalence age matched'.  No data available message shown 
image
.. but what's also strange here is that the selected age is 15+, whereas the default for this indicator in the response is "15-49":
```
{
            "indicator_id": "anc_prevalence_age_matched",
            "x_axis_id": "sex",
            "disaggregate_by_id": "source",
            "selected_filter_options": {
              "area": [
                {
                  "id": "MWI",
                  "label": "Malawi - Demo"
                }
              ],
              "quarter": [
                {
                  "id": "CY2016Q1",
                  "label": "March 2016"
                }
              ],
              "sex": [
                {
                  "id": "female",
                  "label": "Female"
                }
              ],
              "age": [
                {
                  "id": "Y015_049",
                  "label": "15-49"
                }
              ],
...
```
- if you go on to select 'ANC ART coverge age matched' the age filter updates to 15-49, and data is displayed
- if you then change back to 'ANC prevalence age matched' the age filter this time stays as 15-40, and this time data is displayed for the indicator

> This will need fixing in vueChart should we want to always set default each time indicator changes but for now this will do.

Sorry, @LekanAnni  I don't quite follow what you're saying here. Are you saying there's an issue because of the chart emitting further changed events as a result of us now updating the selections when indicator changes (and some of those having inconsistent payload formats)? But your check `if (data.indicatorId..` should guard against this? That shouldn't account for the odd behaviour described above?

EmmaLRussell:
Yeah, just been debugging this a bit, and, yes you're completely right, @LekanAnni  there is an issue with the component inappropriately emitting additional misleading updates because it thinks that it needs to reset its selections when the x-axis filter changes, emits the change, and overwrites the new selections provided by the updated selections. So your initial implementation in `vue-charts` I guess worked better in some cases because it did localise all the changes to the watch on isXAxisOrDisagg (but I think would have had issues if only the value was changing). I think the way to fix this is possibly to combine the two approaches so that we watch both, but have a unified update method which always refreshes selected. I'll quickly make a candidate branch to test if that actually works.

EmmaLRussell:
OK, I just tried it with this change to `vue-charts`, pushed to npm as 0.6.3, and it seems to me to work with this change (although I've broken a unit test in vue-charts in the process). Let's regroup on this in the morning though!
https://github.com/reside-ic/vue-charts/pull/43/files

EmmaLRussell:
..Sorry, another update! I realised I hadn't actually run `npm install` when I tried this before (the other projects I work on do this automatically so I tend to forget 🤦 ). So, scratch everything I said above, as this seems to me to work fine now. Sorry! 

I still think conceptually, the change I've made in vue-charts would makes sense for some cases, but this seems to work for me now for the Malawi dataset.

EmmaLRussell (approved):

r-ash:
I've deployed this to preview and I am not seeing the chart or table change when selecting a different indicator, not sure if some issue related to a dependency updated not being included in the build? Seen this behaviour with Malawi and Rwanda data
Peek 2022-12-07 21-43

EmmaLRussell:
Did you deploy the associated hintr branch to get the defaults metadata? I just tried to do that locally, but getting errors on importing ADR dataset. I think the hintr branch needs to be merged with master too, to get the latest changes around resource types?

r-ash:
>

r-ash:
> Did you deploy the associated hintr branch to get the defaults metadata? I just tried to do that locally, but getting errors on importing ADR dataset. I think the hintr branch needs to be merged with master too, to get the latest changes around resource types?

Ah probably not, will try this now! Thanks

r-ash:
This is looking great on dev! Thanks

PR #789, opened 2022-12-07
Add licence

r-ash:
## Description

This adds a LICENCE file with standard MIT licence text

## Type of version change

None

## Checklist


richfitz (approved):

r-ash:
> Nice, I think `MIT © Imperial College of Science, Technology and Medicine` would be sufficient since we have used the same in some projects.

Had a quick look at other LICENCE and LICENCE.md files in mrc-ide repo and I've updated this to match that format. I think we're using the full text in most places so let's go with that?

richfitz (approved):

LekanAnni:
@r-ash this PR has failing tests.

r-ash:
Thanks I think just some flakiness with failing test, should be good to go now

richfitz (approved):

LekanAnni (approved):
Nice 👍

PR #792, opened 2022-12-08
mrc-3872 Save serialized profile for OAuth2 users

EmmaLRussell:
## Description

This branch adds saving `serializedProfile` information to the database for new OAuth2 users - this is required so that sharing of projects works. (We need the profile to be saved so that pac4j recognises the user.) Possibly not all these fields are strictly required, but we know this format works!

## Type of version change
_Delete as appropriate_

Patches

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 2.30.1
+
+* Save profiles for new OAuth2 users
+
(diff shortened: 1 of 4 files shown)


r-ash (approved):
I've deployed this to dev and looks good to me!

PR #794, opened 2022-12-09
Mrc 3874 - Turn off rounding in display in the 'Review inputs' tab (only)

LekanAnni:
## Description

This PR turns off rounding in review input. in doing so, I have made a props to make rounding values optional.

## Type of version change
_Delete as appropriate_

## Checklist


EmmaLRussell (approved):
This works great for numbers, which are no longer rounded. Percentages are still rounded to 1dp - is this the requirement?

LekanAnni:
Jeff

> This works great for numbers, which are no longer rounded. Percentages are still rounded to 1dp - is this the requirement?

Percentages are fine as is according to Jeff.

PR #796, opened 2022-12-19

LekanAnni:
This is a reverted PR for Setting up Playwright and add Browser test for data exploration login


r-ash:
Fixed in #798

PR #798, opened 2023-01-03

r-ash:
## Description

This PR adds browser test setup and a basic test of data exploration page.

This follows on from https://github.com/mrc-ide/hint/pull/763 which we had to revert in https://github.com/mrc-ide/hint/commit/a9eba78f80cd6875c080de5e2fb502dde7625efe as it was failing on master.

I think this was failing because of a version mismatch between the chromium installed in docker container and the chromium which is found through npm test from the `package.json`. This fixes that by forcing the install of the chromium to be the same as that in the package.json.

## Type of version change

Patches

## Checklist


LekanAnni (approved):
Looks great, works well nice 👍 fix

PR #800, opened 2023-01-04
Mrc-3281 As a user I can see ADR source URL in output zip for input data so I know the source of the files

LekanAnni:
## Description

This PR adds resource_url to session file manager for onward transmission to hintr for output file. 

To test:

* set `USE_MOCK_MODEL` to false in `run-dependencies`
* Run app as normal without to download step
* download outputFile
* check `info/inputs.csv` file for resource_url

## Type of version change
_Delete as appropriate_

## Checklist


r-ash:
So on dev this did not work for me when selecting a specific release
Steps
1. Login as james@example.com
1. Create new project
1. Import Antarctica from Naomi Development Team org
1. Select a specific release "Antarctica v1"
1. Click import, get errors on every download

When I import the current release it pulls the files successfully and their URL in the output zip looks well formed but just redirects to the dashboard and does not download the file. I expect this is an issue in the ADR because the URL we build is the same format as available on the "Download" button in the ADR but that is not working atm. In dev or in prod. I've sent a question to ADR guys and asked why

src/app/src/main/kotlin/org/imperial/mrc/hint/models/AdrResource.kt
```diff
@@ -0,0 +1,3 @@
+package org.imperial.mrc.hint.models
+
+data class AdrResource(val url: String, val id: String? = "", val resourceId: String? = "")
```
EmmaLRussell, line 3: Do we need `url`? I don't think it actually get used, does it? Because you're constructing the resource url from the other bits, rather than using the more generic url that we have in the front end.

src/app/static/src/app/store/baseline/actions.ts
```diff
@@ -134,21 +133,36 @@ export const actions: ActionTree<BaselineState, DataExplorationState> & Baseline
...
     },
 
     async importPopulation(context, url) {
         if (url) {
-            await uploadOrImportPopulation(context, {url: "/adr/population/", payload: qs.stringify({url})},
+            const data = {
+                url,
+                id: context.state.selectedDataset?.id || "",
+                resourceId: context.state.selectedDataset?.resources?.pop?.id || ""
+            }
```
EmmaLRussell, line 152: It might be nice to pull the adr resource payload contruction out into its own method, as this logic is repeated per resource type.

EmmaLRussell (commented):
Code approach looks good to me, a few small comments, 👍 

The behaviour worked as described, and I can see the urls in the csv file in the output zip. However when I attempt to browse to one of these urls, I am redirected to https://dev.adr.fjelltopp.org/dashboard/ and no download starts. What should happen, and how are users supposed to make use of these urls? Might it be a quirk with the dev adr instance? Do we need to ask Fjellltopp about this?

src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
```diff
@@ -138,4 +149,26 @@ class LocalFileManager(
...
+    {
+        val adr = adrClientBuilder.build()
+
+        val response = adr.get("package_activity_list?id=${adrResource.id}")
+
+        val data = objectMapper.readTree(response.body)["data"]
+
+        if (data.isEmpty || adrResource.resourceId.isNullOrEmpty() || filename.isEmpty())
+        {
+            return ""
```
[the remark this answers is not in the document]
LekanAnni: I don't think this should stop the app from progressing, if we can't construct a reliable url then we show no url.

src/app/src/main/kotlin/org/imperial/mrc/hint/models/VersionFileWithPath.kt
```diff
@@ -1,11 +1,18 @@
 package org.imperial.mrc.hint.models
 
-data class VersionFile(val hash: String, val filename: String, val fromAdr: Boolean)
+@Suppress("ConstructorParameterNaming")
```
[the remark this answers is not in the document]
LekanAnni: I could not serialise the POJO without underscore into hintr expected payload so I left it without exploring other options, I have now added JsonProperty annotations.

src/app/src/main/kotlin/org/imperial/mrc/hint/models/AdrResource.kt
LekanAnni: It's used for accessing ADR endpoint, needed.

src/app/static/src/app/store/baseline/actions.ts
LekanAnni: I am not sure repetition can be avoided, resource, url are all unique.

LekanAnni:
> Code approach looks good to me, a few small comments, 👍
> 
> The behaviour worked as described, and I can see the urls in the csv file in the output zip. However when I attempt to browse to one of these urls, I am redirected to https://dev.adr.fjelltopp.org/dashboard/ and no download starts. What should happen, and how are users supposed to make use of these urls? Might it be a quirk with the dev adr instance? Do we need to ask Fjellltopp about this?

Tomek is looking into why the url gets redirected. I believe the url is correct.

src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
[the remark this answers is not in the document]
LekanAnni: The same endpoint affects releases as well. We currently can't fetch released datasets on master.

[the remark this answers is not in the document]
EmmaLRussell: > The same endpoint affects releases as well. We currently can't fetch released datasets on master.

Sorry, not sure what this comment means in relation to the previous one?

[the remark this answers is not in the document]
LekanAnni: It means pulling from released dataset will error as the endpoint is currently down.

also to note that it’s erroneous in production as well.

src/app/src/main/kotlin/org/imperial/mrc/hint/models/AdrResource.kt
EmmaLRussell: Ah yes, it's used to get the original filename.

src/app/static/src/app/store/baseline/actions.ts
EmmaLRussell: I just meant you could have something like
```
const buildData = (url: string, resourceType: string) => ({
    url,
    id: context.state.selectedDataset?.id || "",
    resourceId: context.state.selectedDataset?.resources?[resourceType]?.id || ""
});
...
...
const data = buildData(url, "pop");
```
which would save repetition of drilling down to get dataset id and resource ids

EmmaLRussell (approved):
Fine to merge this while we're waiting for confirmation from Tomek about why urls aren't functioning as expected. See above suggestion about possible helper function in action to build data objects.

src/app/static/src/app/store/baseline/actions.ts
EmmaLRussell: OK, this works for me:
```
const buildData = (context: ActionContext<BaselineState, DataExplorationState>, url: string, resourceType: DatasetResourceType) => ({
    url,
    id: context.state.selectedDataset?.id || "",
    resourceId: context.state.selectedDataset?.resources?.[resourceType]?.id || ""
});
```
if you declare the DatasetResourceType in types.ts:
`export type DatasetResourceType = "pjnz" | "pop"  | "program" | "anc" | "shape" | "survey";`

PR #802, opened 2023-01-12
mrc-3915: Show job ID instead of full error trace 

r-ash:
## Description

This PR will
* Replace error trace with job ID

At Ian and Jeff's request I've removed the stack trace being returned from hintr as we no longer need to show this to users now we can get it from the back end. We still want to return the job ID as we can use this to find the error in hintr, this PR updates from hintr schema changes in https://github.com/mrc-ide/hintr/pull/437 

## Type of version change
_Delete as appropriate_

Patches

## Checklist


src/app/static/src/app/components/ErrorAlert.vue
```diff
@@ -26,7 +19,6 @@
 
     interface Computed {
         message: string,
-        hasTrace: boolean,
         cssClass: string
```
LekanAnni, line 22: `cssClass` and `showTrace` and `toggleTrace` can be removed now.

LekanAnni (changes requested):
Looking good, 

Please see inline for minor suggestions. This isn't working as expected, you may need to add a browser test. Not working because that backend needs updating https://github.com/mrc-ide/hint/blob/706a49fd1b717719b4871fde893230ec2e763e8e/src/app/src/main/kotlin/org/imperial/mrc/hint/models/ErrorDetail.kt#L7

src/app/static/src/app/components/ErrorAlert.vue
r-ash: Thanks! Tidied that up now

LekanAnni (changes requested):
Thanks for removing those. 

You will also need to change the backend. Endpoint still returns trace and no job Id

```
{
    "errors": [
        {
            "detail": "An unexpected error occurred. If you see this message while you are using Naomi at a workshop, please contact your workshop technical support and show them this code: ubf-osz-hpl. Otherwise please contact support at naomi-support@imperial.ac.uk and quote this code: ubf-osz-hpl",
            "error": "OTHER_ERROR",
            "trace": [
                "No handler found for GET /actuator/health"
            ]
        }
    ],
    "data": {},
    "status": "failure"
}
```

r-ash:
> You will also need to change the backend. Endpoint still returns trace and no job Id

Is this the backend in hint? or in hintr? Hintr has been updated in https://github.com/mrc-ide/hintr/pull/437 which this branch is pinned against. If the backend of hint - can you point me towards where that needs updating?

LekanAnni:
> > You will also need to change the backend. Endpoint still returns trace and no job Id
> 
> Is this the backend in hint? or in hintr? Hintr has been updated in [mrc-ide/hintr#437](https://github.com/mrc-ide/hintr/pull/437) which this branch is pinned against. If the backend of hint - can you point me towards where that needs updating?

Hint backend. see link https://github.com/mrc-ide/hint/blob/706a49fd1b717719b4871fde893230ec2e763e8e/src/app/src/main/kotlin/org/imperial/mrc/hint/models/ErrorDetail.kt#L7

r-ash:
> Hint backend, let me know if I should jump in from here but see link
> 
> https://github.com/mrc-ide/hint/blob/706a49fd1b717719b4871fde893230ec2e763e8e/src/app/src/main/kotlin/org/imperial/mrc/hint/models/ErrorDetail.kt#L7

Thanks, I think this should be updated and ready for another look. Have made a ticket https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-3933/Add-error-trace-into-backend-logs for the things we discussed about logging the error trace from the backend

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt
```diff
@@ -106,19 +106,12 @@ class HintExceptionHandler(private val errorCodeGenerator: ErrorCodeGenerator,
         val messageArguments = arrayOf(
                 appProperties.applicationTitle,
                 errorCodeGenerator.newCode(),
-                appProperties.supportEmail
+                appProperties.supportEmail,
+                if (originalMessage != null) " $originalMessage" else ""
```
LekanAnni, line 110:
```suggestion
             originalMessage
```

LekanAnni: There's really no need for this check.

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/IntegrationTests.kt
```diff
@@ -246,39 +246,16 @@ abstract class SecureIntegrationTests : CleanDatabaseTests()
...
-                Assertions.assertThat(responseEntity.headers.contentType!!.toString()).isEqualTo("application/json")
-                Assertions.assertThat(responseEntity.statusCode).isEqualTo(httpStatus)
-                JSONValidator().validateError(responseEntity.body!!, errorCode, errorDetail, errorTrace)
-            }
-            IsAuthorized.FALSE ->
-            {
-                assertUnauthorized(responseEntity)
-            }
-        }
-    }
```
LekanAnni, line 270: This is redundant 👌

LekanAnni (changes requested):
Looking great, `trace` is completely removed from error rendering.

Just a minor code improvement.

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt
r-ash: Without this won't the formatter render this as `null` instead of an empty string? I tried this here https://pl.kotl.in/m_KcGN0mW

LekanAnni: > Without this won't the formatter render this as `null` instead of an empty string? I tried this here https://pl.kotl.in/m_KcGN0mW

```
private fun unexpectedError(status: HttpStatus,
                                request: HttpServletRequest,
                                originalMessage: String? = ""): ResponseEntity<Any>

instead of 
private fun unexpectedError(status: HttpStatus,
                                request: HttpServletRequest,
                                originalMessage: String? = null"): ResponseEntity<Any>
    {
    

```

r-ash: Ah yeah makes sense to default to empty string, but the `originalMessage` can still come in as `null` right? Exception type can have a `null` message. I don't think we can change to
```
originalMessage: String = ""
```
for example?

LekanAnni (approved):
LGTM 👍

PR #803, opened 2023-01-12
Mrc 3913 - As a user I can download input time series data

LekanAnni:
## Description

This PR added a generic download service to hint and implements time series plot data download. Download file contains two sheets, aggregated data and filtered data.

Coverage issue is not correct, line has been tested.

## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/app/store/plottingSelections/getters.ts
```diff
@@ -15,5 +17,12 @@ export const getters = {
             default:
                 return {}
         }
+    },
+
+    downloadFile: () => (data: DownloadPlotData): void => {
```
EmmaLRussell, line 22: Why is this a getter which returns a method, rather than just making this an action which the component dispatches? Download seems like a classic action!

Also, I'm not convinced 'plottingSelections' is the best module for this, as it doesn't have much to do with plotting selections, since all the data is provided in parameters anyway. Could be new module, or put in root actions?

EmmaLRussell (changes requested):
This looks great, works really nicely. 

As discussed in the meeting, let's take out 'page' from the filtered sheet, and rename 'aggregate data' to 'all data'. (Or did we decide to take out filtered completely in the end?)

The file name prefix ("chart-data") is a bit generic, and I think the full time stamp including  milliseconds is a bit verbose. Could we include the iso, as we often do for other files (like the output zip), e.g. have something like: naomi_MWI_data-review_20230119_112045.xlsx?

Button text could just be 'Download data' rather than 'Download plot data'? (The data is in the table too...) (Update: I think Jeff suggested 'Download indicator')

src/app/static/src/app/components/plots/download/downloadButton.vue
```diff
@@ -0,0 +1,40 @@
+<template>
```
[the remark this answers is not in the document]
LekanAnni: Oh yeah PascalCase of course, it was unintentional. Since we will implement download across the charts, I think it's safe to have a generic download button as well.

src/app/static/src/app/components/genericChart/GenericChart.vue
```diff
@@ -334,6 +337,8 @@
                     result[dataSourceId] = filterData(unfilteredData, filters, selectedFilterOptions);
                 }
 
+                this.addFilteredDataWithoutPageNumber(result)
```
EmmaLRussell, line 340: This feels a bit off as it's setting a data value from inside a computed where computeds ought to be derived from data/state. Might get unexpected effects sometimes. It might be better to make `filteredDataWithoutPages` a computed (which does all the above work), and then make `chartData` a further computed which takes `filteredDataWithoutPages` as a base, and adds the page numbers.

EmmaLRussell (commented):
This looks good to me and makes all the changes we discussed - just a bit uneasy about setting the 'without pages' data value in the component from inside a computed function. See comment below.

EmmaLRussell (approved):
Brilliant, thanks

PR #808, opened 2023-01-27
Mrc 3945 Fix project load issue

LekanAnni:
## Description

This PR adds test to the changes made in hintr regarding resourceUrl. Initial implementation does not allow null value, resource_url has must contain data. However, this causes validation issue when loading legacy projects. This is now fixed and tests are added to this effect.

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 2.32.2
+
+* fix project load
+
 # hint 2.32.1
 
 * As a user I can see ADR source URL in output zip for input data so I know the source of the files
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
@@ -75,7 +75,7 @@ class LocalFileManager(
         originalFilename: String,
         type: FileType,
         fromADR: Boolean,
-        resourceUrl: String? = "",
+        resourceUrl: String? = null,
     ): VersionFileWithPath
     {
         val md = MessageDigest.getInstance("MD5")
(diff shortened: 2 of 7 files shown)


EmmaLRussell (approved):
This works for me! I'm still not completely sure why this fixes it, since the methods had nullable parameters anyway. Did something else object to the default empty string, but was ok with nulls?

LekanAnni:
> This works for me! I'm still not completely sure why this fixes it, since the methods had nullable parameters anyway. Did something else object to the default empty string, but was ok with nulls?

The error comes from the schema validation service. `name : string? = ""` does not handle null but empty response. In fact we should not be using snippets such as `name : string? = "" but name : string? = null`.

EmmaLRussell:
> > This works for me! I'm still not completely sure why this fixes it, since the methods had nullable parameters anyway. Did something else object to the default empty string, but was ok with nulls?
> 
> The error comes from the schema validation service. `name : string? = ""` does not handle null but empty response. In fact we should not be using snippets such as `name : string? = "" but name : string? = null`.

Oh, when it's trying to create objects for that data class from the results from the db? (Not the hintr schema?)

LekanAnni:
> > > This works for me! I'm still not completely sure why this fixes it, since the methods had nullable parameters anyway. Did something else object to the default empty string, but was ok with nulls?
> > 
> > 
> > The error comes from the schema validation service. `name : string? = ""` does not handle null but empty response. In fact we should not be using snippets such as `name : string? = "" but name : string? = null`.
> 
> Oh, when it's trying to create objects for that data class from the results from the db? (Not the hintr schema?)

No, the issue was from the schema, I initially thought I would be able to handle it in frontend using `name : string? = ""` but I then realised this does not handle nullable.

PR #809, opened 2023-01-30
Fix encoding of missingPJNZ error message

r-ash:
## Description

Fixes the special characters in error message when PJNZ is missing

Previously looked like
image

Where the missing Spectrum error message was shown correctly
Screenshot_20230130_100941

This updates the encoding to UTF-8 and removes the helper functions which was converting from ISO-8859-1 to UTF-8 as since Java 9 the default for a properties file is UTF-8. I think this was left over from when we upgraded the Java version. See https://docs.oracle.com/javase/9/intl/internationalization-enhancements-jdk-9.htm#JSINT-GUID-974CF488-23E8-4963-A322-82006A7A14C7 section at the bottom

## Type of version change
_Delete as appropriate_

Patches

## Checklist


LekanAnni (approved):
LGTM nice tidying up 👍

LekanAnni (commented):
I think you need to update the NEWS item

EmmaLRussell (approved):
works for me :)

PR #814, opened 2023-02-08
mrc-3921 User doesn't have perms over a resource when the dataset refreshes

LekanAnni:
## Description

This PR implements Java 11 `HttpClient`, a faster and powerful http client. We can now handle requests for inputStreams nicely, it also supports blocking and non-blocking modes. I think we might consider speeding ADR requests by re-writing the way we currently make requests to ADR with `fuelClient` as this only supports blocking mode.

Error messages encountered while making requests to ADR can now be displayed in frontend instead of a generic hint error.

To test

1. upload from ADR
2. Go to ADR and edit selected dataset metadata
3. refresh dataset in Naomi
4. observe the error message.

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
@@ -12,7 +12,9 @@ import java.io.InputStream
 import java.security.DigestInputStream
 import java.security.MessageDigest
 import jakarta.xml.bind.DatatypeConverter
+import org.imperial.mrc.hint.exceptions.AdrException
 import org.imperial.mrc.hint.models.*
+import org.springframework.http.HttpStatus
 import org.springframework.web.util.UriComponentsBuilder
 
 enum class FileType
@@ -65,9 +67,14 @@ class LocalFileManager(
 
         val resourceUrl = getResourceUrl(data, originalFilename)
 
-        val inputStream = adr.getInputStream(data.url)
+        val response = adr.getInputStream(data.url)
 
-        return saveFile(inputStream, originalFilename, type, true, resourceUrl)
+        if (response.statusCode() != 200)
+        {
+            throw AdrException("adrResourceError", HttpStatus.valueOf(response.statusCode()), response.uri())
+        }
+
+        return saveFile(response.body(), originalFilename, type, true, resourceUrl)
     }
 
     private fun saveFile(
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
@@ -75,16 +75,16 @@ class LocalFileManager(
              * When a user is unauthenticated or lack required permission, the user gets redirected
              * to auth0 login page. handleAdrException handles permission and unexpected ADR errors
              */
-            if (response.uri().host.contains("eu.auth0.com"))
+            if (response.statusCode() == 302)
             {
                 throw AdrException(
                     "noPermissionToAccessResource",
                     HttpStatus.valueOf(response.statusCode()),
-                    data.url
+                    arrayOf(data.url)
                 )
             }
 
-            throw AdrException("adrResourceError", HttpStatus.valueOf(response.statusCode()), data.url)
+            throw AdrException("adrResourceError", HttpStatus.valueOf(response.statusCode()), arrayOf(data.url))
         }
 
         return saveFile(response.body(), originalFilename, type, true, resourceUrl)
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
@@ -69,13 +69,13 @@ class LocalFileManager(
 
         val response = adr.getInputStream(data.url)
 
-        if (response.statusCode() != 200)
+        if (response.statusCode() != HttpStatus.OK.value())
         {
             /**
              * When a user is unauthenticated or lack required permission, the user gets redirected
              * to auth0 login page. handleAdrException handles permission and unexpected ADR errors
              */
-            if (response.statusCode() == 302)
+            if (response.statusCode() == HttpStatus.TEMPORARY_REDIRECT.value())
             {
                 throw AdrException(
                     "noPermissionToAccessResource",
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
@@ -71,7 +71,20 @@ class LocalFileManager(
 
         if (response.statusCode() != 200)
         {
-            throw AdrException("adrResourceError", HttpStatus.valueOf(response.statusCode()), response.uri())
+            /**
+             * When a user is unauthenticated or lack required permission, the user gets redirected
+             * to auth0 login page. handleAdrException handles permission and unexpected ADR errors
+             */
+            if (response.uri().host.contains("eu.auth0.com"))
+            {
+                throw AdrException(
+                    "noPermissionToAccessResource",
+                    HttpStatus.valueOf(response.statusCode()),
+                    data.url
+                )
+            }
+
+            throw AdrException("adrResourceError", HttpStatus.valueOf(response.statusCode()), data.url)
         }
 
         return saveFile(response.body(), originalFilename, type, true, resourceUrl)
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
@@ -75,7 +75,7 @@ class LocalFileManager(
              * When a user is unauthenticated or lack required permission, the user gets redirected
              * to auth0 login page. handleAdrException handles permission and unexpected ADR errors
              */
-            if (response.statusCode() == HttpStatus.TEMPORARY_REDIRECT.value())
+            if (response.statusCode() == HttpStatus.FOUND.value())
             {
                 throw AdrException(
                     "noPermissionToAccessResource",
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
@@ -80,11 +80,11 @@ class LocalFileManager(
                 throw AdrException(
                     "noPermissionToAccessResource",
                     HttpStatus.valueOf(response.statusCode()),
-                    arrayOf(data.url)
+                    data.url
                 )
             }
 
-            throw AdrException("adrResourceError", HttpStatus.valueOf(response.statusCode()), arrayOf(data.url))
+            throw AdrException("adrResourceError", HttpStatus.valueOf(response.statusCode()), data.url)
         }
 
         return saveFile(response.body(), originalFilename, type, true, resourceUrl)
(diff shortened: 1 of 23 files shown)


src/app/src/main/resources/application.properties
```diff
@@ -14,4 +14,5 @@ server.compression.mime-types=text/html,text/xml,text/plain,text/css,text/javasc
 # Compress the response only if the response size is at least 1KB
 server.compression.min-response-size=1024
 server.servlet.session.timeout=120m
-logging.config=classpath:logback-spring.xml
\ No newline at end of file
+logging.config=classpath:logback-spring.xml
+spring.messages.encoding=UTF-8
```
LekanAnni, line 18: This is an opportunistic fix, by default *.property files are not UTF-8 encoded, adding this line ensures internationalisation is encoded in UTF-8.

src/app/static/src/tests/integration/adr-dataset.itest.ts
```diff
@@ -206,7 +206,7 @@ describe("ADR dataset-related actions", () => {
         expect(commit.mock.calls[1][0]["type"]).toBe(BaselineMutation.PopulationUpdated);
         expect(commit.mock.calls[1][0]["payload"]["filename"])
             .toBe("population.csv");
-    }, 7000);
+    }, 10000);
```
LekanAnni, line 209: Increased to 10k to avoid intermittent failure. I think ADR server has become slower.

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintExceptionHandler.kt
```diff
@@ -75,12 +75,37 @@ class HintExceptionHandler(private val errorCodeGenerator: ErrorCodeGenerator,
         return translatedError(error.key, error.httpStatus, request)
     }
 
+    @ExceptionHandler(AdrException::class)
+    fun handleAdrException(error: AdrException, request: HttpServletRequest): ResponseEntity<Any>
+    {
+        logger.error(request, error)
+
+        val messageArguments = arrayOf(error.adrUri ?: "")
```
r-ash, line 83: Pretty minor but can we make the error text link to the resource instead of just showing the URL?

Like
Unable to load resource, [check resource in ADR](https://adr.unaids.org/country-estimates-23/antarctica-country-estimates-2023-1/resource/ec4f40f7-601e-4f1c-bcb2-4643da4cc4b5).

src/app/src/main/kotlin/org/imperial/mrc/hint/clients/FuelClient.kt
```diff
@@ -87,4 +93,30 @@ abstract class FuelClient(protected val baseUrl: String)
                 .second
                 .asResponseEntity()
     }
+
+    private fun getRequest(uri: URI): HttpRequest
+    {
+        val entry = standardHeaders().entries.first()
```
r-ash, line 99: Do you not want to include all headers here?

src/app/src/main/kotlin/org/imperial/mrc/hint/clients/FuelClient.kt
```diff
@@ -87,4 +93,30 @@ abstract class FuelClient(protected val baseUrl: String)
...
+    }
+
+    private fun clientBuilder(): Builder
+    {
+        return HttpClient
+            .newBuilder()
+            .followRedirects(HttpClient.Redirect.ALWAYS)
+    }
+
+    protected fun getFile(uri: URI): HttpResponse<InputStream>
```
r-ash, line 116: Is this always a file? Looks like this function is more generic at this level?

src/app/src/main/kotlin/org/imperial/mrc/hint/FileManager.kt
```diff
@@ -65,9 +67,27 @@ class LocalFileManager(
...
+        val response = adr.getInputStream(data.url)
 
-        return saveFile(inputStream, originalFilename, type, true, resourceUrl)
+        if (response.statusCode() != 200)
+        {
+            /**
+             * When a user is unauthenticated or lack required permission, the user gets redirected
+             * to auth0 login page. handleAdrException handles permission and unexpected ADR errors
+             */
+            if (response.uri().host.contains("eu.auth0.com"))
```
EmmaLRussell, line 78: Is this really the only way we can tell if we've had an auth issue? Hardcoding the auth0 host seems really grimy! Could we at least just check if it's a redirect? Do we get a 302 or something here? ADR can't return 4xx?

src/app/src/main/kotlin/org/imperial/mrc/hint/clients/FuelClient.kt
EmmaLRussell: Yeah, what's going on here?

src/app/src/main/kotlin/org/imperial/mrc/hint/clients/FuelClient.kt
EmmaLRussell: It's pulling a stream, so I guess it could be `getStream`..?
So as far as I can tell, this is replacing the old `java.net` implementation with the `java.net.http` equivalent, and moving it from the ADRFuelClient to FuelClient class to make it more generic. Makes sense!

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/HintException.kt
```diff
@@ -1,6 +1,10 @@
...
 
 import org.springframework.http.HttpStatus
+import java.net.URI
 
-open class HintException(val key: String, val httpStatus: HttpStatus = HttpStatus.INTERNAL_SERVER_ERROR) :
-        Exception("HintException with key $key")
+open class HintException(
+        val key: String,
+        val httpStatus: HttpStatus = HttpStatus.INTERNAL_SERVER_ERROR,
+        val adrUri: String? = null
```
EmmaLRussell, line 9: Could we avoid having error-specific information like `adrUrl` in the base exception class? If we need to have some additional info beyond key and httpStatus, could we handle those generically, eg. additionalArgs array? ...But anyway, it looks to me like the only place where adrUri is actually used, is in `HintExceptionHandler.handleAdrException` where the parameter is explicitly an `AdrException` anyway - so could we exclude it from the base class?

src/app/src/main/resources/ErrorMessageBundle_fr.properties
```diff
@@ -12,3 +12,6 @@ unexpectedError=Une erreur inattendue s''est produite. \
 userDoesNotExist=L'utilisateur n'existe pas.
 userExists=L'utilisateur existe déjà.
 unknownInputTimeSeriesType=Type de série temporelle d'entrée inconnu
+noPermissionToAccessResource=Vous nêtes pas autorisé à charger cette ressource à partir dADR. Contactez ladministrateur \
```
EmmaLRussell, line 15: Some missing apostrophes here I think
```suggestion
noPermissionToAccessResource=Vous n'êtes pas autorisé à charger cette ressource à partir d'ADR. Contactez l'administrateur \
```

EmmaLRussell (commented):
Works really nicely, everything working end to end!

The refactor to the http client seems really useful, but was that actually stopping us from showing specific error messages for ADR permission issues? That could just have been address by a change to the error handling, right? Or was the issue with the non-blocking capacity of the old client?

When you refresh from ADR and get one of these new errors, 'Continue' is not enabled. You can enable it by clicking 'remove' on the file - but I'm not sure this will be obvious to users, since it hasn't actually loaded the file at all, so I don't think it's going to be clear that one needs to remove it. 
image
I guess this situation is not really different to one where a file failed to load because it was invalid - but it feels like 'remove' is even less obvious here.

LekanAnni:
> The refactor to the http client seems really useful, but was that actually stopping us from showing specific error messages for ADR permission issues? That could just have been address by a change to the error handling, right? Or was the issue with the non-blocking capacity of the old client?

The old urlConnection could not handle errors, it was initially throwing IOException when permission is not limited. I could have handled this error without changing the client but I would not have able to determine permission issues error. I think upgrading the client is also a necessity as the `UrlConnection` is very old, we can now take advantage of the tools in the new client.


I really do not think continue should be enabled. The issue should be actioned by the user, either contacting ADR or removing the file.

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/AdrException.kt
```diff
@@ -0,0 +1,9 @@
+package org.imperial.mrc.hint.exceptions
+
+import org.springframework.http.HttpStatus
+
+class AdrException(
+    key: String,
+    status: HttpStatus,
+    url: Array<String> = arrayOf("")
```
LekanAnni, line 8: codecov 👎  line is obviously tested.

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/AdrException.kt
```diff
@@ -2,4 +2,8 @@ package org.imperial.mrc.hint.exceptions
 
 import org.springframework.http.HttpStatus
 
-class AdrException(key: String, status: HttpStatus, url: String) : HintException(key, status, url)
+class AdrException(
+    key: String,
+    status: HttpStatus,
+    url: Array<String> = arrayOf("")
+) : HintException(key, status, url)
```
EmmaLRussell, line 9:
```suggestion
class AdrException(
    key: String,
    status: HttpStatus,
    url: String
) : HintException(key, status, arrayOf(url))
```
No need for url to be an array I don't think. And does it require a default? Shouldn't url always be provided?

src/app/src/main/kotlin/org/imperial/mrc/hint/exceptions/AdrException.kt
EmmaLRussell: Is it because none of the tests test the default url parameter - they all provide it in some form I think, via `expectTranslatedAdrException`? But I don't think you should need the default anyway..?

EmmaLRussell (commented):
LGTM, couple of small comments on the exception class.

EmmaLRussell (approved):

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ExceptionHandlerTests.kt
```diff
@@ -58,6 +58,78 @@ class ExceptionHandlerTests : SecureIntegrationTests()
...
+        )
+    }
+
+    @Test
+    fun `adr no permission to load resource error are correctly formatted in French`()
+    {
+        expectTranslatedAdrException(
+            "noPermissionToAccessResource",
+            "Vous nêtes pas autorisé à charger cette ressource à partir dADR. " +
+                    "Contacter ladministrateur de lensemble de données pour obtenir lautorisation",
```
r-ash, line 117: This looks like it has lost `'` punctuation? .e.g `d'ADR`, `l'administrateur`. `l'ensemble' etc. I think for multiline strings in the ErrorMessageBundle.properties these need to be escaped with another `'` e.g. `l''ensemble'`. `unexpectedError` has an example

r-ash (changes requested):

src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ExceptionHandlerTests.kt
LekanAnni: Nice spot. I was unsure why I wasn't getting the punctuations in the test.

PR #815, opened 2023-02-09
Mrc-3974 Naomi issue: model options reset if data set updated

LekanAnni:
## Description

This PR invalidate but retains modelOptions when resources get retrieved. However, I have subscribed to clear model options only when dataset or released gets updated. This is to avoid model options getting corrupted. 

## Type of version change
_Delete as appropriate_

## Checklist


r-ash (changes requested):
This looks good, a couple of things I thought about.

1. Can we add an integration test to cover when there is an option in the state but the data is changed so that the value in the state is no longer one of the possible options?
2. We should update the text in the reset dialog here "Changing this will result in the following steps being discarded:" is no longer true. Perhaps not appropriate to show this dialog anymore at all when only the model options have been set. Probably worth checking the desired behaviour with Jeff.
reset_dialog
3. When an option no longer exists in the data it says "(unknown)" next to the value. This is fairly subtly and indeed when you validate the options in this case it says they are valid. Can we have the validity check fail if there are any options selected which are unknown?
image

src/app/static/src/app/store/root/mutations.ts
```diff
@@ -135,6 +136,13 @@ export const mutations: MutationTree<RootState> = {
         Object.assign(state.modelOptions, initialModelOptionsState());
     },
 
+    [RootMutation.InvalidateOptions](state: RootState) {
```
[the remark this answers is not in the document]
LekanAnni: We should indeed clear warnings whenever we invalidate options. Warnings usually get retrieved upon validation.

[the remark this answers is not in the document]
LekanAnni: As discussed, validation for options should be handled in `hintr`,  this is to ensure consistency with running in CLI i.e by Jeff and on web.

src/app/static/src/app/store/baseline/mutations.ts
```diff
@@ -38,6 +38,11 @@ export const BaselineUpdates = [
     BaselineMutation.PopulationUpdated
 ];
 
+export const BaselineDatasetUpdates = [
+    BaselineMutation.SetDataset,
+    BaselineMutation.SetRelease
```
EmmaLRussell, line 43: So options will be cleared when either of these are updated? Don't we want to retain them on `SetRelease`? That's equivalent to loading a 'different version of the dataset' . Or are we only expecting to retain options when we do a refresh when the latest version changes?

EmmaLRussell (commented):
Been out of the loop on this one, so apologies if I'm completely misinterpreting what's required:

It took me a while to understand that we're only retaining options here if the dataset is being refreshed - if you're pulling a new release, a new dataset or adding/removing input files it makes no attempt to retain options. Is that actually what Jeff is requesting? I'd assumed a new release should retain options, and it would probably be preferable too if simply adding a new file.

That would require some logic to retain old values and use defaults for old ones. However we do already have that logic, and we use it to update options with new versions on step load. Why don't we just use that to enable us to retain any relevant option values?

It also took me a while to understand why the logic of what steps we're warning for in the dialog is now partially embedded in the components rather than located in the getter `changesToRelevantSteps` - it's because only changes via some components result in warnings being required for options, and those are the ones which do not include a 'discardStepWarning` value, right? The fact that those components opting out of issuing the warning happens to coincide with the mutations wiping options (or not) is really quite opaque - I think it's going to make this area the code potentially even more difficult to understand and maintain. 

So I think there should at least be some considerable commenting linking these two parts of the code..

But I'd also suggest considering the approach of attempting to retain options in all cases - that may be useful in a wider range of circumstances, which may actually be what Jeff would prefer. And I think that would be much simpler since we already have the logic for safely updating options - so it should largely just be a case of updating `changesToRelevantSteps` to not include the Model Options step. 

Does that make sense?

LekanAnni:
> Been out of the loop on this one, so apologies if I'm completely misinterpreting what's required:
> 
> It took me a while to understand that we're only retaining options here if the dataset is being refreshed - if you're pulling a new release, a new dataset or adding/removing input files it makes no attempt to retain options. Is that actually what Jeff is requesting? I'd assumed a new release should retain options, and it would probably be preferable too if simply adding a new file.
> 
> That would require some logic to retain old values and use defaults for new ones. However we do already have that logic, and we use it to update options with new versions on step load. Why don't we just use that to enable us to retain any relevant option values?
> 
> It also took me a while to understand why the logic of what steps we're warning for in the dialog is now partially embedded in the components rather than located in the getter `changesToRelevantSteps` - it's because only changes via some components result in warnings being required for options, and those are the ones which do not include a 'discardStepWarning` value, right? The fact that those components opting out of issuing the warning happens to coincide with the mutations wiping options (or not) is really quite opaque - I think it's going to make this area the code potentially even more difficult to understand and maintain.
> 
> So I think there should at least be some considerable commenting linking these two parts of the code..
> 
> But I'd also suggest considering the approach of attempting to retain options in all cases - that may be useful in a wider range of circumstances, which may actually be what Jeff would prefer. And I think that would be much simpler since we already have the logic for safely updating options - so it should largely just be a case of updating `changesToRelevantSteps` to not include the Model Options step.
> 
> Does that make sense?

I don't think you fully grasp the requirement. Lets have a chat.

r-ash:
> As mentioned above, validation should be handled in the hintr hence an integration test.

Can you create a follow on ticket for this? It will be slow to rebuild the possible options in the validation so we should have the validation send the full form data to hintr. Hintr can then check that the control value exists in the possible control options.

> It took me a while to understand that we're only retaining options here if the dataset is being refreshed

As I understand it the use case we are trying to address here is users review their data or try to fit the model but notice an issue in 1 or more of their input files. They fix the issue in excel and then go to step 1, remove the offending file and upload the new version using the file browser. Or they update the version on the ADR and pull it using the "Refresh". We want to retain options in both cases.

src/app/static/src/app/components/resetConfirmation/ResetConfirmation.vue
```diff
@@ -104,7 +110,13 @@
             isGuest: mapGetterByName(null, "isGuest"),
             errorsCount: mapStateProp<ErrorsState, number>("errors", state => {
                 return state.errors ? state.errors.length : 0;
-            })
+            }),
+            showRelevantSteps() {
```
EmmaLRussell, line 114:
```suggestion
            showRelevantSteps() {
            // In special cases when data for a downstream step can be retained, remove the warning for that step
```

src/app/static/src/app/root.ts
```diff
@@ -104,16 +104,21 @@ const resetState = (store: Store<RootState>): void => {
...
-                store.commit(RootMutation.ResetSelectedDataType);
-                store.commit(RootMutation.ResetOptions);
-                store.commit(RootMutation.ResetOutputs);
+            if (type[0] == "baseline") {
+                if (BaselineDatasetUpdates.includes(type[1] as BaselineMutation)) {
+                    store.commit(RootMutation.ResetOptions);
+                }
+
+                if (BaselineUpdates.includes(type[1] as BaselineMutation)) {
+                    store.commit(RootMutation.ResetSelectedDataType);
```
EmmaLRussell, line 113:
```suggestion
                    // If data has been updated without changing the selected dataset, retain model options, but reset their valid status to false to force revalidation
                    store.commit(RootMutation.ResetSelectedDataType);
```

EmmaLRussell (commented):
> I don't think you fully grasp the requirement. Lets have a chat.

Lol, thanks. Chatted with Rob about this ticket on Friday, and we agreed that indeed it wouldn't be appropriate to retain model options when selected dataset changes. However model options should be retained when the user selects a different release / latest version of the same dataset from the Browse ADR dialog. 

> As discussed, validation for (unknown) options should be handled in hintr, this is to ensure consistency with running in CLI i.e by Jeff and on web.

We could handle this in HINT by automatically removing any non-existent values when options are reloaded - it currently removes no longer existent sections and controls, so it does seem like an omission that it doesn't check selectable option values too.  This would be an edit to the [`updateForm` util](https://github.com/mrc-ide/hint/blob/0ff4c4d70e15837a125b045b0059f47defc1c42b/src/app/static/src/app/utils.ts#L289). This means that it would silently remove these options on page load - I believe it currently does that anyway for sections/controls - so would this be desirable behaviour for  selectable options too @r-ash ?

> Lekan: As discussed, validation for (unknown) options should be handled in hintr, this is to ensure consistency with running in CLI i.e by Jeff and on web.
> Rob:  It will be slow to rebuild the possible options in the validation so we should have the validation send the full form data to hintr. Hintr can then check that the control value exists in the possible control options.

We could send the full form metadata back for hintr to validate, but I guess then there wouldn't be consistency with CLI usage since presumably the CLI wouldn't be able to do that? I think if we're just trying to avoid this issue from HINT then removing the options as we already do the fields should be enough.

This isn't just an issue for this ticket - it could already happen if a new version of the options removed some option values. (Although this ticket makes it more likely.)

A couple of suggested comments on the code below to make like a little easier for future maintainers.

LekanAnni:
> We could handle this in HINT by automatically removing any non-existent values when options are reloaded - it currently removes no longer existent sections and controls, so it does seem like an omission that it doesn't check selectable option values too. This would be an edit to the [`updateForm` util](https://github.com/mrc-ide/hint/blob/0ff4c4d70e15837a125b045b0059f47defc1c42b/src/app/static/src/app/utils.ts#L289). This means that it would silently remove these options on page load - I believe it currently does that anyway for sections/controls - so would this be desirable behaviour for selectable options too @r-ash ?

I have added a ticket for the validation here https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4014/Validate-model-options. We can do that in a different ticket.

EmmaLRussell (commented):
I can see why you've added the new flag to the baseline state, but it would be nice to avoid needing that, since it's just another value to maintain...

Possible alternative: since the `baseline/SetDataset` mutation is, I think, only called from the adr action (`getDataset`)[https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/adr/actions.ts#L70], which also calls `SetRelease`, how about 
1. only committing `SetDataset` when the dataset is actually different from the current dataset
2. include `SetRelease` in `BaselineUpdates` so that options are unvalidated (I think we do want that when a release is changed, same as when the Refresh button is used). 

 If this doesn't work for some reason, please could you change the name of `selectedDatasetIsRefreshed` to e.g. `selectedDatasetHasChanged`? I assumed the former meant that the dataset data had been refreshed, but the dataset itself was the same dataset - turned out it meant the opposite!

LekanAnni:
> Possible alternative: since the `baseline/SetDataset` mutation is, I think, only called from the adr action [`getDataset`](https://github.com/mrc-ide/hint/blob/master/src/app/static/src/app/store/adr/actions.ts#L70), which also calls `SetRelease`, how about

Plugin listens wether `BaselineMutation.SetDataset` is called and if so it triggers  necessary action. Your suggestion will not work. I have renamed the prop.

src/app/static/src/app/hintVersion.ts
```diff
@@ -1 +1 @@
-export const currentHintVersion = "2.33.0";
+export const currentHintVersion = "2.33.0";
```
r-ash, line 1:
```suggestion
export const currentHintVersion = "2.33.1";
```

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # hint 2.33.0
 
+* Retain model options when selected dataset is updated data is updated
+
```
r-ash, line 4:
```suggestion
# hint 2.33.1

* Retain model options when selected dataset is updated data is updated

# hint 2.33.0

```

r-ash (approved):
Looks good, left a comment to tidy up version numbers

EmmaLRussell (approved):
Approving since you need to get it deployed, though still not convinced you really need that flag!

EmmaLRussell:
Sorry @LekanAnni , you're absolutely right, I completely failed to note which mutation did the actual data setting. :facepalm: Sorry, brain must have been running on empty last week. 

I can still think of three alternatives which I think would work, but all of them have drawbacks. I would probably still prefer the first one at least to keeping a flag in the store for the ephemeral value of the dataset having changed, but it's getting pretty borderline. In order of preference:

1. Replace `SetDataset` with separate mutations to be called for a refresh of the same dataset vs switching to a new dataset, e.g. `UpdateDataset` and `SwitchDataset`. They would each do exactly the same thing (same code as current `SetDataset`), but in the plugin `SwitchDataset` would trigger resetting the model options, but `UpdateDataset` would just unvalidate them.
Drawback: seems weird having two mutations which do exactly the same thing.

2. Keep `SetDataset` as it is but make a new mutation, e.g. `SwitchDataset` which doesn't actually do anything but triggers the model options reset in the plugin. 
Drawback: seems *really* weird having a mutation which doesn't mutate anything...

3. Let the baseline `getDataset` action detect that dataset id has changed and commit the necessary reset root mutation(s) itself, outside of the plugin mechanism.
Drawback: it would be confusing (and muddling responsibilities) to not have all the resets localised in the plugin any more - although there is some precendent for it, since we already do resets in the `root/validate` action.  

In all cases, an explicatory comment would be required. 
Up to you!

PR #821, opened 2023-03-23
Example define component

M-Kusumgar:
The purpose of this PR is to show an example of the process of changing Vue.extends to defineComponent. The first commit is just for history purposes, you can safely ignore most of it but just have a skim through (most of the stuff from the first commit is changed later on). The second commit "example of define component" has a comment that outlines the process.

### DIFF
--- a/src/app/static/src/components/adr/SelectRelease.vue
+++ b/src/app/static/src/components/adr/SelectRelease.vue
@@ -1,5 +1,5 @@
 <template>
-    <div id="selectRelease" v-if="datasetId && releases.length">
+    <div id="selectRelease" v-if="datasetId && (releases as Release[]).length">
         <div class="pt-2">
             <input
                 type="radio"
@@ -57,7 +57,6 @@
 
 <script lang="ts">
     import {defineComponent} from "vue";
-    import type { PropType } from "vue";
     import { mapActionByName, mapStateProp, mapMutationByName } from "../../utils";
     import TreeSelect from "vue3-treeselect";
     import { ADRState } from "../../store/adr/adr";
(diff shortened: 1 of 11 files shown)


src/app/static/src/components/adr/SelectRelease.vue
```diff
@@ -125,7 +123,7 @@
                 return (this.choiceADR === "useLatest") || !!this.releaseId;
             },
             releaseOptions() {
-                return this.releases.map((d) => ({
+                return (this.releases as Release[]).map((d) => ({
```
hillalex, line 126: it's really odd that you've had to cast this everywhere, suggests the type qualifiers on line 102 aren't working as expected?

src/app/static/src/components/downloadIndicator/DownloadIndicator.vue
```diff
@@ -33,14 +34,14 @@
         downloadingIndicator: boolean
     }
 
-    export default defineComponent<unknown, Methods, Computed, Props>({
```
hillalex, line 36: why have these type qualifiers been removed, whereas in other components they remain?

src/app/static/src/components/downloadResults/DownloadResults.vue
```diff
@@ -187,9 +187,24 @@
...
+            downloadUrl(downloadId: string) {
+                return `/download/result/${downloadId}`;
+            },
             handleUploadModal() {
                 this.uploadModalOpen = true;
             },
+            handleDownloadResult(downloadResults: DownloadResultsDependency) {
+                window.location.assign(this.downloadUrl(downloadResults.downloadId));
+            },
+            downloadSpectrumOutput() {
```
hillalex, line 199: Confused as to what's going on here - these are duplicating methods mapped below via `mapActionByName`

PR #822, opened 2023-03-28
Vue3 Upgrade Redo Base Epic PR

M-Kusumgar:
I realised it was quite difficult to keep track of changes with my initial code and there were a lot of random changes to the config files (basically due to me copying and pasting from a fresh Vue 3 install and trying to get things to work at the start of this project). This will now be a base PR and I will create other PRs to merge into this one. This particular one just has package changes currently!

Merged PRs:
- None for now

Active PRs:
- #823
- #825 
- #829 

Upcoming PRs (not an exhaustive list and not in order):
- defineComponentVue2 in all vue components
- updating reside's Vue charts package
- updating reside's Vue dynamic form package
- recreating required bootstrap vue components using vanilla bootstrap

Proof of Concept:
#831

PR #826, opened 2023-03-28
Define Component Vue 2 proxy

M-Kusumgar:
Sub branch of #822 

Typescript was throwing a lot of errors about the Computed, Methods and Props interfaces that are used throughout the frontend. The types in defineComponent (Vue3) have since moved on to ComputedOptions, MethodOptions and PropOptions which differ in a subtle way (more details in the comments in the file). Our interfaces and the new types are not compatible anymore so this file adjusts our interfaces to the Vue3 compatible types, which was the cause of the majority of the errors in the Vue3 upgrade because if typescript disagrees with the types then it also doesnt know what `this` refers to (e.g. doesnt think `this.variable` exists).

This is completely to do with types and our interfaces so if we were in JavaScript for example then we could just replace Vue.extend to defineComponent and everything would work just fine because the actual javascript is compatible with Vue2 component options.

This is the big idea that hopefully will speed up this upgrade by quite a lot! Please read the comments in the file! They have explanations of everything going on but if you have any questions please do ask!

### DIFF
--- a/src/app/static/vue2vue3/defineComponentProxy.ts
+++ b/src/app/static/vue2vue3/defineComponentProxy.ts
@@ -77,7 +77,21 @@ type ComponentObjectPropsOptions<P> = {
 //    and prop objects but since our Props type that we passed into Vue.extend
 //    already had these I just pass them into this component to bypass the
 //    ExtractPropType function that vue was using.
-type ComponentOptionsWithObjectProps<PropsOptions = ComponentObjectPropsOptions<Data>, RawBindings = {}, D = {}, C extends ComputedOptions = {}, M extends MethodOptions = {}, Mixin extends ComponentOptionsMixin = ComponentOptionsMixin, Extends extends ComponentOptionsMixin = ComponentOptionsMixin, E extends EmitsOptions = EmitsOptions, EE extends string = string, I extends ComponentInjectOptions = {}, II extends string = string, BaseProps = {}, Props = BaseProps & EmitsToProps<E>, Defaults = ExtractDefaultPropTypes<PropsOptions>> = ComponentOptionsBase<Props, RawBindings, D, C, M, Mixin, Extends, E, EE, Defaults, I, II> & {
+type ComponentOptionsWithObjectProps<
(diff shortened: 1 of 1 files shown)


EmmaLRussell (commented):
Following our conversation - about whether you really need to take this approach, or if the standard way of using TypeScript with Vue3 Options API ought to work (as described [here](https://vuejs.org/guide/typescript/options-api.html))...
...I tried taking that approach just on Choropleth - which is about the most complicated component we have - and I did seem to be able to eliminate all the relevant TypeScript errors.

The errors I was initially seeing were all about props being potentially undefined (and those are possibly valid errors... I think we have fudged a bit by assuming that props will be available, whereas really we need to wait for them to be fetched.. that might be handled further up the component parentage - can't remember!) Anyway, I was able to get rid of those errors by firstly specifying that the props were required, and secondly defining the `props` inline in the options passed to `defineComponent` i.e. not in a separate const - that did seem to make a difference. 

Were these the sorts of errors you were seeing when you tried that approach? Or were your errors on different components?

Attaching `Choropleth.vue` with the changes I made - there's still a couple of issues e.g. (implicit any) but I don't think they're showstoppers. (Attaching with txt extension so github accepts it.. )
[Choropleth.vue.txt](https://github.com/mrc-ide/hint/files/11113955/Choropleth.vue.txt)

M-Kusumgar:
Hey emma! Yeah so I just looked at this on the coach. If you try review inputs file you'll probably see what I am talking about

EmmaLRussell:
As discussed, I'd be in favour of avoiding this by instead updating `mapStateProps` etc to something like:
```
export const mapStateProps = <S, I>(namespace: string,
    map: Dict<(this: CustomVue, state: S) => any>) => {
    return mapState<S>(namespace, map) as I;
};
```
..and splitting up `Computed` interfaces where necessary (e.g. `ComputedFromProps`, `ComputedFromGetters`).
Not sure what you meant by a hybrid solution, but do go ahead and put it in another branch and I'll have a look! :)

PR #831, opened 2023-04-27
Proof of concept for Vue3 Hint

M-Kusumgar:
This is the proof of concept branch of the Vue 3 upgrade. Since the app is building now, we can work on fixing the bugs and issues in parallel.

The strategy we have opted for is the Vue 2 proxy `defineComponent`. Once a part of this branch is fully fixed then I will do smaller PR's to the base branch #822 but this just gives us a building app to work from!

Known bugs:
- Choropleth map not rendering properly because the state value for DataSource (in ReviewInputs) is being set after the page loads (I think, check comment in reviewInputs beforeMount hook)
- Need to fix changing detail level in choropleth map
- Bubble plot works for the most part except when you select a smaller area or change detail
- Bar charts resize on update, need to fix their size
- Currently only login page and normal app work. Have not gotten to data exploration/forgot password
- More to come!

There are couple of commented out bits of code and so on. This branch is just a proof of concept, don't worry about reviewing it. The proper PRs will come from small folders from this branch!

To run, just use the same steps as before!


r-ash (approved):
:shipit: :bow: :pray: :tada:

PR #832, opened 2023-05-04
mrc-4226: Stop issue where download button was triggering duplicate download events

r-ash:
## Description

Since upgrade to Vue3 every download button was triggering two files to be downloaded. This is because since the migration there is a click event listener on both the button and its containing div. See https://v3-migration.vuejs.org/breaking-changes/emits-option.html for details. I think for us the most appropriate way to fix is to rename the event on the parent component so it doesn't create the click event listener.

To test this can run up the app and go to the time series or outputs download steps and click the button. You should get 1 file downloaded.

## Type of version change

Patches

## Checklist


LekanAnni (changes requested):
Nice 👍 

Did you try click.prevent ?

r-ash:
> Nice +1
> 
> Did you try click.prevent ?

Just tried that out but causes and error on button click. I'm not quite sure if that is what we want but let me know if I am misunderstanding. The `click.prevent` will stop the browser default action, for the button this is to propagate the event to the parent container. So if we put a `click.prevent` on the button then that download will indeed only be triggered once but the click will be on the containing div which is the total width of the page, not just the size of the button. So I think we want to remove the click event firing from the containing div by renaming the event.

r-ash:
downloadIndicator and downloadResults tests are now passing so I think this is ready for another look

src/app/static/src/app/components/downloadIndicator/DownloadIndicator.vue
```diff
@@ -1,9 +1,9 @@
 <template>
-    <div>
+    <div id="indicator-download">
         <download-button
             :name="'downloadIndicator'"
             :disabled="downloadingIndicator"
-            @click="download"></download-button>
+            @trigger-download="download"></download-button>
```
LekanAnni, line 6:
```suggestion
            @click.stop="download"></download-button>
```

LekanAnni (changes requested):
I don't think we should be introduction custom events for cases where this issue occurs. vue handles event propagation with out of the box modifier. 

Sorry I misled you with `.prevent`. Using `click.stop()` helps propagating innerDiv. Works fine.

src/app/static/src/tests/components/downloadIndicator/downloadIndicator.test.ts
```diff
@@ -72,7 +72,9 @@ describe("download indicator", () => {
 
     it('can trigger download with iso3 country prefix in filename', async() => {
         const wrapper = getWrapper();
-        await wrapper.findComponent(DownloadButton).vm.$emit("click")
+        const button = wrapper.find("#indicator-download").find("button");
+        expect(button.attributes().disabled).toBeUndefined();
```
M-Kusumgar, line 76:
```suggestion
        expect(button.element.disabled).toBe(false);
```

src/app/static/src/tests/components/downloadIndicator/downloadIndicator.test.ts
```diff
@@ -82,9 +84,11 @@ describe("download indicator", () => {
         expect(mockDownloadFileActions.mock.calls[0][1].data).toStrictEqual(mockDownloadIndicatorData())
     });
 
-    it('can use country prefix when iso3 data is empty', () => {
+    it('can use country prefix when iso3 data is empty', async() => {
         const wrapper = getWrapper(createSut({iso3: "", country: "Malawi"}));
-        wrapper.findComponent(DownloadButton).vm.$emit("click")
+        const button = wrapper.find("#indicator-download").find("button");
+        expect(button.attributes().disabled).toBeUndefined();
```
M-Kusumgar, line 90:
```suggestion
        expect(button.element.disabled).toBe(false);
```

M-Kusumgar (changes requested):
Looks good! The relevant tests are passing now too! Emma suggested that nicer way to test if buttons were disabled in the components test PR, I thought it would be nice to be consistent here as well!

src/app/static/src/app/components/downloadIndicator/DownloadIndicator.vue
r-ash: I think the issue here is that since Vue3 upgrade there is an new click event triggered from the `download-button` div. On current production (see click event)

image

Whereas in current `hintpproofofconcept`

Screenshot_20230621_125159

In hintpproofofconcept there is a click event triggered from the `download-button` div which is new and not on production.

I think this is because of changes to fallthrough attributes. See [https://vuejs.org/guide/components/attrs.html#attribute-inheritance the click ](https://vuejs.org/guide/components/attrs.html#v-on-listener-inheritance) so I think what is happening here is when we click it is triggering an event from the button and from the div. You can see that as if we click in the div but not on the button we only get 1 download triggered. So I think what we want to do here is remove the click being triggered from the `download-button` div, and only keep the event trigger from the button - which we do want to propagate up to the `DownloadIndicator` component. 

It looks like we could do that by turning off fallthrough attributes by default and put them onto the button itself instead of the containing div, but that felt like quite a big hammer which could perhaps cause issues elsewhere. So opted for changing the name of the v-on so that click trigger is not created on the `download-button` div.

M-Kusumgar (approved):
Nicely done!

LekanAnni (changes requested):
I think stopping propagation is part of vue event modifiers.  I don't think you need a custom event to stop propagation, you could just add `click.stop()`.

See reference https://vuejs.org/guide/essentials/event-handling.html#event-modifiers

r-ash:
Ok chatted with Lekan about this, he was suggesting using `click.stop` to stop propagation from the button itself and use the click event created on the containing div to manage the download. Then resize the containing div to the size of the button.

I've done this for the time series indicator download, but instead of using `click.stop` just remove the click event entirely as it was not needed if we're using the one from the containing div. And then resized the containing div to the size of the button.

We can't do this for the output downloads because the containing div also has a title in it, so we need this event to come from just the button so I've stuck with renaming the event for this one to avoid the event handler being created on the containing div.

M-Kusumgar (approved):
Just tested it all out, looks good to me!

LekanAnni (approved):
👍

PR #833, opened 2023-05-04
Mrc-3370-clean integrate SSO authorisation  

LekanAnni:
## Description

This PR update ADRFuelClient to accept request with and without ADRkey. Updates widget to reflect different use cases, where login method can either be basic or SSO.

This has been deployed to [preview](https://naomi-preview.dide.ic.ac.uk/)  hence you can proceed to test SSO auth method while testing basic auth method locally.

## Type of version change
_Delete as appropriate_

## Checklist


r-ash (changes requested):
Just checked on preview and when I log in with my account, who should have access to several datasets can only see the public Antarctica dataset

image

And indeed when I import it it only says I have read access to this when it should be read & write. I wonder if some issue with getting permissions from dev ADR?

EmmaLRussell (commented):
Sorry, I'm a bit confused about how this works. I'd assumed that with SSO we'd be passing an oauth access token to ADR via the requests to ADRFuelClient (as described [here](https://www.oauth.com/oauth2-servers/the-resource-server/)) - but as far as I can tell the only difference to those requests is that the ADR key has been removed - but surely we need to include oauth credentials in order to authenticate (and therefore be authorized)? How is the access token getting to ADR?

src/app/src/main/kotlin/org/imperial/mrc/hint/clients/ADRFuelClient.kt
```diff
@@ -28,6 +28,11 @@ class ADRClientBuilder(val appProperties: AppProperties,
         val apiKey = this.encryption.decrypt(encryptedKey)
         return ADRFuelClient(this.appProperties, apiKey, this.logger)
     }
+
+    fun buildSSO(): ADRClient
+    {
+        return ADRFuelClient(this.appProperties, "", this.logger)
```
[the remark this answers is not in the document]
LekanAnni: You are absolutely right, I forgot to include the bearer.

src/app/src/main/kotlin/org/imperial/mrc/hint/clients/ADRFuelClient.kt
```diff
@@ -7,6 +7,7 @@ import org.imperial.mrc.hint.logging.GenericLogger
 import org.imperial.mrc.hint.logging.logADRRequestDuration
 import org.imperial.mrc.hint.security.Encryption
 import org.imperial.mrc.hint.security.Session
+import org.imperial.mrc.hint.security.oauth2.ProfileDefinition.Companion.token
```
r-ash, line 10: Just a question for my understanding, how does this token get set when importing like this? From reading about companion objects it sounds like this will get this like a static value but that looks to be an empty string. https://github.com/mrc-ide/hint/blob/5903eecaf25ed70c10b7b3c276785aa1afb5d0f9/src/app/src/main/kotlin/org/imperial/mrc/hint/security/oauth2/ProfileDefinition.kt#L21 Is kotlin doing something clever here?

LekanAnni:
I think codecov is unhappy but code paths are fairly tested

r-ash:
Have deployed this to preview with latest config changes, getting some issues

sso_bug

src/app/src/main/resources/config.properties
```diff
@@ -30,5 +30,6 @@ oauth2_client_id=fakeId
 oauth2_client_secret=fakeSecret
 oauth2_client_url=fakeUrl
 oauth2_client_audience=naomi
-oauth2_client_adr_url=http://localhost:5000
-oauth2_login_method=false
\ No newline at end of file
+oauth2_client_adr_server_url=http://localhost:5000
+oauth2_login_method=false
```
EmmaLRussell, line 34: I think this is a bit confusing since we're using oauth2 to do the hint login, but this just refers to using oauth2 to pass ADR credentials for fetching resources? So maybe `oauth2_adr_resources` or something similar might be clearer?

src/app/src/main/resources/config.properties
```diff
@@ -30,5 +30,6 @@ oauth2_client_id=fakeId
 oauth2_client_secret=fakeSecret
 oauth2_client_url=fakeUrl
 oauth2_client_audience=naomi
-oauth2_client_adr_url=http://localhost:5000
-oauth2_login_method=false
\ No newline at end of file
+oauth2_client_adr_server_url=http://localhost:5000
```
EmmaLRussell, line 33: Is this not going to be the same as the adr_url in practice. or at least share the same base url? Why do we need to keep this separately?
UPDATE: Thanks for the local test config, looks like they are using an unrelated base url!
I think it would be good to include all of the config for testing here, apart from the secret. which should be stored and fetched from the vault.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/LoginController.kt
```diff
@@ -63,4 +65,10 @@ class LoginController(
     {
         return oauth2RegisterRedirect()
     }
+
+    @GetMapping("/sso")
+    fun isSSOLoginMethod(): ResponseEntity<String>
```
EmmaLRussell, line 70: What's the reason for supporting non-SSO? I thought our intention was to switch over completely to using ADR to authenticate HINT users via their resource server? What configurations won't do that?
If it's to allow testing without exposing the secret, we should be able to do a fetch of the secret from the vault in the CI script. Or is it to allow us an easy way back if it doesn't work out?!

src/app/src/main/kotlin/org/imperial/mrc/hint/security/HintPac4jConfigService.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.imperial.mrc.hint.security.oauth2.clients.OAuth2Client
+import org.pac4j.core.client.Clients
+import org.pac4j.core.client.IndirectClient
+import org.pac4j.core.config.Config
+import org.pac4j.core.http.callback.PathParameterCallbackUrlResolver
+import org.pac4j.http.client.indirect.FormClient
+import org.pac4j.jee.context.session.JEESessionStore
+import org.pac4j.sql.profile.service.DbProfileService
+
+class HintPac4jConfigService(
```
EmmaLRussell, line 14: This class just reproduces the logic that was already in SecurityConfig.getPac4jConfig? Is this just an opportunistic refactor then, rather than necessary for the SSO changes?

src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
```diff
@@ -79,6 +68,17 @@ class Session(
         }) as CommonProfile
     }
 
+    fun getAccessToken(): String
+    {
+        val manager = ProfileManager(webContext, sessionStore)
+
+        val profile = manager.profiles.singleOrNull() ?: OAuth20Profile()
+
+        val token = profile.getAttribute(ACCESS_TOKEN) ?: ""
```
EmmaLRussell, line 77: How does the access token get into the profile? I can't see anywhere that happens - it is handled automatically by Pac4j?

Should be throwing an error here rather then returning empty string if there's no token in the profile? I think we should only be requesting it here it if we're expecting it to exist?

src/app/src/main/kotlin/org/imperial/mrc/hint/security/oauth2/ProfileDefinition.kt
```diff
@@ -37,9 +33,6 @@ class ProfileDefinition(
 
     override fun getProfileUrl(accessToken: Token?, configuration: OAuthConfiguration?): String
     {
-        accessToken.let {
-           token = (accessToken as OAuth2AccessToken).accessToken
-        }
         return "https://${appProperties.oauth2ClientUrl}/userinfo"
     }
```
EmmaLRussell, line 37: Why was this change necessary, and how does it affect behaviour?

EmmaLRussell (commented):
It would have been useful to have some explanation in the description about how the security classes fit together - the flow isn't obvious from the code since (I think?) pac4j handles some parts.

Thanks for sending the local test config over - After granting access to the app, I see this error in backend when it tries to fetch datasets: 
```
{"detail":"User 'auth0|6380ada6c78f5fe0d639efe6' has not yet logged into ADR","error":"ADR_ERROR"}
```

It's working ok for you locally?

EmmaLRussell:
Also, I really think we need some integration tests of this.

src/app/src/main/resources/config.properties
EmmaLRussell: Sorry, I've just realised that this was the config key we were already using for supporting all Auth0 login, as opposed to legacy HINT login, right? So we're assuming that SSO for getting ADR resources is the method to use so long as Auth0 access is what we're using to log in at all? Sorry, I'd forgotten we were still supporting legacy login,

LekanAnni:
> {"detail":"User 'auth0|6380ada6c78f5fe0d639efe6' has not yet logged into ADR","error":"ADR_ERROR"}

Not sure where this error originates. The issue at the moment is resource permission. 401 unauthorised which causes redirection back to login page.

src/app/src/main/kotlin/org/imperial/mrc/hint/security/oauth2/ProfileDefinition.kt
LekanAnni: This was never the right way to retrieve access token. This PR has updated the code to retrieve token from the session. Pac4j OAuth20 stores access and refresh token in sessions.

src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
LekanAnni: For some reason it might not exists, perhaps that's why PAC4j makes it optional. I do not think we should throw an error, authentication should be managed dynamically. For example, if a request is made to retrieve resources without access_token it should return only the public datasets and the app should still be usable.

src/app/src/main/kotlin/org/imperial/mrc/hint/security/HintPac4jConfigService.kt
LekanAnni: Opportunistic refactor, should have commented on it.

src/app/src/main/kotlin/org/imperial/mrc/hint/security/SecurityConfig.kt
EmmaLRussell: That would indicate that something was wrong though, wouldn't it? If we're configured to support SSO and the user is accessing ADR (which should be done by authenticated users only) then there's definitely a problem if there's no access token I think? I guess we can still give access to public datasets, but maybe we should at least log that there's an auth issue rather than just silently passing the empty token.

EmmaLRussell (commented):
Looks good. I saw some ephemeral issues on preview, where I got an error something like "user is not logged into ADR" - however this resolved on next attempt, so I expect it was probably an ADR dev server issue. 

Getting odd behaviour when I run locally, only getting public datasets. Perhaps I've got the config wrong. Could we document the config for running locally with SSO, or enable it with a script? Would just need to make sure we don't include the secret, but manually or automatically fetch it from the vault.

EmmaLRussell:
Also chatted to @r-ash and @M-Kusumgar about automated SSO testing, and we agreed it would be useful to add a test suite to run daily. Ticket here: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4312

M-Kusumgar (approved):
Don't think I will be a huge help with a lot of the code changes! I have tested both locally for the old method working and the preview for the new method and have not found any problems with either! Looks like this is working fine!

r-ash (approved):
Nice one, this looks great to me!

PR #834, opened 2023-05-09
mrc-4218: Fix tooltip width and glitchy display

r-ash:
## Description

This PR fixes a few issues with the tooltip
* Set the `popperTriggers` on the filter select tooltip so this stays open when user highlights it
* Set the max height of the tooltip so that very long tooltip does not extend beyond the viewable area
* Set max-width on a tooltip so that long text wraps instead of going super wide
* Stop weird flickery display of tooltips by floating the tooltip outside of the icon area
* Add some styling

## Type of version change
_Delete as appropriate_
Patches

## Checklist


EmmaLRussell (commented):
Looks like the build is broken!

> Set the popperTriggers on the filter select tooltip so this stays open when user highlights it

This means highlighting works in FilterSelect (Plot type in Time Series) but not elsewhere e.g. on the shorter model options tooltips (I guess those are coming from Vue Dynamic Form anyway?). Not sure how important highlighting in a tooltip is anyway though?

The scrolling of very long tooltips is nice, but could be more obvious? Scrollbar on plot type appears grey on black momentarily then disappears. 

LGTM otherwise. Flicker fix and max dimensions looks like they work nicely.

r-ash:
Thanks

> Looks like the build is broken!

I think expected until we fix the tests, maybe we should try get that going next.

> This means highlighting works in FilterSelect (Plot type in Time Series) but not elsewhere e.g. on the shorter model options tooltips (I guess those are coming from Vue Dynamic Form anyway?). Not sure how important highlighting in a tooltip is anyway though?

I think this mimics the current behaviour where FilterSelect tooltip stays open on hover. We could try make this behaviour consistent and have them stay open on the model options page too. What do you think? I was avoiding doing this because with the `popperTriggers` if you hover over a tooltip and quickly (but not too quickly) move your mouse out you can cause the tooltip to stay open even when not hovering over it. Felt unlikely to occur especially on FilterSelect but maybe more likely on the options page where users might move from one tooltip to another fairly quickly.

> The scrolling of very long tooltips is nice, but could be more obvious? Scrollbar on plot type appears grey on black momentarily then disappears.

What browser are you on? On Firefox I get a subtle scrollbar which I actually thought looked quite good. Chrome's is much more obvious 
image
I could try get firefox to style like this, or maybe change the colour of the scroll to white to make it a bit more obvious?

EmmaLRussell:
> I think expected until we fix the tests, maybe we should try get that going next.

Sorry, yes of course. 

> I think this mimics the current behaviour where FilterSelect tooltip stays open on hover. We could try make this behaviour consistent and have them stay open on the model options page too. What do you think? I was avoiding doing this because with the `popperTriggers` if you hover over a tooltip and quickly (but not too quickly) move your mouse out you can cause the tooltip to stay open even when not hovering over it. Felt unlikely to occur especially on FilterSelect but maybe more likely on the options page where users might move from one tooltip to another fairly quickly.

Yeah, I think it's fine as it is.

> What browser are you on? On Firefox I get a subtle scrollbar which I actually thought looked quite good. Chrome's is much more obvious image I could try get firefox to style like this, or maybe change the colour of the scroll to white to make it a bit more obvious?

Yeah, I was on Firefox, and was initially a bit confused as the content was truncated, with no scrollbar visible (as I hadnt noticed it before it faded). I think it would be good but probably fiddly to improve this, maybe not worth it.

r-ash:
> Yeah, I was on Firefox, and was initially a bit confused as the content was truncated, with no scrollbar visible (as I hadnt noticed it before it faded). I think it would be good but probably fiddly to improve this, maybe not worth it.

Now made that white so hopefully a bit more obvious, that look a bit better?
image

EmmaLRussell (approved):

PR #835, opened 2023-05-09
mrc-4216 Fixing header dropdown delay and href bugs

M-Kusumgar:
## Problems/motivation:
- When you click the dropdown on a different page (navigate to the privacy page from the `online support` dropdown and then try to open the dropdown in `hintpproofofconcept`) it will navigate you to the home page on click. This is because the `<a>` has a link to `#` which navigates the user back to the home page.
- The dropdowns in the navbar take one second's delay to close when you click off them after opening them. I am assuming this was done in the past because the blur event on the `<a>` tag closes the menu and this does not let you click any links in the menu.

## New features/solutions:
- I will replace the `<a>` tag with a `<p>` tag so it does not need to have a link. The `<p>` tag is not focus-able so I will use the `tabIndex` property which will let it be focus-able. This means there is no link to `#` and the user stays on the same page. I will also add scoped styling to the `<p>` tag to make it look like an `<a>` tag.
- I have added event handlers that check whether the mouse is over the menu, in which case keep it open (even if the user clicks on it which means they will now be able to click links without the menu disappearing)

- Clicking navbar dropdowns like `online support` should toggle the dropdown
- Dropdown menus should not disappear when mouse is over them but should disappear after 100ms when mouse is not on them
- You should be able to click links in the menus
- Opening the menus should keep you on the same page

## How to run app:

## Checklist


LekanAnni (changes requested):
Looking good,

I think it's nice that you have added this PR, could you branch out of master instead ? This issue is not depended on vue3 upgrade. While working off master please add some tests.

EmmaLRussell:
> Looking good,
> 
> I think it's nice that you have added this PR, could you branch out of master instead ? This issue is not depended on vue3 upgrade. While working off master please add some tests.

This bug doesn't apply in master I don't think - I'm not seeing it on production.

M-Kusumgar:
@LekanAnni @EmmaLRussell Yes so the bug for navigating back to the home page is after the Vue 3 upgrade. I just got a little carried away and added an extra feature! Imagined us doing testing right at the end because I haven't managed to get the test environment up and running yet! The PRs that are mainly bug fixes and so on will probably not have any tests in them yet

EmmaLRussell (commented):
I think the menu should close when an item is clicked (as it currently does in master..eventually) rather than on blur.

Good to have the menu not be a link I think, but having to handle the mouse events seems like a faff. The WODIN drop-drops don't seem to have this level of drama - it's just using the bootstrap classes, e.g. [here](https://github.com/mrc-ide/wodin/blob/main/app/static/src/app/components/header/AppHeader.vue#L8). Would something similar work here? That does use an `<a>` but with `role="button"` so at least it's admitting what it is!

EmmaLRussell:
> Perhaps a system we can keep is if there's a slightly new feature or different functionality you make a comment in the relevant test section? So we don't forget? Maybe one starting with `// TODO: test` or something like that so we can search for it? What do you guys think of this suggestion?

That's fine by me, but I'd suggest getting the test env sorted sooner rather than later, or you'll be storing up a lot of work for yourself! I'd probably suggest doing another ticket for the missing tests (and list them all in there) as it's surprisingly easy to forget about //TODOs! :laughing:

M-Kusumgar:
> > Perhaps a system we can keep is if there's a slightly new feature or different functionality you make a comment in the relevant test section? So we don't forget? Maybe one starting with `// TODO: test` or something like that so we can search for it? What do you guys think of this suggestion?
> 
> That's fine by me, but I'd suggest getting the test env sorted sooner rather than later, or you'll be storing up a lot of work for yourself! I'd probably suggest doing another ticket for the missing tests (and list them all in there) as it's surprisingly easy to forget about //TODOs! laughing

Actually yes this is much more sensible. I can see this getting lost in the sea of code! Will do a testing ticket and have these as subtasks then!

M-Kusumgar:
> I think the menu should close when an item is clicked (as it currently does in master..eventually) rather than on blur.
> 
> Good to have the menu not be a link I think, but having to handle the mouse events seems like a faff. The WODIN drop-drops don't seem to have this level of drama - it's just using the bootstrap classes, e.g. [here](https://github.com/mrc-ide/wodin/blob/main/app/static/src/app/components/header/AppHeader.vue#L8). Would something similar work here? That does use an `<a>` but with `role="button"` so at least it's admitting what it is!

I will give this a try soon, hopefully it will work because that seems much nicer!

M-Kusumgar:
> Good to have the menu not be a link I think, but having to handle the mouse events seems like a faff. The WODIN drop-drops don't seem to have this level of drama - it's just using the bootstrap classes, e.g. [here](https://github.com/mrc-ide/wodin/blob/main/app/static/src/app/components/header/AppHeader.vue#L8). Would something similar work here? That does use an `<a>` but with `role="button"` so at least it's admitting what it is!

I tried this setup out, but found that it defaulted to the old behaviour of the page refreshing and returning to the homepage, not sure why this is happening! But I have added clicking the menu closes it so I guess this is pretty much equivalent!

M-Kusumgar:
should be ready to review

PR #836, opened 2023-05-10
mrc-4214 Passing data into index.ftl freemarker template

M-Kusumgar:
## Problems/motivation:
- The title of the app (Naomi) and user are not being passed into the `<hint>` tag in `index.ftl`. This is because we are now mounting `Hint.vue` on the `<div>` with app id instead of just creating components (e.g. `UserHeader` and `Error`) and using them as tags in the freemarker template. (You cannot see the title on left of navbar and username/email in the middle of the navbar)
- Div with an unnecessary id of app
- Unused imports commented out in `index.ts` need cleaning up

## New features/solutions:
- A fix is to use data attributes and pass those into the app when creating it
- Remove id in div and comments

- There should be `Naomi` written on the left side of the navbar
- When you log in, should be able to see your email after `Logged in as` in the navbar

## How to run app:

## Checklist


src/app/static/src/app/index.ts
```diff
@@ -43,7 +38,7 @@ export const router = createRouter({
     routes,
 });
 
-const app = createApp(Hint);
+const app = createApp(Hint, {...(mountEl as HTMLDivElement).dataset});
```
EmmaLRussell, line 41: Could we do `createApp({components:{Hint}}).mount("#app");` to save having to do the query selector?

EmmaLRussell (commented):
I guess we need to do this for the data exploration template too?

src/app/static/src/app/index.ts
M-Kusumgar: I might be being a little stupid here but I tried just doing
```JavaScript
const app = createApp({ components: Hint })
```
And I added the hint tag back into the `index.ftl` file. But this did not show anything when you run the app since the Vue app doesn't have a template to render. The query selector is to pass in the data attributes not to mount the app yet. Mounting the app happens at the end of the index file.

M-Kusumgar:
We will have to do this for data exploration in the very near future. I was hoping to get all the bugs and everything ironed out in the main mode and then the data exploration should be very easy because it is just a bunch of copied code as I understand it!

src/app/static/src/app/index.ts
EmmaLRussell: Yeah, but I was hoping you could avoid that by passing in the values from the template. This approach does work for us in WODIN (see [here](https://github.com/mrc-ide/wodin/blob/main/app/server/views/app.hbs and [here](https://github.com/mrc-ide/wodin/blob/main/app/static/src/app/wodin.ts)), but you're right, it's not doing anything here. I think that may be because WODIN uses a different build target? (`--target lib`)

M-Kusumgar: so I just tried this, get a bunch of umd, umd.min, common, etc and when I import the relevant ones into the ftl files it doesn't display anything on the screen, not sure exactly what the problem is here, might be to do with the setup we currently have with hint

LekanAnni (approved):

PR #840, opened 2023-05-18
Cancel polling when state is reset

r-ash:
## Description

We have seen an issue where users browser is sending requests to `/download/status` endpoint without an ID. It isn't 100% clear how users are getting into this state, but we can recreate it one way.

1. Start a new project, progress to calibration step and open network tab of devtools
2. Start calibration and see polling start
3. Go to first step and remove a file to trigger "new version" dialog
4. Confirm to create a new version
5. See the browser continue to poll for `/download/status` without a download ID

This is happening because when we reset the state neither the `ResetOutputs` or `ResetDownload` are cancelling any of the potentially in progress polling.

## Type of version change

Patches

## Checklist

### DIFF
--- a/src/app/static/src/app/store/root/mutations.ts
+++ b/src/app/static/src/app/store/root/mutations.ts
@@ -13,7 +13,7 @@ import {PayloadWithType, Project} from "../../types";
 import {mutations as languageMutations} from "../language/mutations";
 import {initialProjectsState} from "../projects/projects";
 import {router} from '../../router';
-import {initialModelCalibrateState} from "../modelCalibrate/modelCalibrate";
+import {initialModelCalibrateState, ModelCalibrateState} from "../modelCalibrate/modelCalibrate";
 import {initialADRUploadState} from "../adrUpload/adrUpload";
 import {initialDownloadResultsState} from "../downloadResults/downloadResults";
 import {initialGenericChartState} from "../genericChart/genericChart";
@@ -136,10 +136,16 @@ export const mutations: MutationTree<RootState> = {
     },
 
     [RootMutation.ResetDownload](state: RootState) {
+        stopPolling(state.downloadResults.spectrum);
+        stopPolling(state.downloadResults.coarseOutput);
+        stopPolling(state.downloadResults.summary);
+        stopPolling(state.downloadResults.comparison);
         Object.assign(state.downloadResults, initialDownloadResultsState());
     },
 
     [RootMutation.ResetOutputs](state: RootState) {
+        stopPolling(state.modelRun);
+        stopPolling(state.modelCalibrate);
         Object.assign(state.modelRun, initialModelRunState());
         state.modelRun.ready = true;
         Object.assign(state.modelCalibrate, initialModelCalibrateState());
@@ -157,4 +163,9 @@ export const mutations: MutationTree<RootState> = {
     },
     ...languageMutations
 
+};
+
+const stopPolling = (state: any) => {
+    clearInterval(state.statusPollId);
+    state.statusPollId = -1;
 };
--- a/src/app/static/src/tests/root/mutations.test.ts
+++ b/src/app/static/src/tests/root/mutations.test.ts
@@ -37,6 +37,7 @@ import {Language} from "../../app/store/translations/locales";
 import {router} from '../../app/router';
 import {initialModelCalibrateState} from "../../app/store/modelCalibrate/modelCalibrate";
 import {initialDownloadResultsState} from "../../app/store/downloadResults/downloadResults";
+import {ModelCalibrateMutation} from "../../app/store/modelCalibrate/mutations";
 
 describe("Root mutations", () => {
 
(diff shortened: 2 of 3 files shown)


src/app/static/src/app/store/root/mutations.ts
```diff
@@ -158,3 +164,11 @@ export const mutations: MutationTree<RootState> = {
     ...languageMutations
 
 };
+
+const stopPolling = (state: any) => {
```
r-ash, line 168: Is this the idiomatic way to do this? I want to pass in state which could be `ModelCalibrateState` or `ModelRunState` or one of the download states

EmmaLRussell: I think it would be more typical in Typescript to make an interface which all these state types support and define that as being the type passed in here. 
```
interface PollingState {
    statusPollId: number
}
```
... and then update each of the relevant state types such that they extend the interface ...and remove `statusPollId` as defined indivudally in each of those,,

I'm surprised that use of `any` isn't failing linting, I thought we'd disallowed it in non-test code.

src/app/static/src/tests/root/mutations.test.ts
```diff
@@ -3,30 +3,31 @@ import {initialModelRunState} from "../../app/store/modelRun/modelRun";
...
+    mockSurveyResponse
 } from "../mocks";
 import {DataType} from "../../app/store/surveyAndProgram/surveyAndProgram";
 import {RootState} from "../../app/root";
 import {
     BarchartSelections,
-    ScaleType,
-    initialPlottingSelectionsState
+    initialPlottingSelectionsState,
+    ScaleType
```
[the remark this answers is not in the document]
EmmaLRussell: Oh it insists on them being alphabetical? It's a bit distracting, but it is nice to have those in order!

EmmaLRussell (commented):
This looks good to me, see suggestion on the `stopPolling` parameter.
Were you also going to put in the 'last line of defence' check in the actions? In a separate ticket?
Haven't run this up yet as having issues with RDP this morning 🙄 .. will try again shortly..

EmmaLRussell:
OK, got this running and it looks good :+1:

src/app/static/src/app/store/root/mutations.ts
```diff
@@ -158,3 +164,11 @@ export const mutations: MutationTree<RootState> = {
...
+    if (state.statusPollId == -1) {
+        return
+    }
+    clearInterval(state.statusPollId);
+    state.statusPollId = -1;
```
LekanAnni, line 173: There is no need for resetting poll id,  it will get reset when initialState gets applied below.

src/app/static/src/tests/root/mutations.test.ts
```diff
@@ -44,6 +45,11 @@ describe("Root mutations", () => {
         router.push('/login')
     })
 
+    afterEach(() => {
+        // restore the spy created with spyOn
+        jest.restoreAllMocks();
```
LekanAnni, line 50: Why `restoreAllMocks` ? 

Should clear mocks before each.

LekanAnni (changes requested):
Thanks for getting this in, we never actually clearedInterval but reset pollingId. See below for a few suggestions.

src/app/static/src/app/store/root/mutations.ts
M-Kusumgar: I think the type interface `PollingState` is the right approach but instead of changing the individual states of the download results I think it's simpler to use generics and do this for the type definition of `stopPolling`
```TypeScript
const stopPolling = <T extends PollingState>(state: T) => {
```
This will make sure anything passed into it has the `statusPollId` field!

M-Kusumgar (commented):
Looks good! Hopefully this will mean hint is more robust! I agree with all of lekans and emmas comments, have a look at the comment about `stopPolling` function. I think generics is what you're after

src/app/static/src/tests/root/mutations.test.ts
r-ash: From https://dev.to/edwinwong90/jestclearallmocks-vs-jestresetallmocks-vs-jestrestoreallmocks-explained-5aal I think we want to `restoreAllMocks` because it will restore the mocked function from call to `spyOn` back to the original implementation. Whereas `clearAllMocks` just resets the mock usage data. 

I feel like it's generally better to clean up completely after each test to keep each test isolated. Setting a `spyOn` is like modifying global state and that can lead to non-obvious test failures and test ordering issues if not tidied up.

src/app/static/src/app/store/root/mutations.ts
r-ash: mmh I see this is duplicated in this instance. I left this around because was thinking it would make the `stopPolling` function more reusable and matches the impl of `stopPolling` function in other mutations. I feel like this is the place to set it to `-1` as it keeps all the `stopPolling` logic in one place and can be re-used in the future.

r-ash:
> Were you also going to put in the 'last line of defence' check in the actions?

Agh thanks yes, forgot to do this but I've added that now. Including doing a `console.warn` if that path gets executed. I assume we don't want to show anything to the user. At the moment when this happens we return an error but it doesn't get shown to the user, this will at least tell us something went wrong in dev console.

src/app/static/src/tests/root/mutations.test.ts
r-ash: Yeah I think makes sense here to use `restoreAllMocks` because we usethe `spyOn` in some tests but not all so from what I understand I think we should clear it up after we create the spy.

LekanAnni (approved):
LGTM 👍

PR #841, opened 2023-05-18
mrc-4256 Allow retry of project load when validate fails

EmmaLRussell:
## Description
This branch updates project load behaviour so that project state is not automatically rolled back to the last valid step when validation fails. 

`validate` is a root action called during project load, after all data has been loaded (both loaded state JSON and results retrieved from the backend e.g. model run or calibrate results). A project state is considered invalid if there are any steps which are complete (have all expected data which would allow user to proceed to subsequent step) which come after a step which is incomplete. This might typically occur if an error has been encountered when retrieving results from backend e.g. when fetching model run results for a project which had previously fully completed. The previous behaviour was to immediately rollback the state to the last valid step (the step before the first unexpectedly incomplete step), removing all data for subsequent steps. 

In this branch, the `validate` action does not perform the rollback itself but saves `invalidSteps` as an array in the state. The `LoadErrorModal` gives the user the option to Rollback or to Retry, which will trigger another attempt at loading the project. This might be useful if e.g. a temporary connectivity issue meant that the previous load attempt failed. 

The dialog also displays which steps are considered invalid and which step the rollback would roll back to: 

image

I've refactored the `LoadErrorModal` to interact with the store itself rather than receiving props from `UploadNewProject` - there's only one instance of the `LoadErrorModal` which always behaves in the same way, so it doesn't seem necessary to get values from the store via its parent. It's also unclear to me that it should be a child of `UploadNewProject`, as it is also displayed when loading existing projects... but I've left it where it is for now. 

I think it might also be good to provide a third option, to close the project without making changes and return to the projects page. If Retry is failing multiple times but user doesn't want to Rollback, it would be good to give them a way to easily escape the modal. Can do this in a separate ticket if it seems useful. 

We may want to split up the `loadErrorModal` eventually - it's going to get more complex as we add in further feedback to the user on the load errors. But might be worth leaving that until we have the functionality fixed.  

To manually test this behaviour it's useful to be able to engineer one of the result endpoints to fail or succeed as required without restarting the app. I did this by modifying `result` in `ModelRunController` as follows:
```
@GetMapping("/result/{id}")
    @ResponseBody
    fun result(@PathVariable("id") id: String): ResponseEntity<String>
    {
        val lines = java.io.File("/home/emma/dev/hint/test_succeed.txt").readLines()
        val success = (lines.count() > 0) && (lines[0] == "1")

        if (!success) {
            throw Exception("BOOM")
        }
        return apiClient.getResult(id)
    }
```
.. then you can edit whatever file it reads lines in from to have contents "1" or "0" depending on whether you want the next model run result fetch to succeed or fail. Do something similar in e.g. DiseaseController.getSurvey to test loading where the first step is invalid.  

Currently failing in ADR tests only

## Type of version change

Minor

## Checklist


src/app/static/src/app/components/load/LoadErrorModal.vue
```diff
@@ -2,15 +2,48 @@
...
+                <li v-for="step in invalidSteps" :key="step" v-translate="stepTextKey(step)"></li>
+              </ul>
+              <template v-if="lastValidStep >= 1">
+                <span v-translate="'loadErrorStepsFromValidAction'" />
+                <span v-translate="stepTextKey(lastValidStep)" />.
+                <span v-translate="'loadErrorStepsFromValidWarning'" />
+              </template>
+              <span v-else v-translate="'loadErrorStepsAllAction'" />
+            </p>
+            <p v-else>{{ loadError }}</p>
```
EmmaLRussell, line 17: In the case where the state is invalid and there are invalid steps, the `loadError` is not very informative ("There was a problem loading your data. Some data may have been invalid. Please contact support if this issue persists.") so we replace it with the richer UI here.

src/app/static/src/tests/testHelpers.ts
```diff
@@ -96,6 +97,11 @@ export const expectTranslated = (element: Wrapper<any>,
                                  attribute?: string) =>
     expectTranslatedWithStoreType<DataExplorationState>(element, englishText, frenchText, portugueseText, store, attribute);
 
+export const expectHasTranslationKey = (element: Wrapper<any>, translationKey: keyof Translations) => {
```
EmmaLRussell, line 100: Follwing on from Rob's suggestion that testing every translation string is OTT, how about this as an alternative, to check that an element has the expected translation key?

Could be accompanied by a test of the locales strings to check that each key has different value in each language.

src/app/static/src/app/components/load/LoadErrorModal.vue
```diff
@@ -2,15 +2,46 @@
...
+              <template v-else>
+                  <button id="ok-load-error"
+                          type="button"
+                          class="btn btn-red"
+                          data-dismiss="modal"
+                          aria-label="Close"
+                          @click="clearLoadError"
+                          v-translate="'ok'">
+                  </button>
+              </template>
```
M-Kusumgar, line 44: What is this button for? I couldn't get into a state where this would show

M-Kusumgar (commented):
I think this will be helpful for sure. I think the information is worded very well and conveys all the info. When you rollback, is it supposed to do something in the kotlin layer? Because currently if you rollback and then try to load again it shows the load error modal again. I also tried to rollback, do a model run and then load project again and it remembered my original state, not the new one after rollback. I am assuming that this is another ticket though? Although I would argue these things are pretty interconnected!

src/app/static/src/tests/testHelpers.ts
M-Kusumgar: I think this is a better approach!

src/app/static/src/app/components/load/LoadErrorModal.vue
EmmaLRussell: This could be shown if you tried to load a corrupted zip file for example.

EmmaLRussell:
> When you rollback, is it supposed to do something in the kotlin layer? Because currently if you rollback and then try to load again it shows the load error modal again. 

Rollback should delete any input files in the backend that are associated with rolled back state. 

> I also tried to rollback, do a model run and then load project again and it remembered my original state, not the new one after rollback. I am assuming that this is another ticket though? Although I would argue these things are pretty interconnected!

OK, interesting, that should be working. How far had you got with the original project, and which endpoint(s) did you sabotage to trigger the dialog? Do you mean you had, say, run through the project to completion, sabotaged Model Run Result, rolled back to Model Options, repaired the Run endpoint, done a run but not a calibrate, and then tried to reload project, and seen that it was attempting to reload at post-calibrate stage, and showing the error dialog? Was the endpoint sabotaged at that point or not?

M-Kusumgar:
> > I also tried to rollback, do a model run and then load project again and it remembered my original state, not the new one after rollback. I am assuming that this is another ticket though? Although I would argue these things are pretty interconnected!
> 
> OK, interesting, that should be working. How far had you got with the original project, and which endpoint(s) did you sabotage to trigger the dialog? Do you mean you had, say, run through the project to completion, sabotaged Model Run Result, rolled back to Model Options, repaired the Run endpoint, done a run but not a calibrate, and then tried to reload project, and seen that it was attempting to reload at post-calibrate stage, and showing the error dialog? Was the endpoint sabotaged at that point or not?

Okay wait, just for clarification, when you say sabotaged endpoint so you mean have a "0" in the file and repaired endpoint you mean have a "1" in the file? I changed the get result endpoint in the `ModelRunController`. My steps were:
1. pull up the app and pick the antarctica (malawi) file
2. change area level to region in model options
3. run model run with a "1" in the file
4. run calibration
5. click projects page
6. change "1" to "0" in the file
7. try and load project, error happens like normal
8. click rollback
9. click projects page again
10. load project

You get the error again, also if you instead after step 9, fix the endpoint by changing it to a "1" then your old project loads (state before error or rollback)

EmmaLRussell:
> You get the error again, also if you instead after step 9, fix the endpoint by changing it to a "1" then your old project loads (state before error or rollback)

thanks for clarifying! Sounds like it's not saving the rolled back state correctly - will investigate..

EmmaLRussell:
> > You get the error again, also if you instead after step 9, fix the endpoint by changing it to a "1" then your old project loads (state before error or rollback)
> 
> thanks for clarifying! Sounds like it's not saving the rolled back state correctly - will investigate..

Hm, OK, I think I see what's going on, and I think it would always have caused confusion - when the state rolls back, it wipes the current project & version too, so at that point you're in a sort of orphan project which isn't being saved to the back end, but can be reloaded from local storage. Which is why it looked to me like it worked when I tested by reloading the page, but didn't work for you when you tried reloading from the project page... Might need a rethink on this one..

EmmaLRussell:
Just chatted with Rob about this - we're going to modify the rollback slightly so that it actually spawns a new version rather than interfering with the current one. So you'll always have your original project data available to attempt to reload (as problem may be ephemeral), but can optionally rollback to a new version at the 'last known good state' too. 

Will make those changes in this branch so please hold off doing further reviews before then.

EmmaLRussell:
OK, this should be ready for re-review with change where a new version is generated when state is rolled back so the original state is always preserved. Still failing in ADR tests only.

LekanAnni (changes requested):
Looking good,

**Sabotage 1**
* I sabotaged the endpoint by restarting reeds container.
* Modal displayed as expected.
* Retry does not fix the problem after container is backup
* Rollback does the job and rolls back to the known previous step.
* Project could not re-run model, and modal keeps showing up, blocking the error message. 

**Sabotage 2**

 * Bock request to` /model/options/{iso3}` 
 * loading options will continue to spin
 *  modal does not appear. 
 * I think this is a normal behaviour

It does look like users can potentially be stuck with displayed modal without resolving any of the issues.

M-Kusumgar (changes requested):
Look good! Rollback seems to be working and I like the new version approach as that lets us debug better! Could you make the modal close if the user navigates back on their browser? Currently if I load from the projects page and there is an error, the modal appears. If I then click back on the browser I get to the projects page with the modal still up.

EmmaLRussell:
> Look good! Rollback seems to be working and I like the new version approach as that lets us debug better! Could you make the modal close if the user navigates back on their browser? Currently if I load from the projects page and there is an error, the modal appears. If I then click back on the browser I get to the projects page with the modal still up.

Yep, good point. I think that would be a good fix to roll into this ticket as that will involve fixing this issue anyway:
https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4272

EmmaLRussell:
Thanks for the successful sabotages @LekanAnni  ! Can I get a bit more detail as I'm not seeing quite the same behaviour. 


> **Sabotage 1**
> 
>     * I sabotaged the endpoint by restarting reeds container.
> 
>     * Modal displayed as expected.
> 
>     * Retry does not fix the problem after container is backup
> 
>     * Rollback does the job and rolls back to the known previous step.
> 
>     * Project could not re-run model, and modal keeps showing up, blocking the error message.

How did you restart the redis container? Just a literal `docker restart` with the container name? Should that bring up the redis container successfully so that hintr should operate normally after such a restart?

> 
> **Sabotage 2**
> 
>     * Bock request to` /model/options/{iso3}`
> 
>     * loading options will continue to spin
> 
>     * modal does not appear.
> 
>     * I think this is a normal behaviour
> 

So I tried this with the `model/options` endpoint and didn't see infinite spinner, but instead saw error message on the Model options step instead of the form (that's an issue in itself but probably a separate one!) whether the project had progressed past the Model options step or not. But the `model/options` endpoint doesn't have an `{iso3}` part - do you mean the `model/calibrate/options/{iso3}` endpoint?

EmmaLRussell:
OK, I just tried it with the calibrate options endpoint sabotaged and am seeing the behaviour you describe for that. In this case (as with the model options endpoint) the failure of the options endpoint doesn't appear to be considered a loading error - I think this is probably because the selected options in the state itself are intact, it's just that the available options can't be fetched from the state. In both cases the subsequent steps and state are available to navigate to. Clearly these two situations aren't being handled consistently (error message and hidden form vs endless spinner displayed). Not clear that it would make sense to show the load error dialog here, but maybe a 'Retry load options' button on the form itself?

EmmaLRussell:
I've made a ticket to deal more gracefully with errors when loading options - this feels separate from and lower priority than the issues related to inconsistent state, since all project state is automatically preserved in that case anyway: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4296/Provide-one-click-way-for-users-to-retry-load-of-run-calibrate-options

On the issue of the redis container restart, I checked with Rob and it is not expected that results will be available when redis only is restarted, as redis needs to be brought up before hintr. If you restart hintr after restarting redis, you should find that Retry then succeeds.

EmmaLRussell:
> I think we should continue to add these tests. This is because sometimes there are encoding issues which the tests pickup. Also it's a good idea to test the contents.

OK so we have a difference of opinion on this one, as I know @r-ash and @M-Kusumgar feel the opposite way!
I might be wrong, but I feel like most of the encoding issues the tests have picked up have been in the back end? But maybe some in the front end too. 

I can see both points of view:
- it feels insecure to not test the actual text that gets rendered to the page
- it feels verbose and OTT to put in tests that are essentially just a rearrangement of the strings in the locales file, when it should be enough in a unit test to check that the correct translation keys have been applied (testing the translations come out correctly is really an integration test). 

Maybe we could compromise and standardise on keeping translation string tests slightly separate from core component logic tests, even if just in a separate `describe` block? Although that would have drawbacks too as it's useful to share the setup for both...

Anyway, probably not a debate to have in this PR. Happy to put in the fuller translation tests here, though I think would still be useful to keep the translation key check helper in for cases where you want to check the key is correct but full strings are tested elsewhere.

EmmaLRussell:
Apologies for the churn on this. @M-Kusumgar  and @LekanAnni 

In summary:

@LekanAnni 
> Sabotage 1 (redis container restart

This is expected - Retry will eventually resolve once hintr container is also restarted

> Sabotage 2 (broken options endpoints)

This is a slightly different issue, which will be addressed in https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4296/Provide-one-click-way-for-users-to-retry-load-of-run-calibrate-options

> I think we should continue to add these tests. This is because sometimes there are encoding issues which the tests pickup. Also it's a good idea to test the contents

I've now included full tests of the translated strings, but in a separate `describe` block to the logic tests. Expect further debate!

> This isn't getting translated. 

Aria-labels should now be getting translated

@M-Kusumgar 
> Could you make the modal close if the user navigates back on their browser?

This is already an issue for the modal in master, and will be addressed in https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4272

LekanAnni:
> This is expected - Retry will eventually resolve once hintr container is also restarted
> 
> >

Restarting Redis container should not make the app unusable, perhaps you might want to make a ticket to look into that. @r-ash what are our options ?

src/app/static/src/tests/testHelpers.ts
r-ash: I think testing every string isn't necessary. The current test approach means every time a string changes you have to update several tests which seems more pain than it is worth. We have tests that the directive works so we should rely on that, I don't know what testing every string in each language gives us besides loads of churn when we want to update a string. We've seen encoding issues on production even when we're testing every translation so I'm not sure they even guard us against that. Maybe we can try this and if we see missing translations in the app then we can revert back to the old approach of testing every string?

r-ash:
> Restarting Redis container should not make the app unusable, when the container comes back up, all contents are lost. Perhaps you might want to make a ticket to look into that. @r-ash what are our options ?

If you restart the redis container it will break the connection hintr has to it. The contents of redis shouldn't be lost after restarting it, you just won't get any successful responses from hintr because hintr no longer has a working connection to redis. If you restart hintr container too it should reconnect and then work as before.

LekanAnni:
> > Restarting Redis container should not make the app unusable, when the container comes back up, all contents are lost. Perhaps you might want to make a ticket to look into that. @r-ash what are our options ?
> 
> If you restart the redis container it will break the connection hintr has to it. The contents of redis shouldn't be lost after restarting it, you just won't get any successful responses from hintr because hintr no longer has a working connection to redis. If you restart hintr container too it should reconnect and then work as before.

We could keep the connection alive. I think we can fix this.

M-Kusumgar (approved):
I am happy to approve this! @EmmaLRussell @LekanAnni Just a thought about the translation stuff from the Vue 3 upgrade perspective! I have had to add a lot of unnecessary bloat in my opinion with adding the translate directive to mount options of any component testing text. Additionally, our tests will be running slower because every time we test translation, there are a bunch of awaits waiting for the whole component to update vs if we were to just test the translation strings in one go, there would be one 2-3 awaits and would reduce a lot of unnecessary slowness mounting the directive and awaiting every store update!

That being said we should test the translate directive thoroughly so perhaps we do the huge string tests on a custom test Vue component that just has a bunch of `<p>` tags or something like that with each of the translation keys so that we can check all the strings work. Would this not suffice as a test for the strings working?

EmmaLRussell:
> That being said we should test the translate directive thoroughly so perhaps we do the huge string tests on a custom test Vue component that just has a bunch of `<p>` tags or something like that with each of the translation keys so that we can check all the strings work. Would this not suffice as a test for the strings working?

I wondered about having some sort of approach like that, but I think the trouble with decoupling the strings from their actual context like that is that the tests then become even closer to just an automatable rewrite of the locales.ts file.. 

Another thought that occurred to me is that the browser tests might be the most appropriate place to test rendered translations as part of the fully integrated e2e tests. However we still have very few e2e tests, and it's difficult to set up all scenarios e.g. the sort of failure cases covered by the branch...

Full disclosure: I did just spot a bug when I put in the translation string tests, as I hadn't updated the rollback text for the case where there was an error but no invalid steps. So those test may have some practical use!

M-Kusumgar:
>Full disclosure: I did just spot a bug when I put in the translation string tests, as I hadn't updated the rollback text for the case where there was an error but no invalid steps. So those test may have some practical use!

I am assuming you mean that you didn't change a value in the translation object and so it was displaying a wrong test? because surely that type of issue would be caught when we have these `<p>` tag tests too? Or was the bug more involved than that?

EmmaLRussell:
> I am assuming you mean that you didn't change a value in the translation object and so it was displaying a wrong test? because surely that type of issue would be caught when we have these `<p>` tag tests too? Or was the bug more involved than that?

It was more that considering how the actual texts for the various translation keys fitted together in the context of the whole dialog - which conditionally displays different text depending on the situation - made me realise that I hadn't updated the text values for one situation in order to reflect what was now happening (new version rather than lost data). I *may* have spotted that if working in a one-translation-test-to-rule-them-all system, but maybe not!

LekanAnni (approved):
LGTM 

Thanks

PR #842, opened 2023-05-18
Restart polling for calibrate status if stage incomplete but calibration in progress

r-ash:
## Description

Spotted this whilst doing #840, when we load calibrate page we do not start polling again if the users project has a calibration active i.e. `state.calibrating` is true. This can result in users opening the page and the interface shows the last received status which can be an in progress state. Meaning that they see a spinner but it will never complete and the progress won't update, and they can't even restart the calibrate.

I have seen users in this state sometimes, though not clear how they get into that state. Had Ian report he waited for calibration for 5 hours once but I suspect his project was just not polling. I think the large DRC results can appear like this if someone reloads a project after it failed to get the result but not 100% sure. You can recreate reliably however by
1. Start a new project, run fit and start calibration
2. After calibration starts polling go back to first page and remove one of the files
3. Click OK on "new version" dialog to create a new version
4. In a private browser window open v1 of the project
5. See the spinner shown forever but no polling for calibration result

This PR fixes this by adding a new action called on mount which will restart the polling if `state.calibrating` is true but `state.complete` is false. I took a go on this so let me know if approach is not quite right or anything not idiomatic.

## Type of version change

Patches

## Checklist

### DIFF
--- a/src/app/static/src/app/store/modelCalibrate/actions.ts
+++ b/src/app/static/src/app/store/modelCalibrate/actions.ts
@@ -23,6 +23,7 @@ export interface ModelCalibrateActions {
     getResult: (store: ActionContext<ModelCalibrateState, RootState>) => void
     getCalibratePlot: (store: ActionContext<ModelCalibrateState, RootState>) => void
     getComparisonPlot: (store: ActionContext<ModelCalibrateState, RootState>) => void
+    resumeCalibrate: (store: ActionContext<ModelCalibrateState, RootState>) => void
 }
 
 export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrateActions = {
@@ -126,6 +127,13 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
                 selectFilterDefaults(response.data, commit, PlottingSelectionsMutations.updateComparisonPlotSelections)
             }
             commit(ModelCalibrateMutation.SetComparisonPlotData, response.data);
+        }
+    },
+
+    async resumeCalibrate(context) {
+        const {dispatch, state} = context;
+        if (state.calibrating && !state.complete) {
+            await dispatch("poll");
         }
     }
 };
--- a/src/app/static/src/app/store/modelCalibrate/actions.ts
+++ b/src/app/static/src/app/store/modelCalibrate/actions.ts
@@ -132,7 +132,7 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
 
     async resumeCalibrate(context) {
         const {dispatch, state} = context;
-        if (state.calibrating && !state.complete) {
+        if (state.calibrating && !state.complete && state.calibrateId) {
             await dispatch("poll");
         }
     }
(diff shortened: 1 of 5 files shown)


src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -127,6 +128,13 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
...
             commit(ModelCalibrateMutation.SetComparisonPlotData, response.data);
         }
+    },
+
+    async resumeCalibrate(context) {
+        const {dispatch, state} = context;
+        if (state.calibrating && !state.complete) {
+            await dispatch("poll");
+        }
     }
```
EmmaLRussell, line 138: An alternative would be to do the logic check for `state.calibrating && !state.complete` in the component and dispatch `poll` from there to save making another action. I think we've tended to put quite a lot of logic like that into the components, though you could argue it belongs more in the store. Either way is ok I think.

EmmaLRussell (approved):
I guess this could also happen if you reload the page while calibrating, or log out and back in and reload..

Wondered if we needed to do this for model fit too, but manual test suggests that already seems to cope with reload ok.

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -127,6 +128,13 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
             }
             commit(ModelCalibrateMutation.SetComparisonPlotData, response.data);
         }
+    },
+
+    async resumeCalibrate(context) {
+        const {dispatch, state} = context;
+        if (state.calibrating && !state.complete) {
```
LekanAnni, line 135: I think you should check if calibrateId exists before polling.

LekanAnni (changes requested):
Looking good,

See inline comment

M-Kusumgar (approved):
LGTM! Just a minor case convention issue, other than that all good

src/app/static/src/tests/modelCalibrate/actions.test.ts
```diff
@@ -426,4 +428,23 @@ describe("ModelCalibrate actions", () => {
...
+
+        expect(dispatch.mock.calls.length).toBe(0);
+
+        const complete_state = mockModelCalibrateState({calibrating: false, complete: true});
+        await actions.resumeCalibrate({dispatch, state: complete_state} as any);
+
+        expect(dispatch.mock.calls.length).toBe(0);
+
+        const incomplete_state = mockModelCalibrateState({calibrating: true, complete: false});
+        await actions.resumeCalibrate({dispatch, state: incomplete_state} as any);
```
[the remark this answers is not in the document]
r-ash: Oops, thanks! I'm too drilled in snake case from R

src/app/static/src/app/store/modelCalibrate/actions.ts
r-ash: I'm happy to move this into the component if you think that is more standard, but if you're not feeling too strong I'm tempted to leave as is out of laziness

LekanAnni (approved):
LGTM 👍

PR #843, opened 2023-05-19
mrc-4268: Fix mutation and action tests

r-ash:
## Description

This PR fixes the mutations and actions tests. After this PR I expect all the mutations and actions to be passing. I've also fixed the router test for the main app here. This doesn't fix the router tests for data exploration as we've not updated the router for data exploration yet and I thought that would work well together in a separate PR see [mrc-4283](https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4283).

## Type of version change

None

## Checklist

### DIFF
--- a/src/app/static/src/tests/projects/mutations.test.ts
+++ b/src/app/static/src/tests/projects/mutations.test.ts
@@ -4,7 +4,15 @@ import {router} from "../../app/router";
 
 describe("Projects mutations", () => {
     const testNow = Date.now();
-    global.Date.now = jest.fn(() => testNow);
+
+    beforeAll(() => {
+        jest.useFakeTimers('modern');
+        jest.setSystemTime(testNow);
+    });
+
+    afterAll(() => {
+        jest.useRealTimers();
+    });
 
     const mockProject = {
         id: 1,


src/app/static/jest.config.js
```diff
@@ -31,4 +31,7 @@ module.exports = {
     modulePaths: [
         "<rootDir>"
     ],
+    testPathIgnorePatterns : [
+        "src/tests/components"
+    ],
```
r-ash, line 36: TODO: Remove this!

src/app/static/package.json
```diff
@@ -3,9 +3,6 @@
   "version": "1.0.0",
   "description": "",
   "main": "index.js",
-  "jest": {
```
r-ash, line 6: Was getting warnings about this being duplicated with jest config so have removed here

src/app/static/src/app/index.ts
```diff
@@ -1,47 +1,9 @@
...
-    },
-    {path: "/accessibility", component: Accessibility},
-    {path: "/privacy", component: Privacy},
-    {path: "/projects", component: Projects}
-];
-
-export const router = createRouter({
-    history: createWebHashHistory(),
-    routes,
-});
```
r-ash, line 44: Was running into an issue when this was being imported into test files on line `app.mount('#app');` as it couldn't find `#app' so split the router part of this into a separate file which we can import to test the routing

src/app/static/src/tests/dataExploration/dataExploration.test.ts
```diff
@@ -26,6 +26,6 @@ describe("DataExploration", () => {
         const store = new Vuex.Store(module.storeOptions);
         store.commit({type: "baseline/Ready", payload: null});
         expect(mockConsoleLog.mock.calls[0][0]).toBe("baseline/Ready");
-        expect(mockSaveState.mock.calls[0][0]).toBe(module.storeOptions.state);
+        expect(mockSaveState.mock.calls[0][0]).toStrictEqual(module.storeOptions.state);
```
r-ash, line 29: Comparing objects now needs a `toStrictEqual` instead of `toBe`

src/app/static/src/tests/modelCalibrate/actions.test.ts
```diff
@@ -240,7 +240,9 @@ describe("ModelCalibrate actions", () => {
                         indicator_id: "test indicator",
                         x_axis_id: "test_x",
                         disaggregate_by_id: "test_dis",
-                        selected_filter_options: {"test_name": {}}
+                        selected_filter_options: {
+                            type: [{id: "test", label: "test"}]
```
r-ash, line 244: This data is actually representative of the format that is will be received when running the full app

src/app/static/src/tests/genericChart/mutations.test.ts
```diff
@@ -13,13 +12,9 @@ describe("genericChart mutations", () => {
 
     it("sets dataset",  () => {
         const state = mockGenericChartState();
-        const vueSetSpy = jest.spyOn(Vue, "set");
```
r-ash, line 16: The line which was calling Vue.set has been removed in one of the other vue3 migration PRs but I think we don't need to test this anymore? But not 100% sure. Sounds like this call isn't required see e.g. https://stackoverflow.com/questions/69780482/how-do-i-do-vue-set-in-vue-3

src/app/static/src/app/index.ts
EmmaLRussell: Makes sense for router logic to be in another file anyway I think.

src/app/static/src/app/store/modelOptions/mutations.ts
```diff
@@ -66,7 +66,8 @@ export const mutations: MutationTree<ModelOptionsState> = {
     },
 
     [ModelOptionsMutation.ModelOptionsFetched](state: ModelOptionsState, action: PayloadWithType<DynamicFormMeta>) {
-        const newForm = (state.optionsFormMeta?.controlSections.filter((x) => (x?.label)).length === state.optionsFormMeta?.controlSections.length)
+
```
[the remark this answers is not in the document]
EmmaLRussell: Not sure what that change was all about. This is checking if the form has any control section to see if it has already been initialised, and the change from master checks instead if there are no control sections without labels set (but that check would pass if there were no control sections at all, unlike in the logic you've reverted back to...). @M-Kusumgar can you remember what the idea of this change was?

src/app/static/src/tests/genericChart/mutations.test.ts
EmmaLRussell: Yeah, looks like `Vue.set` isn't called any more, but the `state.selectionDatasetId` value is also set in the mutation, so should test that too here.

src/app/static/src/tests/modelCalibrate/actions.test.ts
EmmaLRussell: If you're just testing that some value gets shuffled around correctly by actions and mutations when it doesn't require its internal structure, it's often handy to just use terse dummy values rather than a full test object of the correct type. But this looks pretty terse anyway so fair enough.

EmmaLRussell (commented):
LGTM. Actions and mutations tests now all passing, except for one, Still seeing this failure locally, and on BK:
```
 FAIL  src/tests/projects/mutations.test.ts
  ● Projects mutations › VersionUploadSuccess sets versionTime

    expect(received).toEqual(expected) // deep equality

    Expected: 1685437664216
    Received: 1685437664231

       96 |         const state = mockProjectsState();
       97 |         mutations[ProjectsMutations.VersionUploadSuccess](state);
    >  98 |         expect(state.versionTime!.valueOf()).toEqual(testNow);
          |                                              ^
       99 |     });
      100 |
      101 |     it("sets created version as current", () => {

      at Object.<anonymous> (src/tests/projects/mutations.test.ts:98:46)
```

src/app/static/src/app/store/modelOptions/mutations.ts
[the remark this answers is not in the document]
M-Kusumgar: I do not remember the idea behind this change, I think it is safe to revert it back to how it is in master!

M-Kusumgar (changes requested):
Looks good! That one test that is failing is because the `Date.now()` is running first and the test is running after a bit so the values are out of sync. Try replacing the `testNow` variable with `Date.now()` inside the `toEqual`.

src/app/static/src/tests/genericChart/mutations.test.ts
r-ash: Thanks, added a test for that

EmmaLRussell (approved):
👍

M-Kusumgar (approved):
Sorry for the delay! looks good to me!

PR #844, opened 2023-05-26
Vue 3 Hint Component Tests

M-Kusumgar:
## Problems/motivation:
This PR mainly aims to fix the components tests in Hint since they are not passing.

## How to review:
For your sanity, please review the commits individually (they are grouped like folders in components test folder), they should be bite-size enough. I think it is also probably a good idea to split the workload amongst you guys. Maybe two people take first half and two people take the second half? Not all the tests will be passing, reasons include:
- Bugs that will be fixed in future PRs shortly
- Me not being sure how to fix them
- Them being skipped (see `testPathIgnorePatterns` in jest.config.js) because they will probably be fixed shortly in the future bug fixes

Please read the guide below before reviewing for some explanations of changes and commits!

### "test refactoring" commit (first commit)
- Makes some props optional because typescript was complaining about not supplying all the props when you mount or shallowMount a component, this is not a problem we will have to worry about in the future though!
- Replaces:
  - `propsData` -> `props`
  - `.find` -> `.findComponent`
  - `.findAll` -> `.findAllComponents`
  - `.at(n)` -> `[n]`
  - Note: these are just some changes in `vue-test-utils` however `.find` still exists just not for custom components so for example you should still use `.find("p")` (to find a `<p>`). I will revert this change in fixing all the folders.

### "initial commit" (second commit)
- this is just the changes @LekanAnni made in this #839 , no need to review this commit

### "add mounts with translate and fix helpers" commit
- We cannot register the `v-translate` directive globally anymore and have to register it with `mount` or `shallowMount` so these functions just all it into the mounting options

### Main changes
- `mount` (and similarly `shallowMount`) have changed. I will just provide an example below to show the difference:

Before:
```TypeScript
const wrapper = mount(Component, {
    store,
    stubs,
    props,
    router,
    directives,
    localVue
})
```
After:
```TypeScript
const wrapper = mount(Component, {
    global: {
        plugins: [store, router],
        stubs,
        directives
    },
    props
})
```
As you can imagine this change has been quite annoying even with some clever regex find and replace. `shallowMount` has also changed because in the past it would just stub custom components however now `shallowMount` stubs the first parent component it encounters (i.e. its depth is 1).
- stubbed components `props` are actually `attributes` so if you stub a custom component, you would have to refer to its props as attributes (i.e. use `.attributes` instead of `.props`).
- Typescript can't understand `wrapper.vm.$data` so I have to cast it as `any` but it is still the correct place!
- Many functions are now `async` and require `await`s in front of them. Some include:
  - `setData`, `setProps`, `setValue`
  - `trigger`
- Any time you want to trigger any updates in the wrapper in a non-async function, e.g. `store.commit()` you must wait for the wrapper to update so you have to add an `await nextTick()`
- If an object that you are comparing in `expect().toBe()` serialises to the same string you have to use `expect().toStrictEqual()`
- If a button is disabled the `disabled` attribute is `""` empty string. If a button is enabled then the attribute is undefined!
- `flushPromises()` waits for all async code to finish running (useful for waiting for loading states to disappear)

### How to test
Go to `src/app/static` and run `npm run test components/` and can also run `npm run test translate` for the directive!

Good Luck!!

## Checklist


src/app/static/src/tests/components/accessibility.test.ts
```diff
@@ -30,33 +30,33 @@ describe(`accessibility`, () =>{
     it(`renders accessibility tags as expected`, () => {
         const rendered = getWrapper()
         const store = rendered.vm.$store;
-        const h1 = rendered.find("#accessibility-content")
-            .find("h1");
+        const h1 = rendered.findComponent("#accessibility-content")
+            .findComponent("h1");
```
EmmaLRussell: There are lots of example of this sort of thing - as you mention in your comment, you shouldn't be using `findComponent` or `findAllComponents` which use css selectors - are you saying you've fixed these in subsequent commits?

src/app/static/src/tests/components/downloadIndicator/downloadButton.test.ts
```diff
@@ -14,7 +14,7 @@ describe("download button", () => {
             directives: {
                 translate: mockTranslate
             },
-            propsData: props
+            props: props
```
EmmaLRussell, line 19: Could just be `props` but not a showstopper!

EmmaLRussell (commented):
Review of "test refactoring" commit only:

> Makes some props optional because typescript was complaining about not supplying all the props when you mount or shallowMount a component

That doesn't seem like a good reason for making a prop optional if it really isn't optional. Or are they really optional as far as the component is concerned, but we had wrongly defined them as required?

>..  this is not a problem we will have to worry about in the future though!

Sorry, not quite sure what you mean by that.

Yeah, kind of tricky to review this commit in isolation - i can see that it's made all the blanket changes as you describe, but that some of them will require modification, and a lot of the tests in those modules are failing (at this commit) for other reasons too.

LekanAnni (changes requested):
Looking good,

I have not looked at the tests but I think buildkite should be able to build app images and successfully test backend etc. Could you get this branch to build on buildkite ?

src/app/static/src/tests/components/accessibility.test.ts
M-Kusumgar: Yes just in this particular case, I have reverted all the ones that use css selectors back into normal `find`s or `findAll`s

M-Kusumgar:
Should be passing backend tests and build etc on buildkite now!

M-Kusumgar:
> That doesn't seem like a good reason for making a prop optional if it really isn't optional. Or are they really optional as far as the component is concerned, but we had wrongly defined them as required?

Sorry yh what I meant was they are really optional for the component! Throughout hint some props have been optional and some havent without explicit declaration so i changed them all to required and then followed around errors to make certain props optional that actually needed to be optional. The issue is, in testing since we put in minimal props, it was uncovering even more optional props than the front end.

> .. this is not a problem we will have to worry about in the future though!

What I meant by this was, typescript was complaining since certain props are 'required' but not supplied in the testing. This isn't a problem to worry about because in the later commits there are helper functions `shallowMountWithTranslate` and `mountWithTranslate` and these are not strict about any types so the errors go away

> Yeah, kind of tricky to review this commit in isolation - i can see that it's made all the blanket changes as you describe, but that some of them will require modification, and a lot of the tests in those modules are failing (at this commit) for other reasons too.

So until you get to the commits that start with "fix most" the tests are going to fail since there are lots of reasons for the tests failing

src/app/static/src/tests/components/dataExploration/dataExploration.test.ts
```diff
@@ -132,27 +132,27 @@ describe(`data exploration component`, () => {
         );
         const wrapper = shallowMount(DataExploration, {store});
 
-        expect(wrapper.find("stepper-navigation-stub").props("backDisabled")).toBe(false)
-        expect(wrapper.find("stepper-navigation-stub").props("nextDisabled")).toBe(true)
+        expect((wrapper.findComponent("stepper-navigation-stub") as VueWrapper).props("backDisabled")).toBe(false)
```
M-Kusumgar, line 155: instead of as vue wrapper, just find comopnent by importing vue comp

src/app/static/src/tests/testHelpers.ts
```diff
@@ -68,33 +69,36 @@ export function expectAllMutationsDefined(mutationDefinitions: any, mutationTree
...
-export function expectTranslatedWithStoreType<T extends TranslatableState>(element: Wrapper<any>,
+export async function expectTranslatedWithStoreType<T extends TranslatableState>(element: any,
                                                                            englishText: string,
                                                                            frenchText: string,
                                                                            portugueseText: string,
                                                                            store: Store<T>,
                                                                            attribute?: string) {
+    const value = () => attribute ? element.attributes(attribute) : element.text();
     store.state.language = Language.en;
     registerTranslations(store);
```
M-Kusumgar, line 81: can you await i18n.init? try await in front of expect

src/app/static/src/tests/testHelpers.ts
```diff
@@ -112,14 +116,44 @@ export const expectChangeLanguageMutations = (commit: Mock) => {
...
+    expect(wrapper.findComponent(ErrorReport).props("open")).toBe(true);
 }
 
 export function expectArraysEqual(result: any[], expected: any[]) {
     expect(result).toEqual(expect.arrayContaining(expected));
     expect(expected).toEqual(expect.arrayContaining(result));
 }
+
+// @ts-ignore
+export function shallowMountWithTranslate<T extends TranslatableState, C>(component: C, store: Store<T>, options?: any): VueWrapper<InstanceType<C>> {
```
M-Kusumgar, line 132: try extends partial defineComponent

M-Kusumgar:
@EmmaLRussell @r-ash @LekanAnni  Following the meeting, I have allocated different commits (i.e. test folders) to different people. There is also information for which tests will fail (the rest should pass!)

# Rob
## adr
Fails:
- ADR Key › can add key

## dataExploration
Fails:
- None

## downloadIndicator
Fails:
- download button › it can emit click event

## downloadResults
Fails:
- download › can emit download
- uploadModal  › does not render summary report when summary metadata request failed
- uploadModal  › does not render spectrum report when spectrum metadata request failed
- uploadModal  › does not render comparison report when comparison metadata request failed
- Download Results component › can download comparison file once prepared

## files
Fails:
- File upload component › calls upload when file is selected

## genericChart
Fails:
- None

# Lekan
## header
Fails:
- None

## load
Fails:
- None

## modelCalibrate
Fails:
- Model calibrate component › setting options value commits update mutation

## modelOptions
Fails:
- None

## modelOutput
Fails:
- ModelOutput component › computes barchart selections
- ModelOutput component › commits updated selections from barchart and orders them according to filter
- ModelOutput component › commits updated selections from barchart as normal if no matching xAxis key is provided
- ModelOutput component › commits updated selections from barchart as normal if no barchart filters are provided
- ModelOutput component › commits updated selections from barchart and orders them according to nested filter
- ModelOutput component › renders barchart table
- ModelOutput component › renders barchart table with correct indicator props

## modelRun
Fails:
- None

# Emma
## password
Fails:
- Forgot password component › updates html lang when language changes
- Reset password component › updates html lang when language changes

## plots
Special case! This entire folder is ignored so can't run tests on it but just check that the refactoring points we discussed are okay!

## projects
Fails:
- Projects component › renders as expected with current project
- Project history component › renders as expected
- Project history component › can expand project row
- Project history component › can collapse project row

## reviewInputs
Fails:
- Survey and programme component › does not render choropleth controls if there is no selected data type
- Survey and programme component › renders choropleth as expected
- Survey and programme component › renders filters as expected
- Survey and programme component › data source is not rendered if no selected data type
- Survey and programme component › survey is included in data sources when survey data is present
- Survey and programme component › programme (ART) is included in data sources when programme data is present
- Survey and programme component › ANC is included in data sources when ANC data is present
- Survey and programme component › can select data source
- Survey and programme component › renders table as expected

## uploadInputs
Fails
- None

Note: Just spotted an error! On line 23 of uploadInputs.test.ts it says `import {testUploadComponent} from "./fileUploads."` it should be `import {testUploadComponent} from "./fileUploads"` (last dot in import path needs to be removed!)

# Whoever has gone through their chunk first!
## root
These are files not in a folder e.g. `accessibility.test.ts`, `stepper.test.ts`, `modal.test.ts`, etc. A lot of these are fairly small!
Fails:
- None

M-Kusumgar:
I have just skimmed over most of the tests, there are a small number of `await nextTick()` around from when I did find and replace `Vue.nextTick()` that are not necessarily needed! Can do a future commit to get rid of those

src/app/static/src/tests/testHelpers.ts
M-Kusumgar: Following this, I was wondering why registerTranslations was even needed! I just deleted it from the function and everything works! So now it is just, change the store, await nextTick!

src/app/static/src/tests/testHelpers.ts
M-Kusumgar: shallowMount unfortuantely complains about Partial<DefineComponent> and if I extend as DefineComponent the function is fine but then it complains about the components we put in are not the right type. May need to just add a comment about this!

M-Kusumgar: btw these aren't just components that use the `defineComponentVue2` it even complains about components that use `defineComponent`

src/app/static/src/tests/components/dataExploration/dataExploration.test.ts
M-Kusumgar: This works! will be done in a future commit!

r-ash (changes requested):
I've had a look through folders
* adr
* dataExploration
* downloadIndicator
* downloadResults
* files
* genericChart
and also had a look a the files
* errorAlert
* errorReport
* errors

This looks great to me, left a few comments mostly for my understanding

src/app/static/src/tests/components/projects/projectHistory.test.ts
```diff
@@ -751,22 +767,22 @@ describe("Project history component", () => {
...
 
-        const modal = wrapper.findAll(".modal").at(1);
+        const modal = wrapper.findAll(".modal")[1];
         const input = modal.find("input");
-        const renameBtn = modal.find(".modal-footer").findAll("button").at(0);
+        const renameBtn = modal.find(".modal-footer").findAll("button")[0];
         await input.setValue("");
 
-        expect(renameBtn.attributes("disabled")).toBe("disabled");
+        expect(renameBtn.attributes("disabled")).toBe("");
```
EmmaLRussell, line 778: The "" vs undefined attribute values to check for disabled values is pretty confusing - might be more readable to use `(renameBtn.element as HTMLButtonElement).disabled` which gives you a boolean

src/app/static/src/tests/components/projects/versionStatus.test.ts
```diff
@@ -20,16 +20,17 @@ describe("Version status component", () => {
...
-        return shallowMount(VersionStatus, {store});
+        return shallowMountWithTranslate(VersionStatus, store, {global: {plugins: [store]}});
     };
     const date = new Date(2020, 8, 1, 12, 45, 59);
 
     it("does not render if no current project", () => {
         const wrapper = getWrapper();
-        expect(wrapper.isEmpty()).toBe(true);
+        // just empty element if v-if isn't true
+        expect(wrapper.html()).toBe("<!--v-if-->");
```
EmmaLRussell, line 30: Feels like there should be a nicer way of doing this, but not sure what it would be! I wonder why they got rid of `isEmpty`..

EmmaLRussell (commented):
Looks good, no big surprises! 

There's various modules where it might be nice to put in a `getWrapper` util to avoid too much
```
global: {
                plugins: [store]
            }
``` 
boilerplate - but I think that's could be something we do later as opportunistic clean-ups when new tests are added.

src/app/static/src/tests/components/adr/ADRIntegration.test.ts
```diff
@@ -194,16 +203,19 @@ describe("adr integration", () => {
...
         expect(mockTooltip.mock.calls[0][1].value).toBe("Atualmente não tem permissões de escrita para este conjunto de dados e não poderá carregar ficheiros para o ADR");
     });
 
     it("call getUserCanUpload action if dataset is selected", () => {
         const store = createStore("123", null, {})
-        shallowMount(ADRIntegration, {store});
-        store.state.baseline.selectedDataset = fakeDataset
+        const rendered = shallowMount(ADRIntegration, {global: { plugins: [store] }});
+        // @ts-ignore
+        rendered.vm.$options.watch.selectedDataset.call(rendered.vm)
```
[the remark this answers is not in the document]
M-Kusumgar: sorry this was early days of refactoring tests! I can remove the ts-ignore and instead do `rendered.vm as any` and it works!

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -302,56 +303,75 @@ describe("select dataset", () => {
 
     afterEach(() => {
         jest.resetAllMocks();
+        jest.restoreAllMocks();
```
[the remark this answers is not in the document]
M-Kusumgar: So we need both of these, `resetAllMocks` will replace each `jest.fn()` with a new one and `restoreAllMocks` will reset `jest.spyOn` functions back to their original function as far as I know! Also removing either of those does cause the expected errors (removing `resetAllMocks` leads to mock functions being called more times than expected as they are carried over from previous tests and removing `restoreAllMocks` causes the `spyOn` tests to fail)

src/app/static/src/tests/components/adr/selectDataset.test.ts
```diff
@@ -931,27 +1014,32 @@ describe("select dataset", () => {
...
 
-        expect(rendered.findAll(LoadingSpinner).length).toBe(1);
+        expect(rendered.findAllComponents(LoadingSpinner).length).toBe(1);
 
-        await Vue.nextTick();
-        await Vue.nextTick();
-        await Vue.nextTick();
+        await nextTick();
+        await nextTick();
+        await nextTick();
```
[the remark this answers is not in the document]
M-Kusumgar: yep you are correct! I have replaced anytime there is more than one `await nextTick()` in a row to `await flushPromises()` just seems cleaner, maybe I will do a regex find and replace for those cases around the test files!

src/app/static/src/tests/components/downloadIndicator/downloadButton.test.ts
```diff
@@ -29,21 +31,22 @@ describe("download button", () => {
         expect(mockTranslate.mock.calls[0][1].value).toBe("downloadIndicator")
     });
 
-    it('it does not render button when disabled prop is true ', () => {
+    it('it does not render button when disabled prop is true ', async () => {
         const wrapper = getWrapper({disabled: true, name: "downloadIndicator"})
-        expect(wrapper.find(".btn").attributes("disabled")).toBe("disabled")
+        await wrapper.setProps({disabled: true, name: "downloadIndicator"})
+        expect(wrapper.find(".btn").attributes("disabled")).toBe("")
```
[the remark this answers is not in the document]
M-Kusumgar: you're right! now sure why that line was there, have removed it!

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
```diff
@@ -204,52 +210,60 @@ describe(`uploadModal `, () => {
 
         const store = wrapper.vm.$store
         store.state.language = Language.fr;
-        expect(mockTooltip.mock.calls[4][1].value).toBe("Créer une version étiquetée dans l'ensemble de données ADR, pour inclure tous les fichiers d'entrée et de sortie");
-        expect(mockTooltip.mock.calls[5][1].value).toBe("Télécharger de nouvelles versions des fichiers sélectionnés sans créer de version");
+        await nextTick();
+        expect(mockTooltip.mock.calls[2][1].value).toBe("Créer une version étiquetée dans l'ensemble de données ADR, pour inclure tous les fichiers d'entrée et de sortie");
+        expect(mockTooltip.mock.calls[3][1].value).toBe("Télécharger de nouvelles versions des fichiers sélectionnés sans créer de version");
```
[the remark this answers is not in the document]
M-Kusumgar: I actually have no idea! But I feel like 0,1,2,3,4,5 is better than 0,1,4,5,6,7 anyway haha, my best guess would be in the past since vue seemed to be doing a lot of lifecycles without the `await nextTick()`s it might've done a lifecycle when we try and get the `store` out of the wrapper, which would explain the 2 & 3 going missing in the original sequence

src/app/static/src/tests/components/errors.test.ts
```diff
@@ -21,25 +20,25 @@ describe("Errors component", () => {
...
+        expect(paras[0].text()).toBe("First error");
+        expect(paras[1].text()).toBe("Error detail");
     });
 
     it("renders nothing when no errors", () => {
         const store = createStore([]);
-        const wrapper = shallowMount(Errors, {propsData, store, localVue});
-        expect(wrapper.html()).toBeFalsy();
+        const wrapper = shallowMount(Errors, {props, global: {plugins: [store]}});
+        expect(wrapper.html()).toBe("<!--v-if-->");
```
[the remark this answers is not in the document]
M-Kusumgar: yes, so this is the same as empty wrapper basically so nothing shown because vue uses `<!--v-if-->` to keep track of where the `v-if`s are (conditional rendering) so if the html is just `<!--v-if-->` that means the `hasErrors` variable in `Errors.vue` is false. If it was true, Vue would replace the `<!--v-if-->` comment with our template

src/app/static/src/tests/components/adr/selectDataset.test.ts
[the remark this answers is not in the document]
r-ash: Ah makes sense! Thanks

src/app/static/src/tests/components/downloadResults/uploadModal.test.ts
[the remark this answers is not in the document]
r-ash: Ah yep that makes sense

src/app/static/src/tests/components/plots/filterSelect.test.ts
```diff
@@ -15,38 +16,50 @@ describe("FilterSelect component", () => {
...
                     {id: "3", label: "three", description: "Third option"},
                     {id: "4", label: "four", description: "Fourth option"}
                 ]
             },
-            directives: {
-                tooltip
-            }
         });
-        expect(wrapper.find(HelpCircleIcon).exists()).toBe(true);
+        expect(wrapper.findComponent(VueFeather).exists()).toBe(true);
```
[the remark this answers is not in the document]
M-Kusumgar: agreed! found another place I missed out the type too, fixed them both

src/app/static/src/tests/components/plots/bubble/bubblePlot.test.ts
```diff
@@ -55,59 +55,64 @@ const allAreaIds = ["MWI", "MWI_3_1", "MWI_3_2", "MWI_4_1", "MWI_4_2"];
...
+    return shallowMountWithTranslate(BubblePlot, store, {
+        props: {...props, ...customPropsData},
+        global: {
+            plugins: [store]
+        }
+    });
 };
 
-export const expectIndicatorSelect = (wrapper: Wrapper<Vue>, divId: string, value: string) => {
+export const expectIndicatorSelect = (wrapper: any, divId: string, value: string) => {
```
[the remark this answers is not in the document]
M-Kusumgar: hmmm I think I should just fix it here because I changed it in this PR! But I agree, so I think the reason was because now there is a distinction between DOMWrapper and VueWrapper and so we needed to be either more careful with types and perhaps allow either one in some cases or we could do `any`s but I am happy to tighten up the `any`s!

src/app/static/src/tests/components/plots/mapLegend.test.ts
```diff
@@ -20,26 +19,26 @@ describe("Map legend component", () => {
         min: 1
     };
 
-    const expectLevels = (levels: WrapperArray<Vue>) => {
+    const expectLevels = (levels: any) => {
```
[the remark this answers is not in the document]
M-Kusumgar: done, this one was actually a DOMWrapper! :)

src/app/static/src/tests/components/plots/mapLegend.test.ts
```diff
@@ -20,26 +19,26 @@ describe("Map legend component", () => {
...
+        expect(levels[1].text()).toBe("1.8");
+        expect(levels[2].text()).toBe("1.6");
+        expect(levels[3].text()).toBe("1.4");
+        expect(levels[4].text()).toBe("1.2");
+        expect(levels[5].text()).toBe("1");
     };
 
-    const expectIcons = (icons: WrapperArray<Vue>) => {
-        expect(icons.at(5).element.style
+    const expectIcons = (icons: any) => {
```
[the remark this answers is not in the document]
M-Kusumgar: done

src/app/static/src/tests/components/plots/table/table.test.ts
```diff
@@ -39,44 +40,54 @@ const propsData = {
...
     };
 
     it('renders a table', () => {
         const wrapper = getWrapper();
-        expect(wrapper.contains('div')).toBe(true);
-        expect(wrapper.contains('table')).toBe(true);
-        expect(wrapper.contains('br')).toBe(true);
+        expect(wrapper.find('div')).toBe(true);
+        expect(wrapper.find('table')).toBe(true);
+        expect(wrapper.find('br')).toBe(true);
```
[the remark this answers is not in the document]
M-Kusumgar: yes! thats my bad

src/app/static/src/tests/components/plots/testHelpers.ts
```diff
@@ -162,10 +160,10 @@ export const testData = {
     ]
 };
 
-export const expectFilter = (wrapper: Wrapper<Vue>, divId: string, value: string[], label: string, multiple: boolean, options: FilterOption[]) => {
+export const expectFilter = (wrapper: any, divId: string, value: string[], label: string, multiple: boolean, options: FilterOption[]) => {
```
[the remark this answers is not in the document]
M-Kusumgar: yep used VueWrapper here

src/app/static/src/tests/components/plots/table/table.test.ts
```diff
@@ -88,71 +99,71 @@ describe('Table from testdata', () => {
...
+        expect(wrapper.find('table')).toBe(false);
         expect(wrapper.text()).toContain('No data are available for these selections.')
     });
 
     it('renders correct markup when no data are available for selected filters in French', () => {
         const wrapper = getWrapperFr({
             filteredData: []
         });
-        expect(wrapper.contains('table')).toBe(false);
+        expect(wrapper.find('table')).toBe(false);
```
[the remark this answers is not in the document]
M-Kusumgar: done

src/app/static/src/tests/components/plots/table/table.test.ts
```diff
@@ -88,71 +99,71 @@ describe('Table from testdata', () => {
         const wrapper = getWrapper({
             filteredData: []
         });
-        expect(wrapper.contains('table')).toBe(false);
+        expect(wrapper.find('table')).toBe(false);
```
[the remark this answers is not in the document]
M-Kusumgar: done

src/app/static/src/tests/components/progress.test.ts
```diff
@@ -1,12 +1,12 @@
...
 import ProgressBar from "../../app/components/progress/ProgressBar.vue";
-import {BProgress, BProgressBar} from "bootstrap-vue";
+import {BProgress, BProgressBar} from "bootstrap-vue-next";
 
 describe("progress bar", () => {
 
     it("renders name without helptext", () => {
-        const wrapper = shallowMount(ProgressBar, {
-            propsData: {
+        const wrapper = mount(ProgressBar, {
```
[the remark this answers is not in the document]
M-Kusumgar: ahh I am not sure, I have reverted some of the mounts in the file to shallowMounts! the ones I have kept as mounts are necessary

src/app/static/src/tests/components/projects/projectHistory.test.ts
```diff
@@ -293,42 +309,42 @@ describe("Project history component", () => {
...
+        await expectTranslated(buttons[1], "Cancel", "Annuler", "Cancelar", store);
     });
 
     it("invokes deleteProject action when confirm delete", async () => {
         const wrapper = getWrapper(testProjects);
         const deleteLink = wrapper.find("#p-1").find(".project-cell.delete-cell").find("button");
-        deleteLink.trigger("click");
-        await Vue.nextTick();
+        await deleteLink.trigger("click");
+        await nextTick();
```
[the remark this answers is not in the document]
M-Kusumgar: no I do not! This was just a consequence of two find and replaces, one was the Vue.nextTick to nextTick and the other was finding .triggers and putting awaits in front of them! Have fixed this now

src/app/static/src/tests/components/projects/shareProject.test.ts
```diff
@@ -39,236 +39,259 @@ describe("ShareProject", () => {
...
 
         expect(modal.findAll("input").length).toBe(2);
-        const input2 = wrapper.find(Modal).findAll("input").at(1);
+        const input2 = wrapper.findComponent(Modal).findAll("input")[1];
         input2.setValue("goodemail");
-        input2.trigger("blur");
+        await input2.trigger("blur");
 
-        setTimeout(() => {
+        setTimeout(async () => {
```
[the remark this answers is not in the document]
M-Kusumgar: Yh I have just removed all `setTimeout`s in this file. I prefer async await anyway and cannot really see much advantage of `setTimeout`s in any of our tests!

Note: two tests actually failed when I removed the setTimeouts, they are "if email is valid, validation feedback is not shown and button enabled" and "if email is valid but cased differently, validation feedback is not shown and button enabled" tests. The reason was that they were looking for `.text-danger` and seeing if it had a `d-none` class (bootstrap display none) but this failed because we had it wrapped in a `v-if` so `.text-danger` was not being rendered at all, ive fixed these tests now!

src/app/static/src/tests/components/projects/shareProject.test.ts
```diff
@@ -39,236 +39,259 @@ describe("ShareProject", () => {
...
         expect(modal.findAll("input").length).toBe(2);
-        const input2 = wrapper.find(Modal).findAll("input").at(1);
-        input2.setValue("GOODEMAIL");
-        input2.trigger("blur");
+        const input2 = wrapper.findComponent(Modal).findAll("input")[1];
+        await input2.setValue("GOODEMAIL");
+        await input2.trigger("blur");
 
-        setTimeout(() => {
+        setTimeout(async () => {
```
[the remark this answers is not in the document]
M-Kusumgar: yep all fixed

src/app/static/src/tests/components/projects/versionStatus.test.ts
M-Kusumgar: yh I thought so too, looked around a bit and just found this really, not much online about it! yh kinda annoying that they decided to get rid of that

src/app/static/src/tests/components/projects/projectHistory.test.ts
M-Kusumgar: sounds like a job for regex!! will do this in a separate commit to your other changes as this may affect a fair number of files!

src/app/static/src/tests/components/dataExploration/dataExploration.test.ts
M-Kusumgar: done

r-ash (approved):
Looks great from my end, nice work slogging through this. Let's get this in!

src/app/static/src/tests/components/projects/projectHistory.test.ts
EmmaLRussell: ..I'm pretty sure it lets you cast element in that case, done it lots in WODIN e.g. [here](https://github.com/mrc-ide/wodin/blob/798dedaabf6b4275dcf34207ead26461dfd482b3/app/static/tests/unit/components/run/runTab.test.ts#L116)
Maybe in all those cases the element type is included in the selector so it's smart enough to work out that "button#run-btn" will give you an HTMLButtonElement? So we could update the selectors in those cases - looks like there's only really a handful of examples.

M-Kusumgar: yep I believe the "button" part of the string is making all the difference actually, should be able to get this working!

M-Kusumgar: actually apologies! I was being dumb, you can cast it, I was forgetting the `.element` :disappointed:

src/app/static/src/tests/components/adr/ADRKey.test.ts
```diff
@@ -154,7 +154,7 @@ describe("ADR Key", function () {
 
         await rendered.find("button").trigger("click");
 
-        expect(rendered.find("button").attributes("disabled")).toBe("");
+        expect((rendered.find("button").element as HTMLButtonElement).disabled).toBe(true);
```
r-ash, line 157: Oh this is nicer!

EmmaLRussell (approved):

src/app/static/src/tests/testHelpers.ts
```diff
@@ -67,33 +70,36 @@ export function expectAllMutationsDefined(mutationDefinitions: any, mutationTree
...
 }
 
-export function expectTranslatedWithStoreType<T extends TranslatableState>(element: DOMWrapper<any>,
+export async function expectTranslatedWithStoreType<T extends TranslatableState>(element: DOMWrapper<any>,
                                                                            englishText: string,
                                                                            frenchText: string,
                                                                            portugueseText: string,
                                                                            store: Store<T>,
                                                                            attribute?: string) {
+    const value = () => attribute ? element.attributes(attribute) : element.text();
```
LekanAnni, line 79: Why is this a function ?

src/app/static/src/tests/testHelpers.ts
```diff
@@ -111,14 +117,51 @@ export const expectChangeLanguageMutations = (commit: Mock) => {
...
+
+/*
+This are functions that let us easily put in the translate directive into mount and shallowMount test
+functions. There are ts-ignores as shallowMount/mount require a DefineComponent type as their first
+argument however if we replace the C generic type below to "C extends DefineComponent" then we get
+typescript errors when we use shallowMountWithTranslate with our normal vue components.
+
+Since this is just for testing, ts-ignore was used for this special case.
+*/
+// @ts-ignore
```
LekanAnni, line 139: Could you not ignore at all ?

LekanAnni (changes requested):
Looking really good. Please see inline comments.

src/app/static/jest.config.js
```diff
@@ -31,4 +31,11 @@ module.exports = {
     modulePaths: [
         "<rootDir>"
     ],
+    testPathIgnorePatterns: [
+        "<rootDir>/src/tests/components/header/fileMenu.test.ts",
+        "<rootDir>/src/tests/components/plots/.*.test.ts",
+        "<rootDir>/src/tests/components/plots/.*/.test.ts",
+        "<rootDir>/src/tests/components/app.test.ts",
+        "<rootDir>/src/tests/components/dataExplorationApp.test.ts",
```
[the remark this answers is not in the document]
M-Kusumgar: these will be fixed in future PRs, the ignore was mainly done to reduce noise while testing as most of the tests in these files fail. The plots especially will be done in the next 2-3 PRs

src/app/static/src/tests/testHelpers.ts
M-Kusumgar: not sure, I have not actually changed this at all from when this was originally implemented, can do an opportunistic commit to clean that up!

M-Kusumgar: ahh just tried it and realised why! we need this to be a function because we only want to get the value once the state has updated, if this wasn't a function then it would store the english string in value and when we change the language to french, value will not update!

src/app/static/src/tests/testHelpers.ts
M-Kusumgar: the comment above this line was trying to say that if I remove the ignore then mount has type constraints for the component argument (it needs to extend type `DefineComponent`) however if I add that constraint to the generic type `C` then when I use the `mountWithTranslate` function, typescript complains about the type of our Vue components, since this is testing, I thought the easiest way is just to ts-ignore this and we just have to make sure that we put in vue components into the mount functions

LekanAnni (approved):

src/app/static/src/tests/testHelpers.ts
```diff
@@ -111,14 +117,51 @@ export const expectChangeLanguageMutations = (commit: Mock) => {
...
+            directives: {
+                ...options?.global?.directives,
+                translate: translate(store)
+            }
+        }
+    });
+}
+
+// @ts-ignore
+export function mountWithTranslate<T extends TranslatableState, C>(component: C, store: Store<T>, options?: any): VueWrapper<InstanceType<C>> {
```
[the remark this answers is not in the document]
M-Kusumgar: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4322/Removing-ts-ignores-in-mountWithTranslate-and-shallowMountWithTranslate-functions-in-testing

I have made a ticket about this, I think after the upgrade there can be a bunch of cleanup that we can do incrementally so would be good to remember this issue

PR #846, opened 2023-05-31
mrc-4272 Link to Projects page from Invalid State dialog, and refactor into separate component

EmmaLRussell:
## Description

This branch adds a link to the Projects page from the dialog reporting load failure due to invalid state. This link is only available to logged in users as guest users do not have access to the Projects page. 

Adding this link revealed an issue which meant that the dialog was still displayed after routing to Projects. The best way to deal with this seemed to be to separate out the old 'LoadError' dialog (which *can* be displayed on the Projects page if failure occurs in the early stages of load attempt e.g. couldn't read zip file) from the invalid state dialog behaviour (which should only be displayed in the main stepper page when state has been partially loaded). Invalid state dialog content has been pulled out into a new component, LoadInvalidModal, which is embedded in the Stepper component, so can only be shown when that is mounted, This also makes the dialog templates cleaner, without a lot of `v-if`s separating the two behaviours. 

The new dialog is only displayed iff there are `invalidSteps` in the state, *not* if load state is `LoadFailed`, which triggers the display of the generic Load Failed dialog. 

I also fixed an issue from the previous branch where a new version would be generated on rollback in all cases - this actually shouldnt happen for guest users, who do not have multiple project versions available. Guidance on the dialog has also been updated to reflect this change. 

@M-Kusumgar suggested adding a 'x' close button to the dialog - I haven't done this, partly because we actually don't use the pattern anywhere else in the app. All other dialogs are exited using the main buttons in footer. But I can add a ticket for this if people would still prefer it to be added. 

Can be manually tested in the same way as the previous branch e.g. by sabotaging the ModelRun /result endpoint.

## Type of version change
_Delete as appropriate_

Minor

## Checklist


M-Kusumgar (changes requested):
Hey Emma, looks great! Good idea with putting the modal as part of the stepper so it automatically unmounts! There is an issue I noticed though, if you click the projects page link after getting a load error and then try and load that same project again, the modal is showing at the start and then project load happen! Probably quite a quick fix but yh would be nice not to have this glitchy behaviour!

EmmaLRussell:
> Hey Emma, looks great! Good idea with putting the modal as part of the stepper so it automatically unmounts! There is an issue I noticed though, if you click the projects page link after getting a load error and then try and load that same project again, the modal is showing at the start and then project load happens! Probably quite a quick fix but yh would be nice not to have this glitchy behaviour!

Yeah, I think that's going to be the same issue as we've had for a long time with state momentarily appearing on load before error dialog/spinner is shown. I'll have a dig into it, might just be a case of rearranging things a bit!

EmmaLRussell:
OK, I think this should be better - it no longer shows the dialog when the project is in an unready state as far as the stepper is concerned - you should just see the loading spinner in that case (unready just means it's still sorting itself out - it can be 'ready' and failed). I've also made this apply to the warning alerts too as they were distracting as well.

M-Kusumgar (approved):
Looks good to me! :technologist:

PR #847, opened 2023-06-13
Mrc-4223 Choropleth bugs Vue 3 upgrade

M-Kusumgar:
## Problems/motivation:
- Choropleth map breaks when you tried to change datasource in review inputs
- Changing detail level when you have an area selected (e.g. central) causes weird behavour, some regions outside the selected region can become coloured or regions sometimes have the wrong colour.

## New features/solutions:
- Changing datasource was an issue because the type of datasource passed into `selectDataSource` method was incorrect
- Treeselect for indicator in map control was not updating its value when the datasource changed so had to use a key to force it to re-render when the `indicatorOptions` were changed.
- vnode-updated is an event that fires off after the updated lifecycle hook of the child element, this was used to update the bounds after any change and remove unnecessary watchers
- key is changed to `area_id` in `properties` of the feature as `feature.id` doesn't actually exist for a lot of features. Not sure how we were getting away with all that!

- Should be able to change all opitons
- Should be able to change DataSource
- Map control should update indicator when DataSource is changed
- Should be able to change detail level in conjunction with setting specific area (e.g. northern)
- Coloured regions should be within the specified area
- No data should show if you select area to be `northern` and detail level to be `country` for example

## How to run app:

## Checklist

### DIFF
--- a/src/app/static/src/app/components/plots/MapControl.vue
+++ b/src/app/static/src/app/components/plots/MapControl.vue
@@ -5,6 +5,9 @@
                 <div v-if="showIndicators" class="row form-group">
                     <label class="col-3 col-form-label" v-translate="'indicator'">
                     </label>
+                    <!-- The indicator modelValue was not updating when the
+                    indicator options were changing so added indicator options
+                    as key to force re-render -->
                     <div class="col" :key="JSON.stringify(indicatorOptions)">
                         <treeselect  :modelValue="indicator"
                                      :multiple="false"
--- a/src/app/static/src/app/components/plots/MapControl.vue
+++ b/src/app/static/src/app/components/plots/MapControl.vue
@@ -5,7 +5,7 @@
                 <div v-if="showIndicators" class="row form-group">
                     <label class="col-3 col-form-label" v-translate="'indicator'">
                     </label>
-                    <div class="col">
+                    <div class="col" :key="JSON.stringify(indicatorOptions)">
                         <treeselect  :modelValue="indicator"
                                      :multiple="false"
                                      :clearable="false"
--- a/src/app/static/src/app/components/plots/choropleth/Choropleth.vue
+++ b/src/app/static/src/app/components/plots/choropleth/Choropleth.vue
@@ -6,7 +6,7 @@
                  @update="onFilterSelectionsChange"></filters>
         <div id="chart" :class="includeFilters ? 'col-md-9' : 'col-md-12'">
             <l-map ref="map" style="height: 800px; width: 100%" @ready="updateBounds" @vnode-updated="updateBounds">
-                <template v-for="feature in currentFeatures" :key="feature">
+                <template v-for="feature in currentFeatures" :key="feature.properties.area_id">
                     <l-geo-json :geojson="feature"
                                 :options="options"
                                 :optionsStyle="{...style, fillColor: getColor(feature)}">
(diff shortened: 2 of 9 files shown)


EmmaLRussell (commented):
Looks really good! 

It seems that there are still some issues with initialising the filter options - when choropleth first loads, Sex, Age and Household Survey are all shown as (unknown) - presumably they all do have selected values though, in order to show any data on the choropleth? Similarly when select ART or ANC - in the case of ANC, the year shows the placeholder rather than (unknown).   This also affects the output choropleth. Will this be addressed in a future ticket?

M-Kusumgar:
> Looks really good! 
> 
> It seems that there are still some issues with initialising the filter options - when choropleth first loads, Sex, Age and Household Survey are all shown as (unknown) - presumably they all do have selected values though, in order to show any data on the choropleth? Similarly when select ART or ANC - in the case of ANC, the year shows the placeholder rather than (unknown).   This also affects the output choropleth. Will this be addressed in a future ticket?

Yep the initializing the filters is another ticket! That is consistent across basically all of the filters in other plots too, so should be an independent fix

src/app/static/src/app/components/plots/MapControl.vue
```diff
@@ -5,7 +5,7 @@
                 <div v-if="showIndicators" class="row form-group">
                     <label class="col-3 col-form-label" v-translate="'indicator'">
                     </label>
-                    <div class="col">
+                    <div class="col" :key="JSON.stringify(indicatorOptions)">
```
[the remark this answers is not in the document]
M-Kusumgar: done

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -5,27 +5,31 @@
...
+                                :indicators-metadata="indicators"
+                                :level-labels="featureLevels"
+                                @detail-changed="onDetailChange"
+                                @indicator-changed="onIndicatorChange"></map-control>
+                    <map-legend v-show="!emptyFeature"
+                                :metadata="colorIndicator"
+                                :colour-scale="indicatorColourScale"
+                                :colour-range="colourRange"
+                                @update="updateColourScale"></map-legend>
+                </template>
```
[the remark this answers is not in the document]
M-Kusumgar: so I can remove the template nesting on the `reset-map` part but not this bigger chunk, if you change a filter so that `emptyFeature` is true you actually get an "insertBefore on Node" DOMException. No idea why!

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -417,22 +419,15 @@
...
-                    this.updateBounds();
-                },
+        watch: {
                 filters: function () {
                     this.initialise();
                 },
                 indicators: function () {
                     this.initialise();
                 },
             },
```
LekanAnni, line 429: Do you not need to watch if the value has changed before calling these functions ?

M-Kusumgar: Do you mean the `initialised` and the `selectedAreaFeatures` watchers I have gotten rid of? These I have gotten rid of because the `updateBounds` function is called when the map fires `@ready` and `@vnode-updated` and with those events these watchers become redundant.

EmmaLRussell (approved):

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
LekanAnni: I guess I got confused, the previous implementation does have some like this `filters: function (newVal: boolean) {this.initialise(); }`

So you basically removed the `newVal` because it was not used, right ?

PR #849, opened 2023-06-20
mrc-4315 Include invalid step details in new version note when rollback

EmmaLRussell:
## Description

This branch includes the invalid step details in the new project version generated on rollback to last valid step. I've added a `stepTextKeys` getter to the stepper module to avoid some duplicated logic across the rollback action and loadInvalidModal, which both display invalid step details. 

## Type of version change
Minor

## Checklist


r-ash (approved):
This looks good, stepTextKeys helper looks great

M-Kusumgar (approved):
Works like a charm!

PR #850, opened 2023-06-21
mrc-4338 Display Input Time Series plots with single values

EmmaLRussell:
## Description

With the latest Angola data (see zip file attached to ticket), ART data was not being plotted on the Input Time Series, because it has only a single data point. 

This was because of the Generic Plot config [jsonata](https://jsonata.org/), which is transformed into the full json passed to plotly by injecting the plot data. Jsonata has a tendency to convert arrays with single values in them into scalar values which are not in arrays. However, plotly requires json arrays for its data values. 

You can force jsonata to use an array for single values by wrapping the expression in square brackets, and happily this also works for multi-value arrays - it won't nest the array in a further array. Hence the change here which simply forces the x and y values for each subplot to be an array. 

We don't need to do this for the 'highlighted' values in red, as these are already calculated in pairs of values so you'll always have at least two. 

The resulting single-value plots are pretty useless as time series plots since there isn't a time series as such, but this is probably the expected behaviour, and better than not showing the values at all!

Updated the integration tests to include a further area with a single data point. 

## Type of version change
_Delete as appropriate_

Patches

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 2.37.1
+
+* Fix Input Time Series plots with single values
+
 # hint 2.37.0
 
 * Auth0 single sign on authorization integration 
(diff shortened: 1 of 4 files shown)


LekanAnni (approved):
LGTM 👍

PR #851, opened 2023-06-21
Mrc 4221 Fix treeselect render issues

M-Kusumgar:
## Problems/motivation:
- State is completely fine however the filters don't render the correct values on mount. They work just fine after the user manually selects something.

## New features/solutions:
- Changed `modelValue` to `model-value` to be consistent throughout hint. (opportunistic cleanup)
- Created a wrapper to fix treeselect component. A boolean is used to re-render the treeselect on mount if it requires an update. We don't want to re-render when the user selects something because this causes an error in the treeselect component. This is probably because our re-render happens before the component itself has time to finish all its lifecycle hooks. To bypass this, we only re-render them on mount and not when the user selects something. Using watchers this component will only allow re-renders when the treeselect options change (and this is not possible when the user is selecting something from the options).

- All treeselects in the app should work normally

## How to run app:

## Checklist


src/app/static/src/app/components/TreeSelect.vue
```diff
@@ -0,0 +1,57 @@
...
+        <slot></slot>
+    </treeselect>
+</template>
+<script lang="ts">
+    import { PropType, defineComponent } from "vue";
+    import Treeselect from "vue3-treeselect";
+
+    export default defineComponent({
+        components: {
+            Treeselect
```
EmmaLRussell, line 16: I wonder if it would be better to call this something more obviously different from the internal 3rd party component e.g. HintTreeSelect? `<treeselect>` and `<tree-select>` are pretty easily confused! (Probably not HintTreeSelect if we're going to pull this out into its own package, though not sure if this is on the cards..)

EmmaLRussell (commented):
Nice, I think a wrapper like this is definitely the way to go.  Could do with a couple of comments I think, explaining the situation - this code is really clear but it may not be obvious to a future HINT explorer why this wrapper was required. 

 I think deciding whether to rerender based on the options is not the most general solution - we may not currently do this anywhere in HINT, but you can imagine the selected value changing in the store but the options NOT changing e.g. same set of options shared between filters, but each indicator has a separate selected value which is persisted.  

So rather than using the options at all, could we keep a `lastEmitted` data value, set in the `input` method , then check that against any updated value we get in the `modelValue` watcher? And re-render if those values differ, as we'll know this is an externally triggered value change? I think then we could lose `preventRerender`?

Also, the way the `reRender` key works is that it's simply a boolean toggle, so we force an update in the internal treeselect by changing the toggle value? I find this a little unintuitive as I think of bools as flags - I think I'd tend to use an integer value that increments for a key like this - but happy with a bool toggle (with a comment!) if you'd rather.

src/app/static/src/app/components/TreeSelect.vue
M-Kusumgar: agreed, I have gone with HintTreeSelect for now!

M-Kusumgar:
> Also, the way the `reRender` key works is that it's simply a boolean toggle, so we force an update in the internal treeselect by changing the toggle value? I find this a little unintuitive as I think of bools as flags - I think I'd tend to use an integer value that increments for a key like this - but happy with a bool toggle (with a comment!) if you'd rather.

Correct, I just thought it would be better to use a bool because of memory. It has always seems very icky to me to have an integer without an upper bound increment (hypothetically indefinitely) for something that doesn't actually care about the value of the integer. However I realised I was being dumb by then interpolating into a string, so instead re-render is now a number but instead of incrementing it toggles between 0 and 1 by using function x -> 1-x and this is way cleaner I think

M-Kusumgar:
should be ready to review now, lots of random pain with this one!

src/app/static/src/app/components/HintTreeSelect.vue
```diff
@@ -0,0 +1,81 @@
...
+            modelValue: {
+                handler: function(newVal, oldVal) {
+                    // Multi-select component already seems to work fine so we want to
+                    // not re-render those. We only want to re-render if:
+                    // newVal === null : to replace value with "Not used" in treeselect
+                    // newValStringCase : to re-render when we pass string into treeselect
+                    // newValArrayCase : to re-render when we pass array of length 1 into treeselect
+
+                    const newValStringCase = (typeof newVal === "string") && this.lastEmittedValue !== newVal;
+                    const newValArrayCase = newVal instanceof Array && newVal.length === 1 && this.lastEmittedValue !== newVal[0];
```
EmmaLRussell, line 69: I don't quite see how this works together with the parameter type of `input` being a string[] - if last emitted was ["someValue"] and ["someValue"] is the newVal here, `newValArrayCase` would wrongly be true here since ["someValue"] !== "someValue". Should the `value` parameter type of `input` actually be `string` - or does the interface specify `string[]` but you actually get strings? Should `string` also be the type of `lastEmittedValue`?

M-Kusumgar: Sorry I should have really included an explanation for this in the PR comment. After some investigation, I found the third-party treeselect component *always* emits a `string` on input. When things happen in our store though, some `newVal`s are `string[]` of length 1 and some are `string`. Not sure where this inconsistency came from but equally I would say this is not worth fixing as:
1. There might actually be a reason for this
2. This would involve checking every state that stores a possible treeselect value and potentially altering the type, which seems OTT for a render issue 😆 
So I have included these two cases!

LekanAnni (changes requested):
Hi Mantra, looking at our last conversation regarding this option. We have used multiSelect else where with treeSelect, this works. https://github.com/reside-ic/vue-next-dynamic-form/blob/482517e0e62bf2d236fcc5d37d6bde83e66548ec/src/DynamicFormMultiSelect.vue#L31


I am not convinced that it would not work for this multiSelect. Instead of creating a wrap around treeSelect can we try to make this work ?

M-Kusumgar:
> I am not convinced that it would not work for this multiSelect. Instead of creating a wrap around treeSelect can we try to make this work ? 

Sorry also should've explained this, so there are two problems with the treeselect component:
1. The treeselect value on mount is displayed as (unknown) this is actually a simple thing to fix, you just wait for a value and then render the treeselect, this is the issue that is not a problem in dynamic form because the formdata already exists and then we render dynamic form.
2. The treeselect doesn't update the value when an external update (maybe in store) changes it, for example if you change the data source value on the choropleth map. This second problem is not one the dynamic form has to deal with which is why the treeselects in it work. This PR mainly focuses on fixing this second problem and the first problem is just fixed for free with this method.

LekanAnni (approved):
LGTM

EmmaLRussell (approved):

PR #855, opened 2023-06-26
mrc-4341 Reload on retry if user is guest

EmmaLRussell:
## Description

"Retry" was not working from the Load Failed on invalid state dialog when user was guest. This was because it was attempting to load project by requested id from backend - however this load path is not available to guest users, who can only reload from current session (and current local storage state in front end). We can retry the load process for guest users by simply reloading the browser location, 

## Type of version change
_Delete as appropriate_

Patches

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 2.37.1
+
+* Fix Retry load for guest user
+
 # hint 2.37.0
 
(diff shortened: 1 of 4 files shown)


LekanAnni (approved):
LGTM 👌

r-ash (approved):

PR #856, opened 2023-06-27
Rest of tests

M-Kusumgar:
## Problems/motivation:
- Time to get all of the tests working in Hint

## Guide for reviewers
This branch is still ahead but just putting it up for review. Note we are only focusing on the commits from "data exploration done" commit. The earlier ones are different PRs that just haven't caught up! We will be splitting off the commits because there are a lot of miscellaneous bits of tests so thought it is more efficient to combine into one bigger PR. This is the relevant [diff](https://github.com/mrc-ide/hint/compare/1df91de..47c8525). Perhaps we can split like this:

# Lekan
Commits:
- data exploration done
- forgot password done
- reset password done
- finish off adding all modes

Notes:
- We changed the way the app was mounted to `<div id="app"></div>`  in the passing data to freemarker templates PR ([reminder](https://github.com/mrc-ide/hint/pull/836)). This means that we would need a way of changing what mode was mounted on to the div in the freemarker template but the templates are identical (the forgot password and reset password are basically pure html, really this difficulty was between normal app and data exploration app). Therefore I went with the strategy of giving the divs different IDs to switch which components are mounted. The index file then just checks which div is present and returns that app from the `getApp` function.

# Emma
The rest of the commits (they are smaller but more numerous)

Notes:
- File shenanigans we have discussed. This creates mocks for File related constructors as jest doesn't include that
- Had to mock vue leaflet components since fully mounting them was difficult and returned errors
- Misc tests:
  - UploadModal.vue: if you look at the `outputFileError()` returns you will see that we actually either return a key or a translated string, I think if was working before because v-translate would just display the text as it was if it didn't have the corresponding key but not anymore
  - ProjectHistory.vue: v-b-toggle was broken so I just custom made the arrow icon button to show the versions in the projects history page
  - locales.ts - spacings weren't showing up on "create a new project or return to current project", not sure what the issue was but I have added the spacing in. These keys arent used anywhere else so shouldn't impact anything
  - The `detail: null` bits are because of a type change earlier. Just adjusting test assertions to make sure types matched
- testRegex and testMatch - had to do this as jest doesn't allow you to have both properties at the same time so we have to be consistent (testMatch used in base jest config file)
- Adding as any to defineComponentVue2 functions - the compiler was complaining about types again so I put as any on the inside of the function. The reason it is complaining about types is because I haven't defined the full type with all the generics (<_> having 4 arguments instead of 8-10 that the full version has and the defaults don't match up towards the end or something ). Can have a chat about this but I don't think it's a huge issue as the type assertion on the arguments still means we have the full power of typescript!

## How it should behave
All tests should pass! We already have a green build on buildkite but please do test it on your local too!

## Checklist


EmmaLRussell (commented):
Just on the file shenanigans, it was bugging me to not be able to test the trigger behaviour, so I had a bash around of the test code, and I *think* I got it working using spyOn:
```
jest.spyOn((wrapper.vm.$refs as any).pjnz, "files", "get").mockImplementation(() => [testFile]);
await wrapper.find("input").trigger("change");
...etc
```

Do check that I'm not just hallucinating, but I think that should mean we can do without splitting up the handler in the component, and the global mocking in the test file.

M-Kusumgar:
> Do check that I'm not just hallucinating, but I think that should mean we can do without splitting up the handler in the component, and the global mocking in the test file.

Just tested it out! It works :tada: ! Thanks for that, it was bugging me too but I didn't know about this usage of `jest.spyOn`!

src/app/static/src/tests/components/plots/mapControl.test.ts
```diff
@@ -33,42 +42,44 @@ describe("Map control component", () => {
...
     });
 
 
-    it("emits indicator-changed event with indicator", () => {
+    it("emits indicator-changed event with indicator", async () => {
         const wrapper = mount(MapControl, {props, global: {plugins: [store]}});
-        wrapper.findAllComponents(TreeSelect)[0].vm.$emit("input", "art_coverage");
-        expect(wrapper.emitted("indicator-changed")![0][0]).toBe("art_coverage");
+        wrapper.findAllComponents(HintTreeSelect)[0].vm.$emit("update:model-value", "art_coverage");
+        await nextTick();
```
EmmaLRussell, line 74: Are you sure you need this `nextTick` - I thought `emit`s were synchronous? I tried running locally without it and it passed.

src/app/static/src/tests/components/plots/mapControl.test.ts
```diff
@@ -33,42 +42,44 @@ describe("Map control component", () => {
...
+        expect(wrapper.emitted("indicatorChanged")![0][0]).toBe("art_coverage");
     });
 
-    it("emits detail-changed event with detail", () => {
+    it("emits detail-changed event with detail", async () => {
         const wrapper = mount(MapControl, {props, global: {plugins: [store]}});
-        wrapper.findAllComponents(TreeSelect)[1].vm.$emit("input", 3);
-        expect(wrapper.emitted("detail-changed")![0][0]).toBe(3);
+        wrapper.findAllComponents(HintTreeSelect)[1].vm.$emit("update:model-value", 3);
+        await nextTick();
```
EmmaLRussell, line 81: ...same here

src/app/static/src/tests/components/projects/projectHistory.test.ts
```diff
@@ -256,20 +256,20 @@ describe("Project history component", () => {
         await expectTranslated(headers[7], "Copy to", "Copier", "Copiar para", store);
         await expectTranslated(headers[8], "Share", "Partager", "Partilhar", store);
 
-        expect(true).toBe(false)
+        // expect(true).toBe(false)
```
EmmaLRussell, line 259: extraneous comment?

src/app/static/src/tests/directives/translate.test.ts
```diff
@@ -211,13 +211,17 @@ describe("translate directive", () => {
...
+        expect((renderedMultiple.element as any).__lang_unwatch__).toHaveProperty("value")
         renderedAttribute.unmount();
-        expect((store._watcherVM as any)._watchers.length).toBe(4);
+        expect((renderedAttribute.element as any).__lang_unwatch__).toStrictEqual({})
         renderedText.unmount();
-        expect((store._watcherVM as any)._watchers.length).toBe(3);
+        expect((renderedText.element as any).__lang_unwatch__).toStrictEqual({})
         renderedMultiple.unmount();
-        expect((store._watcherVM as any)._watchers.length).toBe(0);
+        expect((renderedMultiple.element as any).__lang_unwatch__).toStrictEqual({})
```
EmmaLRussell, line 224: Wow, this is even more gribbly than before. Can we really not view these values from the store any more?

EmmaLRussell (commented):
Comments for:
- leaflet map fixes
- misc tests

> I think if was working before because v-translate would just display the text as it was if it didn't have the corresponding key but not anymore

Not sure why that behaviour would have changed? I think that would be useful to retain if possible, especially for error strings etc which might originate in hintr, rather than each component having to check if something is an expected  key or not.

EmmaLRussell (commented):
All the other commits assigned to me look good :+1:

src/app/static/src/tests/components/plots/mapControl.test.ts
M-Kusumgar: youre right! done

src/app/static/src/tests/components/projects/projectHistory.test.ts
M-Kusumgar: removed

src/app/static/src/tests/directives/translate.test.ts
M-Kusumgar: ive tried to explore this a bit more and so far I cannot really see any way of getting watcher info from the store :( do have a go if you have time but I think theyve just change the way they have implemented store watchers because they don't seem to be storing any of it on the store object

LekanAnni (changes requested):
Looking good,

Please see my inline comments. Code coverage is complaining, you might want to improve the test coverage.

src/app/static/src/tests/dataExploration/router.test.ts
```diff
@@ -26,39 +27,68 @@ console.error = jest.fn();
...
+
+jest.mock("../../app/components/dataExploration/DataExploration.vue", () => ({
+    name: "DataExploration",
+    template: "<div id='data-exploration-stub'/>"
+}))
+
+jest.mock("../../app/components/Accessibility.vue", () => ({
+    name: "Accessibility",
+    template: "<div id='accessibility-stub'/>"
+}))
```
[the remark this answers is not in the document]
M-Kusumgar: yes you are correct! have removed it :)

src/app/static/src/tests/components/files/fileUpload.test.ts
```diff
@@ -1,17 +1,39 @@
...
 import {emptyState} from "../../../app/root";
 import registerTranslations from "../../../app/store/translations/registerTranslations";
 import {BDropdown} from "bootstrap-vue-next";
-import {expectTranslatedWithStoreType, shallowMountWithTranslate} from "../../testHelpers";
+import {expectTranslatedWithStoreType, mountWithTranslate, shallowMountWithTranslate} from "../../testHelpers";
 import {DataExplorationState, initialDataExplorationState} from "../../../app/store/dataExploration/dataExploration";
 
 describe("File upload component", () => {
 
+    // @ts-ignore
```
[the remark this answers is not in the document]
M-Kusumgar: So jest does not have `File` or `FormData` classes hence we need to mock them. We however do not need all the info these typically provide such as when a file was last updated. I can remove the `// @ts-ignore` and add an `as any` to the class if that is better?

src/app/static/src/tests/components/files/fileUpload.test.ts
```diff
@@ -1,17 +1,39 @@
...
+    global.File = class MockFile {
+        filename: string;
+        constructor(parts: (string | Blob | ArrayBuffer | ArrayBufferView)[], filename: string, properties ? : FilePropertyBag) {
+          this.filename = filename;
+        }
+    }
```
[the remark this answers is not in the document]
M-Kusumgar: Yh actually this was extraneous, don't need `File` defined in this file specifically, need `FormData` though!

src/app/static/src/tests/components/files/fileUpload.test.ts
```diff
@@ -1,17 +1,39 @@
...
+
+    // @ts-ignore
```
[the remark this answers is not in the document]
M-Kusumgar: yep earlier comment about `as any` option

M-Kusumgar:
> Code coverage is complaining, you might want to improve the test coverage.

I will improve the code coverage but ignore any complains about it from the vue-charts or vue-dynamic-form folders, these will not need testing right now as we will bring in the updated packages from npm in a later pull request

src/app/static/src/tests/directives/translate.test.ts
EmmaLRussell: Fair enough!

M-Kusumgar:
> Not sure why that behaviour would have changed? I think that would be useful to retain if possible, especially for error strings etc which might originate in hintr, rather than each component having to check if something is an expected key or not.

this is a good idea, I think the fix should be simple, just adding the default behaviour but I have created a ticket for it for now, will address this in a couple PRs though! https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4471/Translate-tooltip-default-string

M-Kusumgar:
Code Coverage @LekanAnni :
- ignore HintDataExploration complaints, have fully tested this
- Upload Modal complaints are fixed too
- downloadIndicator lines are in fact covered
- projectHistory complaints are fixed in the next with some tests PR #857

src/app/static/src/tests/directives/translate.test.ts
```diff
@@ -211,13 +211,17 @@ describe("translate directive", () => {
...
         });
-        expect((store._watcherVM as any)._watchers.length).toBe(5);
+        expect((renderedAttribute.element as any).__lang_unwatch__).toHaveProperty("value")
+        expect((renderedText.element as any).__lang_unwatch__).toHaveProperty("innerHTML")
+        expect((renderedMultiple.element as any).__lang_unwatch__).toHaveProperty("innerHTML")
+        expect((renderedMultiple.element as any).__lang_unwatch__).toHaveProperty("placeholder")
+        expect((renderedMultiple.element as any).__lang_unwatch__).toHaveProperty("value")
         renderedAttribute.unmount();
-        expect((store._watcherVM as any)._watchers.length).toBe(4);
+        expect((renderedAttribute.element as any).__lang_unwatch__).toStrictEqual({})
```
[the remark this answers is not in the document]
LekanAnni: Seems like a library test. kind of unnecessary.

buildkite/shared-build-env.dockerfile
```diff
@@ -2,10 +2,10 @@ FROM mrcide/node-docker:master
 
 RUN apt-get install wget
 
-RUN wget -qO - https://adoptopenjdk.jfrog.io/adoptopenjdk/api/gpg/key/public | apt-key add -
-RUN add-apt-repository --yes https://adoptopenjdk.jfrog.io/adoptopenjdk/deb/
+RUN wget -qO - https://packages.adoptium.net/artifactory/api/gpg/key/public | apt-key add -
+RUN add-apt-repository --yes https://packages.adoptium.net/artifactory/deb/
 RUN apt-get update
-RUN apt-get install adoptopenjdk-11-hotspot -y
+RUN apt-get install temurin-11-jdk -y
```
LekanAnni, line 8: For context, adoptopenjdk-11-hotspot was moved, probably renamed, so Mantra updated this line to point to a new location.

LekanAnni (approved):
LGTM 👍 

Thanks for making the changes

EmmaLRussell (approved):
Looks good, with remaining issues now moved to separate tickets.

PR #861, opened 2023-08-03
Update ADR integration to pull 2024 datasets

r-ash:
## Description

Ian has asked us to switch to pull 2024 datasets. ADR have added the new type, they are just working on migrating datasets over now

## Type of version change

None

## Checklist


r-ash:
image

Have seen this working on locally deployed app. I've also created private test dataset https://adr.unaids.org/country-estimates-24/antarctica-country-estimates-2024-1-2 and the public dataset https://adr.unaids.org/country-estimates-24/antarctica-country-estimates-2024-1

LekanAnni (approved):
LGTM

M-Kusumgar (commented):
Looks good I can see the private and public dataset when I use my ADR key. However, should I not be able to see the public one when I use "1" as my ADR key for example?
image

r-ash:
> Looks good I can see the private and public dataset when I use my ADR key. However, should I not be able to see the public one when I use "1" as my ADR key for example? image

Thanks yeah I noticed that too - not sure why that is happening, I guess some permission issue. I'll ask the ADR guys if there is some other setting I need to change.

r-ash:
Ah me just being silly - I forgot to make the resources public.. that'll do it.

Working now

image

PR #864, opened 2023-08-15
add packages + remove dynamic form folder

M-Kusumgar:
## Description
This branch aims to remove the vue-dynamic-form folder from hint and use our refactored vue-next-dynamic-form package instead. It also aims to combine the vue-charts folder into Hint instead of using the reside package.

This one should actually be super small to review :laughing:


M-Kusumgar:
ADR is down, cannot really do anything about it but will open it up for review anyway! i think the code changes ive made are all independent!

M-Kusumgar:
> LGTM
> 
> So the vue-charts change was just to remove it from the package.json? It's still being used within HINT in the same location, right?
> 
> Were the '?' and '(required)' labels always glued together like this or has this crept in recently? image

Yep the vue charts thing was just to remove it from package.json, the "?" and "(required)" were not together, just a visual thing i can fix in vue-next-dynamic-form very quickly but yh can be done at any time!

LekanAnni (approved):
LGTM

PR #866, opened 2023-08-17
mrc-3328 Keep user's language preference outside project state

LekanAnni:
## Description

This removes language state from getting stored in the backend and local storage. Add new key to local storage and monitor language alongside language state. The instruction on the ticket is a lot miss leading, I have been trying to solve this differently, trying to remove language from the state completely. 

We still need language to be in the state for reusability across the app, This PR keeps track of user language outside of projects, with additional key in local storage. 

## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/app/localStorageManager.ts
```diff
@@ -64,6 +63,16 @@ declare const currentUser: string;
 
 export class LocalStorageManager {
 
+    saveLanguageState = (lang: Language) => {
+        localStorage.setItem("language", lang);
+    }
+
+    getLanguageState = (): Language => {
```
EmmaLRussell, line 70:
```suggestion
    getLanguage = (): Language => {
```

src/app/static/src/app/localStorageManager.ts
```diff
@@ -64,6 +63,16 @@ declare const currentUser: string;
 
 export class LocalStorageManager {
 
+    saveLanguageState = (lang: Language) => {
```
EmmaLRussell, line 66:
```suggestion
    saveLanguage = (lang: Language) => {
```

src/app/static/src/app/localStorageManager.ts
```diff
@@ -83,7 +92,12 @@ export class LocalStorageManager {
         const appStateKey = getAppStateKey(dataExplorationMode);
         const item = window.localStorage.getItem(appStateKey);
         if (item) {
-            return JSON.parse(item) as Partial<RootState>;
+            const partialRootState = JSON.parse(item) as Partial<RootState>;
+            /**
+             * For smooth compatibility, code uses local storage persisted language state,
+             * and default to English if needed.
+             */
+            return {...partialRootState, ...{language: this.getLanguageState()}}
```
EmmaLRussell, line 100:
```suggestion
            return {...partialRootState, language: this.getLanguageState()}
```

EmmaLRussell (commented):
Looks great. Correctly reloads a project with the user's new language preference if it has changed since the project was saved. I think leaving the state value in place but loading it from local storage is a good call for minimising the code changes, 

One thing it doesn't do, which was suggested in the ticket, was to save the user's language preference from the Forgot Password page. But I think that;s not a high priority.

M-Kusumgar (changes requested):
Code looks good, little bit confused about functionality. Now the language is just updated in local storage so the app doesn't update when you select a new language from the dropdown? Think we do need to retain that behaviour because currently if i select "fr" from the dropdown, nothing actually happens until i refresh the page

M-Kusumgar (approved):

src/app/static/src/app/localStorageManager.ts
LekanAnni: Done.

news.md.save
```diff
@@ -0,0 +1,1026 @@
+# hint 2.39.0
```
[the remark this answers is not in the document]
LekanAnni: Done

src/app/static/src/app/localStorageManager.ts
LekanAnni: Done

EmmaLRussell (approved):

PR #867, opened 2023-08-18
mrc-4538 break calibrate result response into data and metadata

LekanAnni:
## Description

This PR separates calibrate result response endpoints and update usage in frontend. I added backward compatibility for legacy projects. 

## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/app/store/load/actions.ts
```diff
@@ -102,8 +102,22 @@ export const actions: ActionTree<LoadState, RootState> & LoadActions = {
         //storage then reload the page, to follow exactly the same fetch and reload procedure as session page refresh
         //NB load state is not included in the saved state, so we will default back to NotLoading on page reload.
 
-        // Backwards compatibility fix: projects which calibrated before bug fix in mrc-3126 have empty calibrate options
+
         const {modelCalibrate} = savedState
+
+        // Backward compatibility
```
EmmaLRussell, line 108: As with the other backward compatibility comment, it might be nice to reference the ticket in case it's not obvious what the relevant change was:
```suggestion
        // Backward compatibility fix: projects which calibrated before mrc-4538 have metadata embedded in result
```

src/app/static/src/app/store/modelCalibrate/mutations.ts
```diff
@@ -29,10 +30,10 @@ export enum ModelCalibrateMutation {
     ComparisonPlotStarted = "ComparisonPlotStarted",
     SetPlotData = "SetPlotData",
     SetComparisonPlotData = "SetComparisonPlotData",
-    WarningsFetched = "WarningsFetched",
     CalibrateResultFetched = "CalibrateResultFetched",
     ClearWarnings = "ClearWarnings",
-    ResetIds = "ResetIds"
+    ResetIds = "ResetIds",
+    Metadata = "Metadata"
```
EmmaLRussell, line 36: This might be a little more consistent with the other mutation names:
```suggestion
    MetadataFetched = "MetadataFetched"
```

EmmaLRussell (commented):
LGTM, works great. I was actually imagining that the front end Calibrate state would stay the same data structure (with metadata and data together in the same object) but that the bits of the object would get populated separately by the two api calls. But putting them separately actually results in minimal code changes, and is definitely simpler for the mutations to deal with, so I think this is a good way to go. 

I can't quite remember what the plan for this epic was - so after this branch, the metadata and data are split into separate endpoint calls, and there will be further work in both hintr and HINT to support getting slices of the data (by indicator initially) rather than all the data at once? 

This doesn't seem to be targeting a particular hintr branch so I guess all three endpoints are currently supported in master hintr (and the old metadata+data endpoint will eventually be removed?)

Not sure why this is in the changeset?:  [src/app/static/scripts/master.zip](https://github.com/mrc-ide/hint/pull/867/files#diff-d85f5e82d8382e52709d989f0f55967b19074d8b298800e50bfef2703a8d8e90)

LekanAnni:
> LGTM, works great. I was actually imagining that the front end Calibrate state would stay the same data structure (with metadata and data together in the same object) but that the bits of the object would get populated separately by the two api calls. But putting them separately actually results in minimal code changes, and is definitely simpler for the mutations to deal with, so I think this is a good way to go.
> 
> I can't quite remember what the plan for this epic was - so after this branch, the metadata and data are split into separate endpoint calls, and there will be further work in both hintr and HINT to support getting slices of the data (by indicator initially) rather than all the data at once?
> 
> This doesn't seem to be targeting a particular hintr branch so I guess all three endpoints are currently supported in master hintr (and the old metadata+data endpoint will eventually be removed?)
> 
> Not sure why this is in the changeset?: [src/app/static/scripts/master.zip](https://github.com/mrc-ide/hint/pull/867/files#diff-d85f5e82d8382e52709d989f0f55967b19074d8b298800e50bfef2703a8d8e90)

removed

PR #868, opened 2023-09-14
mrc-4620: Fix download indicator button

r-ash:
## Description

This PR will fix download indicator button so it actually triggers the download

Looks like this might have regressed because of some merge issue as version with a "click" which got merged in PR https://github.com/mrc-ide/hint/pull/832/files#diff-8a738f358069431b5f882d9eaf4de4af15e4f2b4acb3ce74520b88f0c2385005L6 has changed to "trigger-download" so wonder if some merge timing screwed this up. Either way this should be fixed now.

## Type of version change

None

## Checklist


M-Kusumgar (approved):
ignore codecov! I think this is fine and works well when i tested it!

PR #872, opened 2023-10-18
mrc-4634: Fix issue where rehydrated project's model options are always marked as invalid

r-ash:
## Description

At the moment the validation logic in hint is always marking rehydrated projects options as invalid which is leading to problems with broken project state. I want to tighten up the check that we use to see if the model options are still valid.

1. Fit a model and download output zip
1. Upload output zip
1. Go back to model options
1. See that the options are not valid anymore (I'm guessing we're not persisting this?), but later points in workflow are still available
1. Go back to project page
1. Click to reload the project
1. See load error

At the moment when a user goes to model options page the app will

1. Fetch the DynamicFormMeta from the backend
1. Populate the form with any options from the `optionsFormMeta` in the state (which should exist for all non-rehydrated projects) or from the `options` in the state (if it is a rehydrated project)
1. Check that the new form is identical to the `optionsFormMeta` saved in the state. If it isn't then mark this options as invalid i.e. state.valid = false

This is causing problems because in a rehydrated project we do not have an `optionsFormMeta` in the state so it is always marking it as invalid. It also is a bit too strict as if we reload a project after for example adding a new option we are now marking that as invalid, which I don't think it should be.

This PR updates that logic in 2 days
1. On reload populate the form from any options in the state. Not from the `optionsFormMeta`, I think this should cut out one fork in our model load logic as now both rehydrated and reloaded projects will use the `options` from the state.
2. Only mark the form as now invalid if for a select or multiselect the value selected is not one of the available options

We could extend 2. to also mark as invalid
* Numbers where they are outside of the valid ranges (if this control has a range)
* Empty values where the control is required

Very keen for any feedback on this in particular
* I'd love to hear about any code which I could approach in a more idiomatic way, I've not written much typescript before
* Where should new `checkOptionsValid` function live? I started putting it into `vue-next-dynamic-form` itself thinking it might be a thing another user of the library might want but perhaps we should just move it there at the point that we want to use the function somewhere else.
* Do you guys run any formatter? Is there a standard configuration somewhere?

## Type of version change
_Delete as appropriate_

Patches

## Checklist


src/app/static/package.json
```diff
@@ -9,7 +9,7 @@
     "build": "rm -rf public && vue-cli-service build ./src/app/index.ts",
     "serve": "rm -rf public && VUE_CLI_SERVICE_CONFIG_PATH=$PWD/vue.config.js vue-cli-service serve ./src/app/index.ts",
     "copy": "cp -r ./templates/*.ftl ./public && cp -r ./assets/* ./public",
-    "test": "vue-cli-service test:unit -c jest.config.js",
```
r-ash, line 12: Let me know if you want me to back this out @LekanAnni I could just leave it on locally

src/app/static/src/app/store/modelOptions/mutations.ts
```diff
@@ -81,6 +78,56 @@ export const mutations: MutationTree<ModelOptionsState> = {
...
+
+function checkOptionsValid(formMeta: DynamicFormMeta): boolean {
+    let valid = true;
+    formMeta.controlSections.forEach(section => {
+        section.controlGroups.forEach(group => {
+            group.controls.forEach(control => {
+                // We could go further and check that if it is null or empty that this isn't
+                // a required control but just assuming this for now seems ok
+                if (control.value != null && control.value != "" && hasOptions(control)) {
+                    valid = valid && checkControlOptionValid(control);
```
M-Kusumgar, line 93: if you find something that sets valid to false, would you not want to break out of the loops? for this scenario you could replace `.forEach` with `.every` and that will break out of the for loop as soon as you return false

https://stackoverflow.com/questions/34425795/breaking-out-of-an-inner-foreach-loop

just a suggestion though, happy for you to not do it since the forms arent huge

src/app/static/src/app/store/modelOptions/mutations.ts
```diff
@@ -81,6 +78,56 @@ export const mutations: MutationTree<ModelOptionsState> = {
...
+}
+
+function getAllOptions(control: ControlWithOptions): string[] {
+    const options= control.options.map((option: Option) => getOptions(option));
+    return options.flat();
+}
+
+function getOptions(option: Option): string[] {
+    const options = [option.id];
+    if (option.children !== undefined) {
```
M-Kusumgar, line 129: this is for the nested area labels right? i think there were a decent amount with children as [] so makes sense to cut off those possibilities!
```suggestion
    if (option.children !== undefined && option.children.length > 0) {
```

M-Kusumgar (commented):
Looking good, couple of tiny suggestions! also answers to your questions in the PR comment:

src/app/static/src/app/store/modelOptions/mutations.ts
r-ash: Thanks, that's nice!

M-Kusumgar (approved):
works like a charm :D

src/app/static/src/app/store/modelOptions/optionsUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+    } else if (Array.isArray(value)) {
+        valid = value.every(item => options.includes(item));
+    }
+
+    return valid;
+}
+
+function getAllOptions(control: ControlWithOptions): string[] {
+    const options= control.options.map((option: Option) => getOptions(option));
+    return options.flat();
```
EmmaLRussell, line 41: I think this will work

```suggestion
    return control.options.flatMap((option: Option) => getOptions(option));
```

src/app/static/src/app/store/modelOptions/optionsUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+function getOptions(option: Option): string[] {
+    const options = [option.id];
+    if (option.children !== undefined && option.children.length > 0) {
```
EmmaLRussell, line 46:
```suggestion
    if (option.children && option.children.length > 0) {
```

Or you could also do `if (option.children && option.children.length) {` since 0 is falsy

src/app/static/src/tests/utils.test.ts
```diff
@@ -389,59 +388,12 @@ describe("utils", () => {
...
-                                    type: "select",
-                                    required: false,
-                                    value: 4
-                                }
-                            ]
-                        }
-                    ]
-                }
-            ]
+    it("writeOptionsIntoForm writes options into form metadata as expected", () => {
```
EmmaLRussell, line 391: Might want to do a unit test for `checkOptionsValid` by itself too?

EmmaLRussell (commented):
looks good!

I'd be a bit cautious about adding the validation logic to the form since that would affect other project, though it seems pretty benign. 

Am I right in thinking that the idea behind the previous logic was that it would force the user to manually eyeball the form an click "Validate" if the options had changed, and that this is no longer the case? I wonder if that was a Jeff requirement at some point? 

> We could extend 2. to also mark as invalid
>   Numbers where they are outside of the valid ranges (if this control has a range)
>   Empty values where the control is required

Why don't we need to that here? Does something else ensure that the form is invalidated in those cases?

src/app/static/src/app/store/modelCalibrate/mutations.ts
```diff
@@ -51,10 +51,7 @@ export const mutations: MutationTree<ModelCalibrateState> = {
     },
 
     [ModelCalibrateMutation.ModelCalibrateOptionsFetched](state: ModelCalibrateState, action: PayloadWithType<DynamicFormMeta>) {
-        const newForm = state.optionsFormMeta.controlSections.length
-            ? {...updateForm(state.optionsFormMeta, action.payload)}
-            : constructOptionsFormMetaFromData(state, action.payload)
-        state.optionsFormMeta = newForm;
+        state.optionsFormMeta = writeOptionsIntoForm(state.options, action.payload);
```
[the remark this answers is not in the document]
r-ash: So there is no validation on the calibration options at the moment. As long as the options are set users can run calibration.

src/app/static/src/tests/utils.test.ts
r-ash: Thanks yes! Have added that now

r-ash:
> I'd be a bit cautious about adding the validation logic to the form since that would affect other projects e.g. MINT, though it seems pretty benign.

Thanks ok, will leave that in here for now
> Am I right in thinking that the idea behind the previous logic was that it would force the user to manually eyeball the form an click "Validate" if the options had changed, and that this is no longer the case? I wonder if that was a Jeff requirement at some point?

Not really sure where it came from, I think possibly made sense previously to mark this step as invalid. But then this started causing larger issues when we tightened up the project reload logic? Previously I don't think we were triggering a load error if later steps were complete but not the current step? I can't quite remember all the history of this. 

> > We could extend 2. to also mark as invalid
> > Numbers where they are outside of the valid ranges (if this control has a range)
> > Empty values where the control is required
> 
> Why don't we need to that here? Does something else ensure that the form is invalidated in those cases?

Just me being lazy, I've added those now!

src/app/static/src/app/store/modelOptions/optionsUtils.ts
```diff
@@ -21,10 +30,27 @@ function hasOptions(control: Control): control is ControlWithOptions {
...
+}
+
+function checkNumberControlValid(control: NumberControl): boolean {
+    let valid = true;
+
+    if (control.min !== undefined && typeof control.value === "number") {
+        valid = valid && control.value > control.min
+    }
+    if (control.max !== undefined && typeof control.value === "number") {
+        valid = valid && control.value < control.max
```
EmmaLRussell, line 44: Min and max are inclusive I think
```suggestion
        valid = valid && control.value >= control.min
    }
    if (control.max !== undefined && typeof control.value === "number") {
        valid = valid && control.value <= control.max
```

r-ash: Good spot! Thanks

EmmaLRussell (approved):

PR #874, opened 2023-10-23
Make some small miscellaneous fixes

r-ash:
## Description

Started pulling at some threads and ended up here, there are a few different things going on here so give me a shout if you'd like me to carve this up differently

The main things here are
* Fixed some npm build warnings (though there are still many more)
* Fixed some errors from chrome developer console
* Removed the duplicate of filters. Previously these could be shown outside of the `choropleth` component or inside it. It was rendered outside in the review inputs step, I believe because there we have the switch of data source type. I moved this data source instead into a slot as it felt like this would tidy up the duplication of where filters can be and simplify the choropleth logic a little
* One of the chrome warnings I fixed up was a duplicate input ID in forms, this was the input ID in the upload zip and the upload json form. So I started having a look at this and noticed it wasn't working anymore, so I have fixed that up and uncommented the commented out tests

Let me know if I've got the wrong end of the stick on any of these, all this still very new to me so trying to learn as I go.

## Type of version change

Patches

## Checklist


src/app/static/src/app/components/Warning.vue
```diff
@@ -1,7 +1,7 @@
 <template>
     <div v-if="warnings.length > 0">
         <h5 class="alert-heading pt-2">
-            <vue-feather type="alert-triangle" size="1.0x" class="custom-class mr-1 mb-1"></vue-feather>
+            <vue-feather type="alert-triangle" size="1.0" class="custom-class mr-1 mb-1"></vue-feather>
```
r-ash, line 4: This was causing an error in dev console

src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -34,7 +34,7 @@
             <h4 v-if="!loading" v-translate="'browseADR'"></h4>
             <p v-if="loading" v-translate="'importingFiles'"></p>
             <div v-if="!loading">
-                <label for="datasetSelector" class="font-weight-bold" v-translate="'datasets'"></label>
+                <label class="font-weight-bold" v-translate="'datasets'"></label>
```
r-ash, line 37: This was raising an "improvement issue" in chrome dev console as it didn't match up with any input field

src/app/static/src/app/components/load/UploadNewProject.vue
```diff
@@ -1,8 +1,8 @@
 <template>
     <div id="load-project-name">
         <modal id="load" :open="openModal">
-            <label class="h5" for="project-name-input" v-translate="'enterProjectName'"></label>
-            <input id="project-name-input"
+            <label class="h5" :for="inputId" v-translate="'enterProjectName'"></label>
+            <input :id="inputId"
```
r-ash, line 5: Had to set this ID here to avoid chrome dev console showing an error for duplicate input field ID (for upload zip and upload json forms)

r-ash: Perhaps there is a better way to achieve this though?

src/app/static/src/app/store/load/actions.ts
```diff
@@ -74,8 +78,11 @@ export const actions: ActionTree<LoadState, RootState> & LoadActions = {
 
         if (!rootGetters.isGuest) {
             await (dispatch("projects/createProject", {name: state.projectName || null}, {root: true}));
-            savedState.projects.currentProject = rootState.projects.currentProject;
-            savedState.projects.currentVersion = rootState.projects.currentVersion;
```
r-ash, line 78: This was causing rehydrate from JSON to fail as `savedState.projects` was undefined. Looking at the state we save out this must have been removed at some point. I've initialised the state here like we do for project rehydrate

src/app/static/src/app/components/load/UploadNewProject.vue
M-Kusumgar: i think passing it as a prop is completely fine!

src/app/static/src/app/store/load/actions.ts
```diff
@@ -74,8 +78,11 @@ export const actions: ActionTree<LoadState, RootState> & LoadActions = {
...
+            savedState.projects = {
+                currentProject: rootState.projects.currentProject,
+                currentVersion: rootState.projects.currentVersion,
+                previousProjects: []
+            } as any
```
M-Kusumgar, line 85: not 100% sure but could this be defined in any situation? in which case maybe you just want to check first if it is undefined before resetting it completely

M-Kusumgar (commented):
Looks good! happy with most of the changes. One suggestion though, would it be good to just remove the option to load project from a json file considering it is behind a feature flag and no one seems to be using it? thoughts @EmmaLRussell?

r-ash:
> Looks good! happy with most of the changes. One suggestion though, would it be good to just remove the option to load project from a json file considering it is behind a feature flag and no one seems to be using it? thoughts @EmmaLRussell?

FWIW just asked Rachel and she has no objections to ripping it out!

EmmaLRussell:
> > Looks good! happy with most of the changes. One suggestion though, would it be good to just remove the option to load project from a json file considering it is behind a feature flag and no one seems to be using it? thoughts @EmmaLRussell?
> 
> FWIW just asked Rachel and she has no objections to ripping it out!

Fine by me, I guess we don't really need it since people will be loading from the output zip now. Maybe do that in another branch though?

src/app/static/src/app/store/load/actions.ts
EmmaLRussell: I guess we can revert this when we take out the rehydrate from JSON, since we should always be starting from a default project state?

src/app/static/src/tests/components/header/fileMenu.test.ts
```diff
@@ -111,79 +111,79 @@ describe("File menu", () => {
...
-    //     await expectTranslated(link, "LoadJSON", "ChargerJSON", "CarregarJSON", store as any);
-    //     const input = wrapper.find("#upload-file")
-    //     await expectTranslated(input,
-    //         "Select file",
-    //         "Sélectionner un fichier",
-    //         "Selecionar ficheiro",
-    //         store as any,
-    //         "aria-label");
-    // });
+    it("downloads JSON file", async () => {
```
EmmaLRussell, line 114: Wow, when did all this get commented out!? Was it a casualty of the Great Vue3 Campaign?

EmmaLRussell (commented):
Nice, very tidy! :broom:

src/app/static/src/app/components/Warning.vue
r-ash: Oops thanks yes, fixed this now by setting 1.0rem and also set the alignment on this. Now looks like 

Screenshot_20231027_110156

src/app/static/src/app/store/load/actions.ts
r-ash: I don't think it should ever be defined but I guess that could change in the future and should be better to be safe. I've updated that now to add a guard.

src/app/static/src/app/components/adr/ADRKey.vue
```diff
@@ -2,8 +2,7 @@
     <div class="row mb-2">
         <div class="col-8">
             <div class="d-flex d-inline-block">
-                <label for="key"
-                       class="font-weight-bold align-self-center col-form-label"
+                <label class="font-weight-bold align-self-center col-form-label"
```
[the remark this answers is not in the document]
r-ash: Yeah I think upset because when it is first loaded the input field doesn't exist. It's created when a user clicks the "Add button"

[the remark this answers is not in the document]
r-ash: Fixed that one now, have switched from a `v-if` to ` v-show` so the input always exists

src/app/static/src/app/components/adr/SelectDataset.vue
r-ash: I'm struggling with this one a bit, I can't work out how to get the ID onto the actual input field. And having a look through docs here https://vue-treeselect.js.org/ I'm struggling to find anything that might achieve that? Any bright ideas?

As it is with the `for` set that still wouldn't help with accessibility right as it isn't pointing to any input?

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -1,10 +1,13 @@
...
-                 :filters="filters"
-                 :selectedFilterOptions="selections.selectedFilterOptions"
-                 @update:filters="onFilterSelectionsChange"></filters-comp>
-        <div id="chart" :class="includeFilters ? 'col-md-9' : 'col-md-12'">
+        <div class="col-md-3">
+            <slot/>
+            <filters-comp
+                          :filters="filters"
+                          :selectedFilterOptions="selections.selectedFilterOptions"
+                          @update:filters="onFilterSelectionsChange"></filters-comp>
```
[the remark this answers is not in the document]
r-ash: Yep that's right, data source is the only place we include stuff atm

src/app/static/src/tests/components/header/fileMenu.test.ts
r-ash: Yeah I think so! Will be nice to cut it all out after this PR :dagger:

src/app/static/src/app/store/load/actions.ts
r-ash: Yep that's right, will remove that in next PR

src/app/static/src/app/components/adr/SelectDataset.vue
M-Kusumgar: unfortunately, digging through the source code theres no way to pass id onto the input tag https://github.com/megafetis/vue3-treeselect/blob/f577e47c448b3b9c162322c910b97a4a2ce027ef/src/components/SingleValue.vue#L30

src/app/static/src/app/store/load/actions.ts
r-ash: I made that change that in that's causing errors now. This should never be set. It is reading the `savedState` variable from the file and we know for json and zip it is not set. And there is a test which saves a file with projects as an empty object and then reads that in using `setFiles` action and tests that it has the currentProject, currentVersion and previousProjects set so I'm going to put this back in to make that test pass. I think not too big of a deal though as we're going to rip this all out shortly

M-Kusumgar (approved):
Nice one, the console is very happy :laughing:

PR #875, opened 2023-10-24
mrc-4653: Fix issue with tooltips not being updated when filter selections change

r-ash:
## Description

This PR fixes issue where when any filter value is changed (except detail level) the tooltips are not being updated

The issue here is the `onEachFeature` function is only called once when the feature is created and not when the selections change. This adds a new function into the `update` hook which redraws updates the tooltips to current values.

This isn't a problem for the bubble plot, I've not looked deeply into why but I think because the circles are redrawn everytime unlike the choropleth where the map regions are not redrawn just have their colour updated

## Type of version change

Patches

## Checklist


src/app/static/src/tests/components/plots/choropleth/choropleth.test.ts
```diff
@@ -575,10 +591,10 @@ describe("Choropleth component", () => {
...
         onEachFeatureFunction(mockFeature, mockLayer);
         expect(mockLayer.bindTooltip.mock.calls[0][0]).toEqual(`<div>
-                                <strong>Area 1</strong>
-                                <br/>1.00%
-                                <br/>(1.00% - 10.00%)
-                            </div>`);
+                        <strong>Area 1</strong>
+                        <br/>1.00%
+                        <br/>(1.00% - 10.00%)
+                    </div>`);
```
r-ash, line 597: I'm not really sure why the whitespace here has changed but figured it wouldn't make a difference to the rendered tooltip

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -59,7 +60,7 @@
 
     interface Data {
         style: any,
-        fullIndicatorRanges: Dict<NumericRange>
+        fullIndicatorRanges: Dict<NumericRange>,
```
M-Kusumgar, line 63: ive seen this type of change in a couple places in your PRs, i dont mind it at all but just curious, is this an auto formatter?

M-Kusumgar (commented):
code looks good to me! couple of comments though:
1. tooltips still showing 0.0% on areas which are unselected but im assuming this is what the slack conversation was about so another ticket for that?
2. tooltip not showing on the geometric centre especially if detail level is zone and you hover over the northern region, i think it is safe to assume that this should probably be another ticket too?

src/app/static/src/tests/components/plots/choropleth/choropleth.test.ts
r-ash: Oh yes that makes sense. Is that important? This will all be rendered nicely when we display the HTML right. Suppose maybe the test could strip whitespace and then check equality if the whitespace is uninteresting?

M-Kusumgar: oh i mean the html will ignore it, dont think it is important to worry about this whitespace, we dont really worry about it elsewhere and would just complicate testing a bit

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
M-Kusumgar: ohhh okay makes sense haha, yh no worries im not fussed about whether it is there or not

EmmaLRussell (approved):

PR #876, opened 2023-10-25
mrc-4631: Add special error message for issue where user tries to pull from ADR before creating an account

r-ash:
## Description

To recreate the issue
1. Go to Naomi and sign in with a new auth0 account
2. Try to pull data from the ADR
3. See error message

You get this error because the new account has never logged into the ADR and so does not yet have an ADR account. I just want to surface a more useful message in this case which gives the user details of what they can do themselves to fix the problem. It is unlikely that we will actually get this from users as they will (usually always) go to the ADR first

## Type of version change

Patches

## Checklist


src/app/static/src/app/components/adr/SelectDataset.vue
```diff
@@ -201,6 +201,17 @@
                 namespace,
                 (state: ADRState) => state.adrError
             ),
+            adrErrorText() {
+                let errorKey = "errorFetchingDatasetsFromADR";
+                // When we get an error matching this this is because a user has tried to pull data from the ADR
+                // after authenticating with Auth0 before they have logged into the ADR at all.
+                // In this scenario they do not have an ADR account yet, so need to login there first for the
+                // fetch datasets to work.
+                if (this.adrError && this.adrError.detail && this.adrError.detail.match("has not yet logged into ADR")) {
```
M-Kusumgar, line 210:
```suggestion
                if (this.adrError?.detail?.match("has not yet logged into ADR")) {
```
should also work, the ? will just return undefined if the property does not exist and not continue down the chain

M-Kusumgar (approved):
looks good to me! just see comment about potential change!

README.md
```diff
@@ -33,6 +33,20 @@ For more information about developing the front-end see [src/app/static/README](
 
 For more information on Generic Chart, see [/docs/GenericChart.md](https://github.com/mrc-ide/hint/blob/mrc-2537/docs/GenericChart.md)
 
+#### Run up app with Auth0 authentication
+
```
EmmaLRussell, line 37: Thanks for doing this, that's useful!

EmmaLRussell (commented):
I think it would probably be nicer to handle this in the backend rather than the front end, so that knowledge of the vagaries of the ADR API can be restricted to the back end rather than having a component interrogating the raw API error message.

If you did that, you could make this div which is currently hardcoded to a translation key use the message from the error instead (could be a generic ErrorAlert?), allowing the backend to translate in the error resource bundle. So you would handle the error in the AdrController, throw an exception with the "has not yet logged in" key or with a generic "error fetching datasets" key, then add those to the error resource bundle along with their translations. Let's chat about it on Monday if that doesn't make sense!

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -94,14 +95,23 @@ class ADRController(private val encryption: Encryption,
...
         val response = adr.get(url)
-        return if (response.statusCode != HttpStatus.OK)
+        if (response.statusCode != HttpStatus.OK)
         {
-            response
+            var errorKey = "adrFetchingDatasetsError"
+            if (response.body?.contains("has not yet logged into ADR") == true)
+            {
+                errorKey = "adrFetchingDatasetsNoAccountError"
+            }
```
EmmaLRussell, line 104: Something like this would work: 

```suggestion
           val errorKey = if (response.body?contains("has not yet logged into ADR)) "adrFetchingDatasetsNoAccountError" else "adrFetchingDatasetsError"         }
```

EmmaLRussell (commented):
This looks like a better approach to me. The only trouble is that it will effectively swallow any other non-OK ADR responses to throw the generic exception, so perhaps we should only throw in the special case, and pass the response through unchanged (as it did before) for any other ADR error responses).

EmmaLRussell (approved):
Looks good! Do you want to create a ticket for generically logging raw errors throught the exception handler, to replace this special case?

I still think you could probably use an `ErrorAlert`, but no big deal.

M-Kusumgar (approved):
Nice yh apologies for leading you down the path of frontend stuff! this approach is probably better! Looks good to me!

PR #877, opened 2023-10-26
mrc-4654: Update help menu and add link to videos

r-ash:
## Description

This PR tidies up the help links
* Remove the FAQ link as this is out of date and no one is updating it
* Adds a link to the youtube videos - we still need to update these
* Moves basic steps link into the support menu

## Type of version change

Minor

## Checklist


src/app/static/src/app/hintVersion.ts
```diff
@@ -1 +1 @@
-export const currentHintVersion = "3.1.0";
+export const currentHintVersion = "3.1.1";
```
EmmaLRussell, line 1: Should be minor change rather than patch?

EmmaLRussell (approved):
LGTM. I prefer having all help-related stuff in a single menu, but I assume we originally left it outside of the menu to make it easy to find and access...?

r-ash:
> LGTM. I prefer having all help-related stuff in a single menu, but I assume we originally left it outside of the menu to make it easy to find and access...?

Yeah I think so, I moved it in to keep everything in one place which I think I prefer too. I've shown this to Rachel and she likes it too. I could ask Ian for feedback if you want to err on safe side? Or we could deploy it and see if he complains

M-Kusumgar (approved):
looks good to me!

PR #878, opened 2023-10-26
add feature flag + exoskeleton of table tab

M-Kusumgar:
## Description

## Type of version change
_Delete as appropriate_

## Checklist


src/app/static/src/tests/components/modelOutput/modelOutput.test.ts
```diff
@@ -122,6 +121,11 @@ function getStore(modelOutputState: Partial<ModelOutputState> = {}, partialGette
 declare let currentUser: string;
 currentUser = "guest";
 
+// will be easy to just remove this once feature switch enabled
+jest.mock("../../../app/featureSwitches",() => {
+    return {switches: {tableTab: true}}
+});
```
r-ash, line 127: Nice, I like that more than putting it in every test which is what I've done elsewhere!

r-ash (approved):
This looks great to me!

src/app/static/src/tests/components/modelOutput/modelOutput.test.ts
M-Kusumgar: yhh thought since we are definitely adding this in, we just assume it is on in testing XD

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -229,6 +235,12 @@ import { defineComponent } from "vue";
             }
             tabs.push("comparison");
 
+            if (switches.tableTab) {
+                tabs.push("table")
+            }
+
+            console.log(switches)
```
EmmaLRussell, line 242: rogue debug?

src/app/static/src/app/components/outputTable/Filters.vue
```diff
@@ -0,0 +1,3 @@
+<template>
```
EmmaLRussell, line 1: This component is going to contain both the presets drop down and also (probably) a generic `Filters` component, isn't it (from `components/plots/Filters`)? So it might be a bit confusing for this to just be called `Filters` too? Maybe `tableFilters` or `tableSelections`?

src/app/static/src/app/components/outputTable/OutputTable.vue
```diff
@@ -0,0 +1,15 @@
...
+        </div>
+        <div id="chart" class="col-md-9">
+            <table-comp></table-comp>
+        </div>
+    </div>
+</template>
+
+<script setup lang="ts">
+    import Filters from "./Filters.vue"
+    import TableComp from "./Table.vue"
```
EmmaLRussell, line 14: Not a huge fan of naming the imports differently from the component file, why do this? Maybe there's not a huge benefit in this container component anyway, it's probably not going to do anything other than literally contain the two inner components, so maybe those could just sit directly in the tab, and then `Table` could become `OutputTable`?

EmmaLRussell (commented):
works for me!

src/app/static/src/app/components/modelOutput/ModelOutput.vue
M-Kusumgar: yes sorry XD removed

src/app/static/src/app/components/outputTable/Filters.vue
M-Kusumgar: I like table filters! lets do that actually

src/app/static/src/app/components/outputTable/OutputTable.vue
M-Kusumgar: so I noticed the `ModelOutput.vue` file is getting a little big, I think it would be nice to eventually split it up a bit so it can just be concerned with few controls like show this plot if this tab is selected. I was thinking this component could be in charge of filtering the data based on filter selections!

I dont like naming the files differently, basically the html tag for `<table>` is overwriting the vue one, is there a simple way to fix this or do you think it is best to just name it something different? do you think maybe `TableData` is a better name or `TableDisplay` something like that?

M-Kusumgar: gone with `TableData` for now

EmmaLRussell (approved):

PR #879, opened 2023-10-26
mrc-4661 added table selections to state

M-Kusumgar:
## Description

I have added the table selections to state in this PR and decided on the info the state should hold based on the UI of the filter selects in this [mockup](https://docs.google.com/presentation/d/1bAp4mY-8yxUaxWadPIoPPtrIhLNRSLU236h5CDjGqSo/edit#slide=id.g2902438d226_7_0).

## Checklist

### DIFF
--- a/src/app/static/src/app/hintVersion.ts
+++ b/src/app/static/src/app/hintVersion.ts
@@ -1,1 +1,1 @@
-export const currentHintVersion = "3.1.0";
+export const currentHintVersion = "3.1.1";
--- a/src/app/static/src/app/hintVersion.ts
+++ b/src/app/static/src/app/hintVersion.ts
@@ -1,1 +1,1 @@
-export const currentHintVersion = "3.1.1";
+export const currentHintVersion = "3.1.0";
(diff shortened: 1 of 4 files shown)


r-ash (approved):
Nice one

src/app/static/src/app/hintVersion.ts
```diff
@@ -1 +1 @@
-export const currentHintVersion = "3.1.0";
```
EmmaLRussell, line 1: Arguably all of these changes don't merit a version change since they aren't changing the publicly available behaviour of the app. And the third number is generally for patches (bug fixes) anyway, I think? We can put in a minor change when we flip the switch. Might cut down on merge conflicts a bit..

EmmaLRussell (approved):

src/app/static/src/app/hintVersion.ts
M-Kusumgar: ohhh that actually helps a lot, given how many hint PRs there are, i was wondering how to efficiently manage version changes, will remove the change!

PR #882, opened 2023-11-10
add ability for kotlin to query indicators

M-Kusumgar:
## Description

## Type of version change
_Delete as appropriate_

## Checklist


src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
```diff
@@ -51,11 +58,50 @@ class JooqCalibrateDataRepository: CalibrateDataRepository
...
+        age_group,area_id,
+        calendar_quarter,
+        indicator,
+        ROUND(lower, 4) AS lower,
+        ROUND(mean, 4) AS mean,
+        ROUND(mode, 4) AS mode,
+        sex,
+        ROUND(upper, 4) AS upper
+        FROM data WHERE indicator='$indicator'"""
+    }
```
r-ash, line 84: I'm a bit rusty here but I think from when I doing Java a long time ago we would return this as a prepared statement then set the values in it instead of doing the string interpolation.

```
val query = con.prepareStatement(
  """
  SELECT ... from data where indicator = ?
  """)
query.setString(1, indicator)
result = query.executeQuery()
```

I think setting the value in the prepared statement gives protections around sql injection.

src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
```diff
@@ -51,11 +58,50 @@ class JooqCalibrateDataRepository: CalibrateDataRepository
...
+    private fun getDataFromConnection(
+        conn: Connection,
+        indicator: String): List<CalibrateResultRow> {
+        val query: String
+        if (indicator == "all") {
+            query = DEFAULT_QUERY
+        } else {
+            val indicatorStmt: Statement = conn.createStatement()
+            val indicatorResultSet = indicatorStmt.executeQuery(INDICATOR_QUERY)
+            val validIndicators = convertIndicatorToList(indicatorResultSet)
```
r-ash, line 95: I wonder if this is necessary if we're using prepared statement like above? Also feels like we query for an indicator which doesn't exist we should just return 0 rows, the actual query was valid and fine but just no data available? 

Does returning the error make it easier to handle in the front end though?

r-ash (changes requested):
This looks good, couple of comments but happy to be overruled - still very rusty with this stuff

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/CalibrateController.kt
```diff
@@ -41,11 +41,13 @@ class CalibrateController(val apiClient: HintrAPIClient, val calibrateDataServic
         return apiClient.getCalibrateResultMetadata(id)
     }
 
-    @GetMapping("/result/data/{id}")
+    @GetMapping("/result/data/{id}/{indicator}")
```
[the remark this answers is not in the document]
M-Kusumgar: So I did potentially think about doing this, but would this not make it a bit slower? I don't know about the speeds of querying for two indicators values in one go vs querying for one indicator on two different threads because we can simultaneously send requests, plus even if one fails we would still get results for the other indicator but with the multiple query params the whole thing would fail. Those were my thoughts but I'm honestly not sure 🤣🤣 if you think it's better to do it with query params I'm more than happy to do that

src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
M-Kusumgar: Ohhh really? I'll read up on it, I'm not very familiar with these things! But definitely sounds very good! Would make this easier for sure!

src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
M-Kusumgar: No you're right, let's do it with a prepared statement and if it doesn't match we should at least return an empty array, I guess it's not really an error if the indicator doesn't exist but equally this may make our lives difficult in the future when we try and debug why data may not be showing up on plots, I wonder if we should error and the frontend should handle it if it's an "indicator not present" error?

src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
M-Kusumgar: Just read up on it, yh you're definitely right, will refactor to prepared statements when I'm in the office

src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
M-Kusumgar: after more thought on this, it is probably best to not distinguish the two and just return an empty array to the frontend!

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/CalibrateController.kt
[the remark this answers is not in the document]
EmmaLRussell: Just supporting a single indicator is probably fine for now, if that's the only way we're going to be using it, but if we do want to support multiple indicators in future, it shouldn't be using two query string parameters with the same name - we could comma separate instead though?

EmmaLRussell (approved):
lgtm!

r-ash (approved):
This looks great!

PR #884, opened 2023-11-16
mrc-4693: Update output plots to request slices of data per indicator

r-ash:
## Description

This PR will
* Request only slices of data per indicator from the choropleth, bubble plot and barchart. Note the comparison plot pre-fetches some slightly different format data so that is unchanged here.

I experimented a little bit with changing the underlying format of the calibrate data to be a dict with keys as indicators to data values as the data for that indicator. That feels worthwhile to do to me as I think will clean up some of the bits here where I am checking if we already have an indicator or not. And I'd prefer to keep that in 1 item in state rather than have a 2nd item which is tracking what indicators have been fetched. I put that down that as it was getting to quite a big change. So for now this is just the simplest implementation of fetching the slices which works. 

We should probably also think about if we want to show a "loading" spinner as the data changes, but can do that separately.

## Type of version change

Minor

## Checklist


src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -13,7 +13,7 @@
             </div>
         </div>
         <div class="row mt-2">
-            <div v-if="selectedTab==='map'" id="choropleth-container" class="col-md-12">
+            <div v-if="selectedTab === modelOutputTabs.Map" id="choropleth-container" class="col-md-12">
```
r-ash, line 16: I updated this to an enum too as did some switching on it in a later action and thought the enum would be a bit of a tidier way to do that

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -223,28 +221,32 @@
...
             }
-            tabs.push("comparison");
 
-            if (switches.tableTab) {
-                tabs.push("table")
+            tabs.push(ModelOutputTabs.Map);
+            tabs.push(ModelOutputTabs.Comparison);
+
+            if (!inactiveFeatures.includes("BubblePlot")) {
+                tabs.push(ModelOutputTabs.Bubble);
```
r-ash, line 238: Updated the order whilst I was here as something Rachel has wanted to change for a while

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -124,6 +120,39 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
...
+
+        if (!state.status.done || !indicatorId) {
+            // Don't try to fetch data if the calibration hasn't finished
+            return
+        }
+
+        let indicatorKnown = false;
+        if (state.result) {
+            indicatorKnown = state.result.data.some(row => {
+                return row.indicator === indicatorId
```
M-Kusumgar, line 137: as we discussed earlier you could have an array in state that stores which indicators are fetched and then have a mutation that pushes to the data array and pushes to the indicatorsFetched array to make sure they are in sync but that can be done in the next iteration of this!

src/app/static/src/app/store/modelCalibrate/mutations.ts
```diff
@@ -146,7 +146,15 @@ export const mutations: MutationTree<ModelCalibrateState> = {
 
     [ModelCalibrateMutation.ResetIds](state: ModelCalibrateState) {
         stopPolling(state)
-    }
+    },
+
+    [ModelCalibrateMutation.CalibrateResultFetched](state: ModelCalibrateState, action: PayloadWithType<CalibrateResultWithType>) {
```
M-Kusumgar, line 151: theres another mutation with the same name here that you forgot to delete i think

src/app/static/src/app/store/modelCalibrate/mutations.ts
```diff
@@ -146,7 +146,15 @@ export const mutations: MutationTree<ModelCalibrateState> = {
...
+        if (!state.result) {
+            state.result = action.payload
+        } else {
+            state.result.data = [...state.result.data, ...action.payload.data]
```
M-Kusumgar, line 155:
```suggestion
            state.result = structureClone(action.payload);
        } else {
            state.result.data.push(...action.payload.data);
```
So the `...` spread operator creates a shallow copy (copies over references to a new reference and doesnt create new memory addresses for all the content) but it is inefficient especially as a the size of `state.result.data` grows. Best to do a .push on it which wont create a copy.

The structure clone is there because something is setting the `action.payload` in the first if branch as inextensible (cant push to it basically) and this might explain your earlier struggles with just trying to push stuff, a structureClone will create a deep copy (new memory location for everything) and will also make it extensible, since we are only doing this deep copy once on load it should be completely fine!

M-Kusumgar (changes requested):
Looks good, nice one! we should do loading components asap though because the delay is noticeable even with malawi, see couple of small comments

src/app/static/src/app/store/modelCalibrate/mutations.ts
r-ash: oops, thanks!

src/app/static/src/app/store/modelCalibrate/mutations.ts
r-ash: :pray: awesome thank you! I was fighting with .push

src/app/static/src/app/store/modelCalibrate/actions.ts
r-ash: Thanks yep - I've put this in as seems worth doing here and wasn't too bad in the end.

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -404,7 +404,7 @@ const en: Translations = {
     back: "back",
     both: "Both",
     browseADR: "Browse ADR",
-    bubble: "Bubble Plot",
+    bubble: "Bubble",
```
EmmaLRussell, line 407: User request? 😆

src/app/static/src/app/store/plottingSelections/actions.ts
```diff
@@ -0,0 +1,40 @@
...
+import {PayloadWithType} from "../../types";
+import {DataExplorationState} from "../dataExploration/dataExploration";
+
+export interface PlottingSelectionsActions {
+    updateBarchartSelections: (store: ActionContext<PlottingSelectionsState, DataExplorationState>, payload: PayloadWithType<BarchartSelections>) => void
+    updateChoroplethSelections: (store: ActionContext<PlottingSelectionsState, DataExplorationState>, payload: PayloadWithType<Partial<ChoroplethSelections>>) => void
+    updateBubblePlotSelections: (store: ActionContext<PlottingSelectionsState, DataExplorationState>, payload: PayloadWithType<Partial<BubblePlotSelections>>) => void
+}
+
+export const actions: ActionTree<PlottingSelectionsState, DataExplorationState> & PlottingSelectionsActions = {
```
EmmaLRussell, line 16: Yeah, I can't think of a better way than this, though it seems quite long winded! If it hadn't been for the Bubble plot anomaly of having two indicator ids, I might have suggested having another plugin to monitor mutations, but since there's only three and one of them's different this is probably the sanest approach.

EmmaLRussell (commented):
Looks good! I agree with Mantra that the delay while fetching is quite noticable so would be good to show loading indication in future ticket.

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -312,9 +314,12 @@
...
             ...mapMutationsByNames("plottingSelections",
-                ["updateBarchartSelections", "updateComparisonPlotSelections", "updateBubblePlotSelections",
-                "updateOutputChoroplethSelections", "updateOutputColourScales", "updateOutputBubbleSizeScales"] as const),
-            tabSelected: mapMutationByName<keyof Methods>("modelOutput", ModelOutputMutation.TabSelected),
+                ["updateComparisonPlotSelections", "updateOutputColourScales", "updateOutputBubbleSizeScales"] as const),
+            tabSelected: mapActionByName<keyof Methods>("modelOutput", "updateSelectedTab"),
+            updateBarchartSelections: mapActionByName<BarchartSelections>("plottingSelections", "updateBarchartSelections"),
+            updateOutputChoroplethSelections: mapActionByName<ChoroplethSelections>("plottingSelections", "updateChoroplethSelections"),
+            updateBubblePlotSelections: mapActionByName<BubblePlotSelections>("plottingSelections", "updateBubblePlotSelections"),
+            getResultData: mapActionByName<ModelCalibrateState>("modelCalibrate", "getResultData"),
```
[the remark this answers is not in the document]
r-ash: Thanks, removed it

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -124,6 +120,39 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
...
+        }
+
+        if (!indicatorKnown) {
+            const response = await api<ModelCalibrateMutation, ModelCalibrateMutation>(context)
+                .ignoreSuccess()
+                .withError(ModelCalibrateMutation.SetError)
+                .freezeResponse()
+                .get<CalibrateDataResponse["data"]>(`calibrate/result/data/${calibrateId}/${indicatorId}`);
+
+            if (response) {
```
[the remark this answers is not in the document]
r-ash: Oh yep that's right that is a throwback. I am using it now since adding the separate bit in state to track the previously fetched indicators

src/app/static/src/app/store/translations/locales.ts
r-ash: Yeah, noticed this wasn't consistent with the other tabs. Not sure if came from Rachel originally or us..

EmmaLRussell (commented):
I think we've missed a piece out here, which is resetting the result data and fetched indicators when calibration is re-run. I noticed that the fetched indicators weren't being reset when run re-calibrate, but it looks like the result isn't either - I was surprised to see results for previously fetched indicators after re-calibrate but I think I must have been seeing old data.. 

We should probably reset both in the `CalibrateStarted` mutation, where everything else is reset.

r-ash:
> I think we've missed a piece out here, which is resetting the result data and fetched indicators when calibration is re-run. I noticed that the fetched indicators weren't being reset when run re-calibrate, but it looks like the result isn't either - I was surprised to see results for previously fetched indicators after re-calibrate but I think I must have been seeing old data..
> 
> We should probably reset both in the `CalibrateStarted` mutation, where everything else is reset.

Oh good spot, thanks. I've updated that

EmmaLRussell (approved):

M-Kusumgar (approved):

PR #885, opened 2023-11-16
Mrc 4714 New FilterSelects and Filters components

M-Kusumgar:
## Description
This PR lays the foundation for the table filters PR coming next. Since we are using the new treeselect the events emitted are slightly different than the old one. I will explain the differences below. The testing was copied and pasted from the normal `Filters` and  `FilterSelect` components without any changes except a small one to a `testHelper.ts` function call `expectFilter` so we can ignore codecov for this PR.

(Note this is probably a future branch of the treeselect component but I will outline the events that are emitted and the type of values it takes into the prop, this info should be sufficient for this PR. Also ignore any CSS weirdness with the treeselect component, the styles will be updated but mainly focusing on functionality with this PR).

## Guide: Treeselect API
* modelValue: The new treeselect takes in `string` for single select and `string[]` for multiselect (the old treeselect could also take in ["id"] for single select but I think it is better to be simple within the component and not have to put ifs for this)
* placeholder: `string`
* options: same between old and new treeselect
* @update:modelValue: this event is the most different. For the single select it will output the node of options that is clicked and will have this structure:
```
type Option = {
  id: string,
  label: string,
  children?: Option
}
```
This is the same as the previous behaviour. The multi select however will output all the top level option nodes and is of type `Option[]`. This mirrors the tags shown in the multi select. Previously we have used the @select and @deselect events to filter/push on an array but this does not work for nested options because if you pick option with id "parentId" and deselect its child with id "childId" then the old `FilterSelect` would try and filter out the node with id "childId" from the array (which doesnt do anything), i.e. ["parentId"].filter(x => x !== "childId") is still equal to ["parentId"] and this is not the behaviour we want.

So instead we let the treeselect component handle this logic and just update filter selections based on the @update:modelValue event instead of pushing and filtering an array ourselves. (if you compare `FilterSelect` and `NewFilterSelect` you will notice it doesnt have the select or deselect functions). That is the main difference for this PR

## Checklist


EmmaLRussell (commented):
I'm not seeing any filters in the Table tab when I run this up - should I?

Is the idea with the `New..` components that we'll eventually replace all the old `FilterSelect` and `Filter`s and rename the `New..` components as the new `FilterSelect` and `Filter`? Because I think those should definitely be temporary names! Alternative would be to rename the existing ones `Old..` now I guess. 

A lot of the conditionals around `disabled` seems a bit of a faff - do those events actually get triggered anyway if the component is disabled? Assuming that's inherited from the old component anyway.

M-Kusumgar:
> I'm not seeing any filters in the Table tab when I run this up - should I?

Not yet! this is just a groundwork, I split it up to isolate the logic in the filter select component which will be coming next

> Is the idea with the `New..` components that we'll eventually replace all the old `FilterSelect` and `Filter`s and rename the `New..` components as the new `FilterSelect` and `Filter`? Because I think those should definitely be temporary names! Alternative would be to rename the existing ones `Old..` now I guess.

Yep exactly, once we are done with these big features I would like to do a big cleanup of plots in general so combine the vue charts and generic plot into the plots folder, rework some logic so it is using these new filter selects and remove the old ones!

> A lot of the conditionals around `disabled` seems a bit of a faff - do those events actually get triggered anyway if the component is disabled? Assuming that's inherited from the old component anyway.

Yh agreed, for now i have just copied this from the old components, this new filter select is only in the table tab and we dont expect any disabled treeselects in that so have just left it for now, these will be addressed and used once we use these filter selects for all the plot components

src/app/static/src/app/components/outputTable/NewFilters.vue
```diff
@@ -0,0 +1,48 @@
...
+        },
+        methods: {
+            getSelectedFilterValues(filterId: string) {
+                return (this.selectedFilterOptions[filterId] || []).map((o: FilterOption) => o.id);
+            },
+            onFilterSelect(filter: DisplayFilter, selectedOptions: FilterOption[]) {
+                const newSelectedFilterOptions = {...this.selectedFilterOptions};
+                newSelectedFilterOptions[filter.id] = selectedOptions;
+
+                console.log(newSelectedFilterOptions)
```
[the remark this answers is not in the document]
M-Kusumgar: removed!

r-ash:
> The multi select however will output all the top level option nodes and is of type Option[]. This mirrors the tags shown in the multi select. Previously we have used the @select and @deselect events to filter/push on an array but this does not work for nested options because if you pick option with id "parentId" and deselect its child with id "childId" then the old FilterSelect would try and filter out the node with id "childId" from the array (which doesnt do anything), i.e. ["parentId"].filter(x => x !== "childId") is still equal to ["parentId"] and this is not the behaviour we want.

What does this means in terms of behaviour change?

currentMultiselect

At the moment looks like something like
1. User selects high level item e.g. "Central"
2. System shows "Central" as token and all lower level items as ticked
4. User deselects a low level item from within "Central" e.g. "Nkhotakota"
5. System removes "Central" token, and shows tokens for all other values at the same level as "Nkhotakota" within the multiselect. It shows ticks next to these individuals in the drop down but shows "-" next to the higher level items

src/app/static/src/app/components/outputTable/NewFilterSelect.vue
```diff
@@ -0,0 +1,104 @@
...
+        computed: {
+            treeselectValue() {
+                return this.disabled ? null : this.multiple ? this.value : this.value[0];
+            },
+            currentLanguage: mapStateProp<RootState, Language>(null,
+                (state: RootState) => state.language),
+            placeholder() {
+                const key = this.disabled ? "notUsed" : "select";
+                return i18next.t(key, this.currentLanguage)
+            },
```
r-ash, line 79: This over a v-translate? Or is there some reason that doesn't work in this context?

src/config/hintr_version
```diff
@@ -1 +1 @@
-master
+mrc-4673
```
r-ash, line 1: Should we merge this hintr PR?

r-ash (approved):
Changes look good to me, is it worth adding a test covering the selecting/deselecting from an option with children? I see you add an area select with children in the `newFilters` test but then don't edit the filters for that as far as I can see? Though perhaps that is covered in the tree select components tests itself?

M-Kusumgar:
> What does this means in terms of behaviour change?

This should actually be exactly the same behaviour for the user will explain more just underneath!

> At the moment looks like something like
> 
> 1. User selects high level item e.g. "Central"
> 2. System shows "Central" as token and all lower level items as ticked
> 3. User deselects a low level item from within "Central" e.g. "Nkhotakota"
> 4. System removes "Central" token, and shows tokens for all other values at the same level as "Nkhotakota" within the multiselect. It shows ticks next to these individuals in the drop down but shows "-" next to the higher level items
> 
> Edited note:
> 
> I realise your talking about the events above. How is that working in this new filter select when selecting/deselecting nested options? e.g. above in step 4 when we deselect an item from within "Central" what ids will the event have?

So this is good behaviour and the treeselect will be replicating this behaviour, however in your example video, if you look at the choropleth, it doesnt actually update to remove the district you have deselected in the treeselect. This is the behaviour we want to fix.

Essentially every time there is an update to the component, we fire off an @update:modelValue event, instead of us handling select and deselect events (which involved pushing and removing from an array) we are letting the treeselect do all of that logic for us. The @update:modelValue event just makes sure the selections in our state match the tags that are shown in the treeselect. This is currently not happening because (I have written the steps above in terms of the events):
1. @select event from the old treeselect fires so in our state we have selection as ["central"] or something like this
2. -
3. @deslect event happens but we deselect "Nkhotakota" so the component gives up the id "nkhotakota". Our old filter select component tries to remove "nkhotakota" from our selections which are ["central"] and it doesnt actually do anything because "nkhotakota" is not part of our selections because our selections do not take the nested structure into account.
4. Old treeselect shows updated tags but our state is mismatched and still thinks we have only "central" selected

Now same steps with the new treeselect and filter selects (events are simplified but the essence is there):
1. @update:modelValue fires ["central"] so out state has selections as ["central"]
2. -
3. user deselects "Nkhotakota" and @update:modelValue fires [...everyDistrictInCentralApartFromNkhotakota] and we **replace** ["central"] with [...everyDistrictInCentralApartFromNkhotakota] so now our state reflects that we have every district in central without Nkhotakota.
4. update event from new treeselect ensures our state and the tags it shows on screen are in sync

src/config/hintr_version
M-Kusumgar: i think we should merge it just after we flip the feature switch, to make sure there arent any problems, i dont see why there would be but just to be on the safe side

src/app/static/src/app/components/outputTable/NewFilterSelect.vue
M-Kusumgar: i think v-translate would probably work, not sure why, just copied and pasted the old component and changed the bits i needed, will focus more on cleanup when i do the big plots cleanup project!

M-Kusumgar:
> Changes look good to me, is it worth adding a test covering the selecting/deselecting from an option with children? I see you add an area select with children in the `newFilters` test but then don't edit the filters for that as far as I can see? Though perhaps that is covered in the tree select components tests itself?

yh this will be covered in the treeselect, we actually only really look at the @update:modelValue event in this component so probably should be the only thing we test

PR #887, opened 2023-11-21
mrc-4731: Fix issue where comparison plot defaults were not being set

r-ash:
## Description

This PR follows on from #884 and should be reviewed after that

We made a change a while ago to the comparison plot so that when a user changes an indicator this updates the selected filter options. It looks like this has regressed at some point.

The metadata coming back looks good so what is going on.

To recreate

1. Open comparison plot
1. Change the indicator
1. The x-axis will change and the filters will change from multiselect to singleselect or the other way as appropriate but it won't update the actual selected options
1. It looks like what is happening is the update event gets fired but then another event gets fired which is resetting the options.

I think this is because when we update the filter selections this is going to the FilterSelect component and that has 2 watchers. 1 on the value which updates the selected data and one on isXAxisOrDisagg, which manages converting from a multiselect to a single select using the selected data. isXAxisOrDisagg fires an event which updates the value. So the issue is the isXAxisOrDisagg watcher is running before the value watcher. So flow is

1. State updates
1. This updates the value data
1. Runs isXAxisOrDisagg watcher which fires update event
1. That event updates the value data
1. So the value is getting set correctly and then reset to the old value.

Can fix this by making sure isXAxisOrDisagg calls our update to make sure selected is up to date before firing the event.

## Type of version change
_Delete as appropriate_

Patches

## Checklist


src/app/static/src/app/vue-chart/tests/bar/filterSelect.test.ts
```diff
@@ -126,8 +128,25 @@ describe("FilterSelect component", () => {
...
+        expect(wrapper.emitted("update:filter-select")![0][0]).toStrictEqual([]);
+
+        // In production the event above would commit a mutation which clears the value, simulating that here
+        // for the testing
+        await wrapper.setProps({value: []});
+
         await wrapper.setProps({isXAxis: false, isDisaggregateBy: false});
+        expect(wrapper.emitted("update:filter-select")!.length).toBe(2)
+        expect(wrapper.emitted("update:filter-select")![1][0]).toStrictEqual([{id: "fo1", label: "option 1"}]);
+    });
```
r-ash, line 141: Note the updates to this test, previously it would automatically select the first in the list of options, but now it will first reverts to the first of whatever is in the `value`. But if the value is empty this will then use the first from the list of options. In production there is a listener on the parent component which commits a mutation which updates the `value` whenever the `selection` is changed. So here the deselect would remove `id2` from the value. I've added another `setProps` which recreates this behaviour so we can still test that the first option is picked if the `value` prop is empty.

M-Kusumgar (approved):
LGTM! I compared it against prod and yep definitely made a difference, especially when you go from prevalence to art number attending

EmmaLRussell (approved):

PR #891, opened 2023-11-24
Mrc 4664 Data filtering for table selections

M-Kusumgar:
## Description

I think this is a slight difference to how you guys normally do the filtering maybe, im not sure, I just wanted to have a go at it myself, all this component does is get all the data, go through it once and appends a row of the data if all the filters selections and indicator selection match.

There is a comment explaining why the filters are being pulled, it is just to map `filter.id` to `filter.column_id` since `filter.column_id` are the ids that correspond to the field in the data

## Checklist


src/app/static/src/app/components/outputTable/TableData.vue
```diff
@@ -1,3 +1,65 @@
...
+            if (!result) {
+                return []
+            } else {
+                const data = result.data;
+                const selectedOptions = selections.value.selectedFilterOptions;
+                const filterKeys = Object.keys(selectedOptions);
+                const selectedOptionIds: Record<string, (string | number)[]> = {};
+                for (const k in selectedOptions) {
+                    if (k === "area_level") {
+                        selectedOptionIds[k] = selectedOptions[k].map(op => parseInt(op.id));
```
EmmaLRussell, line 42: That's a shame!
I guess the area level values are in the data as numbers, so we do need to deal with them as numbers, but it's pretty gross needing to have a baked in special that the component is aware of, given that we're pretty metadata-driven other than this. 
Could have a valueType in the metadata I guess (which is also a bit eurgh) Or could actually get these as strings from the database, if we're not using it for anything else.

src/app/static/src/app/components/outputTable/TableData.vue
```diff
@@ -1,3 +1,65 @@
...
+                    } else {
+                        selectedOptionIds[k] = selectedOptions[k].map(op => op.id);
+                    }
+                }
+                const filteredData = [];
+                for (let i = 0; i < data.length; i++) {
+                    const row = data[i];
+                    if (row.indicator !== selections.value.indicator) continue;
+                    const valid = filterKeys.every(key => {
+                        return selectedOptionIds[key].includes(row[filtersIdToDataId.value[key]]);
```
EmmaLRussell, line 52: And this works for us because for the row/column filters we've shoved all values into `selectedOptions` so we don't need to treat those differently here..?

src/app/static/src/app/components/outputTable/TableData.vue
```diff
@@ -1,3 +1,65 @@
...
+                        selectedOptionIds[k] = selectedOptions[k].map(op => parseInt(op.id));
+                    } else {
+                        selectedOptionIds[k] = selectedOptions[k].map(op => op.id);
+                    }
+                }
+                const filteredData = [];
+                for (let i = 0; i < data.length; i++) {
+                    const row = data[i];
+                    if (row.indicator !== selections.value.indicator) continue;
+                    const valid = filterKeys.every(key => {
```
EmmaLRussell, line 51: "valid" seems an odd term

```suggestion
                    const includeRow = filterKeys.every(key => {
```

EmmaLRussell (commented):
Works nicely!

So the idea here is to separate out the filtering logic of the component from the display part? That seems like a good idea to me - these components can get very complicated, and I think that separation is probably nicer than shoving unshared logic into some utils file in some random place.

src/app/static/src/app/components/outputTable/TableData.vue
```diff
@@ -1,3 +1,65 @@
...
+            });
+            return idToDataId;
+        });
+
+        const filteredData = computed(() => {
+            const result = store.state.modelCalibrate.result;
+            if (!result) {
+                return []
+            } else {
+                const data = result.data;
```
r-ash, line 36: Could drop the else here to remove a level of nesting? Or do you have to have a bunch of `?` to appease the type checking in your IDE without it?

src/app/static/src/app/components/outputTable/TableData.vue
r-ash: I could also return the filter option values as integers if that helps. Though perhaps that would be a bigger change as I think they're always string at the moment

r-ash: Worth adding a test which will check this path too?

r-ash (commented):
This looks great, couple of small comments

src/app/static/src/app/components/outputTable/TableData.vue
M-Kusumgar: I think it might be a good idea to return detail level as a string actually to the frontend and put like one exception maybe in the backend? not right now but potentially later on when we do the plots cleanup epic?

M-Kusumgar: will add a test after standup

M-Kusumgar: done

src/app/static/src/app/components/outputTable/TableData.vue
M-Kusumgar: yep yep exactly

src/app/static/src/app/components/outputTable/TableData.vue
M-Kusumgar: no no, can drop nesting looks cleaner!

src/app/static/src/app/components/outputTable/TableData.vue
M-Kusumgar: done

r-ash (approved):

EmmaLRussell (approved):

PR #892, opened 2023-11-24
Mrc 4665 reshape table data for table display

M-Kusumgar:
## Description

This PR reshapes the filtered data

## Checklist

### DIFF
--- a/src/app/static/src/app/components/outputTable/TableReshapeData.vue
+++ b/src/app/static/src/app/components/outputTable/TableReshapeData.vue
@@ -1,0 +1,92 @@
+<template>
+    <table-display :data="reshapedData" :header-name="tableLabelHeader"/>
+</template>
+
+<script lang="ts">
+import { computed, defineComponent } from 'vue';
+import TableDisplay from './TableDisplay.vue';
+import { useStore } from 'vuex';
+import { RootState } from '../../root';
+import { Filter, FilterOption } from '../../generated';
+
+export default defineComponent({
+    components: {
+        TableDisplay
+    },
+    props: {
+        data: Object
+    },
+    setup(props) {
+        const store = useStore<RootState>();
+
+        const features = computed(() => store.state.baseline.shape?.data.features);
+        const selections = computed(() => store.state.plottingSelections.table);
+        const filterDataIdtoId = computed(() => {
+            const filters: Filter[] = store.getters["modelOutput/tableFilters"] || [];
+            const dataIdToId: Record<string, string> = {};
+            filters.forEach(f => {
+                dataIdToId[f.column_id] = f.id;
+            });
+            return dataIdToId;
+        });
+        
+        const presetMetadata = computed(() => {
+            const currentPreset = store.state.plottingSelections.table.preset;
+            const presetMetadata = store.state.modelCalibrate.metadata?.tableMetadata.presets;
+            const metadata = presetMetadata?.find(m => m.defaults.id === currentPreset);
+            if (currentPreset && presetMetadata && metadata) {
+                return { row: metadata.defaults.row, column: metadata.defaults.column };
+            }
+            return null;
+        });
+
+        const reshapedData = computed(() => {
+            const data = props.data;
+            if (presetMetadata.value && data) {
+                const reshapedData: any[] = [];
+                const rowKey = presetMetadata.value.row;
+                let rowOptions: FilterOption[] = [];
+                if (rowKey === "area_id") {
+                    if (features.value) {
+                        rowOptions = features.value.map(f => {
+                            return { id: f.properties.area_id, label: f.properties.area_name };
+                        });
+                    }
+                } else {
+                    rowOptions = selections.value.selectedFilterOptions[filterDataIdtoId.value[rowKey]];
+                }
+                const columnKey = presetMetadata.value.column;
+                for (let i = 0; i < data.length; i++) {
+                    const currentRow = data[i];
+                    const rowVal = currentRow[rowKey];
+                    const rowLabel = rowOptions.find(op => op.id === rowVal)?.label;
+                    const index = reshapedData.findIndex(d => d.label === rowLabel);
+                    if (index === -1) {
+                        reshapedData.push({
+                            "label": rowLabel,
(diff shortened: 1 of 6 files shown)


src/app/static/src/app/components/outputTable/TableReshapeData.vue
```diff
@@ -0,0 +1,92 @@
+<template>
+    <table-display :data="reshapedData" :header-name="tableLabelHeader"/>
```
EmmaLRussell, line 2: Hmm, starting to be a bit less keen on this pattern if it's going to get very nested..
It feels like it maybe it would make more sense to use the Composable pattern here, and keep the filtering and reshaping logic in separate modules, (if we want to keep the file size small) but locally to the actual display component - but they wouldn't be components themselves.

src/app/static/src/app/components/outputTable/TableReshapeData.vue
```diff
@@ -0,0 +1,92 @@
...
+            return null;
+        });
+
+        const reshapedData = computed(() => {
+            const data = props.data;
+            if (presetMetadata.value && data) {
+                const reshapedData: any[] = [];
+                const rowKey = presetMetadata.value.row;
+                let rowOptions: FilterOption[] = [];
+                if (rowKey === "area_id") {
```
EmmaLRussell, line 49:
```suggestion
                // If rows are per area, generate pseudo filter options from all features in shape data
                if (rowKey === "area_id") {
```

src/app/static/src/app/components/outputTable/TableReshapeData.vue
```diff
@@ -0,0 +1,92 @@
...
+            if (currentPreset && presetMetadata && metadata) {
+                return { row: metadata.defaults.row, column: metadata.defaults.column };
+            }
+            return null;
+        });
+
+        const reshapedData = computed(() => {
+            const data = props.data;
+            if (presetMetadata.value && data) {
+                const reshapedData: any[] = [];
```
EmmaLRussell, line 46: It's a bit confusing having the local and computed name the same here.

src/app/static/src/app/components/outputTable/TableReshapeData.vue
```diff
@@ -0,0 +1,92 @@
...
+                        reshapedData[index][`upper_${currentRow[columnKey]}`] = currentRow.upper;
+                        reshapedData[index][`lower_${currentRow[columnKey]}`] = currentRow.lower;
+                    }
+                }
+                return reshapedData;
+            }
+            return [];
+        });
+
+        const tableLabelHeader = computed(() => {
```
EmmaLRussell, line 82: Does the corresponding table filter not have a name that could be used here?

src/app/static/src/app/components/outputTable/TableReshapeData.vue
M-Kusumgar: We can refactor this into more of a composable pattern, in fact, ive been wanting to use composable pattern to tidy up all of the plots in hint, perhaps this can be part of that big cleanup which is probably going to be the next big thing i would like to do in hint, i think this nested component pattern emerged because of the way i am doing these PRs, i feel like I can keep a clearer picture in my head and work on independent components, but yes will probably change to a nicer pattern on the plots cleanup epic!

src/app/static/src/app/components/outputTable/TableReshapeData.vue
M-Kusumgar: thanks! definitely deserved a comment

src/app/static/src/app/components/outputTable/TableReshapeData.vue
M-Kusumgar: yep changed it to `tableData` for now? not sure I am not very good at naming these things haha

src/app/static/src/app/components/outputTable/TableReshapeData.vue
M-Kusumgar: ah so the reason this is hardcoded in is because we would have to have a filter config for "Area" to get that header however for "Sex by Area" the filter is `area_level` which means if we try to get the header in a programmatic way, we would end up with a header called "Area Level"

r-ash: In the choropleth on filtering area level we call this filter "Detail", probably good to be consistent with that? Can we also translate the header name here

M-Kusumgar: for the choropleth, we have made a custom dropdown for it and hard coded in that we want these options so we called it detail, for the table tab, this is happening completely generically based on the metadata, i think if you change the label property in filters that you return with the preset for "Sex by Area" to "Detail" it will be consistent

src/app/static/src/app/components/outputTable/TableReshapeData.vue
EmmaLRussell: Up to you, but I feel like it wouldn't be any less clean to start with the Composables as a first version, since they would separate out as nicely as components do. But I guess we want to get this stuff merged soon so you don't want to be doing a big refactor of a live PR now..

src/app/static/src/app/components/outputTable/TableReshapeData.vue
EmmaLRussell: > which means if we try to get the header in a programmatic way, we would end up with a header called "Area Level"

Right, so we're filtering by `area_level` but the data we're showing in the column is for `area` - and we do have `area` filter elsewhere that we could pull the name from but we don't have one as part of this preset (and can't necessarily assume that it will always be present unless we make some other assumptions...)
I think I'd still prefer it to come in with the metadata somehow, but I guess this is fine for now. 

As Rob says, we do still need to translate though.

src/app/static/src/app/components/outputTable/TableReshapeData.vue
M-Kusumgar: yh, i think its probably best not to refactor at this stage, but i am definitely for the composables soon in the future

r-ash (approved):

EmmaLRussell (approved):

PR #893, opened 2023-11-24
mrc-4734: Fix issue with model options showing as empty

r-ash:
## Description

This PR fixes two issues
* With DRC if you used default model options, ran a fit and returned to model options the step would never load. That is because error throwing from `getAllOptionIds` with stack trace
```
TypeError: Cannot read properties of undefined (reading 'flatMap')
    at getAllOptionIds (optionsUtils.ts:66:28)
    at checkControlOptionValid (optionsUtils.ts:53:21)
    at eval (optionsUtils.ts:21:29)
    at Proxy.every (<anonymous>)
    at eval (optionsUtils.ts:14:35)
    at Proxy.every (<anonymous>)
    at eval (optionsUtils.ts:13:38)
    at Proxy.every (<anonymous>)
    at checkOptionsValid (optionsUtils.ts:12:37)
    at Store.eval (mutations.ts:72:55)
```
   This is because that code was expecting options to always be set but seen examples where that is not the case in the payload coming back. I've added explicit handling to this step to return empty if the options don't exist.
* With a multiselect when the value was coming in as "" this was showing a token when first rendering the options. If you submitted a fit and went back it had correctly updated the value to an empty array and so wasn't showing a token. I've fixed this by adding a parsing step to the model options response. It will now turn value for a multiselect from "" into `[]`. For a select or a multiselect if the options are missing it will also create these as an empty array.

This fixes the issue with missing options as well as showing the "unknown" token. See the ticket for an example of what was shown

## Type of version change
_Delete as appropriate_

Patches

## Checklist


src/app/static/src/app/utils.ts
```diff
@@ -368,19 +368,30 @@ export const validateEmail = (test: string): boolean => {
...
+    const stateHasOptions = Object.keys(options).length > 0;
+    optionsForm.controlSections.forEach(section => {
+        section.controlGroups.forEach(group => {
+            group.controls.forEach(control => {
+                // Write existing option into controls
+                if (stateHasOptions && control.name in options) {
+                    control.value = options[control.name];
+                }
+                // Ensure options exist
+                if (hasOptions(control) && !control.options) {
```
EmmaLRussell, line 385: This line suggests that `hasOptions` is poorly named! Should be `isSelect`?

M-Kusumgar: it could be select or multiselect for `hasOptions` so perhaps `isDropdown` might be a more suitable name? `isSelect` sounds like the function is checking `control.type === "select"` only

r-ash:
> Good to make the code more defensive, but is this a metadata problem? Why would we have a model option with no options? And why was that only happening sometimes? Is there a way to reproduce the problem so we can test this fixes it?

So the options can be empty if the data upload is missing for that indicator. So we get this when survey data is uploaded but say doesn't have ART data included, which is the case for some countries. 

I wonder if hintr was always returning with no options but this has only become a problem since the changes I made in https://github.com/mrc-ide/hint/pull/872 but I'm not too sure.

Either way I agree the metadata shouldn't be returning without options. I tightened up the schema for that here https://github.com/mrc-ide/hintr/pull/481 just got some weird codecov failure on that which I have sorted out in another PR

EmmaLRussell (approved):

PR #895, opened 2023-11-27
Mrc 4667 Download filtered data button for table tab

M-Kusumgar:
## Description

This PR gives the user a button to download the filtered data. We have had to pull in features to get the area names and parent ids and we use options to make sure the order of the headers is what we conventionally give to the users

## Checklist

### DIFF
--- a/src/app/static/src/app/components/outputTable/TableData.vue
+++ b/src/app/static/src/app/components/outputTable/TableData.vue
@@ -82,7 +82,7 @@ export default defineComponent({
 
         const handleDownload = () => {
             const prefix = store.state.baseline.iso3 || store.state.baseline.country;
-            const data = { filteredData: getFullData(filteredData.value), unfilteredData: [] };
+            const data = { filteredData: getDownloadData(filteredData.value), unfilteredData: [] };
             const filename = `${prefix}_naomi_table-data_${appendCurrentDateTime()}.xlsx`;
             exportService({ data, filename, options: { header }})
                 .addFilteredData()


src/app/static/src/app/components/outputTable/TableData.vue
```diff
@@ -55,9 +68,31 @@ export default defineComponent({
             }
             return filteredData
         });
+
+        const getFullData = (filteredData: any[]) => {
```
EmmaLRussell, line 72: Suggested rename - 'full data' sounds like it's going to be unfiltered to me. 
```suggestion
        const getDownloadData = (filteredData: any[]) => {
```

src/app/static/src/app/components/outputTable/TableData.vue
```diff
@@ -8,10 +11,19 @@ import { RootState } from '../../root';
 import { useStore } from 'vuex';
 import { Filter } from '../../generated';
 import TableReshapeData from './TableReshapeData.vue';
+import DownloadButton from '../downloadIndicator/DownloadButton.vue';
+import { exportService } from '../../dataExportService';
+import { appendCurrentDateTime } from '../../utils';
+
+// defines the order of headers on the excel download
+const header = [ "area_id", "area_name", "area_level",
```
EmmaLRussell, line 19: Again, it feels like this could maybe be metadata'ed a bit, but I think that's something we could add later. 
They want 'parent area id' - just the next level up, not the whole hierarchy?

src/app/static/src/app/dataExportService.ts
```diff
@@ -10,17 +10,24 @@ export interface ExportService {
...
         this._data = payload.data
         this._workbook = XLSX.utils.book_new();
+        this._options = payload.options;
     }
 
     addFilteredData = (): ExportService => {
-        const worksheet = XLSX.utils.json_to_sheet(this._data.filteredData);
+        let worksheet;
+        if (this._options) {
+            worksheet = XLSX.utils.json_to_sheet(this._data.filteredData, this._options);
```
EmmaLRussell, line 27: So this will only include columns for the specified headers in `options`, not any other fields which may be in the data?

EmmaLRussell (approved):
Looks great, one tiny rename suggestion. 
Codecov errors are baffling :confused:

src/app/static/src/app/components/outputTable/TableData.vue
r-ash: Maybe we need some layer where "business" logic lives in hint? I don't think we need to necessarily get this back from hintr if it isn't something science want to change. But putting it outside of the display component is usually good practice right? I have similar thoughts on the missing data PR where I'm putting some business-like logic into the Plotly component

M-Kusumgar: hmm not sure we want this in the metadata, i guess we could put it there but its just about the ordering of the excel sheet, i think maybe we should leave it in some random file because as rob said this isnt something that will change so can just be hardcoded in. Perhaps something to think about for the plot cleanup?

The parent_area_id was a think rachel asked, yh she didnt want the whole hierarchy, not sure why honestly haha

src/app/static/src/app/dataExportService.ts
M-Kusumgar: hmm i think so yh, but these are actually all the fields in the data, it just lets me reorder them!

src/app/static/src/app/components/outputTable/TableData.vue
EmmaLRussell: > Maybe we need some layer where "business" logic lives in hint? I don't think we need to necessarily get this back from hintr if it isn't something science want to change. But putting it outside of the display component is usually good practice right? I have similar thoughts on the missing data PR where I'm putting some business-like logic into the Plotly component

I guess that's the idea of the 'service' layer, though not sure we always think of them 'services' exactly.

PR #898, opened 2023-11-29
mrc-4140: Add visual indication of aggregations which include missing data

r-ash:
## Description

This PR will add a visual indication that aggregate data in the time series contains missing data

* If this region is a "missing" value instead of a true 0 => will appear as grey with a tooltip "This value is missing from the uploaded data"
* If the region is an aggregate with a missing value included => will appear as grey with a tooltip "Aggregate value missing data for n regions" where n is the number of regions missing
* If any of these are also > 25% change (atm highlighted red) then they will be a light pink with the same tooltips as above

Still TODO but probably in another ticket
* In tooltip show the area names which are missing

## Type of version change

Patches

## Checklist


M-Kusumgar (commented):
Looks really good! left a couple of small comments! Also a suggestion for points not chopping off, where you define `baseLayout[`yaxis${i + 1}`]`, remove the "autorange" and "rangemode" properties, we can define our own range. First in that loop define the max of the data:
```typescript
const maxOfData = Math.max(...dataByArea[areaIds[i]].map(d => d.value || 0));
```
then define range property in the layout:
```typscript
"range": [-maxOfData * 0.1, maxOfData * 1.1]
```
this should give a padding in the chart

EmmaLRussell (approved):
LGTM! Good idea @M-Kusumgar to sort out adding a bit of padding.

This module is a horribly confusing bit of code in general, for historical reasons, but this isn't the time to overhaul it..

M-Kusumgar (approved):
Nice one! (pending build success)

PR #900, opened 2023-11-30
Mrc 4744 Refactor table tab with id, label objects for column and row in preset metadata

M-Kusumgar:
## Description

Note: still have to increment version number + news, should we do a bigger increment? like instead of 3.4.2 -> 3.4.3, should we do 3.4.2 -> 3.5.0? Table tab is fairly big.

This is officially the last PR for tables epic I think! Should be fairly quick to review, main things changed include:
* Hintr branch is now back to master
* Every place we were using `.column` or `.row` from the preset defaults we are now using `.column.id` and `.row.id` except for one place which is the `tableLabelHeader` computed value in `TableReshapeData` where we use the value
* Removed the feature switch for the table completely

## Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# hint 3.5.0
+
+* Add new table tab to Model Outputs page
+
 # hint 3.3.0
 
 * Keep user's language preference outside project state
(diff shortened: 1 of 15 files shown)


EmmaLRussell (approved):
> Removed the feature switch for the table completely

Bold move. :laughing:  UNAIDs have seen demo and are happy with it?

PR #904, opened 2023-12-05
fix projects page bug

M-Kusumgar:
## Description

Floating vue tooltips take a long time to unmount when you have 400 of them all rendered. So we are only displaying tooltips for 25 projects or less.

## Checklist


r-ash (approved):
Looks great, seen this working on staging

EmmaLRussell (approved):
Looks good! It means lots of repetition, but that should be temporary if we're going to replace with paging later.

PR #907, opened 2023-12-06
mrc-4761: Fix rehydrate bug where fetchedIndicators are incorrect type

r-ash:
## Description

Found another issue when trying to add a script for testing, the rehydrate is not working since we added `fetchedIndicators`. If you rehydrate a project with current step set to output plots then for some reason this is an object in state and so `.add` function doesn't exist which is throwing an error. After refreshing the page it loads fine and recognises it as a set so something a little tricky here.

I had a look at where we do the rehydration and the indicatorsFetched gets set to an empty set correctly there so I think something going on after the reload

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/scripts/run-dependencies.sh
+++ b/scripts/run-dependencies.sh
@@ -41,7 +41,8 @@ docker run --rm -d \
   -e REDIS_URL=redis://redis:6379 \
   -e USE_MOCK_MODEL=true \
   $HINTR_IMAGE \
-  --results-dir=/results
+  --results-dir=/results \
+  --inputs-dir=/uploads
 
 # Need to give the database a little time to initialise before we can run the migration
 docker exec -it $DB wait-for-db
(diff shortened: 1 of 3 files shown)


scripts/run-dependencies.sh
```diff
@@ -41,7 +41,8 @@ docker run --rm -d \
   -e REDIS_URL=redis://redis:6379 \
   -e USE_MOCK_MODEL=true \
   $HINTR_IMAGE \
-  --results-dir=/results
+  --results-dir=/results \
+  --inputs-dir=/uploads
```
r-ash, line 45: This was the main change I wanted to get in, will need this to rehydrate locally. Working on a script to be able to run a fit and put it into a local result

M-Kusumgar (commented):
I cannot reproduce the error but i think i know what is happening from your comment, basically sets cannot be serialised into json see: https://stackoverflow.com/questions/35440290/es6-set-does-not-serialize-to-array

In short this is because the set could be a set of anything so JSON.stringify and JSON.parse doesnt work with this so this is probably why you are getting an object.

If you are in a rush to merge this no problem, happy for you to merge this, however the longer term nicer fix for this is to just store it in an array and before you push an indicator to this array just check if it exists in the array, that should behave nicely with the JSON.stringify. It should take like 5 mins to do that! but up to you :P

PR #908, opened 2023-12-07
mrc-4762 fetch first n indicators

M-Kusumgar:
## Description

For the first iteration let's go for a simpler approach! Let's just block the user from going to the model outputs page until we have fetched the first n indicators on the calibrate plot

## Checklist

### DIFF
--- a/src/app/static/src/app/store/modelCalibrate/actions.ts
+++ b/src/app/static/src/app/store/modelCalibrate/actions.ts
@@ -1,10 +1,11 @@
 import {ModelCalibrateState} from "./modelCalibrate";
-import {ActionContext, ActionTree, Commit} from "vuex";
+import {ActionContext, ActionTree, Commit, Dispatch} from "vuex";
 import {DynamicFormData, DynamicFormMeta} from "@reside-ic/vue-next-dynamic-form";
 import {api} from "../../apiService";
 import {RootState} from "../../root";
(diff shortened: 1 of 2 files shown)


src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -180,6 +189,10 @@ export const getResultMetadata = async function (context: ActionContext<ModelCal
 
         selectFilterDefaults(data, commit, PlottingSelectionsMutations.updateBarchartSelections)
         commit(ModelCalibrateMutation.Calibrated);
+
+        const indicators = data.plottingMetadata.barchart.indicators;
+        await fetchFirstNIndicators(dispatch, indicators, 5);
+
```
EmmaLRussell, line 195: Shouldn't this be happening before the commit of `ModelCalibrateMutation.Calibrated`? I think that's setting the flag that the stepper uses to permit the user advancing to the model outputs page. We maybe get away with it because of awaits, but conceptually it feels like this should come first. 

Oh, in fact I did just try it with a throttled network, and if I clicked on Model outputs step as soon as it was available, I did briefly see no data on the barchart.

M-Kusumgar: youre right! my bad, should be fixed now

EmmaLRussell (approved):
Works for me!

PR #910, opened 2023-12-07
mrc-4768 fix double re-rendering of components

M-Kusumgar:
## Description

The components for the plots are re-rendering. This is because in the updating plot selection actions, we fetch the data and then we commit the actual selections to state. These two commits at different times causes a double re-compute of the filtered data in the various components, this fixes that. @r-ash and I did some profiling to find this out. Here are some of the results for Malawi. For the bubble plot (our most render intensive plot), before:

image

After:

image

Scaled comparison (times shown are nothing to do with what we are interested in, they are just the durations of the snapshots):

image

Rough timing changes:
  - With double mutation (master):
    - Bar: 360 - 430ms
    - Table: 290 - 350ms
    - Map: 420 - 500ms
    - Bubble 490 - 750ms

  - With this PR:
    - Bar: 220 - 290ms
    - Table: 188 - 215ms
    - Map: 228 - 400ms
    - Bubble: 420 - 580ms

## Checklist

### DIFF
--- a/src/app/static/src/app/store/modelCalibrate/actions.ts
+++ b/src/app/static/src/app/store/modelCalibrate/actions.ts
@@ -18,10 +18,20 @@ import {CalibrateResultWithType, Dict, ModelOutputTabs} from "../../types";
 import {DownloadResultsMutation} from "../downloadResults/mutations";
 import {PlottingSelectionsMutations} from "../plottingSelections/mutations";
 import { ModelOutputMutation } from "../modelOutput/mutations";
+import { BarchartSelections, BubblePlotSelections, ChoroplethSelections, TableSelections } from "../plottingSelections/plottingSelections";
 
-type ResultDataPayload = {
-    indicatorId: string,
-    tab: ModelOutputTabs
+export type ResultDataPayload = {
+    payload: Partial<BarchartSelections>,
+    tab: ModelOutputTabs.Bar
+} | {
+    payload: Partial<TableSelections>,
+    tab: ModelOutputTabs.Table
+} | {
+    payload: Partial<ChoroplethSelections>,
+    tab: ModelOutputTabs.Map
+} | {
+    payload: Partial<BubblePlotSelections>,
+    tab: ModelOutputTabs.Bubble
 }
 
 export interface ModelCalibrateActions {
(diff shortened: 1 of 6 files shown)


r-ash (commented):
This looks good, I wonder if we can simplify this a bit to remove one of the actions? See other comment

src/app/static/src/app/store/modelOutput/actions.ts
```diff
@@ -13,31 +14,45 @@ export const actions: ActionTree<ModelOutputState, DataExplorationState> & Model
...
             case ModelOutputTabs.Bubble:
-                currentIndicators.push(rootState.plottingSelections.bubble.colorIndicatorId);
-                currentIndicators.push(rootState.plottingSelections.bubble.sizeIndicatorId);
+                currentIndicatorPayloads.push({
+                    tab,
+                    payload: { colorIndicatorId: rootState.plottingSelections.bubble.colorIndicatorId }
+                });
+                currentIndicatorPayloads.push({
+                    tab,
+                    payload: { sizeIndicatorId: rootState.plottingSelections.bubble.sizeIndicatorId }
```
EmmaLRussell, line 32: I'm a bit confused about why you're making multiple payloads here, when the `getResultData` method has the ability to handle both color and size indicators in a single payload.

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -18,10 +18,20 @@ import {CalibrateResultWithType, Dict, ModelOutputTabs} from "../../types";
 import {DownloadResultsMutation} from "../downloadResults/mutations";
 import {PlottingSelectionsMutations} from "../plottingSelections/mutations";
 import { ModelOutputMutation } from "../modelOutput/mutations";
+import { BarchartSelections, BubblePlotSelections, ChoroplethSelections, TableSelections } from "../plottingSelections/plottingSelections";
 
-type ResultDataPayload = {
-    indicatorId: string,
-    tab: ModelOutputTabs
+export type ResultDataPayload = {
+    payload: Partial<BarchartSelections>,
```
EmmaLRussell, line 24: `payload.payload` feels a bit weird, how about just renaming to `selections` for these types?
```suggestion
    selections: Partial<BarchartSelections>,
```

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -18,10 +18,20 @@ import {CalibrateResultWithType, Dict, ModelOutputTabs} from "../../types";
 import {DownloadResultsMutation} from "../downloadResults/mutations";
 import {PlottingSelectionsMutations} from "../plottingSelections/mutations";
 import { ModelOutputMutation } from "../modelOutput/mutations";
+import { BarchartSelections, BubblePlotSelections, ChoroplethSelections, TableSelections } from "../plottingSelections/plottingSelections";
 
-type ResultDataPayload = {
-    indicatorId: string,
-    tab: ModelOutputTabs
+export type ResultDataPayload = {
```
EmmaLRussell, line 23: Possibly slightly clearer, as `ResultDataPayload` sounds like it's going to be the type of the result data itself

```suggestion
export type GetResultDataPayload = {
```

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -155,12 +178,38 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
...
                 const payload = {
                     data: response.data,
-                    indicatorId: indicatorId
+                    indicatorId
                 } as CalibrateResultWithType
                 commit({type: ModelCalibrateMutation.CalibrateResultFetched, payload: payload});
             }
             commit(`modelOutput/${ModelOutputMutation.SetTabLoading}`, {payload:{tab, loading: false}}, {root: true});
         }
+        updatePlotSelections(commit, payload);
```
EmmaLRussell, line 187: I agree with Rob that `getResultData` now seems a misleading name for this method since it's committing the selections update as well, so either a rename of at least a comment here explaining why the update plot selections happens here would be good.

src/app/static/src/app/store/modelCalibrate/actions.ts
```diff
@@ -155,12 +178,38 @@ export const actions: ActionTree<ModelCalibrateState, RootState> & ModelCalibrat
...
+        case ModelOutputTabs.Map:
+            commit(`plottingSelections/${PlottingSelectionsMutations.updateOutputChoroplethSelections}`, {
+                payload: payload.payload
+            }, {root: true});
+            break;
+        case ModelOutputTabs.Bubble:
+            commit(`plottingSelections/${PlottingSelectionsMutations.updateBubblePlotSelections}`, {
+                payload: payload.payload
+            }, {root: true})
+            break;
```
EmmaLRussell, line 212: Could shorten this a bit. Something like:
```suggestion
    let mutation: PlottingSelectionsMutations;
    switch (payload.tab) {
        case ModelOutputTabs.Bar:
           mutation = PlottingSelectionsMutations.updateBarchartSelections;
            break;
        case ModelOutputTabs.Table:
            mutation = PlottingSelectionsMutations.updateTableSelections;
            break;
        case ModelOutputTabs.Map:
            mutation = PlottingSelectionsMutations.updateOutputChoroplethSelections;
            break;
        case ModelOutputTabs.Bubble:
            mutation = PlottingSelectionsMutations.updateBubblePlotSelections;
            break;
     }
     commit(`plottingSelections/${mutation}`, {payload: payload.payload}, {root: true});       
```

Or could make a dictionary of tab to mutation and select from that - Nick always liked that trick as a way of avoiding switches!

EmmaLRussell (commented):
Seems like a good temporary measure, before the Brave New World sweeps all this away.

r-ash:
This will become obsolete after we change the data filtering

PR #914, opened 2023-12-15
add formatted value columns to table download

M-Kusumgar:
## Description

Ian would like nicely formatted values for the data points rather than the raw values so Jeff agreed that we can have some more columns for formatted data.

To test:
Download the filtered data from the table tab and there should be formatted columns in the sheet. Try for different indicators since they have different formatting.

## Checklist


EmmaLRussell (approved):
LGTM.

Any reason why this doesn't have a ticket? :confused:

r-ash (approved):

PR #915, opened 2023-12-15
resize fix

M-Kusumgar:
## Description

See mrc-4833

## Type of version change
_Delete as appropriate_

## Checklist


r-ash (changes requested):
Barchart changes look good as an interim, though a bit ugly.. but we can address that at a later point. This is changing the calibrate plot though making that look worse. I think we want to keep that at a fixed height if possible.

r-ash (approved):

PR #918, opened 2023-12-20
Mrc 4823 Vite to Hint

M-Kusumgar:
## Description

We will be merging other branches into this. Currently the testing will not work. This only gets the app up and running. We will also add hot reloading in a PR that will merge into this. This is the base for the vite overhaul. To run the app:

1. delete the `tests` folders at these paths: `./src/app/static/src/tests/` and `./src/app/static/src/app/vue-chart/tests/` (just so it can compile)
1. usual `./scripts/run-development-dependencies.sh`
2. switch node version to 20 using `nvm use 20` in another terminal
3. run `npm ci --prefix=src/app/static` in that terminal (reason i am saying to use ci instead of just normal install is because you probably have outdated packages in your node_modules folder)
4. run `src/gradlew -p src app:bootrun` in that terminal

Check out `localhost:8080` should be exactly the same

Couple of small changes:
* little bit stricter with prop types, so have had to add a couple of `as any`s
* hint tree select was acting up in a slightly different way again, this `setTimeout` trick with 0 delay makes the update async which means we let the treeselect component finish all its little bits of code and then we will give it a quick refresh, we were basically doing this in sync before and that works but now it doesnt for some reason. Doesnt matter too much, this component is going to go anyway!
* plotly was messing up in a very small and weird way. If you pick zone as area level for the first time (first time only) then the subplots were stretched so not sure why but just kept it to redrawing the whole thing for now. Will make a ticket addressing this issue shortly.
* `structuredClone` was not working in browser. I am not sure why but `JSON.parse(JSON.stringify())` is equivalent to this and it's how we used to do it back in the day so happily switched to that
* `type="module"` in the freemarker templates is because vite builds code as esm modules i believe and we were stuck on commonjs format in the past for some reason - probably a legacy thing? only difference between these that matters is in commonjs you use `module.exports = {...}` and in esm modules you use `export {...}` which is the modern way

It is nice to simplify some of the config too.

## Checklist


src/app/static/src/app/components/header/FileMenu.vue
```diff
@@ -2,7 +2,7 @@
     <div style="flex:auto">
         <drop-down text="file">
             <p class="dropdown-item mb-0" tabindex="0"
-               @click="$refs.loadZip.click()">
+               @click="($refs.loadZip as any).click()">
```
EmmaLRussell, line 5: Would this work?
```suggestion
               @click="($refs.loadZip as HTMLElement).click()">
```

src/app/static/src/app/components/header/FileMenu.vue
```diff
@@ -18,7 +18,7 @@
                 </a>
                 <a style="display:none" ref="save"></a>
                 <p class="dropdown-item mb-0" tabindex="0"
-                   @click="$refs.loadJson.click()">
+                   @click="($refs.loadJson as any).click()">
```
EmmaLRussell, line 21: Similarly...

EmmaLRussell (commented):
Looks good!

I see you've taken out linting, is the idea to add that back in later with the vite plugin?

Probably not related to this ticket, but search isn't currently working on the plots table (Input and Output) - do I have a vague memory of you mentioning that before? Can't remember what we decided about whether we were ultimately going to remove that, or add a proper table tab in the Review Inputs step?

image
Was "Review results below" always stuck way out the side like that before? Doesn't look quite right!

r-ash:
> image Was "Review results below" always stuck way out the side like that before? Doesn't look quite right!

Agree this looks a bit funny, but this is how it is on prod atm.

r-ash (approved):
Looking good.. I think

M-Kusumgar:
unfortunately right now, I dont see a way to integrate vite and our hot reloading, so i decided to only refactor tests to vitest and keep vue-cli-service instead of vite for our bundler, can always try this upgrade later!

PR #924, opened 2024-01-03
mrc-4857 add kotlin endpoint for full filtered data slicing

M-Kusumgar:
## Description

I am not 100% sure if this is the best way to do this but the general strategy for constructing the SQL statement that I have gone for is to build the prepared SQL statement based on the request.

I have added an iterator operator to the FilterQuery data class so that we can loop over its values easily.

We should be able to test this soon with the next PR (frontend support for this endpoint) but for this PR just adding a new endpoint. You could curl if you want to really test it or postman but it'll probably be easier to rely on the unit tests for this PR and manually test next PR. If you really want to test it the request body should look like:

```JSON
{
  "indicator": [],
  "calendar_quarter": [],
  "age_group": [],
  "sex": [],
  "area_id": [],
  "area_level": []
}
```

and fill in whatever you would like to filter by (remember all of them have string values **except** `area_level` which has an integer value)

## Checklist

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
@@ -3,6 +3,7 @@ package org.imperial.mrc.hint.db
 import org.jooq.tools.json.JSONArray
 import org.springframework.stereotype.Component
 import org.imperial.mrc.hint.models.CalibrateResultRow
+import org.imperial.mrc.hint.models.FilterQuery
 import java.sql.Connection
 import java.sql.DriverManager
 import java.sql.Statement
@@ -35,11 +36,28 @@ ROUND(upper, 4) AS upper,
 area_level,
 FROM data"""
 
+const val FILTER_TEMPLATE = """SELECT
+age_group,area_id,
+calendar_quarter,
+indicator,
+ROUND(lower, 4) AS lower,
+ROUND(mean, 4) AS mean,
+ROUND(mode, 4) AS mode,
+sex,
+ROUND(upper, 4) AS upper,
+area_level,
+FROM data WHERE 
+"""
+
 interface CalibrateDataRepository
 {
     fun getDataFromPath(
         path: String,
         indicator: String): List<CalibrateResultRow>
+    
+    fun getFilteredCalibrateData(
+        path: String,
+        data: FilterQuery): List<CalibrateResultRow>
 }
 
 @Component
@@ -87,6 +105,34 @@ class JooqCalibrateDataRepository: CalibrateDataRepository
         return arrayList
     }
 
+    private fun getFilteredDataFromConnection(
+        conn: Connection,
+        filterQuery: FilterQuery): List<CalibrateResultRow> {
+        val resultSet: ResultSet
+        var query = FILTER_TEMPLATE
+
+        for ((field, values) in filterQuery) {
+            if (values.size > 0) {
+                val questionMarks = values.map { "?" }.joinToString(",")
+                query += "$field IN ($questionMarks) AND "
+            }
+        }
+        query = query.dropLast(5)
+
+        val stmt: PreparedStatement = conn.prepareStatement(query)
+        var stmtIndex = 1;
+        for ((_, values) in filterQuery) {
+            values.forEach { it ->
+                stmt.setString(stmtIndex, it as String)
+                stmtIndex++
+            }
+        }
+
+        resultSet = stmt.executeQuery()
+        val arrayList = convertDataToArrayList(resultSet)
+        return arrayList
+    }
+
     private fun getDBConnFromPathResponse(path: String): Connection {   
         val readOnlyProp = Properties()
         readOnlyProp.setProperty("duckdb.read_only", "true")
@@ -100,6 +146,15 @@ class JooqCalibrateDataRepository: CalibrateDataRepository
     {
         getDBConnFromPathResponse(path).use { conn ->
             return getDataFromConnection(conn, indicator)
+        }
+    }
+
+    override fun getFilteredCalibrateData(
+        path: String,
+        data: FilterQuery): List<CalibrateResultRow>
+    {
+        getDBConnFromPathResponse(path).use { conn ->
+            return getFilteredDataFromConnection(conn, data)
         }
     }
 }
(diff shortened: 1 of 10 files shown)


src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
```diff
@@ -87,6 +105,39 @@ class JooqCalibrateDataRepository: CalibrateDataRepository
...
+        filterQuery: FilterQuery): List<CalibrateResultRow> {
+        val resultSet: ResultSet
+        var query = FILTER_TEMPLATE
+
+        for ((field, values) in filterQuery) {
+            if (values.size > 0) {
+                val questionMarks = values.map { "?" }.joinToString(",")
+                query += "$field IN ($questionMarks) AND "
+            }
+        }
```
r-ash, line 120: I wonder about using jOOQ or Querydsl to build this instead of manually creating the SQL. Probably not a big deal as only building a small amount here. But it might be a bit nicer.

src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
```diff
@@ -87,6 +105,39 @@ class JooqCalibrateDataRepository: CalibrateDataRepository
...
+        query = query.dropLast(5)
```
r-ash, line 121: What does the dropLast here do?

src/app/src/main/kotlin/org/imperial/mrc/hint/models/CalibrateResultRow.kt
```diff
@@ -17,3 +17,28 @@ data class CalibrateResultRow(
...
+    val calendarQuarter: List<String>,
+    @field:JsonProperty("age_group")
+    val ageGroup: List<String>,
+    val sex: List<String>,
+    @field:JsonProperty("area_id")
+    val areaId: List<String>,
+    @field:JsonProperty("area_level")
+    val areaLevel: List<Int>)
+{
+    operator fun iterator(): Iterator<Pair<String, List<Any>>>
```
EmmaLRussell, line 33: This feels a bit like wanting to have your cake and eat it - defining what the fields are as a type, but then trying to treat the fields as if they're just any map to build the query. I think it would be better to have a slightly more verbose query that treats the fields separately. But that could be made neater by not building the SQL query string directly.

EmmaLRussell (commented):
Looks like it works well but I tend to agree with Rob that rolling out own SQL here directly is starting to get a bit icky, and would be nice to use a dsl.

src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
```diff
@@ -35,11 +36,28 @@ ROUND(upper, 4) AS upper,
...
+
 interface CalibrateDataRepository
 {
     fun getDataFromPath(
         path: String,
         indicator: String): List<CalibrateResultRow>
+    
+    fun getFilteredCalibrateData(
+        path: String,
+        data: FilterQuery): List<CalibrateResultRow>
```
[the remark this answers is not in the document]
M-Kusumgar: done

src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
M-Kusumgar: so i tried installing exposed, said it needed java 17, then i was going to use jooq which ended up being a little bit of a headache so i searched online and found a pretty nice one called ktorm: https://www.ktorm.org/, I have gone with this one, it is quite nice but i thought the iterator definition was cool in the FilterQuery stuff and that will not work with this unfortuantely, either way I think this is cleaner overall albeit a bit more verbose

M-Kusumgar:
Unfortuantely we seem to have reached some sort of critical mass of tests with jest, im seeing this a couple of times where javascript runs out of heap memory and cannot allocate more and the front end tests fail! I think they should all pass but CI might not reflect that right now, ive re-tried it but can also test on local to check for the time being before vite

M-Kusumgar:
have merged this branch and main into the gross one #926 and deployed it on preview

src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
```diff
@@ -105,37 +96,49 @@ class JooqCalibrateDataRepository: CalibrateDataRepository
...
+        filterQuery: FilterQuery): List<CalibrateResultRow>
+    {
+        val dataSource = SingleConnectionDataSource(conn)
+        return Database.connect(dataSource)
+            .from(ResultData)
+            .select()
+            .whereWithConditions {
+                if (filterQuery.indicator.size > 0) {
+                    it += ResultData.indicator inList filterQuery.indicator
+                }
```
EmmaLRussell, line 111: Could you make this a little less verbose with a util, something like:
```
fun conditionalAddWhere(conditions: MutableList, queryConfition: List<String>, column: Column)
{
    if (queryCondition.size > 0) {
        conditions += column inList queryConfition
    }
}
```
then call with 
```
.whereWithConditions {
    conditionalAddWhere(it, filterQuery.indicator, ResultData.indicator)
   ...etc
```

src/app/src/main/kotlin/org/imperial/mrc/hint/db/CalibrateDataRepository.kt
```diff
@@ -105,37 +96,49 @@ class JooqCalibrateDataRepository: CalibrateDataRepository
...
-                stmtIndex++
             }
-        }
-
-        resultSet = stmt.executeQuery()
-        val arrayList = convertDataToArrayList(resultSet)
-        return arrayList
+            .map { it ->
+                CalibrateResultRow(
+                    it.getString("indicator")!!,
```
EmmaLRussell, line 130: Shouldn't you be able to use something like `it[ResultData.indicator]` here and avoid the unsafe call and magic strings?

EmmaLRussell (commented):
Thanks, this seems nicer to me!

Could rename the repo class to KtormCalibrateRepository to avoid confusion?

r-ash:
Merged this in a separate pr into #933

PR #926, opened 2024-01-10
Mrc 4867 Hook up full data slicing with the front end

M-Kusumgar:
## Description

## Type of version change
_Delete as appropriate_

## Checklist


EmmaLRussell:
I've had a play with this locally and haven't been able to break it! :hammer: :+1:

r-ash:
Deployed this on preview and it is looking good! Even with DRC data at max seeing this take ~250ms. Most being done in under 150ms

PR #927, opened 2024-01-11
mrc-4678: Remove rehydrate from JSON feature

r-ash:
## Description

This feature has been hidden behind a feature switch and not used for past year or so now can rehydrate from the output zip. Removing this as no longer required!

## Type of version change
_Delete as appropriate_

None

## Checklist


src/app/static/src/app/hintVersion.ts
```diff
@@ -1 +1 @@
-export const currentHintVersion = "3.5.4";
+export const currentHintVersion = "3.6.0";
```
EmmaLRussell, line 1: Arguably this isn't changing functionality so shouldn't need a new version... and if it is, shouldn't the news file be getting updated too?

EmmaLRussell (approved):
Nice!

src/app/static/src/app/hintVersion.ts
r-ash: Good point, I've reverted this version change. Forgot completely about the NEWS, will note that for other PRs

M-Kusumgar (approved):
LGTM! feels good to remove some lines from hint

PR #928, opened 2024-01-17
mrc-4923: Add endpoint to fetch packages and releases from ADR with specific resource

r-ash:
## Description

We want to add a feature to allow users to load an output zip from the ADR. To do this we are going to have to list the datasets in the ADR that have this resource available on them. So this PR does the backend work we need for this
1. Add a new endpoint to list datasets with an output zip
2. Add a new endpoint to list the releases for a package which have an output zip available

## Type of version change

Patches

## Checklist


src/app/src/test/kotlin/org/imperial/mrc/hint/integration/IntegrationTests.kt
```diff
@@ -34,7 +34,10 @@ abstract class SecureIntegrationTests : CleanDatabaseTests()
     @BeforeEach
     fun beforeEach(info: TestInfo)
     {
-        userRepo.addUser("guest", "guest")
+        val userExists = userRepo.getUser("guest")
+        if (userExists == null) {
+            userRepo.addUser("guest", "guest")
+        }
```
r-ash, line 40: Added this check as it was causing the tests to fail if I had previously used the running app or run a test before. Let me know if this is not ok?

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -84,6 +85,24 @@ class ADRController(private val encryption: Encryption,
...
+    fun getDatasetsWithResource(@RequestParam resourceType: String, @RequestParam showInaccessible: Boolean = false): ResponseEntity<String>
+    {
+        val data = listDatasets(showInaccessible)
+        val fileTypeMappings = getApiToFileTypeMappings()
+        val filteredData = data.filter { dataset ->
+            dataset["resources"].any { resource ->
+                resource["resource_type"]?.asText() == fileTypeMappings[resourceType]
+            }
+        }
+
```
r-ash, line 101: @EmmaLRussell we chatted about if the ADR can do this for us. I think from looking - at the moment they can't. Similarly I thought we could pass a filter in to the initial `package_search` request to filter on only those with a resource but I can't get that to work. Even though the docs seem to suggest it should, so I am doing the fetch and filtering in kotlin after the fact.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -403,4 +431,18 @@ class ADRController(private val encryption: Encryption,
         val hash = objectMapper.readTree(response.body!!)["data"]["hash"].asText()
         return hash != newDatasetHash
     }
+
+    private fun getApiToFileTypeMappings(): Map<String, String> {
```
M-Kusumgar, line 435: couldnt this just be a val instead of a function?

M-Kusumgar (approved):
LGTM, just a small comment about function vs val

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -138,21 +155,26 @@ class ADRController(private val encryption: Encryption,
...
+    fun getReleasesWithResource(@PathVariable id: String, @RequestParam resourceType: String): ResponseEntity<String>
+    {
+        val fileTypeMappings = getApiToFileTypeMappings()
+        val releases = objectMapper.readTree(getReleases(id).body!!)["data"]
+        val filteredReleases = releases?.filter { release ->
+            val dataset = getDataset(id, release["id"].asText())
+            val resources = objectMapper.readTree(dataset.body!!)["data"]["resources"]
+            resources.any { resource ->
+                resource["resource_type"]?.asText() == fileTypeMappings[resourceType]
+            }
```
EmmaLRussell, line 168: Could pull this little bit of repeated logic out into a util?

EmmaLRussell (commented):
Could add an integration test for your new import endpoint `/adr/output`?

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
r-ash: Moved this into app properties in fact, not sure if that is appropriate? Let me know if it should be somewhere else

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
```diff
@@ -27,7 +28,7 @@ import javax.servlet.http.HttpServletRequest
 
 @RestController
 @RequestMapping("/adr")
-@Suppress("LongParameterList")
+@Suppress("LongParameterList", "TooManyFunctions")
```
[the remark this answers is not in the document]
r-ash: Split this up now into the listing/pulling and pushing

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ADRController.kt
M-Kusumgar: i think thats fine!

M-Kusumgar (approved):

EmmaLRussell (commented):
looks good! still don't see integration test for /output endpoint though? Might be tricky to implement though?

r-ash:
> looks good! still don't see integration test for /output endpoint though? Might be tricky to implement though?

Ah thanks, misunderstood the last comment. I've added that now (and fixed a bug which it revealed), I've also added an integration test for VMMC upload as I missed that off when I added that in a previous PR.

EmmaLRussell (approved):

PR #930, opened 2024-02-05
Update resource fetched from ADR and update file naming

r-ash:
## Description

This is currently deployed on dev if you want to have a look

Couple of small changes here requested by UNAIDS
* UNAIDS have requested the ADR to include both the inputs and the outputs for the DMPPT2 (VMMC) tool, I have updated the resource name we fetch to match the new name they used to disambiguate the two
* I have also updated the text used to refer to the file to match what the ADR are using, this name was specifically requested from Ian even though I think it is probably worse than what we had before
* I updated other input file type strings so we are consistent on casing, previously we had a mix of "Input file" and "Input File", should all be "Input file" now 

## Type of version change

Patches

## Checklist


M-Kusumgar (approved):
Works for me!

PR #934, opened 2024-02-08
Mrc 5039 new state for plot selection + commit defaults into state

M-Kusumgar:
## Description

For now just take a look at functionality changes, will add tests once all the choices are settled (we have changed approaches a couple times already).

### Key changes

* Add new `plotSelections` state
* Commit default selections into state as soon as metadata is received (this involved computing all effects logic too)

Note: We initially were going to do empty plotControls for choropleth since it doesn't have any plot controls but we want certain effect to be default such as the area being multiselect so have included a plot control with id "default" but this approach is up for debate!

For reference of the metadata structure see https://github.com/mrc-ide/hintr/pull/491, I will address the comments in that PR soon too!

## Checklist

### DIFF
--- a/src/app/static/src/app/store/plotSelections/utils.ts
+++ b/src/app/static/src/app/store/plotSelections/utils.ts
@@ -1,5 +1,5 @@
 import { Commit } from "vuex";
-import { CalibrateMetadataResponse, FilterOption, FilterTypes } from "../../generated";
+import { CalibrateMetadataResponse, FilterOption, FilterRef, FilterTypes, PlotSettingOption } from "../../generated";
 import { PlotName } from "./plotSelections";
 import { PlotSelectionUpdate, PlotSelectionsMutations } from "./mutations";
 import { RootState } from "../../root";
@@ -46,6 +46,54 @@ const getFullNestedFilters = (filterOptions: NestedFilterOption[]) => {
     return fullFilterOptions;
 };
 
+export const selectedFiltersFromPlotSettings = (
+    settings: PlotSettingOption[],
+    defaultFilterTypes: FilterRef[] | undefined,
+    filterTypes: FilterTypes[]
+) => {
+    let filterRefs = defaultFilterTypes;
+    let multiFilters: string[] = [];
+    let filterValues: Record<string, string[]> = {};
+    settings.forEach(setting => {
+        // track which effects need to occur
+        const effect = setting.effect;
+        if (effect.setFilters) {
+            filterRefs = effect.setFilters;
+        }
+        if (effect.setMultiple) {
+            // remove dupes just in case
+            multiFilters = [...new Set([...multiFilters, ...effect.setMultiple])];
+        }
+        if (effect.setFilterValues) {
+            filterValues = {...filterValues, ...effect.setFilterValues};
+        }
+    });
+
+    const filterSelections: PlotSelectionUpdate["selections"]["filterSelections"] = {};
+    filterRefs!.forEach(f => {
+        const filter = filterTypes.find(filterType => filterType.id === f.filterId)!;
+        
+        filterSelections[f.stateFilterId] = multiFilters.includes(f.stateFilterId) ?
+                                            getFullNestedFilters(filter.options) :
+                                            [filter.options[0]];
+
+        if (f.stateFilterId in filterValues) {
+            const optionIds = filterValues[f.stateFilterId];
+            const filterOptions = [];
+            for (let i = 0; i < filter.options.length; i++) {
+                if (optionIds.includes(filter.options[i].id)) {
+                    filterOptions.push(filter.options[i]);
+                }
+                if (filterOptions.length === optionIds.length) {
+                    break;
+                }
+            }
+            filterSelections[f.stateFilterId] = filterOptions;
+        }
+    });
+    return filterSelections;
+}
+
 export const commitPlotDefaultSelections = (metadata: CalibrateMetadataResponse, commit: Commit, rootState: RootState) => {
     const plotControl = metadata.plotSettingsControl;
     const filters = filtersAfterUseShapeRegions(metadata.filterTypes, rootState);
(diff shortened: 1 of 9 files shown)


src/app/static/src/app/store/plotSelections/utils.ts
```diff
@@ -0,0 +1,145 @@
+import { Commit } from "vuex";
+import { CalibrateMetadataResponse, FilterOption, FilterRef, FilterTypes, PlotSettingOption } from "../../generated";
+import { PlotName } from "./plotSelections";
+import { PlotSelectionUpdate, PlotSelectionsMutations } from "./mutations";
+import { RootState } from "../../root";
+
+export const filtersAfterUseShapeRegions = (filterTypes: FilterTypes[], rootState: RootState) => {
+    const filters = [...filterTypes];
+    const area = filters.find(f => f.id == "area");
+    if (area && area.use_shape_regions) {
```
r-ash, line 10: I make a couple of changes to this in my branch but I think no need to check filter type here is "area". If the flag is set for `use_shape_regions` then I think we should switch it out.

src/app/static/src/app/store/plotSelections/utils.ts
```diff
@@ -0,0 +1,145 @@
...
+        const currentOption = q.dequeue();
+        fullFilterOptions.push(currentOption);
+        if (currentOption.children && currentOption.children.length > 0) {
+            currentOption.children.forEach(op => q.enqueue(op));
+        }
+    }
+    return fullFilterOptions;
+};
+
+export const filtersInfoFromPlotSettings = (
```
r-ash, line 49: Does this need to be exported?

src/app/static/src/app/store/plotSelections/utils.ts
```diff
@@ -0,0 +1,145 @@
...
+    filterOptions.forEach(op => q.enqueue(op));
+    while (!q.isEmpty()) {
+        const currentOption = q.dequeue();
+        fullFilterOptions.push(currentOption);
+        if (currentOption.children && currentOption.children.length > 0) {
+            currentOption.children.forEach(op => q.enqueue(op));
+        }
+    }
+    return fullFilterOptions;
+};
```
r-ash, line 47: Does the queue add much here? Might be simpler just to use the functions from the array instead of adding the extra class? 

But also, do we have this logic duplicated in the tree select? Is there a way we can use the logic from the treeselect here instead of repeating it?

src/app/static/src/app/store/plotSelections/utils.ts
r-ash: Maybe for testing, we should write a test before we merge this as it's fairly complicated but I think fine to leave that until we've settled on how we're representing it like you say.

r-ash (commented):
This looks good, main comment is I think it might be a bit nicer to change the metadata a little like we talked through. I don't love the special "default" effect type.

At the moment we have in metadata
```
{
  "filterTypes:" [
    {
      "column_id": ...,
      "id": ...,
      "options": ...,
      "use_shape_regions": ...
    },
    ...
  ],
  "indicators": ...
  "plotSettingsControl": {
    "barchart": {
      "defaultFilterTypes": [
        {
          "filterId": "id",
          "label": "label",
          "stateFilterId": "stateId"
        },
        ...
      ],
      "plotSettings": [
        {
           "id": "id",
           "label": "label123",
           "options": [
             {
                "id": ...,
                "label": ...,
                "effect": { setFilters, setMultiple, setFilterValues }
              },
              ...
            ]
        },
        ...
      ]
  },
  "warnings": [...]
}
```

And in the plotSelections state we have
```
"barchart": {
  "controls": [
    {
      "id": "id",
      "label:" ...,
      "selection": [..]
    },
    ...
  ],
  "filterSelections": [
    {
      "filterId": "id",
      "label": ...,
      "stateFilterId": ...,
      "multiple": false,
      "selection": [...]
    },
    ...
  ]
}
```

I think the representation in state is really good and clear, I wonder if we lean on that
1. Have the metadata return "filterSelections" as "defaultFilterState", this is the initial state for that plot. So has what is multiselect, what filters and shown, what labels are used, the initial default selection for them. This will replace the "default" effect in the `plotSettings` in favour of the metadata being more declarative
2. Keep the controls in the metadata - this should be perfectly static once the "default" effect stuff is removed. That way we don't have to dupe it into the plotSelections state. Then the separation is the selections state contains those things which are dynamic (what filters are shown, what are multiselect) but the static stuff stays in the metadata.

I think that might simplify some of the logic then too. Though we will still have to deal with nested options when top level is selected. But other than that I think it means the logic we need here is when an effect is applied to the "filterSelections" state which we will need anyway going forward but the default state is very obvious from the metadata.

src/app/static/src/app/store/plotSelections/utils.ts
M-Kusumgar: i thought it might be useful to reuse this for the changing plot controls mutations that is coming next!

src/app/static/src/app/store/plotSelections/utils.ts
M-Kusumgar: the queue class doesnt add much, can just do an array, i can potentially just expose this function from the tree select but i wonder if we shouldnt integrate the package and out state setup that tightly, this one might potentially change in the future or the treeselect one might change

M-Kusumgar:
> I think the representation in state is really good and clear, I wonder if we lean on that
> 
> 1. Have the metadata return "filterSelections" as "defaultFilterState", this is the initial state for that plot. So shows what is multiselect, what filters are shown, what labels are used and the initial default selection for them. This will replace the "default" effect in the `plotSettings` in favour of the metadata being more declarative
> 2. Keep the controls in the metadata - this should be perfectly static once the "default" effect stuff is removed. That way we don't have to dupe it into the plotSelections state. Then the separation is the selections state contains those things which are dynamic (what filters are shown, what are multiselect) but the static stuff stays in the metadata.

Definitely agree on number 2, i like the separation of static stuff and stuff that will change!

for number 1, i dislike the "default" setting in `plotSettings` too, having thought about it a bit more im for something similar but instead of a default state in metadata i was thinking like a "defaultEffect" that is always applied in the plot, so choropleth would for example have
```
defaultEffect: {
  setFilters: ["area", "period", ...],
  setMultiple: ["area"]
}
```
what do you guys think @r-ash @EmmaLRussell

PR #935, opened 2024-02-12
mrc-5055: Add filters and plot controls UI

r-ash:
## Description

This PR will add plot controls and filters into the model outputs UI. It will
* Initialise all drop downs for the tab you're on
* Set their available options
* Set them to multi/single select (for filters)
* Set their initial values

## Type of version change

None

## Checklist


src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -1,11 +1,51 @@
...
+import {useStore} from "vuex";
 
-export default defineComponent({})
+export default defineComponent({
+
+    setup() {
+        const store = useStore()
+
+        const plotNames = computed(() => {
+            return Object.keys(store.state.plotSelections) as PlotName[]
```
M-Kusumgar, line 32: Can just get these from the array defined in plotSelections.ts, there's an array of what you're looking for, i don't think this needs to be computed

src/app/static/src/app/components/plots/Filter.vue
```diff
@@ -0,0 +1,70 @@
...
+import i18next from "i18next";
+import {useStore} from "vuex";
+import {RootState} from "../../root";
+import {FilterOption} from "../../generated";
+
+export default defineComponent({
+    props: {
+        filterId: String,
+        multiple: Boolean,
+        selectedOptions: Object as PropType<FilterOption[]>
```
M-Kusumgar, line 28: multiple and selectionOptions can be pulled out of state in this component, might be cleaner as this component would be in charge of that filter select

src/app/static/src/app/components/plots/Filter.vue
```diff
@@ -0,0 +1,70 @@
...
+            console.log("Updating selection to " + JSON.stringify(newSelection));
+            if (Array.isArray(newSelection)) {
+                selected.value = newSelection.map(newOption => newOption.id);
+            } else {
+                selected.value = [newSelection.id];
+            }
+        };
+
+        const placeholder = computed(() => {
+            return i18next.t("select", "en");
```
M-Kusumgar, line 49: should you be pulling out what language it currently is in the state for this?

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -1,11 +1,51 @@
...
+        <ul class="nav nav-tabs">
+            <li v-for="plotName of plotNames" :key="plotName">
+                <a class="nav-link"
+                   :class="selectedPlot === plotName ? 'active': ''"
+                   v-translate="plotName"
+                   @click="switchTab(plotName)"></a>
+            </li>
+        </ul>
+        <div class="mt-2">
+            <PlotControlSet :active-plot="selectedPlot"/>
```
M-Kusumgar, line 12: I think we've stuck to kebab case for component names in hint, same in other files

src/app/static/src/app/store/plotSelections/plotSelections.ts
```diff
@@ -27,8 +27,18 @@ export const initialPlotSelectionsState = (): PlotSelectionsState => {
...
 }
 
+export const plotSelectionsGetters = {
+    plotControls: (state: PlotSelectionsState) => (plotName: PlotName) => {
+        return state[plotName].controls;
+    },
+    outputFilters: (state: PlotSelectionsState) => (plotName: PlotName) => {
+        return state[plotName].filters;
+    },
+};
```
M-Kusumgar, line 37: this logic is quite simple can be put into the component itself, i also just realised, while we were discussing I was a little hesitant on getters but i couldnt figure out why so i left it, i now realise why, they are typed as `any`s so you may have to type cast them in the component, i personally just like putting this logic in the computed in the component if it is this simple. I think most getters are like 2 lines ish in this PR, might be worth putting into components

M-Kusumgar (changes requested):
Looking good, got a couple of comments, mainly about interaction with store and props in the components!

src/app/static/src/app/components/modelOutput/ModelOutput.vue
```diff
@@ -1,11 +1,51 @@
...
+export default defineComponent({
+
+    setup() {
+        const store = useStore()
+
+        const plotNames = computed(() => {
+            return Object.keys(store.state.plotSelections) as PlotName[]
+        })
+
+        const selectedPlot = ref<PlotName>(plotNames.value[0]);
```
r-ash, line 35: Put this into state and have each component read it from state

src/app/static/src/app/store/modelCalibrate/modelCalibrate.ts
```diff
@@ -59,15 +60,18 @@ export const initialModelCalibrateState = (): ModelCalibrateState => {
 };
 
 export const modelCalibrateGetters = {
-    indicators: (state: ModelCalibrateState): BarchartIndicator[] => {
-        return state.calibratePlotResult!.plottingMetadata.barchart.indicators;
+    outputFilterOptions: (state: ModelCalibrateState, getters: any, rootState: RootState) => (filterId: string) => {
+        const type = state.metadata!.filterTypes.find(f => f.id === filterId);
+        return filterAfterUseShapeRegions(type as FilterTypes, rootState).options;
```
r-ash, line 65: Put this into the metadata fetch logic, as a shallow copy of the regions in the features. Then we can remove the getter here and just use the store in the component

r-ash:
Thanks for v useful comments, I've made those changes - it's looking much tidier now.

M-Kusumgar (approved):
Nice one! all looks good to me! will start using this now and let you know if any problems arise but from the testing seems all good

src/app/static/src/app/components/plots/Filter.vue
```diff
@@ -0,0 +1,69 @@
...
+            // TODO: dispatch event and update the selectedOptions in state
+            console.log("Dispatching action to update selection in state " + JSON.stringify(newSelection));
+        };
+
+        const placeholder = computed(() => {
+            return i18next.t("select", store.state.language);
+        });
+
+        return {
+            multiple: filter.value.multiple,
```
M-Kusumgar, line 55: I have changed this in my branch, we couldnt test it until the plot settings dropdowns changed values but in my branch this doesnt update, apparently we have to return computed values only as return in setup, so have to return `filter` here and do `.multiple` on that

PR #943, opened 2024-03-13
mrc-5067: Add choropleth plot

r-ash:
## Description

This PR will add the Choropleth plot back into the app using the new filters. I've updated it to composition API too and tried to split up the logic a bit where appropriate. Though I expect this will change a little bit when I add the bubble plot.

Went through a fair few iterations of this, some of which is in the commit history. I will squash that before we merge but wanted to leave it around in case we want to use anything from that.

I settled on using watchers and passing props and update events through to child components. Initially had this pulling things from state and using computed to do updated but could see quite a lot of duplicated updates because the computed are based on multiple refs. Also had a bit of a stuttering issue with that where when changing a selection, the colour scales were updating, then the async fetch for new plot data completed which caused another update. Meaning you could see the plot change very briefly, then do the final update. 

I've instead chosen to use a watcher and hang the update off when the `plotData` is changed. I hope this will work for us with input plot too but probably something worth chatting through.

Also I'm still all quite new to this to very happy to hear if there is a nicer way to do this! 

## Type of version change
_Delete as appropriate_

None

## Checklist


src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        const updateBounds = () => {
+            if (featureRefs.value.length > 0) {
+                map.value?.leafletObject.fitBounds(featureRefs.value.map(f => f.leafletObject.getBounds()));
+            }
+        };
+        watch(featureRefs.value, () => {
+            updateBounds();
+        });
```
EmmaLRussell, line 111: Maybe move this up next to the other watch to keep those together.

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -0,0 +1,160 @@
...
+import ResetMap from "./ResetMap.vue";
+import MapLegend from "./MapLegend.vue";
+import {getColourRange, getScaleLevels} from "../utils";
+import {IndicatorValuesDict, NumericRange} from "../../../types";
+import {ChoroplethIndicatorMetadata} from "../../../generated";
+import { ScaleSettings } from "../../../store/plotState/plotState";
+import {useMapTooltips} from "../useMapTooltips";
+import {useUpdateScale} from "../useUpdateScale";
+
+export default defineComponent({
```
EmmaLRussell, line 42: You could use script setup tag if you wanted to be really cutting edge ;)
https://vuejs.org/api/sfc-script-setup.html

src/app/static/src/app/components/plots/useUpdateScale.ts
```diff
@@ -0,0 +1,22 @@
...
+import {ScaleSettings} from "../../store/plotState/plotState";
+import {computed} from "vue";
+
+export const useUpdateScale = () => {
+    const store = useStore<RootState>();
+    const selectedIndicator = computed(() => {
+        return store.getters["plotSelections/selectedIndicator"];
+    });
+
+    const updateOutputColourScale = (newScaleSettings: ScaleSettings) => {
```
EmmaLRussell, line 12: I'm getting a bit confused between input/output data here. Are you going to add an `updateInputColourScale` too?

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -0,0 +1,160 @@
...
+    setup() {
+        const store = useStore<RootState>();
+        const plotData = computed<PlotData>(() => store.state.plotData.choropleth);
+
+        const {updateOutputColourScale} = useUpdateScale();
```
EmmaLRussell, line 47: Are we not going to use the same choropleth for input data too? Do we not need to tell the choropleth which data and updates to use, or will that come in a future ticket?

EmmaLRussell (commented):
Looks good to me - but we had already talked about this as an approach!

> Also had a bit of a stuttering issue with that where when changing a selection, the colour scales were updating, then the async fetch for new plot data completed which caused another update. Meaning you could see the plot change very briefly, then do the final update.

Could show a spinner while updating to avoid that issue - though I think we did talk about that idea and rejected it! 

Haven't gone through all the other components (MapLegend etc) - I'm assuming those are basically the same as before?

Not sure MapAdjustScale, MapLegend and ResetMap should sit in the 'choropleth' folder since presumably those will be used by bubble plot too. Could go under 'plots' or intermediate 'maps' folder.

src/app/static/src/app/components/plots/useUpdateScale.ts
r-ash: Yeah so I am expecting this will change when we add the input back in. You can see this a bit in the next PR https://github.com/mrc-ide/hint/pull/944/files#diff-f004fdd6d2702a582a3fac1b2710c566ab7e15bcc8b53691b59229a3f76d8f25R41 but I am expecting that this will change to be `updateColourScale` and it will pull out the "active" plot from state and then commit a mutation to update the relevant scale based on this.

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
r-ash: Yes we will need to, just wanted to leave that for a later ticket

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
r-ash: I tried this but it causes weird behaviour with the bounds updating. I think if we put the `updateBounds` into the earlier watcher it runs before the `featureRefs` have been updated. You can see this if you 
1. Open the map
2. Select "district" area level
3. Deselect "Northern"
4. Switch been deselect and selecting "Central", you will see the map bounds update the previously selected set of features instead of the current one

So I think we need this in a separate watcher which is run after those featureRefs update. 

There is still a bit of stuttering with this in this PR but I have fixed this in the next one.

r-ash:
> > Also had a bit of a stuttering issue with that where when changing a selection, the colour scales were updating, then the async fetch for new plot data completed which caused another update. Meaning you could see the plot change very briefly, then do the final update.
> 
> Could show a spinner while updating to avoid that issue - though I think we did talk about that idea and rejected it!

I've got a fix for this in the next PR which looks like it works quite well.

> Haven't gone through all the other components (MapLegend etc) - I'm assuming those are basically the same as before?

MapLegend is very similar. I've moved the scale level calculation into a util function instead of doing it in the component. But otherwise the same.

The MapAdjustScale is a little different in that we are now committing the new selection into state before emitting an event to update the colour scales in the parent component.

> Not sure MapAdjustScale, MapLegend and ResetMap should sit in the 'choropleth' folder since presumably those will be used by bubble plot too. Could go under 'plots' or intermediate 'maps' folder.

Agreed, that was just bad forward thinking. I move those in the next PR.

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
EmmaLRussell: I didn't really mean combine them into the same watch, I literally just meant to put them next to each other in the file to make the code a bit easier to navigate/understand! :)

src/app/static/src/app/components/plots/useUpdateScale.ts
EmmaLRussell: OK, so the active plot with then potentially be an input one rather than an output one, depending on the current step?

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -0,0 +1,135 @@
...
+    } else {
+        //show a lighter grey than the outlines if no data
+        //so unselected regions are still distinguishable
+        return "rgb(200,200,200)";
+    }
+};
+
+const style = {
+    className: "geojson"
+};
```
M-Kusumgar, line 128: minor comment, this doesnt really change, can we not just put this into the `options-style` instead of spreading it into every feature

or you can make a `getStyle` method that accepts feature and returns a callback that returns
```javascript
{
  className: "geojson",
  fillColor: getColor(feature)
}
```

src/app/static/src/app/components/plots/choropleth/MapAdjustScale.vue
```diff
@@ -0,0 +1,166 @@
+<template>
+    <div class="p-3 ml-2">
+        <div class="form-check dynamic-filtered">
+            <label class="form-check-label">
+                <input id="type-input-dynamic-filtered" class="form-check-input" type="radio" :name="scaleTypeGroup"
+                       :value="scaleType.DynamicFiltered"
+                       v-model="selectedScaleType">
+                <span v-translate="'filteredDataset'"></span>
```
M-Kusumgar, line 8: all these are misaligned, i think the css property you want is `vertical-align: sub`, you can create a class for each of the spans here and add that css prop to the class

src/app/static/src/app/components/plots/choropleth/MapAdjustScale.vue
```diff
@@ -0,0 +1,166 @@
...
+            required: true
+        },
+        selectedScale: {
+            type: Object as PropType<ScaleSettings>,
+            required: true
+        },
+    },
+    setup(props, { emit }) {
+        const store = useStore<RootState>();
+        const scaleStep = computed(() => {
```
M-Kusumgar, line 84: i think
```suggestion
        const scaleStep = computed<number>(() => {
```

src/app/static/src/app/components/plots/choropleth/MapAdjustScale.vue
```diff
@@ -0,0 +1,166 @@
...
+            get() {
+                return props.selectedScale.customMax
+            },
+            set(newMax: number) {
+                const newScaleSetting = {...props.selectedScale};
+                newScaleSetting.customMax = newMax;
+                updateOutputColourScale(newScaleSetting)
+                emit('update:selectedScale')
+            }
+        });
```
M-Kusumgar, line 151: this feels weird, you get an object, than you break it apart and then you glue it back together with set functions that look the same, think you could split up the `v-model` into the `model-value` and update event and the model-value can just reference `selectedScale.<prop>` and update function can be `() => handleChange({...selectedScale, <prop>: $event})` and the `handleChange` could just do `updateOutputColourScale` and `emit`

however i think this could potentially be cleaned up by not doing commit at all, havent tested it out but if you just ditch all the emits (then your update model value function becomes `() => updateOutputColourScale({...selectedScale, <prop>: $event})`) and add a watcher to the correct part of the store that might work

M-Kusumgar (changes requested):
love the new component, feels much cleaner! couple of comments!

src/app/static/src/app/components/plots/useUpdateScale.ts
r-ash: Hmm yeah either have the active plot be an input one. Or just both the "activePlot" and the step they're on to determine where to pull data from state

src/app/static/src/app/components/plots/choropleth/MapAdjustScale.vue
r-ash: Thanks, I've tried to improve the styling here a bit more generally
* Make width at either side more even
* Aligned the number inputs with the "Custom" label to hopefully make it more clear they belong to that
* Sorted out the vertical alignment as you suggest
* Made the number selection boxes wider to handle large numbers better

src/app/static/src/app/components/plots/choropleth/MapAdjustScale.vue
r-ash: Thanks - have tidied this up after chatted through it. No need to use the emit anymore, we can have a watch in the parent on the changes to the colour scale in state

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -0,0 +1,141 @@
...
+import {ChoroplethIndicatorMetadata} from "../../../generated";
+import { ScaleSettings } from "../../../store/plotState/plotState";
+import {useMapTooltips} from "../useMapTooltips";
+import {useUpdateScale} from "../useUpdateScale";
+
+const store = useStore<RootState>();
+const plotData = computed<PlotData>(() => store.state.plotData.choropleth);
+
+const {updateOutputColourScale} = useUpdateScale();
+const indicatorMetadata = ref<ChoroplethIndicatorMetadata>(store.getters["modelCalibrate/indicatorMetadata"]);
```
r-ash, line 45: Instead of a getter, make this a computed which updated when filter drop down changes. And put the logic into a function instead of a getter.

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -0,0 +1,141 @@
...
+    );
+}
+
+const updateFeatures = () => {
+    indicatorMetadata.value = store.getters["modelCalibrate/indicatorMetadata"];
+    const selectedLevel = store.state.plotSelections.choropleth.filters
+            .find(f => f.stateFilterId === "detail")!.selection;
+    const selectedAreas = store.state.plotSelections.choropleth.filters
+            .find(f => f.stateFilterId === "area")!.selection;
+    currentFeatures.value = getVisibleFeatures(features, selectedLevel, selectedAreas);
```
r-ash, line 93: Do we want to hide areas here? This is not what the current UI is doing, it is just greying out unselected regions

src/app/static/src/app/components/plots/choropleth/MapAdjustScale.vue
```diff
@@ -0,0 +1,132 @@
...
+        },
+        selectedScale: {
+            type: Object as PropType<ScaleSettings>,
+            required: true
+        },
+    },
+    setup(props) {
+        const store = useStore<RootState>();
+        const scaleStep = computed<number>(() => {
+            return store.getters["modelCalibrate/scaleStep"];
```
r-ash, line 89: computed can just have the function you are calling down the getter chain

src/app/static/src/app/components/plots/choropleth/MapLegend.vue
```diff
@@ -0,0 +1,79 @@
...
+            <div class="legend-element map-control p-3">
+                <label>{{indicatorMetadata.name}}</label>
+                <div class="legend" v-for="(level, index) in scaleLevels" v-bind:key="index">
+                    <i v-bind:style="level.style"></i>
+                    <span class="level">{{ level.val }}</span>
+                    <span class="hidden" style="display: none">{{ level.style }}</span>
+                    <br/>
+                </div>
+                <div v-if="!!colourScales" class="adjust-scale mt-1">
+                    <a @click="toggleAdjust" href="">
```
r-ash, line 20:
```suggestion
                    <a @click.prevent="toggleAdjust" href="">
```

src/app/static/src/app/store/plotState/mutations.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export enum PlotStateMutations {
+    updateOutputColourScales = "updateOutputColourScales"
+}
+
+export const mutations: MutationTree<PlotState> = {
+    [PlotStateMutations.updateOutputColourScales](state: PlotState, action: PayloadWithType<UpdateColourScale>) {
+        const {indicatorId, newScaleSettings} = action.payload;
+        const newColourScales = {...state.output.colourScales}
```
r-ash, line 17:
```suggestion
        const newColourScales = structuredClone(state.output.colourScales)
```

PR #944, opened 2024-03-14
mrc 5161: Port bubble plot to composition API and add back in

r-ash:
## Description

This PR will add bubble plot back in

## Type of version change

None

## Checklist


src/app/static/src/app/components/plots/MapAdjustScale.vue
```diff
@@ -1,11 +1,11 @@
 <template>
-    <div class="p-3 ml-2">
+    <div class="p-3">
         <div class="form-check dynamic-filtered">
             <label class="form-check-label">
                 <input id="type-input-dynamic-filtered" class="form-check-input" type="radio" :name="scaleTypeGroup"
                        :value="scaleType.DynamicFiltered"
                        :checked="selectedScale.type === scaleType.DynamicFiltered"
-                       @change="updateOutputColourScale(indicatorMetadata.indicator, {...selectedScale, type: parseInt($event.target.value)})">
+                       @change="updateOutputScale(name, {...selectedScale, type: parseInt($event.target.value)})">
```
EmmaLRussell, line 8: Was this change just to make things a little less verbose..?

src/app/static/src/app/components/plots/MapLegend.vue
```diff
@@ -12,7 +12,7 @@
                 <label>{{indicatorMetadata.name}}</label>
                 <div class="legend" v-for="(level, index) in scaleLevels" v-bind:key="index">
                     <i v-bind:style="level.style"></i>
-                    <span class="level">{{ level.val }}</span>
+                    <span class="level">{{ level.label }}</span>
```
EmmaLRussell, line 15: Why this level change?

src/app/static/src/app/root.ts
```diff
@@ -68,8 +68,11 @@ export interface RootState extends DataExplorationState {
     modelCalibrate: ModelCalibrateState,
     modelOutput: ModelOutputState,
     plottingSelections: PlottingSelectionsState,
+    // Output data used for plotting
     plotData: PlotDataState,
+    // Control and filter selections for output plots
     plotSelections: PlotSelectionsState,
+    // Colour and size scales selected by user in output plots
```
EmmaLRussell, line 75: Nice!

EmmaLRussell (approved):
Looks good!

I was wondering if there was more scope to share code between the Choropleth and Bubble plot via composables, but actually I think it would be so fiddly because of the difference of Bubble having the size+colour rather than just colour that it wouldn't be worth it (and there is a lot of code being shared in the utils anyway). 

As a slightly separate issue, I find Country a bit weird as a default level, for both Choropleth and Bubble (where you initially see one huge blob!) - would it make more sense to make the default level the Area Level selected in the model options?

src/app/static/src/app/components/plots/MapAdjustScale.vue
r-ash: And because this now updates the size or the colour scale depending on the "name" prop.

src/app/static/src/app/components/plots/MapLegend.vue
r-ash: No actual change to what is shown here. Previously we were just returning without an explicit type, but I have added one and thought "label" was a better description of what this actually is over "val". See https://github.com/mrc-ide/hint/pull/944/files#diff-74523534d090d132b38de56ec6d1076c4c901addb21445974859b8494fff774fL51

r-ash:
> As a slightly separate issue, I find Country a bit weird as a default level, for both Choropleth and Bubble (where you initially see one huge blob!) - would it make more sense to make the default level the Area Level selected in the model options?

Agreed! We have a bit of work to do to go through the default selections here and set them back to the previous.

src/app/static/src/app/components/plots/useUpdateScale.ts
```diff
@@ -1,18 +1,82 @@
...
-    const updateOutputColourScale = (indicator: string, newScaleSettings: ScaleSettings) => {
+    const getScaleStep = (name: string) => {
+        if (name === "colour") {
+            return colourScaleStep.value;
+        } else if (name === "size") {
+            return sizeScaleStep.value;
+        } else {
+            throw new RangeError(`Scale type must be one of 'name' or 'colour', got ${name}`)
+        }
+    };
```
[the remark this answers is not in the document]
r-ash: Kept this in as otherwise typescript was giving me a warning about no return from this. And if I return undefined or something then I end up with having to manage that type elsewhere

M-Kusumgar (approved):
LGTM

PR #945, opened 2024-03-14
Update duckdb version to stable release

r-ash:
## Description

DuckDB have released version 0.10.0 which brings backwards compatibility and forwards compatibility! See https://duckdb.org/2024/02/13/announcing-duckdb-0100.html

This will help with quite an annoying dependency which we have pinned in the R impl, if we update everything to 0.10.0 should hopefully be simpler to install and going forwards mean we don't have to pin our version in R to the same as in the backend

## Type of version change

None

## Checklist


EmmaLRussell (approved):
Works for me!

PR #950, opened 2024-04-03
mrc-5162: Share project with support directly from error report endpoint

r-ash:
## Description

When a user submits an issue report, we submit the report and then submit another request to share the project with the support user. We've had a few issue reports where the message has been submitted in teams but the project has not been shared. I wonder if this is because potentially the user clicks submit report then shuts their browser before the 2nd request has been sent. 

This PR does 2 things
* Move the issue report and sharing with support user under a single backend endpoint
* Add some more logging around project sharing so we can investigate better should this occur again

## Type of version change

Patches

## Checklist

### DIFF
--- a/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ProjectTests.kt
+++ b/src/app/src/test/kotlin/org/imperial/mrc/hint/integration/ProjectTests.kt
@@ -27,7 +27,7 @@ class ProjectTests : VersionFileTests()
     fun `can create new project`(isAuthorized: IsAuthorized)
     {
         val result = createProject()
-        assertSecureWithSuccess(isAuthorized, result,null)
+        assertSecureWithSuccess(isAuthorized, result, null)
 
         if(isAuthorized == IsAuthorized.TRUE)
         {


scripts/run-dependencies.sh
```diff
@@ -51,3 +51,4 @@ docker run --rm --network=$NETWORK \
   -url=jdbc:postgresql://$DB/hint
 
 $HERE/add-test-user.sh
+$HERE/add-support-user.sh
```
r-ash, line 54: Without this testing out the issue report on local deployment is raising an error as support user doesn't exist

src/app/src/main/kotlin/org/imperial/mrc/hint/service/ProjectAndVersionService.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+
+interface ProjectVersionService
+{
+    fun cloneProjectToUser(projectId: Int, emails: List<String>)
+}
+
+@Service
+class ProjectAndVersionService (
```
EmmaLRussell, line 19: I'm a bit baffled by the respective names of the interface and implementation!
Does it really need to be an interface? There's only ever going to be one implementation, and we can still mock the class for testing. 
I think `ProjectService` would be fine personally!

src/app/src/test/kotlin/org/imperial/mrc/hint/database/ProjectRepositoryTests.kt
```diff
@@ -309,8 +306,8 @@ class ProjectRepositoryTests
         assertThat(p2.sharedBy).isEqualTo(null)
         assertThat(p2.versions.count()).isEqualTo(1)
         assertThat(p2.versions[0].id).isEqualTo("v2s1")
-        assertThat(p2.versions[0].created).isEqualTo(format(ago_3h))
-        assertThat(p2.versions[0].updated).isEqualTo(format(ago_1h))
+        assertThat(toDateTime(p2.versions[0].created)).isCloseTo(ago_3h, within(1, ChronoUnit.MICROS))
```
EmmaLRussell, line 309: Another tiny DRY suggestion - maybe make a little extension function for this isCloseTo - e.g. `isSameTime`. Could wrap the cast `toDateTime` into that too.

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ErrorReportController.kt
```diff
@@ -1,20 +1,55 @@
...
+    private val projectVersionService: ProjectVersionService,
+    private val appProperties: AppProperties,
+)
 {
+    private val logger = GenericLoggerImpl(LoggerFactory.getLogger(ErrorReportController::class.java))
+
     @PostMapping("/error-report")
     @ResponseBody
-    fun postErrorReport(@RequestBody errorReport: ErrorReport): ResponseEntity<String>
+    fun postErrorReport(@RequestBody errorReport: ErrorReport,
```
EmmaLRussell, line 24: Could just include projectId in the `ErrorReport` model I guess, rather than including in the qs. I'm a bit surprised it isn't in there already!

EmmaLRussell (commented):
This seems much nicer!

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ErrorReportController.kt
```diff
@@ -1,20 +1,55 @@
...
+                        @RequestParam("projectId", required = false) projectId: Optional<Int>
```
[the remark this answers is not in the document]
r-ash: Thanks! That's just lack of kotlin knowledge on my side. I've updated that

src/app/src/main/kotlin/org/imperial/mrc/hint/controllers/ErrorReportController.kt
r-ash: Thanks, yeah not sure why I didn't put it there in the first place. I've updated this too

M-Kusumgar (approved):
Looks much better!

PR #951, opened 2024-04-05
mrc-4969: Use relative path to read calibrate result data

r-ash:
## Description

This PR will update calibrate data to use path relative to the mounted results dir. Should make testing and deployment a bit simpler as we no longer have to have the results mounted at the same place.

Before this is merged we need to merge
* https://github.com/mrc-ide/hintr/pull/471
* https://github.com/mrc-ide/hint-deploy/pull/76
* Remove branch pin here

Getting this moving again now as I think we will want it to pass through the path to the downloads so that we can have a progress bar on them. See https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5218/Show-native-browser-progress-on-download for some context

## Type of version change
_Delete as appropriate_

Patches

## Checklist


r-ash:
I pushed this with the wrong ticket ID, re-opened in #955

PR #957, opened 2024-04-11
Mrc 4839 Add vitest mini epic base branch

M-Kusumgar:
## Description

You do not need to look at the whole PR just a couple of changes:
* delete old jest config files and uninstall jest packages
* update to node 20
* install vitest, add its types to tsconfig and add `vitest.config.js`
* add `global.currentUser` to `tests/setup.ts file`
* replace all `jest.` in tests folder with `vi.`

You can already run `npm run test` if you want, a surprising number already pass, but many will fail, future PRs that point to this branch will fix these issues

Note: for now we are not adding vite into hint, just vitest until we figure out how vite can replicate the hot reloading we have with vue-cli-service


EmmaLRussell (approved):
Good to get the lid off the can o worms! :worm:

Will the equivalent of the separate jest config files for adr vs non-adr come back in future branches?

r-ash (approved):
Looks good to me too, agree interested in hhow the separate integration or ADR tests etc will work. But perhaps that will all become clear in a later PR?

PR #958, opened 2024-04-11
Mrc 4840 Fix tests part 1

M-Kusumgar:
## Description

The diff is big but the changes are largely programmatic. Changes of interest:
* vi does not contain types `Mock` and `Mocked` (which jest did) so we have to replace `vi.Mock` to `Mock` and imoprt `Mock` from vitest
* done is deprecated and it is slow, this also removes all dones (these were used in timeouts so instead we are awaiting and advancing fake timers)
* mock plotly in plotly.test.ts (vi.mock is different from jest.mock)

### DIFF
--- a/src/app/static/package.json
+++ b/src/app/static/package.json
@@ -9,7 +9,7 @@
     "build": "rm -rf public && vue-cli-service build ./src/app/index.ts",
     "serve": "rm -rf public && VUE_CLI_SERVICE_CONFIG_PATH=$PWD/vue.config.js vue-cli-service serve ./src/app/index.ts",
     "copy": "cp -r ./templates/*.ftl ./public && cp -r ./assets/* ./public",
-    "test": "vitest --run",
+    "test": "vitest",
     "browser-test": "npx playwright test",
     "integration-test": "jest --testPathIgnorePatterns adr-dataset -c jest.integration.config.js",
     "adr-integration-test": "jest --testPathPattern adr-dataset -c jest.integration.adr.config.js",
--- a/src/app/static/src/tests/components/genericChart/genericChart.test.ts
+++ b/src/app/static/src/tests/components/genericChart/genericChart.test.ts
@@ -197,7 +197,7 @@ describe("GenericChart component", () => {
         return shallowMountWithTranslate(GenericChart, store, {global: {plugins: [store]}, props});
     };
 
-    it("renders as expected without chart data", (done) => {
+    it("renders as expected without chart data", async () => {
         mockAxios.onGet(`/dataset1`)
             .reply(200, mockSuccess({metadata: {defaults: {}}}));
         mockAxios.onGet(`/dataset2`)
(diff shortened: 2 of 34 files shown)


src/app/static/package.json
```diff
@@ -9,7 +9,7 @@
     "build": "rm -rf public && vue-cli-service build ./src/app/index.ts",
     "serve": "rm -rf public && VUE_CLI_SERVICE_CONFIG_PATH=$PWD/vue.config.js vue-cli-service serve ./src/app/index.ts",
     "copy": "cp -r ./templates/*.ftl ./public && cp -r ./assets/* ./public",
-    "test": "vitest --run",
+    "test": "vitest",
```
EmmaLRussell, line 12: This runs tests with watch?

src/app/static/src/tests/components/genericChart/genericChart.test.ts
```diff
@@ -753,16 +731,13 @@ describe("GenericChart component", () => {
...
-        setTimeout(() => {
-            wrapper.findAllComponents(DataSource)[0].vm.$emit("update", "dataset2");
-            setTimeout(() => {
-                expect(mockAxios.history.get.length).toBe(0);
-                done();
-            });
-        });
+        await flushPromises();
+        wrapper.findAllComponents(DataSource)[0].vm.$emit("update", "dataset2");
+        await nextTick();
```
EmmaLRussell, line 739: Should this be a flushPromises? It is in the above test, where you're testing for the positive case where an API fetch *is* done.

src/app/static/src/tests/components/genericChart/plotly.test.ts
```diff
@@ -1,12 +1,19 @@
...
+        react: vi.fn()
+    }
+});
 vi.mock("plotly.js-basic-dist", () => ({
-    newPlot: vi.fn(),
-    react: vi.fn()
+    default: {
+        newPlot: mocks.newPlot,
+        react: mocks.react
+    }
```
EmmaLRussell, line 15: I'm a bit confused about what's going on here. `newPlot` and `react` are being mocked twice?

r-ash (approved):
Just had a quick skim through this but looks good to me! :)

src/app/static/package.json
M-Kusumgar: yh was just for while i was fiddling around with tests and debugging, this is reverted to `--run` later. But perhaps i should also have a "test:dev" script so that it is easier for us? Can do that in a future PR

src/app/static/src/tests/components/genericChart/genericChart.test.ts
M-Kusumgar: yes nice spot! as time went on i started leaning more towards `flushPromises` so hopefully should not be a frequent problem! will change it

M-Kusumgar: done

src/app/static/src/tests/components/genericChart/plotly.test.ts
M-Kusumgar: yh i realised i can just do one of them, should be good now, removed the hoisted stuff

EmmaLRussell (approved):
Nice, thanks!

PR #959, opened 2024-04-11
mrc-4841 Fix tests part 2

M-Kusumgar:
## Description

This PR is a little messy, random misc changes:
* more vi.mocks because it is different from jest.mock (mainly mocking leaflet)
* add disabled, clearable, placeholder and searchable props to hint tree select because before we were testing the base treeselect as the properties were being passed on automatically in the component, but vitest doesnt seem to detect the base treeselect hence added those props to hint tree select to make it testable
* bootstrap expanding and collapsing rows in projects page was being slow, taking 1s in tests so just testing that the bootstrap animation has begun
* I have no idea what is going on with reset password localhost url, there is a comment i left about it

We are almost there actually, all front end tests work, just need to add in all the other configs like integration tests, etc

### DIFF
--- a/buildkite/shared-build-env.dockerfile
+++ b/buildkite/shared-build-env.dockerfile
@@ -1,4 +1,4 @@
-FROM mrcide/node-docker:master
+FROM mrcide/node-20-docker:master
 
 RUN apt-get install wget
 
--- a/buildkite/shared-build-env.dockerfile
+++ b/buildkite/shared-build-env.dockerfile
@@ -1,4 +1,4 @@
-FROM mrcide/node-20-docker:master
+FROM mrcide/node-20-docker:main
 
 RUN apt-get install wget
 
--- a/src/app/static/src/tests/components/outputTable/tableData.test.ts
+++ b/src/app/static/src/tests/components/outputTable/tableData.test.ts
@@ -1,9 +1,16 @@
-const mockAddFilteredData = vi.fn();
-const mockDownload = vi.fn();
-const mockExportService = vi.fn().mockReturnValue({
-    addFilteredData: mockAddFilteredData.mockReturnValue({
+const mocks = vi.hoisted(() => {
+    const mockAddFilteredData = vi.fn();
+    const mockDownload = vi.fn();
+    const mockExportService = vi.fn().mockReturnValue({
+        addFilteredData: mockAddFilteredData.mockReturnValue({
+            download: mockDownload
+        })
+    });
+    return {
+        exportService: mockExportService,
+        addFilteredData: mockAddFilteredData,
         download: mockDownload
-    })
+    }
 });
 
 import { shallowMount } from "@vue/test-utils";
@@ -14,7 +21,7 @@ import TableReshapeData from "../../../app/components/outputTable/TableReshapeDa
 import DownloadButton from "../../../app/components/downloadIndicator/DownloadButton.vue";
 
 vi.mock("../../../app/dataExportService", () => {
-    return { exportService: mockExportService };
+    return { exportService: mocks.exportService };
 });
 
 const mockFilters = [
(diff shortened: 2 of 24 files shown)


buildkite/shared-build-env.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mrcide/node-docker:master
+FROM mrcide/node-20-docker:main
```
r-ash, line 1: :heart:

r-ash (commented):
Where is the comment about the localhost stuff? Can't see that in this PR. I think fine with those changes if these tests are passing OK

src/app/static/src/tests/components/password/resetPassword.test.ts
```diff
@@ -93,7 +93,7 @@ describe("Reset password component", () => {
             "Por favor, solicite outra ligação aqui.",
             store);
         expect((wrapper.find("#request-new-link a").element as HTMLLinkElement).href)
-            .toEqual("http://localhost/password/forgot-password");
+            .toEqual("http://localhost:3000/password/forgot-password");
```
EmmaLRussell, line 96: I think this might just be because vitest is assuming a base url of http://localhost:3000 rather than http://localhost - I see you took out the url environment option from config, maybe that made a difference?

src/app/static/src/tests/components/plots/bubble/sizeLegend.test.ts
```diff
@@ -7,11 +7,15 @@ import MapAdjustScale from "../../../../app/components/plots/MapAdjustScale.vue"
 import {ScaleType} from "../../../../app/store/plottingSelections/plottingSelections";
 import { mountWithTranslate, shallowMountWithTranslate } from "../../../testHelpers";
 
-vi.mock("@vue-leaflet/vue-leaflet", () => {
+vi.mock("@vue-leaflet/vue-leaflet", async () => {
+    const actual = await vi.importActual("@vue-leaflet/vue-leaflet")
```
EmmaLRussell, line 11: Why do you need this?

src/app/static/src/tests/components/outputTable/tableData.test.ts
```diff
@@ -1,9 +1,16 @@
...
+            download: mockDownload
+        })
+    });
+    return {
+        exportService: mockExportService,
+        addFilteredData: mockAddFilteredData,
         download: mockDownload
-    })
+    }
 });
```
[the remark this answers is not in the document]
M-Kusumgar: yep exactly

src/app/static/src/tests/components/password/resetPassword.test.ts
M-Kusumgar: so i had it in the config to begin with and it didnt make a difference, and then on further investigation i realised the environment options are per environment so you have to do
```javascript
environmentOptions: {
  jsdom: {
    url: "http://localhost"
  }
}
```
and then it works! so mystery solved

src/app/static/src/tests/components/plots/bubble/sizeLegend.test.ts
M-Kusumgar: so this lets you partially mock the import, unfortunately if i remove it then it errors saying that `LMap` is not defined in the Choropleth component even though we are only testing the SizeLegend, i have no idea but the cli hints at doing something like this so i just followed it

M-Kusumgar:
> Where is the comment about the localhost stuff? Can't see that in this PR. I think fine with those changes if these tests are passing OK

sorry forgot to add a comment, link is here: https://github.com/mrc-ide/hint/pull/959#discussion_r1565761367 it is fixed now though! just had to tweak the config in the correct way

EmmaLRussell (approved):

r-ash (approved):

PR #960, opened 2024-04-11
Mrc 5243 Add integration test command back

M-Kusumgar:
## Description

Just added integration tests back in with vitest. Some integration tests actually need to wait for the backend so we cannot advance timers like in the unit tests. So I have added an optional interval arg that we can use when we call the polling functions to speed up testing.

There are a couple of changes which will be immediately changed in the next PR:
* we are no longer doing the docker version of tests such as `integration-test-docker` or the docker ADR one, so those config will be removed and the commands removed from `package.json`
* our version of gradle support java 18 so i upgrade it in the next PR
* polling interval here is quite small so have increase it to 400ms instead in the next PR


src/app/src/test/kotlin/org/imperial/mrc/hint/database/ProjectRepositoryTests.kt
```diff
@@ -309,8 +310,8 @@ class ProjectRepositoryTests
         assertThat(p2.sharedBy).isEqualTo(null)
         assertThat(p2.versions.count()).isEqualTo(1)
         assertThat(p2.versions[0].id).isEqualTo("v2s1")
-        assertThat(p2.versions[0].created).isEqualTo(format(ago_3h))
-        assertThat(p2.versions[0].updated).isEqualTo(format(ago_1h))
+        assertThat(roundToSec(toDateTime(p2.versions[0].created))).isEqualTo(roundToSec(ago_3h))
```
M-Kusumgar, line 313: @r-ash did a better job of this in one of his PRs, this will be a merge conflict down the line and I can replace this with his implementation, keeping it in for now so the tests pass

src/app/static/package.json
```diff
@@ -11,10 +11,10 @@
...
     "test": "vitest --run",
     "browser-test": "npx playwright test",
-    "integration-test": "jest --testPathIgnorePatterns adr-dataset -c jest.integration.config.js",
-    "adr-integration-test": "jest --testPathPattern adr-dataset -c jest.integration.adr.config.js",
-    "integration-test-docker": "jest --testPathIgnorePatterns adr-dataset -c jest.integration.docker.config.js",
-    "adr-integration-test-docker": "jest --testPathPattern adr-dataset -c jest.integration.docker.adr.config.js",
+    "integration-test": "vitest --config ./test-integration-configs/integration.ts --run",
+    "adr-integration-test": "vitest --config ./test-integration-configs/adr-integration.ts --run",
+    "integration-test-docker": "vitest --config ./test-integration-configs/docker-integration.ts --run",
+    "adr-integration-test-docker": "vitest --config ./test-integration-configs/adr-docker-integration.ts --run",
```
r-ash, line 17: Do you remember why we run some of these under docker and other not? I can't remember why the 2 configurations here

r-ash (approved):
Thanks, this looks great

src/app/static/package.json
M-Kusumgar: i think it was just to test that they can connect via docker i guess, as the code is run in docker containers, i think a future smoke test might be sufficient though if we even want to test that, the only things that are different is the connection url and the fact that we start the server in a docker container i think

EmmaLRussell: I think the feeling was possibly that the integration tests ought to run in docker on CI to reflect a real deployment, but we wanted to keep option to run on metal for local dev. Suspect there might have been at least one gotcha where tests passed on metal but there was an issue in docker which only showed up in manual testing on deployed instance. 
Not sure we should ditch them tbh, they don't really cost us anything. Of course if we had a fuller e2e test suite...

src/app/static/src/tests/integration/downloadResults.itest.ts
```diff
@@ -143,10 +135,11 @@ describe(`download results actions integration`, () => {
...
 
-        actions.poll({commit, dispatch, state, rootState: root} as any, DOWNLOAD_TYPE.SPECTRUM);
-        vi.advanceTimersByTime(3000);
-        await flushPromises();
-        expect(commit.mock.calls.length).toBe(2);
+        actions.poll({commit, dispatch, state, rootState: root} as any, DOWNLOAD_TYPE.SPECTRUM, 100);
+        await vi.waitUntil(() => commit.mock.calls.length >= 2, {
+            interval: 100,
+            timeout: 6000
+        })
```
EmmaLRussell, line 142: This code block is largely repeated 5 times - could pull it into a separate util method, especially is shared time values might need tweaking?
...or maybe just the waitUntil bit, which also shows up in other modules?..

EmmaLRussell (approved):
Looks good - one tiny suggestion..

src/app/static/src/tests/integration/downloadResults.itest.ts
M-Kusumgar: I did try this, was running into issues and i am not sure why honestly, but it seems to work when i have vi.waitUntil in the test function, as a compromise I have abstracted out the interval and timeout in the test suites that they were used multiple times in!

PR #961, opened 2024-04-12
Mrc 5246 Move testing to github actions

M-Kusumgar:
## Description

So this one is a little large but a large amount of changes also come from deleting the old dockerfiles and shell files and create github actions files with largely the same logic. So hopefully this isn't super bad to review!

Main changes:
* composite action for making a shared env between builds, little bit of trickiness here with getting the git branch and git commit sha because github actions exposes different context objects in the event of a push (merge to `main`) or a pull request.
* Caching `node_modules` folder, there were some warnings online to not do this because packages might break between node versions so I am caching based on node version too
* Some kotlin tests were optimised, instead of waiting for 5 model runs we only wait for one before all the tests and then do our tests (also moved a metadata test that did something similar into download tests). In the adr tests we were importing geo files multiple times but can just be condensed into one test that tests if we can save all our files
* Removed docker test commands and their configs
* Added run kotlin server script to run the server in a background process and wait for it to get ready
* Increase intervals on integration tests

## Times
Without npm cache: ~6m 30s
With npm cache: ~6m


.github/workflows/test-back-end.yml
```diff
@@ -0,0 +1,36 @@
...
+                install-packages: 'false'
+
+            - name: Run deps
+              run: scripts/run-dependencies.sh
+
+            - name: Prep back end tests
+              run: |
+                mkdir -p src/app/static/public
+                mkdir -p src/app/static/public/css
+                yes ".class {height: 500px;}" | head -n 200 >> src/app/static/public/css/app.css
```
M-Kusumgar, line 29: Basically here I am making a dummy build of the app, turns out that the backend tests only need the freemarker template files and the css file, this css file needs to be big enough for gzip to activate, so I am creating a very bad css file that has `.class {height: 500px;}` written 200 times

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,38 @@
...
+              run: src/gradlew -p src :app:bootDistTar
+
+            - name: Build docker image
+              run: docker build -f ./Dockerfile --tag "${CI_ORG}/hint:${CI_SHA:0:7}" src
+
+            - name: Push image
+              run: |
+                docker tag "${CI_ORG}/hint:${CI_SHA:0:7}" "${CI_ORG}/hint:${CI_BRANCH}"
+                docker push "${CI_ORG}/hint:${CI_SHA:0:7}"
+                docker push "${CI_ORG}/hint:${CI_BRANCH}"
```
r-ash, line 38: Can we push these to github container registry too? See https://github.com/mrc-ide/hintr/blob/master/.github/workflows/docker.yaml

And fairly pedantic but can we name this .yml something to do with docker? `docker.yml` or `build-and-push-docker`?

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,38 @@
...
+                username: ${{ secrets.DOCKERHUB_USERNAME }}
+                password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+            - name: Make shared env
+              uses: ./.github/actions/make-shared-env
+
+            - name: Build app
+              env:
+                NODE_ENV: production
+              run: src/gradlew -p src :app:bootDistTar
```
r-ash, line 29: Why d we need to build the app here?

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,38 @@
...
+
+            - name: Build docker image
+              run: docker build -f ./Dockerfile --tag "${CI_ORG}/hint:${CI_SHA:0:7}" src
+
+            - name: Push image
```
r-ash, line 34: Are we able to use the docker/build-push-action here? I think there is support for tagging

.github/workflows/test-back-end.yml
r-ash: Can we document this in the yml? Think you can do this at the top of a run block using
```
: # Create dummy build of app...
```

r-ash (changes requested):
This looks great, couple of small notes.

.github/workflows/build-and-push.yml
M-Kusumgar: the dockerfile just copies the build files, we dont build the app at any point before this

.github/workflows/build-and-push.yml
M-Kusumgar: yep done

.github/workflows/test-back-end.yml
M-Kusumgar: done

PR #962, opened 2024-04-17
mrc-5182: Add barchart back in

r-ash:
## Description

This PR will
* Migrate barchart to composition API and add it back in

I think is is mostly looking good, I had a bit of a fight with this one so up for any suggestions for tidying it up at all. One bit left which still feels a little janky is the updating of the error bars when you click on the legend. It doesn't remove the error bars when you do this, just animates between them. I tried to get the standard hide the error bars, and have them re-appear once the update has completed but I couldn't get this working smoothly. I think it acts weirdly because of timing pains with when the legend onClick is run. And I couldn't see another hook we could use to hide the error bars before this.

## Type of version change

None

## Checklist


src/app/static/src/app/store/plotSelections/getters.ts
```diff
@@ -0,0 +1,16 @@
+import {ControlSelection, FilterSelection, PlotName, PlotSelectionsState} from "./plotSelections";
+import {FilterOption} from "../../generated";
+
+export const getters = {
+    controlSelectionFromId: (state: PlotSelectionsState) => (plotName: PlotName, controlId: string): FilterOption | undefined => {
+        return state[plotName].controls.find((f: ControlSelection) => f.id === controlId)?.selection[0]
+    },
+    filterSelectionFromId: (state: PlotSelectionsState) => (plotName: PlotName, filterId: string): FilterOption[] => {
+        const filterSelection = state[plotName].filters.find((f: FilterSelection) => f.filterId === filterId);
```
M-Kusumgar, line 9: ??? probably, we think
```suggestion
        const filterSelection = state[plotName].filters.find((f: FilterSelection) => f.stateFilterId === filterId);
```

src/app/static/src/app/components/plots/bar/Barchart.vue
```diff
@@ -0,0 +1,200 @@
...
+        const controlSelectionFromId = store.getters["plotSelections/controlSelectionFromId"];
+        const filterSelectionFromId = store.getters["plotSelections/filterSelectionFromId"];
+        const filterIdToColumnId = store.getters["modelCalibrate/filterIdToColumnId"];
+
+        const chartData = ref<BarChartData>({datasets:[], labels: [], maxValuePlusError: 0});
+        const chartOptions = ref({});
+        const displayErrorBars = ref<boolean>(false);
+        const showNoDataMessage = computed(() => {
+            return chartData.value.datasets.length == 0
+        });
```
M-Kusumgar, line 61: could also just do this in the v-if in the template

PR #963, opened 2024-04-25
change from default plot setting to default effect

M-Kusumgar:
## Description

Instead of relying on the default plot control and `defaultFilterTypes` props, we now have a new property called `defaultEffect` and this will contain any effect you want to occur for all plot settings in a given plot

Note: If you would like to see how the metadata changes, see the description here: https://github.com/mrc-ide/hintr/pull/493, I am keeping the description up to date


r-ash (approved):
Looks great, thanks

PR #965, opened 2024-04-30
mrc-4856: Add calibration barchart back in

r-ash:
## Description

This PR adds the calibration plot back into hint using new metadata format, barchart component and filter component

See https://github.com/mrc-ide/hintr/pull/510 for the metadata

## Type of version change

None

## Checklist


src/app/static/src/app/store/plotSelections/getters.ts
```diff
@@ -12,5 +15,59 @@ export const getters = {
         } else {
             return filterSelection.selection
         }
+    },
+    chartData: (state: PlotSelectionsState, getters: any, rootState: RootState, rootGetters: any) =>
```
r-ash, line 19: This is a bit nasty but felt better than reworking all of the individual getters, this way the hardcoding for the plot is just in this once place which I don't think is too bad. There might be a nicer way to write this though...

src/app/static/src/app/components/plots/FilterSet.vue
```diff
@@ -15,6 +15,7 @@ import {useStore} from "vuex";
 import {RootState} from "../../root";
 import Filter from "./Filter.vue";
 import {PlotName} from "../../store/plotSelections/plotSelections";
+import {getMetadataFromPlotName} from "../../store/plotSelections/actions";
```
M-Kusumgar, line 18: might not be used

src/app/static/src/app/components/plots/FilterSet.vue
```diff
@@ -1,12 +1,12 @@
...
-
     <div class="form-group" v-for="f of filters" :key="f.stateFilterId">
-        <label class="font-weight-bold">{{f.label}}</label>
-        <!-- For some reason using filter is saying no component registered
-        with this name, no idea why so using capital Filter -->
-        <Filter :state-filter-id="f.stateFilterId" :plot="plot"/>
+        <div v-if="f.visible">
+            <label class="font-weight-bold">{{f.label}}</label>
+            <!-- For some reason using filter is saying no component registered
+            with this name, no idea why so using capital Filter -->
```
M-Kusumgar, line 6: i believe this happens when vue gets confused with the name, `filter` might be a special name for vue, i know in vue 2 there was a `Vue.filter` so it might be confusing it with that special filter, happy to keep it `Filter` i think that works, if you do want to stick with consistent casing, you can call it something else
```ts
import FilterComponent from "./Filter.vue";
```
and use it as `<filter-component/>` and you should not get the error, but up to you!

src/app/static/src/app/components/plots/bar/utils.ts
```diff
@@ -4,6 +4,7 @@ import {AnnotationOptions} from "chartjs-plugin-annotation";
 import {formatOutput} from "../utils";
 import {PlotData} from "../../../store/plotData/plotData";
 import {Dict} from "../../../types";
+import {PlotName} from "../../../store/plotSelections/plotSelections";
```
M-Kusumgar, line 7: is this used?

src/app/static/src/app/components/plots/bubble/Bubble.vue
```diff
@@ -63,10 +63,10 @@ const getSizeIndicator = () => {
 const colourIndicator = ref<string>(getColourIndicator());
 const sizeIndicator = ref<string>(getSizeIndicator());
 const colourIndicatorMetadata = computed<ChoroplethIndicatorMetadata>(() => {
-    return getIndicatorMetadata(store, colourIndicator.value)
+    return getIndicatorMetadata(store, "bubble", colourIndicator.value)
```
M-Kusumgar, line 66: would probably extract these out into a top level variable called `plotName` or something like that (i have done that in my branch of the input choropleth too) just so its in one variable although no biggy

M-Kusumgar (commented):
Looks good, couple of small comments but works like a charm, is awesome to see it all come together!

r-ash:
Thanks, updated re those comments!

PR #966, opened 2024-05-01
mrc-5305: Fix issue with projects not reloading properly because of setTimeout interval issue

r-ash:
## Description

We updated this for testing but that highlighted a problem where we were passing some options as a payload :see_no_evil: this was causing issues with project reloading as state was being overwritten because the setTimeout poll interval was being set very low. This fixes that.

I had a look elsewhere for other instances of the new `interval = null` but I can't see any which are setting any options in the `dispatch` call.

## Type of version change

Patches

## Checklist


M-Kusumgar (approved):
had a look around too, did not find any more intervals either, looks good to me!

PR #972, opened 2024-05-22
Mrc 5286 Add input choropleth in

M-Kusumgar:
## Description

Basically started again and copied and pasted bits of code into this new local branch. It's been a while so I've forgotten all the different considerations I was making at the time but I think this works.

Main changes:
* pass the plot prop everywhere, just to make the choropleth more generic
* fixed some typescript issues
* extracted all the filtering out in its own file, i have a filterData function that hopefully will replace all the repeated bits of logic but perhaps we leave that extraction to another PR
* small simplification of the update scale composable

### DIFF
--- a/src/app/static/src/app/components/plots/choropleth/Choropleth.vue
+++ b/src/app/static/src/app/components/plots/choropleth/Choropleth.vue
@@ -13,7 +13,8 @@
                 <reset-map @reset-view="updateBounds"></reset-map>
                 <map-legend :indicator-metadata="indicatorMetadata"
                             :scale-levels="scaleLevels"
-                            :selected-scale="selectedScale!"></map-legend>
+                            :selected-scale="selectedScale!"
+                            :plot="plot"></map-legend>
             </template>
         </l-map>
     </div>
@@ -21,7 +22,7 @@
 </template>
 
 <script setup lang="ts">
-import {computed, onMounted, ref, watch} from "vue";
+import {PropType, computed, onMounted, ref, toRefs, watch, defineProps} from "vue";
 import {useStore} from "vuex";
 import {RootState} from "../../../root";
 import { LMap, LGeoJson } from "@vue-leaflet/vue-leaflet";
@@ -42,15 +43,24 @@ import { ScaleSettings } from "../../../store/plotState/plotState";
 import {useChoroplethTooltips} from "./useChoroplethTooltips";
 import {getFeatureData} from "./utils";
 import MapEmptyFeature from "../MapEmptyFeature.vue";
+import { PlotName } from "../../../store/plotSelections/plotSelections";
+
+const props = defineProps({
+    plot: {
+        type: String as PropType<PlotName>,
+        required: true
+    }
+});
+const { plot } = toRefs(props);
 
 const store = useStore<RootState>();
-const plotData = computed(() => store.state.plotData.choropleth as CalibrateDataResponse["data"]);
+const plotData = computed(() => store.state.plotData[plot.value] as CalibrateDataResponse["data"]);
 
 const selectedIndicator = computed<string>(() => {
-    return store.state.plotSelections.choropleth.filters.find(f => f.stateFilterId === "indicator")!.selection[0].id
+    return store.state.plotSelections[plot.value].filters.find(f => f.stateFilterId === "indicator")!.selection[0].id
 });
 const indicatorMetadata = computed<ChoroplethIndicatorMetadata>(() => {
-    return getIndicatorMetadata(store, "choropleth", selectedIndicator.value)
+    return getIndicatorMetadata(store, plot.value, selectedIndicator.value)
 });
 const colourRange = ref<NumericRange | null>(null);
 const scaleLevels = ref<ScaleLevels[]>([]);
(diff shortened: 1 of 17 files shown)


src/app/static/src/app/components/plots/utils.ts
```diff
@@ -161,6 +162,8 @@ export const formatOutput = function (value: number | string, format: string, sc
 };
 
 const getPlotDataForIndicator = (indicatorMetadata: ChoroplethIndicatorMetadata, plotData: CalibrateDataResponse["data"]) => {
+    // if plotData doesn't have indicator key then we don't filter by it
+    if (!plotData?.at(0)?.hasOwnProperty("indicator")) return plotData;
```
r-ash, line 166: When do we not have an indicator?

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
```diff
@@ -146,7 +156,7 @@ const getStyle = (feature: Feature) => {
 }
 
 const emptyFeature = computed(() => {
-    return selectedAreaIds.value.length == 0
+    return plotData.value.length === 0
```
r-ash, line 159: I think this change has broken the boundary changes. You can recreate if you
1. Open the plot, set the detail level to something lower than country e.g. district
2. Update the area to exclude Northern -> no boundary update
3. Update the areas to include Northern -> boundary update to box containing just central and southern districts

If this is "selectedAreaIds" then should update correctly

src/app/static/src/app/store/metadata/actions.ts
```diff
@@ -23,18 +25,21 @@ export const actions: ActionTree<MetadataState, RootState> & MetadataActions = {
...
         }
         if (!sap.programError && sap.program) {
-            fileTypes.push(FileType.Programme)
+            fileTypes.push(FileType.Programme);
+            addAreaLevelsToSAPData(rootState, "program", commit);
         }
         if (!sap.surveyError && sap.survey) {
-            fileTypes.push(FileType.Survey)
+            fileTypes.push(FileType.Survey);
+            addAreaLevelsToSAPData(rootState, "survey", commit);
```
r-ash, line 42: Do we have to worry at all about how much we're saving into local storage here? Or is that not a problem?

src/app/static/src/app/store/plotData/filter.ts
```diff
@@ -0,0 +1,215 @@
...
+import { InputTimeSeriesKey } from "./plotData";
+import { Dict } from "@reside-ic/vue-next-dynamic-form";
+import { SurveyAndProgramState } from "../surveyAndProgram/surveyAndProgram";
+
+type FilteredDataContext = {
+    commit: Commit,
+    rootState: RootState
+}
+
+export const getOutputFilteredData = async (plot: OutputPlotName, selections: PlotSelectionUpdate["selections"]["filters"], context: FilteredDataContext) => {
```
r-ash, line 20: Great to have this all in one place

r-ash (changes requested):
This looks great, one change requested with the map boundary updating. We should be probably get the default metadata updated too to closer match what we have in current production. But we can do that down the line

src/app/static/src/app/components/plots/utils.ts
M-Kusumgar: survey data has indicators but art and anc data in the inputs does not have indicator

src/app/static/src/app/store/metadata/actions.ts
M-Kusumgar: this just replaces them in state, so we should be using basically the same amount of memory right?

src/app/static/src/app/components/plots/choropleth/Choropleth.vue
M-Kusumgar: this is a big problem actually, have been messing around with it, this should really be based off `plotData` and not `selectedAreaIds` because:
* currently if you select area level as "Country" and selected area as "Northern" there is no empty map feature which is not the behaviour on production
* since ART data (at least for `MWI_demo`) is only for "District + Metro" area level, there will be `selectedAreaIds` when you are on area level "Country" but there is no plot data so you again dont see an empty map feature but have an empty map

I think the issue is because `plotData` is being fetched so is an async action, but so is the debounced `updateBounds` theres 2 levels of async here and something is basically going wrong with the debounced function, at least that is my working theory

I have reverted this to `selectedAreaIds` for now but we have broken empty map cases we should fix!

r-ash: Good spot, thanks, I've made a ticket here https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5427/Fix-issues-with-no-empty-data-message-showing

src/app/static/src/app/components/plots/utils.ts
r-ash: Oh yep sorry being little slow here.

src/app/static/src/app/store/metadata/actions.ts
r-ash: Oh sorry, I wasn't looking at this properly. Was thinking this was writing area hierarchy into the data for each of these. But that isn't the case.

r-ash (approved):
Looks excellent! Thanks

PR #977, opened 2024-08-23
Fix remaining tests

r-ash:
## Description

This PR should fix up all remaining tests

In next PR I will remove all old code

## Type of version change

None

## Checklist

### DIFF
--- a/src/app/static/src/app/components/modelCalibrate/CalibrationResults.vue
+++ b/src/app/static/src/app/components/modelCalibrate/CalibrationResults.vue
@@ -1,4 +1,7 @@
 <template>
+    <hr />
+    <h3 v-translate="'calibrateResultsHeader'"></h3>
+    <p class="text-muted" v-translate="'calibrateResultsDesc'"></p>
     <div class="row">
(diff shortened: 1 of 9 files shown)


src/app/static/src/app/components/modelCalibrate/CalibrationResults.vue
```diff
@@ -1,4 +1,7 @@
 <template>
+    <hr />
+    <h3 v-translate="'calibrateResultsHeader'"></h3>
+    <p class="text-muted" v-translate="'calibrateResultsDesc'"></p>
```
M-Kusumgar, line 4: Bit confused as to why this is needed, does it fix any tests?

M-Kusumgar (approved):
Looks good just a quick question

src/app/static/src/app/components/modelCalibrate/CalibrationResults.vue
r-ash: Nope, just brings this up to same as current prod

PR #978, opened 2024-08-23
mrc-5396: Remove old code

r-ash:
## Description

The main goal of this PR is to remove plottingSelections state as we no longer need this. It does a few additional things
* Ensure that metadata is fetched afresh when the language changes

I've checked that metadata is also fetched afresh if some input data has changed. This is the case as we fetch it on mount so it is retrieved every time they view the app. We should think about if this is what we want. Previously we fetched this metadata as soon as they uploaded a PJNZ file. We could do similar and fetch this when all required input files are available. But that involves a bit more complicated logic. 

Few more things to look at
* I noticed that input metadata is taking ~1.5 s to respond, this is a bit too slow I think. I will have a look at if I can speed this up.
* We need to a loading indicator to input plots
* We should update default filters for input plots too
* Display an error from reviewInputMetadata fetch in the UI somewhere

## Type of version change

None

## Checklist


r-ash:
> We should have this mark step 2 as complete as soon as required files have been uploaded

Actually thought about this more, if there is an error when the review input metadata is fetched, we don't want users to be able to progress. So keep the logic as is, they can only continue when that data has been fetched successfully

PR #980, opened 2024-08-27
Fix holes in test coverage

r-ash:
## Description

This PR addresses some of the missing code coverage

## Type of version change

None

## Checklist

### DIFF
--- a/src/app/static/src/app/store/shared/readyStateMutations.ts
+++ b/src/app/static/src/app/store/shared/readyStateMutations.ts
@@ -1,7 +1,0 @@
-import {ReadyState} from "../../root";
-
-export const readyStateMutations = {
-    Ready(state: ReadyState) {
-        state.ready = true;
-    }
-};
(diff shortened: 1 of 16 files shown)


src/app/static/src/app/store/shared/readyStateMutations.ts
```diff
@@ -1,7 +0,0 @@
-import {ReadyState} from "../../root";
-
-export const readyStateMutations = {
```
r-ash, line 3: This was never called

src/app/static/src/tests/plotSelections/utils.test.ts
```diff
@@ -554,4 +554,78 @@ describe("Plot selections utils", () => {
...
+                }
+            ],
+            setMultiple: ["area"]
+        }];
+
+        const update = filtersInfoFromEffects(effects, rootState, metadata);
+
+        // TODO: this is a bit of a weird result, is this really what we want?
+        // Selected top level, and 1 step of children too but not below?
+        expect(update).toStrictEqual([
```
r-ash, line 620: This looks a little weird to me, is it what you're expecting @M-Kusumgar. Let's chat about this when we meet up tomorrow

r-ash: This is just slightly wrong type! Children needs to be an array

M-Kusumgar (approved):

PR #982, opened 2024-08-28
remove reviewinputfetched and add loading state

M-Kusumgar:
## Description

I removed the specific review input fetched boolean and replaced it with a more general loading boolean in generic chart


r-ash (approved):
This is great, thanks! Just merge conflict to sort and then good to go

PR #983, opened 2024-08-29
mrc-4856: Add playwright test for input time series

r-ash:
## Description

This PR does some initial setup for input plot testing. Not much actual testing here but I think this is enough code and can chat about general approach
This will
1. Update the playwright setup to projects, this is the recommended approach in current docs as it produces a trace if setup fails
2. Added a fixture for controlling projects
3. Adding a test that a project can be created and deleted
4. Added a test with a screenshot that files can be uploaded and input time series is rendered

Next steps after this thinking about tests
Time series
* Can change data source

Choropleth
* Can initialise plot
* Can change data source
* Can change indicator
* Can adjust scale
* Can set a specific area (see that part of map alone gets highlighted)

Think that should be enough? But will have a think.

## Type of version change

None

## Checklist


r-ash:
I've spotted an issue here running these locally in parallel. When we run with multiple workers, there can be a race condition where 
1. review-input spec runs, logs in and creates a new version
2. login spec starts, creates a new project
3. review-input uploads first file
4. login spec completes, deletes project, which clears version ID from the session
5. review-input uploads another file, this sees there is no version ID so creates a random one (FileManager saveFIle calls session.getVersionId() which generates a new ID if it is null in the session store) tries to save and then throws a foreign key constraint error as this project version doesn't exist

So we need to run these sequentially for now, but I wonder if we can fix this? Is this because session is scoped to the browser? I wonder if worth scoping it to a specific tab? Is that something we can do?

M-Kusumgar (changes requested):
Awesome, projectPage fixture is actually pretty neat. Couple comments about general playwright things

M-Kusumgar (approved):
Nice one

PR #988, opened 2024-09-03
Input review e2e

r-ash:
## Description

This PR adds end to end tests for input time series and input map :)

## Type of version change

None

## Checklist


src/app/static/src/tests/e2e/review-input.spec.ts
```diff
@@ -10,12 +11,137 @@ test('can view review input plots', async ({ projectPage }) => {
...
+    const waitForPlotUpdate = await buildWaitForUpdateCallback(page.locator(".leaflet-container"));
+
+    // When I change indicator
+    await page.getByRole("button", { name: "HIV prevalence" }).click();
+    await page.locator("a").filter({ hasText: "Viral load suppression" }).click();
+
+    // Updated map is rendered
+    await expect(page.locator("#review-inputs")).toHaveScreenshot("input-map-vls.png");
+
+    // When I change data source
```
M-Kusumgar, line 97:
```suggestion
    // When I open data source menu
```

src/app/static/src/tests/e2e/review-input.spec.ts
```diff
@@ -10,12 +11,137 @@ test('can view review input plots', async ({ projectPage }) => {
...
 
     // When I click continue
     await continueButton.click();
 
     // Review inputs page matches screenshot after data fetched
-    await expect(page.locator("#review-loading")).not.toBeVisible();
-    await expect(page.locator("#review-inputs")).toHaveScreenshot("review-input-landing.png");
+    await expect(page.locator("#review-loading")).toHaveCount(0, {timeout: 10000});
+    await expect(page.locator("#review-inputs")).toHaveScreenshot("time-series-landing.png");
+    const waitForPlotUpdate = await buildWaitForUpdateCallback(page.locator(".plot-container.plotly"), 100, 20);
```
M-Kusumgar, line 22: can probably also do below but i think your method will be faster since youre just checking for the innerHTML, i would still vote for the way youve done it but just in case you wanted a more playwright-y alternative
```ts
await expect.poll(() => locator.screenshot()).toMatchScreenshot([SS])
```

M-Kusumgar (commented):
Nice this basically lays out all the prereq for the output plots + calibration, looks realy nice. Do you know if the choropleth plot had that weird pixel diff btw? would be good to know if its just plotly and we can have a 0 tolerance for everything else or if its just the browser and we have to bake in tolerance for all the plots.

One suggestion:
* both the input plots have a section that uploads test files and continues to the review inputs, i was thinking it would be nice if there was just a general function that will do all the previous steps and just drop you to whatever tab you want, so if there was `goToStep` function or something and it had the logic that we use to skip past the form in it

src/app/static/src/tests/e2e/review-input.spec.ts
r-ash: Nice! So I couldn't get this working but I did get it to work using expect.toPass https://playwright.dev/docs/test-assertions#expecttopass which I think addresses what we need. That looks good to me, and means we don't have to do this plot update every time we change something. Can just wrap the offending slow updating requests in an expect.toPass.

r-ash:
> Nice this basically lays out all the prereq for the output plots + calibration, looks realy nice. Do you know if the choropleth plot had that weird pixel diff btw? would be good to know if its just plotly and we can have a 0 tolerance for everything else or if its just the browser and we have to bake in tolerance for all the plots.

So just had a quick look at this and looks like it does see https://github.com/mrc-ide/hint/actions/runs/10698342428/job/29657515568 for output. Again small text difference, I've also seen failures locally because the caret was the wrong way in the drop downs. So looks like definitely possible we'll have small differences. 

Saw a couple of suggestions about running in docker or using a full page screenshot clipped to just the locator of interest instead of using a locator screenshot. Might be worth a try if we see real failures getting missed in the future.

> One suggestion:
> 
>     * both the input plots have a section that uploads test files and continues to the review inputs, i was thinking it would be nice if there was just a general function that will do all the previous steps and just drop you to whatever tab you want, so if there was `goToStep` function or something and it had the logic that we use to skip past the form in it

Yeah looks good, I've added a first impl of that for the time being. I do also wonder about bootstrapping a completed project. Would be fast then to just clone it, open it and go to step of interest. Definitely still some merit in a test which steps through but could do this down the line if things get slow.

PR #989, opened 2024-09-04
Fix issue with review input metadata setting latter steps invalid

r-ash:
## Description

This PR fixes an issue I spotted with stepper and state you can recreate before this PR by
1. Starting a new project, progress to completed model options stage (or later)
2. Go back to review inputs
3. See later steps set to incomplete

This was a little fiddly, this is because when we call the action `getReviewInputMetadata` we were setting area level on each of the areas. This then commits a mutation e.g. SurveyAndProgramMutation.ANCUpdated, SurveyAndProgramMutation.ProgramUpdated or SurveyAndProgramMutation.SurveyUpdated.

We also have a subscriber on the store (in root.ts) which is listening on all mutations. When it sees one of the above mutations it was triggering a bunch of resets, to enable the behaviour of invalidating later steps if input data changes. So we were accidentally setting later steps as incomplete, namely via the ModelOptionMutation.UnValidate.

So bit of a fiddly timeline one. I have instead moved setting the area level in the data to immediately after the data fetch. This should then mean it is only done once. It works time wise around when areas are available in memory as that logic already exists on the file upload page.

## Type of version change
None

## Checklist


src/app/static/src/app/store/surveyAndProgram/getters.ts
```diff
@@ -33,83 +8,4 @@ export const getters = {
...
-    data: (state: SurveyAndProgramState) => {
-        const res = response(state);
-        return res ? res.data : null;
-    },
-
-    countryAreaFilterOption: (state: SurveyAndProgramState, getters: any, rootState: RootState): FilterOption => {
-        return rootState.baseline.shape!.filters!.regions as FilterOption;
-    },
-
-    filters: (state: SurveyAndProgramState, getters: any, rootState: RootState): DisplayFilter[] => {
```
r-ash, line 46: Found this and removed it as I realise we weren't using it anymore. I think we can also remove `selectedDataType` stuff everywhere, which will simplify a fair bit of the logic. I think this was used to open the same map in the input plots again to what they were currently viewing. But we have removed that and I don't really think it is worth the complexity. I will remove that in another PR to avoid noise here.

M-Kusumgar (approved):
Urgh that mustve been an annoying one, but yh honestly makes sense to add area levels as soon as we have all the data

PR #993, opened 2024-09-09
Add end to end test for calibration plot

r-ash:
## Description

This PR adds end to end test for calibration plot. This also regens types from https://github.com/mrc-ide/hintr/pull/523. We should merge that one before this

## Type of version change

None

## Checklist


M-Kusumgar (approved):
LGTM!

PR #996, opened 2024-09-10
Stop project version state being updated too frequently

r-ash:
## Description

This PR fixes two issues
1. We were seeing version update occur much more frequently than on production, this is because the `dispatch` was being called with `timeout` arg set instead of the options :see_no_evil: have fixed this now so it sets the options. Can see this working nicely.
2. Also configure some setup to avoid a warning about "__VUE_PROD_HYDRATION_MISMATCH_DETAILS__" not being explicitly defined in the JS console

## Type of version change

None

## Checklist

### DIFF
--- a/src/app/static/vue.config.js
+++ b/src/app/static/vue.config.js
@@ -1,5 +1,7 @@
 const path = require("path");
 const { defineConfig } = require("@vue/cli-service");
+const webpack = require('webpack');
+
 
 module.exports = defineConfig({
     transpileDependencies: true,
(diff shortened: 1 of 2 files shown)


src/app/static/vue.config.js
```diff
@@ -35,5 +37,12 @@ module.exports = defineConfig({
             hints: false
         },
         devtool: 'eval-source-map',
+        plugins: [
+            new webpack.DefinePlugin({
+                // Vue CLI is in maintenance mode, and probably won't this PR to fix this in their tooling
```
M-Kusumgar, line 42: not sure i fully understand this comment, do you mean you want that PR in?

M-Kusumgar (approved):
Nice catch, must've been annoying to find, not sure i understand the comment fully but happy to approve!

src/app/static/vue.config.js
r-ash: o lol no, I will fix the English!

PR #997, opened 2024-09-12
make state backwards compatible

M-Kusumgar:
## Description

I have removed the check on review metadata being fetched for the step completion because this was harder to do than expected. We need the `iso` and whether anc and art files were uploaded which relies on us waiting for the initial request to finish. This does not happen in the action we were targeting so I took the simplest route for now and deleted the `modelOuput` state where we had the plotname hardcoded.


src/app/static/src/app/store/stepper/getters.ts
```diff
@@ -20,7 +20,7 @@ export const getters: StepperGetters & GetterTree<StepperState, RootState> = {
     complete: (state: StepperState, getters: any, rootState: RootState, rootGetters: any) => {
         return {
             1: rootGetters['baseline/complete'] && rootGetters['surveyAndProgram/complete'],
-            2: rootGetters['baseline/complete'] && rootGetters['metadata/complete'] && rootGetters['surveyAndProgram/complete'],
```
r-ash, line 23: So this change will mean they can advance straight to step 3 from step 1 right? No need to go to 2nd step and wait for data to be fetched?

r-ash (approved):
Thanks for this, this looks great. Couple of small comments but feel free to take or leave

src/app/static/src/app/store/stepper/getters.ts
M-Kusumgar: yep, but that shouldnt break anything right, was just the behaviour before

PR #1007, opened 2024-09-16
Update URLs after moving repo to hivtools org

r-ash:
## Description

Update URLs to refer to hivtools org after moving repos. Note naomi not moved yet so these are still pointing to mrc-ide.

Still using the dockerhub version of hint user CLI. We need to get a version of that built on main before we can swap this out

## Type of version change

None

## Checklist


M-Kusumgar (approved):
LGTM!

PR #1008, opened 2024-09-16
Fix table

M-Kusumgar:
## Description

Basically just a table lifecycle bug, it is handled slightly differently in the new version of the package i guess so hooking into their row data updated event to resize the columns


r-ash (approved):
This looks great, thank you!

PR #1010, opened 2024-09-19
Modify a few scripts to work in Windows environment

krisdoyon:
## Description

I made the following changes to get the client running locally in a Windows environment:

- Prevent automatic conversion of Unix paths to Windows paths (was causing Docker issues). Scripts work when run in Git Bash environment on Windows.
- Replace 'npm' with 'npm.cmd' in Gradle build config file to allow Windows to locate local npm installation.
- Add 'env' command in front of environment variable declaration for npm serve script in package.json.

### DIFF
--- a/src/app/build.gradle
+++ b/src/app/build.gradle
@@ -1,3 +1,5 @@
+import org.gradle.internal.os.OperatingSystem
+
 buildscript {
     ext.kotlin_version = '1.6.10'
 }
@@ -21,6 +23,9 @@ tasks.jacocoTestReport {
 group = 'org.imperial.mrc'
 sourceCompatibility = JavaVersion.VERSION_11
 def tomcatVersion = '9.0.65'
+
+// Determine npm command based on operating system. Windows needs 'npm.cmd'.
+def npmCommand = OperatingSystem.current().isWindows() ? 'npm.cmd' : 'npm'
 
 dependencies {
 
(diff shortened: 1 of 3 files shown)


src/app/build.gradle
```diff
@@ -90,15 +90,15 @@ task compileFrontEnd() {
     }
     doLast {
         exec {
-            commandLine "npm", "run", "build", "--prefix=$projectDir/static/"
+            commandLine 'npm.cmd', "run", "build", "--prefix=$projectDir/static/"
```
r-ash, line 93: So this doesn't work on unix. Can we do something like this? (I'm not quite sure the exact syntax we'll need here)

```
String npm = 'npm';
if (Os.isFamily(Os.FAMILY_WINDOWS)) {
    npm = 'npm.cmd'
}
doLast {
    exec {
        commandLine npm, "run", "build", "--prefix=$projectDir/static/"
    }
}
```

r-ash (changes requested):
Thanks for this! This all looks good to me, just one note about the gradle changes so it works on unix and windows

src/app/build.gradle
krisdoyon: Added a check to determine npm command using the isWindows method on gradle's OperatingSystem class. I was able to run it locally and tests are now passing so that should cover both cases.

r-ash (approved):
Looks great!

PR #1011, opened 2024-09-23
[NM-43] Refactor current dataset select approach to make logic more reusable

r-ash:
## Description

Sorry fair bit of churn in this one.

We want to add a feature to allow users to load an output zip from the ADR. For some background, when users fit the model they typically import their input data from the [ADR](adr.unaids.org). After the model has fit one of the downloads is an "output zip" and they can choose to download this locally or push it to the ADR. On the projects page we have a "New project" button, under this users can select a local output zip to load the project from. We want to add a new feature to allow users to import this output zip straight from the ADR.

I have already created a PR where we can list ADR datasets which have an output zip in the resources.

We already have an ADR dataset import component. So this PR lays the groundwork to enable us to re-use that component for this new import. There should be no functionality change.

I have:
* Refactored the adr state. We previously had `datasets`, `releases`, `fetchingDatasets` and `adrError`. These are now part of a named object so there can be multiple. Expecting we have `input` which are the current datasets we fetch and `output` for the new type
* Updated actions and mutations to take the `datasetType`
* Split the Modal out of the component into a new one `SelectDatasetModal` which we can re-use. This will be initialised with a prop for the `datasetType` so will manage the switching.
* Split ADR login stuff into a composable `useADR`

## Type of version change

Patches

## Checklist

### DIFF
--- a/src/app/static/src/app/components/adr/ADRInputDataset.vue
+++ b/src/app/static/src/app/components/adr/ADRInputDataset.vue
@@ -30,59 +30,13 @@
                 @click="toggleModal">
             {{ selectText }}
         </button>
-        <modal id="dataset" :open="open">
-            <h4 v-if="!loading" v-translate="'browseADR'"></h4>
-            <p v-if="loading" v-translate="'importingFiles'"></p>
-            <div v-if="!loading">
-                <label class="font-weight-bold" v-translate="'datasets'"></label>
-                <hint-tree-select id="datasetSelector"
-                             :multiple="false"
-                             :searchable="true"
-                             :options="datasetOptions"
-                             :placeholder="select"
-                             :disabled="fetchingDatasets"
-                             v-model="newDatasetId">
-                    <template v-slot:option-label="{node}">
-                        <label v-html="node.raw.customLabel">
-                        </label>
-                    </template>
-                </hint-tree-select>
-                <select-release :dataset-id="newDatasetId" :open="open"
-                                @selected-dataset-release="updateDatasetRelease"
-                                @valid="updateValid">
-                </select-release>
-                <div :class="fetchingDatasets ? 'visible' : 'invisible'"
-                     style="margin-top: 15px"
-                     id="fetching-datasets">
-                    <loading-spinner size="xs"></loading-spinner>
-                    <span v-translate="'loadingDatasets'"></span>
-                </div>
-                <div v-if="adrError" id="fetch-error">
-                    <div v-if="adrError.detail">{{adrError.detail}}</div>
-                    <button @click="getDatasets"
-                            class="btn btn-red float-right"
-                            v-translate="'tryAgain'">
-                    </button>
-                </div>
-            </div>
-            <div class="text-center" v-if="loading" id="loading-dataset">
-                <loading-spinner size="sm"></loading-spinner>
-            </div>
-            <template v-slot:footer v-if="!loading">
-                <button id="importBtn"
-                        type="button"
-                        :disabled="disableImport"
-                        class="btn btn-red"
-                        v-translate="'import'"
-                        @click.prevent="confirmImport">
-                </button>
-                <button type="button"
-                        class="btn btn-white"
-                        v-translate="'cancel'"
-                        @click="toggleModal">
-                </button>
-            </template>
-        </modal>
+        <select-dataset-modal
+            :open="open"
+            :loading="loading"
+            :dataset-type="AdrDatasetType.Input"
+            @confirm-import="confirmImport"
+            @close-modal="toggleModal">
+        </select-dataset-modal>
         <reset-confirmation
             v-if="showConfirmation"
             :discard-step-warning="selectedDatasetIsRefreshed ? modelOptions : null"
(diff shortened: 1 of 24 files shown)


src/app/static/src/app/components/adr/SelectDatasetModal.vue
```diff
@@ -0,0 +1,130 @@
...
+import {useStore} from "vuex";
+import {RootState} from "../../root";
+import i18next from "i18next";
+import {computed, PropType, ref, watch, watchEffect} from "vue";
+import {AdrDatasetType} from "../../store/adr/adr";
+import {Release} from "../../types";
+
+// eslint-disable-next-line no-undef
+defineEmits<{
+    (e: "confirmImport", datasetId: string, release: Release): void
```
krisdoyon, line 73: I believe there may be a type mismatch here. It looks like the confirmImport method on the ADRInputDataset component is expecting a releaseId as string for the second param, where the custom event definition defines the second param as type "Release".

src/app/static/src/app/components/adr/ADRInputDataset.vue
```diff
@@ -379,17 +286,10 @@
...
-                }
-            },
-            preSelectDataset() {
-                const selectedDatasetId = this.selectedDataset?.id
-                if (selectedDatasetId && this.datasets.some((dataset: any) => dataset.id === selectedDatasetId)) {
-                    this.newDatasetId = selectedDatasetId;
-                }
             },
-            confirmImport() {
+            confirmImport(datasetId: string, releaseId: string | null) {
```
krisdoyon, line 290: confirmImport method on ADRInputDataset component is expecting releaseId as a string as second param

r-ash: Oh thanks, good spot. Fixed that now

krisdoyon (approved):

src/app/static/src/app/components/adr/useADR.ts
M-Kusumgar, line 1: we are probably planning to use the modal elsewhere and its definitely good to extract out stuff but are we actually planning on using this composable anywhere else or will that also be used again in the output zip adr upload bit

src/app/static/src/app/store/adr/actions.ts
```diff
@@ -61,41 +68,54 @@ export const actions: ActionTree<ADRState, RootState> & ADRActions = {
...
+    async getDatasets(context, datasetType: AdrDatasetType) {
+        context.commit({type: ADRMutation.SetFetchingDatasets, payload: {datasetType, data: true}});
+        context.commit({type: ADRMutation.SetADRError, payload: {datasetType, data: null}});
         await api<ADRMutation, ADRMutation>(context)
-            .withError(ADRMutation.SetADRError)
-            .withSuccess(ADRMutation.SetDatasets)
+            .withErrorCallback((failure: Response) => {
+                const error = APIService.getFirstErrorFromFailure(failure);
+                context.commit({type: ADRMutation.SetADRError, payload: {datasetType, data: error}});
+            })
```
M-Kusumgar, line 78: i dont think this works, i tried importing the only release for ethiopia but from my console logs it was using the original `this._onError` function

M-Kusumgar (commented):
looks mostly good, left a couple of comments, i think theres some problem with the error handling that we probably want to sort out, general comments otherwise

src/app/static/src/app/components/adr/useADR.ts
```diff
@@ -0,0 +1,34 @@
...
+    const ssoLogin = computed<boolean>(() => {
+        return store.state.adr.ssoLogin
+    });
+    const key = computed<string | null>(() => {
+        return store.state.adr.key
+    });
+
+    onBeforeMount(() => {
+        if (!isGuest.value && !ssoLogin.value) {
+            store.dispatch("adr/fetchKey").then();
```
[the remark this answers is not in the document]
r-ash: This was just to appease intellij warning "Promise returned from dispatch is ignored" and doesn't actually do anything. But looks like I can achieve the same thing, more explicitly by putting a void at the front. https://stackoverflow.com/questions/41278900/intellij-idea-warning-promise-returned-is-ignored-with-aysnc-await

That sound right?

src/app/static/src/app/components/adr/SelectDatasetModal.vue
```diff
@@ -0,0 +1,129 @@
...
+import LoadingSpinner from "../LoadingSpinner.vue";
+import SelectRelease from "./SelectRelease.vue";
+import Modal from "../Modal.vue";
+import {useStore} from "vuex";
+import {RootState} from "../../root";
+import i18next from "i18next";
+import {computed, PropType, ref, watch, watchEffect} from "vue";
+import {AdrDatasetType} from "../../store/adr/adr";
+
+// eslint-disable-next-line no-undef
```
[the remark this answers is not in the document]
r-ash: Oh nice one! Thanks

src/app/static/src/app/components/adr/useADR.ts
r-ash: Yep, so I will use this on the projects page. To get if they are sso login, or not (I won't show the rehydrate from ADR option if not SSO) and then to get the key.

src/app/static/src/app/components/adr/useADR.ts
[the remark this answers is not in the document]
M-Kusumgar: hmmm i guess that turning that arrow function into async and putting an await in front of the `store.dispatch` might please intellij, cos this is very normal, it might be freaking out cos the function is synchronous

src/app/static/src/app/store/adr/actions.ts
M-Kusumgar: rip :laughing:

src/app/static/src/app/components/adr/useADR.ts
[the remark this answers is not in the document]
r-ash: Ah yep that works too!

PR #1014, opened 2024-09-26
Addresses remaining comments from [NM-43]

r-ash:
## Description

This PR just fixes up issues mentioned in nm-43. Mainly around fixing issues in error handling. See https://github.com/hivtools/hint/pull/1011#pullrequestreview-2328100097

## Type of version change

None

## Checklist

### DIFF
--- a/src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
+++ b/src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
@@ -94,11 +94,19 @@ fun Response.asResponseEntity(logger: Log = LogFactory.getLog(HintExceptionHandl
     {
         if (this.body().asString(null).contains("504 Gateway Time-out"))
         {
-            //Special case of ADR Gateway Timeouts returning HTML responses which cannot be parsed as JSON
+            // Special case of ADR Gateway Timeouts returning HTML responses which cannot be parsed as JSON
             val message = "ADR request timed out"
             logger.error(message)
             ErrorDetail(HttpStatus.GATEWAY_TIMEOUT, message)
                     .toResponseEntity<String>()
+        }
+        else if (this.body().asString(null).contains("502 Bad Gateway"))
+        {
+            // Special case of ADR being flaky and returning HTML responses which cannot be parsed as JSON
+            val message = "Failed to get response from ADR, try again later."
+            logger.error(message)
+            ErrorDetail(HttpStatus.BAD_GATEWAY, message)
+                .toResponseEntity<String>()
         }
         else
         {
@@ -175,4 +183,3 @@ fun getUri(url: String): URI
         .build()
         .toUri()
 }
-
--- a/src/app/static/src/app/components/adr/ADRInputDataset.vue
+++ b/src/app/static/src/app/components/adr/ADRInputDataset.vue
@@ -69,6 +69,7 @@ interface Data {
         newDatasetReleaseId: string | null;
         pollingId: number | null;
         modelOptions: number;
+        AdrDatasetType: any;
     }
 
     const names: { [k in keyof DatasetResourceSet]: string } = {
@@ -91,7 +92,8 @@ interface Data {
                 newDatasetId: null,
                 pollingId: null,
                 newDatasetReleaseId: null,
-                modelOptions: Step.ModelOptions
+                modelOptions: Step.ModelOptions,
+                AdrDatasetType: AdrDatasetType,
             };
         },
         components: {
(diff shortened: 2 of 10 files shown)


src/app/src/main/kotlin/org/imperial/mrc/hint/Extensions.kt
```diff
@@ -94,12 +94,20 @@ fun Response.asResponseEntity(logger: Log = LogFactory.getLog(HintExceptionHandl
...
             ErrorDetail(HttpStatus.GATEWAY_TIMEOUT, message)
                     .toResponseEntity<String>()
         }
+        else if (this.body().asString(null).contains("502 Bad Gateway"))
+        {
+            // Special case of ADR being flaky and returning HTML responses which cannot be parsed as JSON
+            val message = "Failed to get response from ADR, try again later."
+            logger.error(message)
+            ErrorDetail(HttpStatus.BAD_GATEWAY, message)
+                .toResponseEntity<String>()
```
r-ash, line 109: I'm now getting these from ADR dev which was showing "Could not parse API response" errors in the front end :( adding some better specific handling.

src/app/static/src/app/components/adr/ADRInputDataset.vue
```diff
@@ -198,7 +197,11 @@ interface Data {
                     await this.deleteBaselineFiles()
                 }
 
-                await this.getDataset({id: this.newDatasetId!, release: this.newDatasetReleaseId});
+                await this.getDataset({
+                    id: this.newDatasetId!,
+                    release: this.newDatasetReleaseId,
+                    datasetType: AdrDatasetType.Input
```
r-ash, line 203: Oops! Missed the datasetType here in the last PR :see_no_evil:

src/app/static/src/app/components/adr/ADRInputDataset.vue
```diff
@@ -69,6 +69,7 @@ interface Data {
         newDatasetReleaseId: string | null;
         pollingId: number | null;
         modelOptions: number;
+        AdrDatasetType: any;
```
M-Kusumgar, line 72: i believe you want this
```suggestion
        AdrDatasetType: typeof AdrDatasetType;
```

src/app/static/src/app/components/adr/ADRInputDataset.vue
M-Kusumgar: ahh is that why everything was breaking?

src/app/static/src/app/components/adr/SelectDatasetModal.vue
```diff
@@ -1,45 +1,49 @@
...
+                    </button>
+                </div>
             </div>
-        </div>
-        <div class="text-center" v-if="loading" id="loading-dataset">
-            <loading-spinner size="sm"></loading-spinner>
-        </div>
-        <template v-slot:footer v-if="!loading && datasetId">
+        </template>
+        <template v-slot:footer v-if="!loading">
```
M-Kusumgar, line 46: wait, couldnt this be up with the `v-else` because that above template is already not loading? also do we not need to check if `datasetId` exists anymore?

M-Kusumgar (commented):
nice, thanks for addressing those! got a couple of comments here as well as some of the things we said in our chat

src/app/static/src/app/components/adr/SelectDatasetModal.vue
r-ash: So this did not work for me inside the v-else. Got an error "ESLint: 'v-slot' directive must be owned by a custom element, but 'template' is not.(vue/ valid-v-slot)". So I think we need this to be nested directly in the modal, not inside another template.

We don't and actually was getting test failures due to this. Actually a bit confused how this was passing before, I possibly some regression from other changes in this PR but not sure exactly what. With the `datasetId` I could only close the dialog by picking a dataset. Pick a dataset, datasetId gets a value and now the footer with the buttons shows up. Which seemed a bit weird to me. We should be able to close the modal without selecting a dataset.

M-Kusumgar: so template with v-slot attribute is one was to define slots the other way is using the `<slot>` tag, and this way it doesnt complain about it:
```ts
            <slot name="footer">
                <div style="float: right;">
                    <button id="importBtn"
                            type="button"
                            :disabled="disableImport"
                            class="btn btn-red"
                            v-translate="'import'"
                            @click="$emit('confirmImport', datasetId!!, datasetReleaseId)">
                    </button>
                    <button type="button"
                            class="btn btn-white"
                            v-translate="'cancel'"
                            @click="$emit('closeModal')">
                    </button>
                </div>
            </slot>
```
you can put that at the end of that `template v-else`

i see the button is disabled if `datasetId` is falsy so fair enough, i think thats a sensible decision

M-Kusumgar: also sorry can replace `float: right;` inline style with class `"float-right"`

PR #1015, opened 2024-09-26
Fix types

M-Kusumgar:
## Description

Vue components have some type errors in vscode. This PR fixes them, have left comments in the code explaining each fix.

### DIFF
--- a/src/app/static/src/app/components/Warning.vue
+++ b/src/app/static/src/app/components/Warning.vue
@@ -28,6 +28,7 @@
     import { Language } from "../store/translations/locales";
     import { Warning } from "../generated";
     import { PropType, defineComponent, nextTick } from "vue";
+import { StyleValue } from "vue";
 
     export default defineComponent({
         name: "Warning",
@@ -79,7 +80,7 @@
             // If the height of the warnings rendered are more than this number, each warning 
             // will be truncated to one line and the show more button will appear. If there is 
             // only one warning message, it will be truncated after the max number of lines
-            lineStyling(){
+            lineStyling(): StyleValue {
                 if (this.warningsLengthy && !this.showFullBox) {
                     if (this.warnings.length === 1) {
                         return {
--- a/src/app/static/src/app/components/Warning.vue
+++ b/src/app/static/src/app/components/Warning.vue
@@ -27,8 +27,7 @@
     import { RootState } from "../root";
     import { Language } from "../store/translations/locales";
     import { Warning } from "../generated";
-    import { PropType, defineComponent, nextTick } from "vue";
-import { StyleValue } from "vue";
+    import { PropType, defineComponent, nextTick, StyleValue } from "vue";
 
     export default defineComponent({
         name: "Warning",
--- a/src/app/static/src/app/components/adr/SelectDatasetModal.vue
+++ b/src/app/static/src/app/components/adr/SelectDatasetModal.vue
@@ -49,7 +49,7 @@
                     :disabled="disableImport"
                     class="btn btn-red"
                     v-translate="'import'"
-                    @click="$emit('confirmImport', datasetId, datasetReleaseId)">
+                    @click="$emit('confirmImport', datasetId!!, datasetReleaseId)">
             </button>
             <button type="button"
                     class="btn btn-white"
--- a/src/app/static/src/app/components/plots/bar/Barchart.vue
+++ b/src/app/static/src/app/components/plots/bar/Barchart.vue
@@ -11,7 +11,7 @@
 
 <script lang="ts">
 import {Bar} from 'vue-chartjs';
-import {BarElement, CategoryScale, Chart as ChartJS, ChartType, Legend, LinearScale, Title, Tooltip} from 'chart.js';
+import {BarElement, CategoryScale, Chart as ChartJS, Legend, LinearScale, Title, Tooltip} from 'chart.js';
 import annotationPlugin from "chartjs-plugin-annotation";
 import {useStore} from "vuex";
 import {RootState} from "../../../root";
--- a/src/app/static/src/app/components/plots/bar/Barchart.vue
+++ b/src/app/static/src/app/components/plots/bar/Barchart.vue
@@ -11,7 +11,7 @@
 
 <script lang="ts">
 import {Bar} from 'vue-chartjs';
-import {BarElement, CategoryScale, Chart as ChartJS, Legend, LinearScale, Title, Tooltip} from 'chart.js';
+import {BarElement, CategoryScale, Chart as ChartJS, ChartType, Legend, LinearScale, Title, Tooltip} from 'chart.js';
 import annotationPlugin from "chartjs-plugin-annotation";
 import {useStore} from "vuex";
 import {RootState} from "../../../root";
@@ -50,7 +50,7 @@ export default defineComponent({
         });
 
         const chartDataGetter = store.getters["plotSelections/barchartData"];
-        const chartData = ref<BarChartData>({datasets:[], labels: [], maxValuePlusError: 0});
+        const chartData = ref<BarChartData<number[]>>({datasets:[], labels: [], maxValuePlusError: 0});
         const chartOptions = ref({});
         const displayErrorBars = ref<boolean>(false);
 
--- a/src/app/static/src/app/components/plots/bar/Barchart.vue
+++ b/src/app/static/src/app/components/plots/bar/Barchart.vue
@@ -50,7 +50,7 @@ export default defineComponent({
         });
 
         const chartDataGetter = store.getters["plotSelections/barchartData"];
-        const chartData = ref<BarChartData<number[]>>({datasets:[], labels: [], maxValuePlusError: 0});
+        const chartData = ref<BarChartData<(number | null)[]>>({datasets:[], labels: [], maxValuePlusError: 0});
         const chartOptions = ref({});
         const displayErrorBars = ref<boolean>(false);
 
--- a/src/app/static/src/app/components/plots/bar/utils.ts
+++ b/src/app/static/src/app/components/plots/bar/utils.ts
@@ -1,5 +1,5 @@
 import {ChartData, ChartDataset, DefaultDataPoint} from "chart.js";
-import {BarchartMetadata, IndicatorMetadata, FilterOption} from "../../../generated";
+import {IndicatorMetadata, FilterOption} from "../../../generated";
 import {AnnotationOptions} from "chartjs-plugin-annotation";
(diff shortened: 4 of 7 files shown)


src/app/static/src/app/components/Warning.vue
```diff
@@ -79,7 +79,7 @@
             // If the height of the warnings rendered are more than this number, each warning 
             // will be truncated to one line and the show more button will appear. If there is 
             // only one warning message, it will be truncated after the max number of lines
-            lineStyling(){
+            lineStyling(): StyleValue {
```
M-Kusumgar, line 82: Before the inferred type of this was probably just any object so we explicitly say it returns something that can go into the style tag. This also gives us type safety if we change this in the future as we wouldn't be able to add keys that weren't css props

src/app/static/src/app/components/adr/SelectDatasetModal.vue
```diff
@@ -49,7 +49,7 @@
                     :disabled="disableImport"
                     class="btn btn-red"
                     v-translate="'import'"
-                    @click="$emit('confirmImport', datasetId, datasetReleaseId)">
+                    @click="$emit('confirmImport', datasetId!!, datasetReleaseId)">
```
M-Kusumgar, line 52: this button is disabled (`disableImport`) if `datasetId` doesn't exist anyway so asserting that `datasetId` is truthy

src/app/static/src/app/components/plots/bar/Barchart.vue
```diff
@@ -50,7 +50,7 @@ export default defineComponent({
         });
 
         const chartDataGetter = store.getters["plotSelections/barchartData"];
-        const chartData = ref<BarChartData>({datasets:[], labels: [], maxValuePlusError: 0});
+        const chartData = ref<BarChartData<number[]>>({datasets:[], labels: [], maxValuePlusError: 0});
```
M-Kusumgar, line 53: I have put in a generic that lets us define the `data` type of the `datasets` key in `ChartData` type that `chart-js` exposes and we always give it a `number[]` from console logging the barchart data

src/app/static/src/app/components/plots/bar/Barchart.vue
```diff
@@ -50,7 +50,7 @@ export default defineComponent({
         });
 
         const chartDataGetter = store.getters["plotSelections/barchartData"];
-        const chartData = ref<BarChartData>({datasets:[], labels: [], maxValuePlusError: 0});
+        const chartData = ref<BarChartData<(number | null)[]>>({datasets:[], labels: [], maxValuePlusError: 0});
```
M-Kusumgar, line 53: this is the type of `data` in the `datasets` array (`datasets` is a key of `ChartData` more on this later)

src/app/static/src/app/components/plots/bar/utils.ts
```diff
@@ -1,14 +1,13 @@
...
 import {PlotData} from "../../../store/plotData/plotData";
 import {Dict} from "../../../types";
 
-export type BarchartIndicatorMetadata = BarchartMetadata["indicators"][0];
-export interface BarChartData extends ChartData {
-    maxValuePlusError: number;
-}
+type BarChartDataset<Data> = ChartDataset<"bar", Data>
+
+type BarChartDefaultData = DefaultDataPoint<"bar">
```
M-Kusumgar, line 10: these are both basically just types from `chart-js` but ive specified the generic of `ChartType` to be `"bar"` for use later

src/app/static/src/app/components/plots/bar/utils.ts
```diff
@@ -17,9 +16,11 @@ export interface ErrorBars {
     }
 }
 
-export type ChartDataSetsWithErrors =  ChartDataset & {
-    errorBars?: ErrorBars
-}
+type ChartDataSetsWithErrors<Data = BarChartDefaultData> = BarChartDataset<Data> & { errorBars?: ErrorBars }
```
M-Kusumgar, line 19: this is basically the same as before but now specifies that the `ChartType` is `bar`

src/app/static/src/app/components/plots/bar/utils.ts
```diff
@@ -17,9 +16,11 @@ export interface ErrorBars {
...
+
+type ChartDataWithErrors<Data = BarChartDefaultData> = Omit<ChartData<"bar", Data, string>, "datasets"> & { datasets: ChartDataSetsWithErrors<Data>[] }
```
M-Kusumgar, line 21: this is cool, this performs surgery, basically the `ChartData` type doesnt allow datasets to have this `errorBars` prop that we carry around in the `chartData` for our ease of use so I `Omit` the `datasets` prop from the normal `ChartData` type and add in our custom datasets type with `errorBars`, `ChartDataSetsWithErrors` type (while also exposing the `Data` generic so that we can specify the type of data)

src/app/static/src/app/components/plots/bar/utils.ts
```diff
@@ -69,7 +70,7 @@ export const plotDataToChartData = function (plotData: PlotData,
     const orderedXAxisLabels = visibleXAxis.map(opt => opt.label);
 
     let maxValuePlusError = 0;
-    const datasets: any[] = [];
+    const datasets: ChartDataSetsWithErrors[] = [];
```
M-Kusumgar, line 73: No more any!

src/app/static/src/app/components/plots/bar/utils.ts
```diff
@@ -17,9 +16,11 @@ export interface ErrorBars {
...
 }
 
-export type ChartDataSetsWithErrors =  ChartDataset & {
-    errorBars?: ErrorBars
-}
+type ChartDataSetsWithErrors<Data = BarChartDefaultData> = BarChartDataset<Data> & { errorBars?: ErrorBars }
+
+type ChartDataWithErrors<Data = BarChartDefaultData> = Omit<ChartData<"bar", Data, string>, "datasets"> & { datasets: ChartDataSetsWithErrors<Data>[] }
+
+export interface BarChartData<Data = BarChartDefaultData> extends ChartDataWithErrors<Data> { maxValuePlusError: number }
```
M-Kusumgar, line 23: this is same as before but now with our `ChartDataWithErrors` type. It also has the `BarChartDefaultData` type which is what `chart-js` uses as the default for the data generic

src/app/static/src/app/components/plots/bar/utils.ts
```diff
@@ -109,14 +110,14 @@ export const plotDataToChartData = function (plotData: PlotData,
         }
         dataset.data[labelIdx] = value;
 
-        dataset.errorBars[xAxisLabel] = {};
+        dataset.errorBars![xAxisLabel] = { minus: undefined, plus: undefined };
```
M-Kusumgar, line 113: this is a necessary evil XD just so the type is happy, i mean we could make them optional keys but i think that would require more changes and assertions because the minus and plus keys are always defined after this

src/app/static/src/app/components/plots/bar/utils.ts
```diff
@@ -133,7 +134,7 @@ export const getErrorLineAnnotations = function(chartData: BarChartData,
     // amount of padding chart js uses by default for each bar
     const barPercentage = 0.8;
 
-    const datasets = chartData.datasets as any[];
+    const { datasets, labels } = chartData;
```
M-Kusumgar, line 137: slicker

src/app/static/src/app/components/plots/bar/utils.ts
```diff
@@ -154,10 +155,10 @@ export const getErrorLineAnnotations = function(chartData: BarChartData,
         if (!dataset) {
             return
         }
-        const errorBarData = dataset.errorBars;
-        const label = dataset.label;
+        const errorBarData = dataset.errorBars!!;
+        const label = dataset.label || "";
```
M-Kusumgar, line 159: `dataset.label` has type `string | undefined` so providing default value for error bar label in case `undefined`

src/app/static/src/app/components/plots/table/TableReshapeData.vue
```diff
@@ -1,5 +1,6 @@
 <template>
-    <table-display :plot="plot"
+    <table-display v-if="tableMetadata"
```
M-Kusumgar, line 2: type of `tableMetadata` could be `undefined` and we dont want to render that anyway so v-if here

src/app/static/src/app/components/plots/timeSeries/downloadTimeSeries/DownloadTimeSeries.vue
```diff
@@ -28,7 +28,7 @@ const unfilteredData = computed(() => {
 
 const chartData = computed(() => store.state.plotData.timeSeries);
 
-const hasData = computed(() => !!unfilteredData.value & !!chartData.value);
+const hasData = computed(() => !!unfilteredData.value && !!chartData.value);
```
M-Kusumgar, line 31: javascript notation XD

src/app/static/src/app/types.ts
```diff
@@ -203,7 +203,7 @@ export interface ReviewInputDataset {
     warnings: Warning[]
 }
 
-export interface StepWarnings {
+export interface StepWarnings extends Dict<Warning[]> {
```
M-Kusumgar, line 206: `Stepper.vue` was complaining about this so added an extends to tell it that this is in fact a `Dict<Warning[]>` even though really typescript shouldve been able to figure this out

src/app/static/src/app/components/plots/bar/utils.ts
```diff
@@ -177,29 +180,30 @@ export const getErrorLineAnnotations = function(chartData: BarChartData,
             */
             const errorBarWidth = (numOfBars / (numOfBars + 10)) * halfBarWidth * 0.3;
 
-            if (errorBarData[xLabel].minus && errorBarData[xLabel].plus) {
+            const { minus, plus } = errorBarData[xLabel];
+            if (minus && plus) {
```
M-Kusumgar, line 184: for some reason typescript could not figure out that we were checking for minus and plus properties in this if statement already, had to define them as variables and then it worked? this one is a mystery to me

r-ash:
Thanks so much for sorting this, so speedy too!

src/app/static/src/app/components/plots/bar/utils.ts
r-ash: Typescript is crazy

r-ash (approved):
Thanks for sorting this, looks good to me

PR #1017, opened 2024-10-18
[NM-27] Add base map to Choropleth

krisdoyon:
## Description

This PR adds a base map to the Choropleth component using a Leaflet tile layer. A "show base map" checkbox now appears on the map, and toggling it will turn the base map on/off. Updated Playwright tests for reviewInputs accordingly.

## Type of version change

Need to discuss versioning changes with Rob before merging.

## Checklist


r-ash (commented):
Code and app looks great. I wonder if we want to add this to the bubble plot too? A question for Rachel (I'll tag her once she joins this org).

We've chatted about tests, I think let's try run these within a docker container. That should hopefully give us a bit more stability between running in windows vs linux. I'm happy if you want to do that in a separate PR or here?

r-ash (approved):
Amazing, thanks for this

PR #1021, opened 2024-10-25
Update to pull from 2025 ADR datasets

r-ash:
## Description

Bit of background: every year the ADR creates a new datasets type for the estimates round. We run a migration to create new year dataset from the old one, meaning in theory we have an archived history of the final complete run from previous years. So every year we run this migration to create new datasets, and then update our ADR integration to pull the new datasets.

This PR will
1. Update ADR input dataset import to pull from 2025 datasets
2. Update the rehydrate to include 2024 & 2025 datasets

Required a bit of a structural change as previously rehydrate was just listing the same year as the input dataset pull. I had to update this a bit to allow it to pull both last years and this years datasets. So I have changed from 2 separate endpoints `/adr/datasets` and `/adr/datasetsWithResource?resourceType="outputZip"` to just `/adr/datasets?type=input` or `/adr/datasets?type=output`. So it hides some of the business logic about what resource type we are pulling for output vs input and the actual years shown in the backend, which feels a little cleaner to me.

I've updated the integration tests too (as dev ADR is looking much more stable), the front end ones should now be passing. The backend ones are still failing, this is because 2025 datasets have been created without a VMMC data type, causing one of the backend tests to fail. This should pass once that is added but I will add a note to check that down the line, I don't think we should block deployment on this.

## Type of version change

Patches

## Checklist


M-Kusumgar (approved):
Looks good to me!

PR #1023, opened 2024-11-12
[NM-59] Add input comparison table

r-ash:
## Description

This PR will add a table for showing the input comparison data. Should be reviewed and merged after #1022 

There are a couple of general features this has added
1. For a table we can now have multiple rows specified in the custom metadata.
2. The table disaggregation works generally now where previously it was specific to sex

Other features
1. Table is shown with groupings for input comparison data
2. Difference column has an up or down arrow to show change

Next steps
1. Add a download for this data, it shouldn't be hard to add this we just need to make sure it is formatted nicely. I have made a ticket for this.

## Type of version change

Minor

## Checklist


src/app/static/src/app/components/plots/table/DifferenceColumnRenderer.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+    <span class="d-inline-flex">
+        <!--Setting a transform to push icon up 1 pixel to align slightly
```
r-ash, line 3: I was fighting with this a bit, not sure what it looks like on your guys end? Just looks like to me the actual character ▲has more white space above than below and so is rendering kind of ugly on firefox. Couldn't really work out a nice way to make it align.

Actually looking at the screenshots now it looks like from chromium perhaps I didn't need to push it up a pixel and that chromium is displaying it nicely.

I tried looking to see if there was an appropriate vue-feather or font-awesome icon I could use instead which might be easier to centre. But vue-feathers was a big rounded triangle so not quite right and all the similar looking ones on font-awesome were premium.

Also considered just drawing svg path in manually but that seemed like madness. 

Curious on your thoughts and how you would solve this?

src/app/static/src/app/components/plots/table/TableDisplay.vue
```diff
@@ -11,22 +11,23 @@
...
+import { PlotName } from "../../../store/plotSelections/plotSelections";
+import { IndicatorMetadata, TableMetadata } from "../../../generated";
+import {getColumnDefs, TableHeaderDef} from "./utils";
 
 const defaultColDef = {
     // Set the default filter type
     filter: 'agNumberColumnFilter',
     // Floating filter adds the dedicated row for filtering at the bottom
-    floatingFilter: true,
+    floatingFilter: false,
```
r-ash, line 30: Chatted with Rachel about this, she thinks turn it off for now to save some vertical height. We can put it back on if people keep missing them

src/app/static/src/app/components/plots/table/TableDisplay.vue
```diff
@@ -38,11 +39,14 @@ const defaultColDef = {
     // Stop the columns from being draggable to rearrange order or remove them
     suppressMovable: true,
 };
+const ROW_HEIGHT = 35;
```
r-ash, line 42: I reduced the row height here too, think it looks a little better

src/app/static/src/app/components/plots/table/TableDisplay.vue
M-Kusumgar: yh think this looks slicker without it

src/app/static/src/app/components/plots/table/TableDisplay.vue
M-Kusumgar: definitely agree

src/app/static/src/app/components/plots/table/TableDisplay.vue
M-Kusumgar: yh think this looks slicker without it

src/app/static/src/app/components/plots/table/DifferenceColumnRenderer.vue
krisdoyon: spacing looks ok to me on both Firefox and Chrome; on my end I am seeing the same icon size in both browsers, the larger size that Mantra showed in the second screenshot

krisdoyon: If consistent sizing is an issue and you do end up deciding to replace with an icon, looks like the "play" icon from vue-feather might be suitable. Could do a 90 degree rotation in either direction to get it oriented correctly.

r-ash: Thanks! Rotated play icon is a great idea. If this looks ok to you both then let's go with it, if we see some issues down the line we can switch to the play button icon. I don't mind the difference in size on linux from chrome to firefox, as long as they are centered :D

krisdoyon (approved):

PR #1025, opened 2024-11-14
Hot fix for state being rewritten on a project

r-ash:
## Description

## Type of version change
_Delete as appropriate_

## Checklist

### DIFF
--- a/src/app/static/src/app/root.ts
+++ b/src/app/static/src/app/root.ts
@@ -81,13 +81,21 @@ export interface WarningsState {
 
 const persistState = (store: Store<RootState>): void => {
     store.subscribe((mutation: MutationPayload, state: RootState) => {
-        // console.log(mutation.type);
-        localStorageManager.saveState(state);
+        //console.log(mutation.type);
+        // We need to be careful with this, if someone has triggered something which
+        // starts a slow action, then they go back to projects page and load a new project
+        // we can get into a state where the project state has been replaced in local storage
+        // with the new project, then the action completes and commits some mutations nuking
+        // loaded project state. If a project load has been initiated, block any further
+        // mutations from happening. This will be reset when the browser refreshes.
+        if (!state.projects.loadingProject) {
+            localStorageManager.saveState(state);
 
-        const {dispatch} = store;
-        const type = stripNamespace(mutation.type);
-        if (type[0] !== "projects" && type[0] !== "errors") {
-            dispatch("projects/queueVersionStateUpload", null, {root: true});
+            const {dispatch} = store;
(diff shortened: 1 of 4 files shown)


src/app/static/src/app/store/surveyAndProgram/actions.ts
```diff
@@ -332,7 +332,7 @@ export const actions: ActionTree<SurveyAndProgramState, RootState> & SurveyAndPr
     async setSurveyResponse(context: ActionContext<SurveyAndProgramState, RootState>, response: SurveyResponse) {
         const {commit, rootState} = context;
         const shapeData = rootState.baseline.shape;
-        if (shapeData) {
+        if (shapeData?.data) {
```
r-ash, line 335: I was also occasionally getting errors from shapeData.data.features being null in the call to getDataWithAreaLevel. So making this check cover that case too.

M-Kusumgar (commented):
this is definitely the correct place to target this bug, i managed to recreate it however theres a bigger issue with all this save state into local storage stuff, local storage is shared between tabs so your goal of making saving state atomic means that you must block saving state at any point theres a change, not just when a project is loading, but also when it decided to save state every 2 seconds for example, so one thing i was able to do after a couple attempts is:
1. create a DRC project, stop at review inputs
2. create a MWI project, go to output plots
3. load DRC and MWI on two separate tabs
4. wait for state saving of MWI (2 sec interval thing)
5. reload DRC (this is a bit flaky) and you can get into a state that DRC is actually on the output plots step because somewhere we are saving a **partial** state with just the stepper state i am guessing

like shown here, note the zomba city error (city in malawi) and DRC area filter
image

if you want to further avoid these issues, i would save full states in local storage only and i would also potentially change the structure of the saved states in localstorage to be an object with keys equal to sessionId and value their local storage, which means that you can actually have things open on multiple tabs without them interacting with each other, 

note: you could even skip saving only full states with this because saving partial states makes sense if you have the correct sessionId but honestly i always prefer to wipe and reconstruct because it costs almost no extra time and is cleaner

but yh since local storage is shared between tabs, its tricky with stuff like this!

r-ash:
Thanks @M-Kusumgar chatted about this

Follow ups which I will make a ticket for
1. Backend should prevent creating a project with the same name
2. We will update local storage key to be specific to the tab which is open. We should add a limit to the number of these which we keep in local storage (20?) and clean up the oldest entry. Remove the metadata from local storage, this is always refetched
3. Down the line, rethink what we save in local storage. We could just save an ID for a session and have the backend manage returning state to us.

M-Kusumgar (approved):

PR #1029, opened 2024-11-18
Ensure that download result polling is stopped when outputs are reset

r-ash:
## Description

We can get into a state where the app will poll the download status without an ID repeatedly. To recreate

1. Go to the “Save resutls” stage on a project, wait for status poll to begin
2. Go back to an earlier step and change something which invalidates the later steps e.g. change a model option
3. See that the app will continue to poll `/download/status` without an ID

When the download step is invalidated, we should cancel any polling

## Type of version change

Patches

## Checklist

### DIFF
--- a/src/app/static/src/app/hintVersion.ts
+++ b/src/app/static/src/app/hintVersion.ts
@@ -1,1 +1,1 @@
-export const currentHintVersion = "3.11.2";
+export const currentHintVersion = "3.11.3";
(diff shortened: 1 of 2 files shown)


M-Kusumgar (approved):
LGTM! happy to approve this for now, i want to test it out but slow internet XD but yh will async run up the app and test

edit: can confirm i dont see the polling

PR #1030, opened 2024-11-18
[NM-73] fix timing by waiting for baseline data

M-Kusumgar:
## Description

This was a weird one to find, basically the way loading state works is we save the details of where the files are in local storage and then we just refresh the page, the `Hint.vue` on before mount fetches everything in parallel. This was causing an issue because later parts of our code assume that `state.baseline.shape` exists for example.

I could replicate this pretty easily and consistently if you throttle chrome network speed to "slow 3g" then it happens basically every time with even malawi, it does not occur anymore with this PR

So theres a handy `state.baseline.ready` boolean that i am watching and triggering the rest of the data fetches and that solves it. Theres equivalent "ready"s for `surveryAndProgram`, `modelRun` and `modelCalibrate` states so we can put similar watches in for those if theres some sort of dependency between these, for now just blocking until `baseline` state is fully populated!

PS: other test failures are occuring on main too, see most recent merge to main, i just re ran those, the E2E test failure is correct, seems like the time 1 in "Include ART data" is not yes, the metadata may have changed in a recent naomi/hintr PR and the backend test is just the wrong error message potentially but yh also occuring on main

### DIFF
--- a/src/app/static/src/app/components/Hint.vue
+++ b/src/app/static/src/app/components/Hint.vue
@@ -1,7 +1,11 @@
 <template>
     <user-header :title="title" :user="user"></user-header>
-    <div class="container mb-5">
+    <div class="container mb-5" v-if="baselineReady">
         <router-view></router-view>
+    </div>
+    <div v-else class="text-center">
+        <loading-spinner size="lg"></loading-spinner>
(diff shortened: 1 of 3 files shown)


r-ash (approved):
This is perfect, thank you!

Test failures are because of default options and test data being updated (working on fixing atm, just slow going!)

PR #1032, opened 2024-11-19
[NM-75] reset filter buttons

M-Kusumgar:
## Description

Have added reset buttons for each filter on hover and reset on all filters always showing. I tried to break it up by having 2 different icons, one for all filters and another icon for the individual filters, it looks weird with the same icon

I have also extract the component

general strategy for this was to calculate the default filter options and transplant the relevant filters, had to add a way to also update multiple filters in one go, because before i guess the assumption was the user would be changing one filter at a time but yh with this reset stuff we require setting multiple filters at once so also added a test for this


r-ash (changes requested):
Functionality wise, this looks really great. Could you bump the version number please?

I wonder if we can put that duplicated code from the plots into some action and re-use it everywhere?

But also couple of things I am not sure about styling wise
1. Can you align the spinner icon with the text? 
image
2. Can we have the cursor show this is a button? Just showing as a normal arrow for me when I hover over it
3. Can you add a tooltip that says what this button does
4. I wonder about styling it to make it more obvious it is a button e.g. in leaflet reset we highlight the button on hover, can we do something like that?
5. Not 100% sure about the icon either, but happy to be overruled. I wonder if we put this on preview and ask Jeff and Rachel to reivew. Or review it on Thursday

M-Kusumgar:
To remove code duplication, ive moved the logic of resetting the filters into the new component, it switches on whether it should reset a single filter or all filters by whether or not youve specified a `stateFilterId` i think its fairly neat now

I have also moved the class from `style.scss` to a scoped style tag, rob lmk if that styling still doesnt show up, we may have to investigate that because it should

i wasnt sold on the different icons either, i have done show on hover for everything now, lmk what you guys think of that!

r-ash:
> To remove code duplication, ive moved the logic of resetting the filters into the new component, it switches on whether it should reset a single filter or all filters by whether or not youve specified a `stateFilterId` i think its fairly neat now
> 
> I have also moved the class from `style.scss` to a scoped style tag, rob lmk if that styling still doesnt show up, we may have to investigate that because it should, i wonder if it was a browser cache issue? if you get that problem try disabling cache, if it still persists lets call
> 
> i wasnt sold on the different icons either, i have done show on hover for everything now, lmk what you guys think of that!

That looks great, working beautifully on my end. Sorry if that was just a browser cache issue before :( 

I'll deploy this on preview and let's get a quick look from Jeff and Rachel but I really like this. I think nice call on having refresh all filters show on hover too. Looks clean!

krisdoyon:
Everything looks and is working nicely for me in the preview deploy that Rob posted. Styling looks good and is consistent for me in Firefox and Chrome. I think the button display on hover is a good compromise; users would need to make a filter selection before the reset button becomes relevant to them, and I think it's obvious enough to notice as a part of the process of changing filters. Looks much cleaner than showing them all by default.

If we do find this setup is too hidden for users, one idea would be to show the "reset all filters" button by default, maybe even greyed out a bit, but still keep the individual ones hidden until hover. IMO the current setup functions nicely though, and my vote would be to roll with it as is if Jeff and Rachel also agree.

r-ash:
Reviewed this with Jeff, he asked if instead of a tooltip we could show the text next to the icon? 

So when you hover over, it says "Reset filter" or "Reset all filters", still with the icon on the RHS

krisdoyon (approved):

r-ash (approved):

PR #1033, opened 2024-11-19
[NM-76] Update test data

r-ash:
## Description

Jeff has updated the test data, this PR goes through and updates that and its uses in tests

## Type of version change

None

## Checklist


M-Kusumgar (approved):
LGTM, had a look at the adr test failures, assuming that is on the ADR end of things

PR #1035, opened 2024-11-20
[NM-52] Add base map to bubble plot

krisdoyon:
## Description

Adds the base map and associated checkbox that are currently in use with choropleth to output bubble plot. Refactored the checkbox into a separate component for consistent use in both maps. Checkbox now uses an emit which is handled in the parent component. Added logic to playwright test for output plots to test functionality similar to choropleth.

## Type of version change

Patch

## Checklist


r-ash (approved):
This looks really great, and code changes look great too. I wonder, can we make the text of "Show base map" also toggle the checkbox? I always go to click the text first but that doesn't work

r-ash (approved):
Thanks, this looks really good!

PR #1036, opened 2024-11-20
[NM-82] download results state refactor

M-Kusumgar:
## Description

This one was painful. Ill remove AGYW in the next PR to see how easy/hard it is to modify this new construction, almost 2000 lines gone, damn.

There was a handy enum in the `types` file called `DownloadType` so i thought way not roll with that, everything is constructed from `downloadConfig.ts`, using `Object.values` on a string enum (does not work well on a usual enum, can try it if you want to know a stupid typescript decision) gives all the values and we can loop over it to do everything

The tests were especially painful, i suspect there'll be small simplifications when we remove AGYW

The other thing is I had to add a payload handler to the api service because i needed finer control over the payload the api need to commit to the state (needed to add `type` so that it knows which `DownloadType` to commit to)

ideally, to add another button you just need to extend the 3 things in `downloadConfig.ts`

Edit: the e2e test failure is a real one! i am removing AGYW in #1037 which im merging into here so will just fix the e2e test there


src/app/static/src/app/components/downloadResults/DownloadResults.vue
```diff
@@ -2,47 +2,17 @@
...
-                <div id="spectrum-download">
-                    <download :translate-key="translation.spectrum"
-                              @trigger-download="downloadSpectrumOutput"
-                              :disabled="!spectrum.downloadId || spectrum.preparing"
-                              :file="spectrum"/>
-                    <div class="pb-2">
-                    <error-alert v-if="spectrum.downloadError" :error="spectrum.downloadError"></error-alert>
+                <template v-for="type in Object.values(DownloadType)" :key="type">
+                    <div :id="`${type}-download`" v-if="switches[type]">
+                        <download :translate-key="`${type}Download`"
```
r-ash, line 7: I wonder if we should put this into the `downloadConfig.ts` like the `switches` a map of type to the key we want to use. I do like all the download stuff being in the single config file.

src/app/static/src/app/store/adrUpload/actions.ts
```diff
@@ -111,23 +112,23 @@ export const actions: ActionTree<ADRUploadState, RootState> & ADRUploadActions =
                 requestParams["resourceId"] = resourceId
             }
             if (resourceType === rootState.adr.schemas?.outputSummary) {
-                downloadId = rootState.downloadResults.summary.downloadId
+                downloadId = rootState.downloadResults[DownloadType.SUMMARY].downloadId
```
r-ash, line 115: What do you also think about handling this through the `downloadConfig`? Have some list of files which need to go to the ADR, and here look through the `DownloadType` and include those which are in the list?

r-ash: Could use it above in `UploadModal`

src/app/static/src/app/store/root/mutations.ts
```diff
@@ -141,6 +136,12 @@ export const mutations: MutationTree<RootState> = {
 
 };
 
+const stopPollingAllDownloadTypes = (state: RootState) => {
```
r-ash, line 139: :heart:

src/app/static/src/app/apiService.ts
```diff
@@ -70,9 +70,10 @@ export class APIService<S extends string, E extends string> implements API<S, E>
         return this;
     };
 
-    withError = (type: E, root = false) => {
+    withError = (type: E, root = false, payloadFunc?: (e: Error) => any) => {
         this._onError = (failure: Response) => {
-            this._commit({type: type, payload: APIService.getFirstErrorFromFailure(failure)}, {root});
+            const error = APIService.getFirstErrorFromFailure(failure);
+            this._commit({ type, payload: payloadFunc ? payloadFunc(error) : error }, { root });
```
r-ash, line 76: So this is what I had previously added `withErrorCallback` for. But I think what you've done is better. Could you remove the `withErrorCallback` helper and tidy up uses of that to use this instead?

But actually, I wonder if we should have 2 functions `withError` and `withErrorHandler`? Or some better name? Whatever you think!

r-ash (commented):
This looks really great, so good to see so much duplication removed.

Couple of comments on a few bits would be nice to get your thoughts on. But I think this looks good to go

src/app/static/src/app/store/downloadResults/actions.ts
```diff
@@ -3,273 +3,83 @@ import {RootState} from "../../root";
 import {DownloadResultsState} from "./downloadResults";
 import {api, ResponseWithType} from "../../apiService";
 import {DownloadResultsMutation} from "./mutations";
-import {ModelStatusResponse} from "../../generated";
-import {DOWNLOAD_TYPE} from "../../types";
+import {Error, ModelStatusResponse} from "../../generated";
```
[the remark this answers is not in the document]
M-Kusumgar: yep removed, neither was switches before it! you can tell i did a couple of drafts before settling on the final version haha

src/app/static/src/app/store/downloadResults/downloadConfig.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+// we do posts to /download/submit/<url>/<calibrate-id> with the body function called
+type PostObject = { url: string, body: (store: ActionContext<DownloadResultsState, RootState>) => object };
+export const downloadTypePrepareUrls: Record<DownloadType, PostObject> = {
+    [DownloadType.SPECTRUM]: { url: "spectrum", body: (store) => store.rootGetters.projectState },
+    [DownloadType.COARSE]: { url: "coarse-output", body: () => ({}) },
+    [DownloadType.SUMMARY]: { url: "summary", body: () => ({}) },
+    [DownloadType.COMPARISON]: { url: "comparison", body: () => ({}) },
+    [DownloadType.AGYW]: { url: "agyw", body: () => ({}) }
```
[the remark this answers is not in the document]
M-Kusumgar: agreed, in fact lets do a bit more generally, lets just make both the body and url functions that take in the store, that would be good!

src/app/static/src/app/store/downloadResults/mutations.ts
```diff
@@ -1,251 +1,96 @@
...
-    SummaryOutputDownloadError = "SummaryOutputDownloadError",
-    PreparingComparisonOutput = "PreparingComparisonOutput",
-    ComparisonOutputStatusUpdated = "ComparisonOutputStatusUpdated",
-    ComparisonDownloadError = "ComparisonDownloadError",
-    ComparisonError = "ComparisonError",
-    PreparingAgywTool = "PreparingAgywTool",
-    AgywError = "AgywError",
-    AgywStatusUpdated = "AgywStatusUpdated",
-    AgywDownloadError = "AgywDownloadError",
+    Preparing = "Preparing",
```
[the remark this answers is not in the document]
M-Kusumgar: yh these were definitely satisfying files!

src/app/static/src/app/components/downloadResults/DownloadResults.vue
M-Kusumgar: hmmm do you mean because theres duplication between this and the `locales.ts` type? because otherwise everything is auto generated, also the types in the locales file ensures we have the matching translation key

r-ash: No, only that when you're adding a new file type you have to look here to see what the key is. It is `{type}Download" instead of just being able to see the full string in the `downloadConfig.ts` ? That make sense?

src/app/static/src/app/store/adrUpload/actions.ts
M-Kusumgar: the reason i didnt do this here is because this requires me to refactor the ADR upload stuff because we need some way of matching up if "outputSummary" exists, then include `DownloadType.SUMMARY`, i say we do a smaller refactor in a separate PR and sort out the adrUpload stuff and there we can put these mappings, infact some of the adrUpload stuff can probably just come from the downloadConfig itself

in the `UploadModal` we have this same mapping issue and also have `outputFileError` which has a weird control flow which does something different if both spectrum and summary have metadata errors vs just one of them, i would like to properly investigate that and see if theres a pattern that we can extract

i can create a ticket for these two if you agree!

src/app/static/src/app/apiService.ts
M-Kusumgar: ohhhh damn sorry i didnt even see that, i was looking for that type of functionality, yh im happy to remove it, i think its nice to declare the mutation you want and all that, i personally think its nicer if we just have one function with an optional arg, it feels like exactly the same responsibility, just with pre commit step or not

src/app/static/src/app/store/adrUpload/actions.ts
r-ash: Yep sounds great doing that separately, yes please create the tickets.

src/app/static/src/app/apiService.ts
r-ash: Nope that's fine, happy with 1 function if you want.

src/app/static/src/app/store/downloadResults/downloadConfig.ts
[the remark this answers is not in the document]
M-Kusumgar: done

M-Kusumgar:
have addressed comments in this commit: https://github.com/hivtools/hint/pull/1037/commits/481b209187f025a78c929ac990875417017ee64d

ADR upload refactor ticket is created!: https://www.notion.so/unaids-naomi/Upload-ADR-state-refactor-f1c71f2f341c49aa98bb9f3bad88e99c?pvs=4

r-ash (approved):
Thanks for that, I just merged in #1034 and then removed `downloadOutput` action. We don't need that any more as we're now downloading via an href instead of an action

PR #1037, opened 2024-11-20
[NM-84] Remove AGYW

M-Kusumgar:
## Description

this was a bigger diff because AGYW has a switch associated with it and also it doesnt have adr metadata support which means i had to put some if statements with it, but the diff would be even smaller when we remove buttons.

to test out how easy adding a button is (given the adr endpoint) you can just add a new value to the enum and follow the typescript errors, for the `url` in the `downloadPostConfig` you can just write one of the existing ones like `"coarse-output"`, it should be about 6 lines of change (3 in `downloadConfig.ts` and 3 in locals if you want all the translations) to add a fully working new button


src/app/static/src/app/featureSwitches.ts
```diff
@@ -1,12 +1,10 @@
-const urlParams = new URLSearchParams(window.location.search);
+const urlParams = new URLSearchParams(globalThis.location?.search);
```
M-Kusumgar, line 1: playwright now loads this code because we import in the `DownloadType` into a playwright file, however it runs in a "node" environment not a browser env which means theres no `window` object. `globalThis` object is just the global this object so in a browser it is `window` and in a "node" env its just the `global` object, which means we can use this with playwright

r-ash: (Can you stick that in a comment above, I'll definitely forget about this)

r-ash (approved):
This looks great, though let's leave open until we merge #1036

PR #1038, opened 2024-11-21
Add a script to gather strings for sending to translator

r-ash:
## Description

This PR adds a script which reads the translation keys from the locales and writes them into a csv we can send to the translator.

## Type of version change

None

## Checklist


krisdoyon:
Ran locally and was able to generate the csv file, but with one oddity: script kept failing due to not being able to import locales with this error:

SyntaxError: The requested module '../src/app/store/translations/locales' does not provide an export named 'locales'

I changed the file extension on collectTranslations from .mts to .ts and ran it that way and it worked as expected.

krisdoyon (approved):

r-ash:
> Ran locally and was able to generate the csv file, but with one oddity: script kept failing due to not being able to import locales with this error:
> 
> SyntaxError: The requested module '../src/app/store/translations/locales' does not provide an export named 'locales'
> 
> I changed the file extension on collectTranslations from .mts to .ts and ran it that way and it worked as expected.

Thanks for this, I've updated that to `.ts` and runs nicely on my machine too.

PR #1039, opened 2024-11-25
Fix issue with link to privacy and accessibility from login page not working

r-ash:
## Description

On the current app if you try to go to privacy or accessibility from the login page you just get redirected back to login page again.

This PR fixes that by switching the history mode, I think this was what was causing the issue. See https://router.vuejs.org/guide/essentials/history-mode.html#HTML5-Mode for details, I think we are safe to switch because of the way we are serving the frontend from freemarker templates. At least comparing this to production, I can't notice any reason not to switch the history mode. Behaviour is the same, and we no longer have the ugly `#` that gets put in the URL.

## Type of version change

Patches

## Checklist

### DIFF
--- a/src/app/static/templates/404.ftl
+++ b/src/app/static/templates/404.ftl
@@ -16,15 +16,15 @@
 </div>
 <hr>
 <div class="d-flex justify-content-center">
-    <a href="/"><img src="/public/images/unaids_logo.png" class="large-logo mt-5"/></a>
+    <a href="/" target="_blank"><img src="/public/images/unaids_logo.png" class="large-logo mt-5"/></a>
 </div>
 
 <div id="partner-logos" class="logos d-flex justify-content-center mt-5">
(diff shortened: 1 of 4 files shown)


src/app/static/templates/404.ftl
```diff
@@ -16,15 +16,15 @@
 </div>
 <hr>
 <div class="d-flex justify-content-center">
-    <a href="/"><img src="public/images/unaids_logo.png" class="large-logo mt-5"/></a>
+    <a href="/"><img src="/public/images/unaids_logo.png" class="large-logo mt-5"/></a>
 </div>
 
 <div id="partner-logos" class="logos d-flex justify-content-center mt-5">
-    <a href="https://www.fjelltopp.org"><img src="public/images/fjelltopp_logo.png" class="small-logo"></a>
```
r-ash, line 23: This was broken on production too, you can check by going to a non-existent root `naomi.unaids.org/not/a/path`

r-ash: Good catch, thanks. Have updated that in this template and the login template.

PR #1040, opened 2024-11-26
[NM-93] Add cascade plot

r-ash:
## Description

We want to add a cascade plot into the app. Cascade plot shows 95-95-95 targets (95% of people living with HIV are aware of the status, 95% of those aware are on treatment and 95% of those on treatment are virally suppressed)

At the moment we don't have viral suppression data from the model, but this shows the first 3 bars of the cascade. In the future we'd also like to add this as a plot of proportions instead of just pure counts.

## Type of version change

Minor

## Checklist


krisdoyon:
Noticed something when playing with the filters running locally and wondering if this is expected behavior or not: If you have area level set to country, then drill down into the area filter and turn off a single district (or any other sub-level), you get the no data available message. Should user be able to show some subset of data at the country level for example, or is this working as intended?

krisdoyon (approved):

PR #1043, opened 2024-12-02
[NM-95] Update naomi population pyramid to not filter aggregated proportions on selected area

r-ash:
## Description

See https://teams.microsoft.com/l/message/19:83f74f91587b4848a35c45bca0c3a2b1@thread.tacv2/1733081977154?tenantId=2b897507-ee8c-4575-830b-4f8267c3d307&groupId=e24ac85e-dd9e-4505-8b9e-047b3a5fa6f0&parentMessageId=1733081977154&teamName=HIV%20Inference%20Group%20-%20WP&channelName=Naomi&createdTime=1733081977154 for some context, the national boundaries are being calculated after filtering. So if they are filtering the area level down to a few districts it won't be the correct boundary.

This fixes that by calculating aggregated national data removing the area filter. We don't want to apply the area filter. I'd like to get this in as a quick fix, but I think there are some opportunities for tidying up this approach.

## Type of version change

Patches

## Checklist

### DIFF
--- a/src/app/static/src/app/components/plots/table/Table.vue
+++ b/src/app/static/src/app/components/plots/table/Table.vue
@@ -18,7 +18,7 @@ import DownloadButton from '../timeSeries/downloadTimeSeries/DownloadButton.vue'
 import { exportService } from '../../../dataExportService';
 import { appendCurrentDateTime } from '../../../utils';
 import { formatOutput, getIndicatorMetadata } from '../utils';
-import { PlotData } from "../../../store/plotData/plotData";
+import { SquarePlotData } from "../../../store/plotData/plotData";
 import { PlotName } from "../../../store/plotSelections/plotSelections";
 
 // defines the order of headers on the excel download
@@ -45,7 +45,7 @@ export default defineComponent({
     },
     setup(props) {
         const store = useStore<RootState>();
-        const plotData = computed<PlotData>(() => store.state.plotData[props.plotName]);
+        const plotData = computed<SquarePlotData>(() => store.state.plotData[props.plotName] as SquarePlotData);
         const filterSelections = computed(() => store.state.plotSelections[props.plotName].filters);
         const indicatorMetadata = computed<IndicatorMetadata>(() => {
             const indicator = filterSelections.value.find(f => f.stateFilterId === "indicator")!.selection[0].id;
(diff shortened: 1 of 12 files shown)


src/app/static/src/app/store/plotData/plotData.ts
```diff
@@ -10,7 +10,13 @@ import {
...
 import { mutations } from "./mutations"
 
-export type PlotData = CalibrateDataResponse["data"] | InputTimeSeriesData | CalibratePlotData | ComparisonPlotData | InputComparisonData | PopulationResponseData;
+export type PopulationPyramidData = {
+    data: PopulationResponseData;
+    nationalLevelData: PopulationResponseData;
+}
+
+export type PlotData = CalibrateDataResponse["data"] | InputTimeSeriesData | CalibratePlotData | ComparisonPlotData | InputComparisonData | PopulationPyramidData;
+export type SquarePlotData = Exclude<PlotData, PopulationPyramidData>;
```
r-ash, line 19: I think this is quite ugly... but not sure if there a better way to do this? Thought about committing the national data to some other place in state but that seemed quite ugly too. Or doing the national filtering later, but that it a big gross... very happy to hear if you have any good ideas, I am a little stumped

src/app/static/src/app/components/plots/table/Table.vue
```diff
@@ -45,7 +45,7 @@ export default defineComponent({
     },
     setup(props) {
         const store = useStore<RootState>();
-        const plotData = computed<PlotData>(() => store.state.plotData[props.plotName]);
+        const plotData = computed<SquarePlotData>(() => store.state.plotData[props.plotName] as SquarePlotData);
```
krisdoyon, line 48: I think we may need to change the type for the data prop in TableReshapeData.vue to match with the new SquarePlotData type. Getting a TS compiler error in this file on line 3.

krisdoyon: Looks good, no compiler errors now. Agreed on the name change as well

src/app/static/src/app/store/plotData/plotData.ts
krisdoyon: I agree it is a bit awkward but I think it makes most sense to do the aggregating of national data and all other data in the same place as we have currently. Only other thing I can think is, as you said, making two state commits in the getPopulationFilteredData function and committing the national level data to a different place in state. The two data sets are so closely tied together though and are really only relevant to the plot so I'm inclined to say just go with the method you have of coupling the two data sets together in an object. We could rename the lowest "data" field to something like "subPlotData" to avoid the data.data repetition but not entirely necessary

krisdoyon (approved):

PR #1046, opened 2024-12-04
[NM-97] Add expanded time series plot view

r-ash:
## Description

This PR adds an expanded time series view. For certain indicators, you can click on a button in the top right of the time series and see an expanded view of the indicator. This is the denominator and numerator and anything which goes into it.

This should hopefully be done in a fairly generic way so we can add new ones in the future.

For now just for ANC prevalence and ANC art coverage.

Also makes some changes to the modal. We had some users and this was rendering quite large with their particular view port so they could not scroll down the modal to click the close button. I also saw this for someone trying to share the project with lots of users, so many that the send/close buttons went off the limits of the page.

So I have changed a few things
* Added a slot in the modal for header, if you use this, you will get a close button in the top right that emits an event
* If you click the grey background, it will emit the same `close-modal` event, have implemented this for all modals in the app
* Also set a height limit for the modal content, and it will scroll if it exceeds this

## Type of version change

Patches

## Checklist


M-Kusumgar (commented):
Nice, code looks good, the modal is much better now, definitely agree it should close when clicked outside it.

I think the colors are hard to distinguish in the modal, especially the formula with 2 terms on the numerator and 3 terms on the denominator, with the darkgreen + green and dark blue + blue pairs.
brainstorming a couple of thoughts:
* playing around with different line styles might help, like maybe denominator is dotted or something like that and then the colors can be more distinct 
* warm color palettes vs cold color palettes between numerator and denominator might also be something to consider
* one interesting thing i thought of was (probably a different PR if we even can do this) is when you hover over a term, all the other lines in the plot fade out, that can also help the user distinguish between the terms in a way that is independent of the colors

PR #1047, opened 2024-12-10
[NM-104] download table ui updated

M-Kusumgar:
## Description

So this PR does a UI update to the download table and lets you configure the icons in the table however you want, it also includes a status icon + text, im not 100% sure about the layout or colors so let me know what you guys think!

You should still be able to download the files and the behaviour should be the same as before for the download buttons


r-ash (changes requested):
This looks great, looking way nicer than what we have at the moment. I particularly like the status indicators. Couple of things I noticed
1. I can't see errors, it shows an error has happened in the status but I cannot see the reason 
image
2. Can we change the naomi download and coarse age group downloads to zip files? 
3. I'm with you that I'm not 100% sold on spacing and colours. But the colour on the status indicators is great, and using it in the file types does distinguish them from each other. Not sure if I have a good solution for that. What about for the spacing, making it look a bit more table like? I wonder if a single line between each download would help connect the Button to the name? And titles on the columns? Maybe too old school or will look funny with the big red button?
4. Reduce the size of the buttons a bit? Or try an icon only one?
5. Can we align the title "Download" on the left hand side with the icons?
6. Would be nice also in the future to show the file size, and potentially also add checkboxes and a "download all selected" button to get them all at the same time
7. Try status icons next to the button? They feel a little floaty

Also note I just merged datapack download, sorry for the clash that will cause.

For fun I asked chatgpt to give me a mockup and this is its terrifying suggestion
image

krisdoyon:
I think the functionality here looks really good. The status icons/text are a big improvement from a UI perspective. Highlighting the row in gray on hover is a nice touch too. As far as the spacing and layout, I think the "Download" header and the individual file names might be a little too close in styling, and the rows might look better overall with a little less bold. This still doesn't look quite right to me but here's a quick mock up of what I mean (reduced file name font size, normal font weight on text/icons, a bit more spacing between rows):

image

One other small addition to consider is a "Download" tooltip on the icon buttons, though the icon is quite recognizable and they're already under a "Download" header so probably not entirely necessary.

M-Kusumgar:
> I think the "Download" header and the individual file names might be a little too close in styling, and the rows might look better overall with a little less bold

agreed, have replaced with a p tag with appropriate font size and made title bold, file extensions a little less bold, looks much better thanks! I have added a tiny bit of space between the rows to split them up a little bit but not quite as much as that example because i thought it looked too spaced out but happy to be overruled!

> One other small addition to consider is a "Download" tooltip on the icon buttons, though the icon is quite recognizable and they're already under a "Download" header so probably not entirely necessary.

I havent done this one because i think the tooltip actually looks weird, like as you hover over each row, could quite think about the placement because it will either cover the table which i think looks bad or itll go out of our hint container if we put the tooltip on the right hand side

PR #1051, opened 2024-12-12
Add missing translations for comparison table tooltips

r-ash:
## Description

This PR adds translations for comparison table tooltips

## Type of version change

None

## Checklist


src/app/static/src/app/components/plots/table/utils.ts
```diff
@@ -161,7 +162,7 @@ export class InputComparisonTooltip implements ITooltipComp {
 }
 
 
-const inputComparisonTooltipCallback = (params: ITooltipParams) => {
+const inputComparisonTooltipCallback = (params: ITooltipParams & { language: Language }) => {
```
M-Kusumgar, line 165: definitely up for more accurate types but you dont actually use `params.language` anywhere, i guess `i18next` has some sort of global language variable and it just knows the language? in which case is it even necessary to pass in this extra language param?

src/app/static/src/app/store/translations/locales.ts
```diff
@@ -714,6 +716,8 @@ const en: Translations = {
     showBaseMap: "Show base map",
     sizeIndicator: "Size Indicator",
     spectrum: "Spectrum",
+    spectrumReallocation: "Spectrum rellocation:",
+    spectrumAdjustmentRatio: "Naomi:Spectrum adjustment ratio:",
```
M-Kusumgar, line 720: i found the `<something>:<something>:` a little bit harder to parse, perhaps `<something>:<something> - `? so
```
spectrumAdjustmentRatio: "Naomi:Spectrum adjustment ratio - "
```

M-Kusumgar (commented):
got a couple of comments, mostly optional though

src/app/static/src/app/components/plots/table/utils.ts
r-ash: You're right, we don't need this. Thanks! I have taken it out.

M-Kusumgar (approved):
Nice :)

PR #1052, opened 2024-12-20
fix bubble plot bugs - one indicator missing + scaleSettings undefined

M-Kusumgar:
Quick one here, I found a bug where if one of the two indicators doesn't have data then the bubble plot still thinks it isnt empty but of course these two indicators are linked so it needs both indicators to be present, so have included that check.

Now i guess the other thing to consider is what happens if the two indicators are there but theres incomplete data for each of them? i am not sure but for now a simple fix

Also i am pretty sure this isnt what was happening in south africa, i think it was something else because in this bug the entire size legend doesnt show and neither do the bubbles so its different behaviour, i think the scale settings mightve been undefined, i managed to get that error once with Lesotho in the naomi support account but then havent managed to recreate it since?? no clue what is going on but also including logic for that just to be on the safe side

### DIFF
--- a/src/app/static/src/app/components/plots/bubble/Bubble.vue
+++ b/src/app/static/src/app/components/plots/bubble/Bubble.vue
@@ -33,7 +33,6 @@
 import {computed, onMounted, ref, watch} from "vue";
 import {useStore} from "vuex";
 import {RootState} from "../../../root";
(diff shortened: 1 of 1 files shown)


r-ash (approved):
Thanks for this, this looks great

PR #1053, opened 2024-12-23
[NM-97] Improve time series expanded styling

r-ash:
## Description
This PR will address styling comments from https://github.com/hivtools/hint/pull/1046 namely
1. Make the colours a bit less noisy by using some line styling instead to distinguish plots
2. Add some highlighting when hovering over the legend to show the line in the plot (I also added this if you hover over the the equation) 

I also tried to improve the layout in the legend to use a grid, or just centre it if there is only 1 item in the legend. Very open for feedback on the approach and look here, as I found that super fiddly to get right.

I will make a ticket to add a "save plot" button with the legend included

## Type of version change

Patches

## Checklist


M-Kusumgar (commented):
Love the hover and fade stuff and the different line styles! however i do prefer a simpler legend styling, had a play around with it and i preferred just `display: grid` and `justify-content: center` on it which looks like this:
image
and on smaller screens it looks like this:
image

the reason is because while the original design was alright on bigger screens:
image
on smaller screens it looks a bit messy:
image

i also think stacked legends are clearer

r-ash:
> i also think stacked legends are clearer

Thanks! Agreed, that does look better. I've updated that

PR #1054, opened 2024-12-24
[NM-106] Add tests for page number component

r-ash:
## Description

During workshops I added a page number component to allow users to manually enter the page of the plots they wanted to see. I didn't write any automated tests in the hope of getting it in quickly. This PR adds tests for the logic for the paging components.

See #1044 for the original PR.

## Type of version change

None

## Checklist

### DIFF
--- a/src/app/static/src/app/components/plots/timeSeries/PageControl.vue
+++ b/src/app/static/src/app/components/plots/timeSeries/PageControl.vue
@@ -3,7 +3,7 @@
         <button id="previous-page"
                 class="btn btn-sm mr-2 btn-red"
                 v-translate:aria-label="'previousPage'"
-                :disabled="pageNumber === 1"
+                :disabled="pageNumber == 1"
                 @click="pageUpdate(pageNumber - 1)">
             <vue-feather type="chevron-left" size="20" class="pagination-icon"></vue-feather>
         </button>
--- a/src/app/static/src/app/components/plots/timeSeries/PageControl.vue
+++ b/src/app/static/src/app/components/plots/timeSeries/PageControl.vue
@@ -12,7 +12,6 @@
                       class="mr-1"
                       :value="pageNumber"
                       @set-value="pageUpdate"
-                      type="number"
                       name="page-number-input"
                       :min="1"
                       :max="totalPages"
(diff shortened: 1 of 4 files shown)


src/app/static/src/app/components/plots/timeSeries/PageControl.vue
```diff
@@ -3,7 +3,7 @@
         <button id="previous-page"
                 class="btn btn-sm mr-2 btn-red"
                 v-translate:aria-label="'previousPage'"
-                :disabled="pageNumber === 1"
```
M-Kusumgar, line 6: i think this can still remain `===` no? would prefer not to use `==`

M-Kusumgar (commented):
tests look good, couple of thoughts on the number input itself:
* being able to enter floats feels weird and can be easily avoided via regex check
* once i am on a page, e.g. page 3 out of 3 and then i enter 4, which should just get clamped to 3, it doesnt update, because im already on 3, it would be nice if it updated everytime so that if i enter 4 while on page 3 (the max) then it updates to 3
* would be nice if while i am typing the page number, i actually knew what page i was on, perhaps in the future it is more like "Page <number> of <total>" like originally and then a "Go to page <number-input>" on the side or below? something like that but yh would just be nice to know the current page number

src/app/static/src/app/components/plots/timeSeries/PageControl.vue
r-ash: umm yes, not very sure why I did this

r-ash:
> tests look good, couple of thoughts on the number input itself:
> 
>     * being able to enter floats feels weird and can be easily avoided via regex check
> 
>     * once i am on a page, e.g. page 3 out of 3 and then i enter 4, which should just get clamped to 3, it doesnt update, because im already on 3, it would be nice if it updated everytime so that if i enter 4 while on page 3 (the max) then it updates to 3
> 
>     * would be nice if while i am typing the page number, i actually knew what page i was on, perhaps in the future it is more like "Page  of " like originally and then a "Go to page " on the side or below? something like that but yh would just be nice to know the current page number

Fixed the 2nd point here, not fixed the first as it's super fiddly (we chatted about this). Agree with 3rd point, for the future!

M-Kusumgar (approved):
Nice

PR #1056, opened 2025-01-09
Fix modal overflow

krisdoyon:
## Description

Removes overflow: auto property from css for modal component. This was causing the dropdown used to select the dataset in the ADR modal to get cut off within the modal. Tested locally with varying screen sizes on the detailed time series modal and scrollbars still appear there as necessary.

## Type of version change

Patches

## Checklist


M-Kusumgar:
> I'm not sure what the best way to fix this is, we could put `overflow: visible` on just this modal but we could potentially see the same issue again with other modals containing a dropdown. I wonder if there is a way to fix this by styling the dropdown options so they sit on top of whatever containing div they're in? Tagging @M-Kusumgar too in case he has any bright ideas.

im not sure that is possible, the way this type of behaviour is usually done is to have the modal as normal and on top of it (not in it) you place the dropdown, so the dropdown menu and dropdown float above the modal and arent restricted by overflows, if the treeselect is in the modal i think this behaviour will continue

i did some fiddling around on this branch and it got me considering, why is that dropdown in a modal? like is it necessary? we could just have the dropdowns on the page and once they select a release/dataset we collapse that part of the ui to show them a summary? i dont think we will ever need a dropdown in a modal on this app. an example of this below:

summary
image

expanded
image

PR #1058, opened 2025-01-27
[NM-109] Add selections to time series plot

krisdoyon:
## Description

This PR adds checkboxes to the input time series plot. After selecting plots the button in the bottom left can be used to toggle showing just selected plots vs all plots. I had to do this with absolute positioning of the checkboxes over the plotly plot since plotly unfortunately doesn't support html elements like checkboxes (at least not that I could figure out). I'm getting the rendered size for the plot layer and then working out some calculations for absolute position depending on which row/column the plot falls in. It looks fine on different screen sizes but still feels a bit hacky to me, if any suggestions for a better way to do this please let me know and I'm happy to look at a refactor if useful.

Currently selections will be preserved for any filter changes while remaining on the time series tab. Navigating away and coming back will reset them as state for the selected plots is within the component. If we do end up wanting to preserve user selections I think it's just a matter of lifting up the state into the store and then adding an accompanying store action or two. Also wondering if we want to maybe add some kind of "reset" button to allow users to start from scratch with their checkbox selections; maybe something similar to the existing filter reset?

Hopefully this at least gets the desired functionality in for the workshop and we can reassess if needed after getting some user feedback.

## Type of version change

Patches

## Checklist


src/app/static/src/app/components/plots/timeSeries/TimeSeries.vue
```diff
@@ -109,36 +157,108 @@ export default defineComponent({
...
+            if (container) {
+                const { width, height } = container.getBoundingClientRect();
+                chartContainerSize.value.width = width;
+                chartContainerSize.value.height = height;
+            }
+        };
+
+        watch(showChecked, () => {
+            pageNumber.value = 1;
+        });
```
r-ash, line 246: Could stick this `pageNumber.value = 1` into `handleShowChecked`?

src/app/static/src/app/components/plots/timeSeries/TimeSeries.vue
```diff
@@ -109,36 +157,108 @@ export default defineComponent({
...
+
+        onMounted(() => {
+            updateChartContainerSize();
+            window.addEventListener("resize", updateChartContainerSize);
+        });
+
+        onUnmounted(() => {
+            window.removeEventListener("resize", updateChartContainerSize);
+        });
```
r-ash, line 261: Agreed this is kind of gross, but maybe necessary. I don't have any bright ideas for working around it, perhaps @M-Kusumgar does.

I do wonder though if worth us refactoring this to instead of using the plotly multiplot, instead build the multiple plots ourselves like we do for population pyramids. Then we could have a component for each plot, draw this checkbox and the "expanded view" button, might be a bit easier to position nicely with the title if we do it like this. The expanded view is also quite fiddly. Not for now though.

src/app/static/src/app/components/plots/timeSeries/Plotly.vue
```diff
@@ -122,6 +122,11 @@ export default defineComponent({
             if (!props.layout.isModal) drawChart()
         });
 
+        watch(() => [...props.chartData], (newData, oldData) => {
+          if (newData.length !== oldData.length) {
+               drawChart();
+          }});
```
r-ash, line 128: What does this do? Seemed to work ok for me without it?

r-ash (approved):
Thanks Kris, this looks really great. I'm happy to merge as is but left a couple of questions. Also tagged @M-Kusumgar, he might have some better ideas than me for arranging the check boxes. I agree the set up there is (necessarily) gross. I wonder if worth switching to drawing the plots individually, and us managing the grid arrangement ourselves instead of using the plotly multiplot layout. Might simplify some of this and the "expanded view" button too.

I wonder if we could move the checkbox stuff into a composable just to break up the time series component a bit. Maybe not worth it.. it's not that big

Agree lets put this into state in a later PR (if Rachel still wants it). I also agree makes sense to get it in and get some user feedback.

src/app/static/src/app/components/plots/timeSeries/TimeSeries.vue
krisdoyon: Yes good spot I will change this

src/app/static/src/app/components/plots/timeSeries/TimeSeries.vue
krisdoyon: I think you are spot on with the individual plots vs multiplot being a good solution to the layout issues. I can keep this in mind when adding the same functionality to population pyramids to make it easier if we do decide to refactor.

src/app/static/src/app/components/plots/timeSeries/Plotly.vue
krisdoyon: I was having issues when doing the following:

1. Select several checked plots
2. Click show selected button
3. Uncheck one of the selected plots

Before adding the watch, in some cases, this was causing the checkbox to be removed from the screen but not the unchecked plot. Specifically with the MWI test data it was happening when the steps above occurred with the area level filter set to region.

